PR #1399, opened 2022-12-10
(WIP) test: add back error message test

h-a-n-a:

- [ ] No


hardfist:
close for stale

PR #1681, opened 2023-01-20
refactor(crates/node_binding): use concurrent compiler error instead of locking

h-a-n-a:
## Summary

Align with webpack's concurrent compiler error if compiler run is called multiple times

- [ ] No


hyf0 (dismissed):

underfin (dismissed):

h-a-n-a:
Closed for stale

PR #1720, opened 2023-02-01
feat: add styled-jsx support

jkzing:
## Summary

Add [styled-jsx](https://github.com/vercel/styled-jsx) transform support via [styled_jsx](https://crates.io/crates/styled_jsx).

A new switch added to `builtins` to control this behavior: 

```js
module.exports = {
  builtins: {
    styledJsx: true
  }
}
```

- [ ] No


hardfist:
styled-jsx is not widely used like styled-components and emotion, so it would be better to put it into swc-loader other than rspack core

PR #1878, opened 2023-02-16
feat: async-node target support

ScriptedAlchemy:
## Summary

Work in progress of adding the "async-node" target to rspack

ReadFileChunkLoading is based off this: 
https://github.com/ScriptedAlchemy/webpack/blob/49dab4cc5ec8507ab4cad766e2a37a23d9f8a8d1/lib/node/ReadFileChunkLoadingRuntimeModule.js

and would require commonjs chunk plugin to be aware of if its async-node or node
https://github.com/ScriptedAlchemy/webpack/blob/2df826722016f844e7a618f5f787ffa6de43c9b7/lib/node/CommonJsChunkLoadingPlugin.js

this lets you use fs.readFile instead of require.

This is very useful as a modified version of this enabled module federation on node servers

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ English | [简体中文](./README.zh-CN.md)
 ## ✨ Features
 
 - 🚀 **Fast**: Based on Rust, the build speed is extremely fast, bringing you the ultimate development experience.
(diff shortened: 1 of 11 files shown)


hardfist:
@ScriptedAlchemy do you still want to work on this, if not I will work on this since it's needed for our test

hardfist:
since async-node target is supported in #3468  this pr could be closed

PR #2119, opened 2023-03-10
ci: 🎡 using public CI resource

IWANABETHATGUY:
## Summary
1. The only metric we care about in benchmark CI (not benchmark history CI) is the diff ratio, Since we always run twice (pr, and main branch) in the same CI machine, so using public CI to test benchmark is safe.

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


.changeset/heavy-countries-drop.md
```diff
@@ -0,0 +1,21 @@
+---
```
h-a-n-a, line 1: We don’t have to add changeset in this scenario, as it is not supposed to create version bump.

IWANABETHATGUY: Done

PR #2153, opened 2023-03-10
feat: create-rspack add template-vue

### LINKED ISSUE #2107: [Feature Request]: add template-vue for create-rspack

Currently `create-rspack` only has react template, can't create other project with rspack

add prompt for users to select their preferred framework, and I'm willing work for it.

faga295:
## Summary
closed #2107 
It has some differences from `examples/vue`:
* remove less loader
* remove `builtins/define` and `devServer` in `rspack.config.js`(I'm not entirely sure are these necessary)

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


hardfist:
hold until we supports vue-loader

packages/create-rspack/template-vue/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,42 @@
+<script >
```
fwqaaq, line 1:
```suggestion
<script>
```

faga295:
@hardfist  cc

packages/create-rspack/index.js
```diff
@@ -45,7 +45,8 @@ yargs(hideBin(process.argv))
 				message: "Project template",
 				choices: [
 					{ title: "react", value: "react" },
-					{ title: "react-ts", value: "react-ts" }
+					{ title: "react-ts", value: "react-ts" },
+					{ title: "vue", value: "vue" }
```
hardfist, line 49: maybe we should use "vue3" here

faga295: Currently vue3 is the default version of vue, I don't think we need to point out the vue version, vue2 is the one that needs to point out

hardfist: I'm not a vue user, it that's the case then it's ok for me

hardfist (approved):

PR #2171, opened 2023-03-11
feat: added the handling of duplicate project names

Nirvana-Jie:
## Summary

Added the handling of duplicate project names

## Related issue (if exists)
image

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/create-rspack/index.js
```diff
@@ -4,23 +4,29 @@ const { hideBin } = require("yargs/helpers");
...
+					targetDir = formatTargetDir(state.value) || defaultTargetDir;
+				}
 			}
 		]);
-		const { projectDir } = result;
-		const root = path.resolve(process.cwd(), projectDir);
+		const root = path.resolve(process.cwd(), targetDir);
 		if (fs.existsSync(root)) {
-			throw new Error("project directory already exists");
+			emptyDir(root);
```
h-a-n-a, line 25: I guess we're not supposed to delete the user directory by accident, so it will be better to prompt up and give users other shots until there are no conflicts.

.changeset/friendly-mice-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-rspack": major
```
h-a-n-a, line 2: This should be a 'patch' change.

packages/create-rspack/utils/index.js
```diff
@@ -0,0 +1,28 @@
...
+exports.emptyDir = function (dir) {
+	if (!fs.existsSync(dir)) {
+		return;
+	}
+	for (const file of fs.readdirSync(dir)) {
+		if (file === ".git") {
+			continue;
+		}
+		fs.rmSync(path.resolve(dir, file), { recursive: true, force: true });
+	}
```
fwqaaq, line 27:
```suggestion
	for (const file of fs.readdirSync(dir)) {
		fs.rmSync(path.resolve(dir, file), { recursive: true, force: true });
	}
```

To determine if the `.git` folder is unnecessary, you should check if this folder is empty (i.e., only containing .git), and use the `break` keyword instead of `continue` to exit the loop. Alternatively, you can write an `isEmpty()` function.

```js
function isEmpty(path) {
  const files = fs.readdirSync(path)
  return (
    files.length === 0 ||
    (files.length <= 2 &&
      files.every((f) => f === '.git' || f.toUpperCase() === 'LICENSE'))
  )
}
```

packages/create-rspack/index.js
Nirvana-Jie: Yes，I will optimize it.Thank you

Nirvana-Jie: done

.changeset/friendly-mice-taste.md
Nirvana-Jie: I have made changes

packages/create-rspack/index.js
```diff
@@ -4,29 +4,45 @@ const { hideBin } = require("yargs/helpers");
...
+				}
 			}
 		]);
-		const { projectDir } = result;
-		const root = path.resolve(process.cwd(), projectDir);
-		if (fs.existsSync(root)) {
-			throw new Error("project directory already exists");
+		let root = path.resolve(process.cwd(), targetDir);
+		while (fs.existsSync(root)) {
+			console.log(`${targetDir} is not empty,please reset the projrct name`);
```
h-a-n-a, line 25:
```suggestion
			console.log(`${targetDir} is not empty, please choose another project name`);
```

Nirvana-Jie: Done,Thank you

PR #2186, opened 2023-03-12
fix(dev-middleware): compatibility with native node handler

danielroe:
## Summary

Currently there are two issues I'm facing with dev-middleware:

1. it uses `express`-only handlers like `req.path` and `res.send()`
2. it does not respect a public path (I may be misunderstanding how this is meant to be used though - should this only be registered to handle paths beginning with the publicPath?

This PR uses node-compatible methods (`req.url` and `res.end()`) which will also work in express, expanding the compatibility of this library. It also normalises any URLs beginning with `publicPath` so they can be handled in this middleware.

(I am happy to remove the change to publicPath handling if this is intended.)

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/eleven-falcons-approve.md
+++ b/.changeset/eleven-falcons-approve.md
@@ -1,0 +1,5 @@
+---
+"@rspack/dev-middleware": patch
+---
+
+Use native Node compatible methods in middleware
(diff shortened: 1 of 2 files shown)


ahabhgk (approved):
Thank you!

PR #2218, opened 2023-03-13
perf(rspack_core): Improve code splitting

Boshen:
## Summary

Previously we were using `HashMap<usize, _>` for id mappings inside the module graph. Performance degrades heavily when there are lots of modules (100,000+) due to hashing inside hot paths.

In this PR, I changed the backing storage with a `Vec`, the index becomes the id so lookups become O(1).

As a bonus, both `AtomicUsize`s are removed, because `ModuleGraph` algorithms are single threaded.

image

### DIFF
--- a/crates/rspack_core/src/module_graph.rs
+++ b/crates/rspack_core/src/module_graph.rs
@@ -2,7 +2,6 @@ use std::borrow::Cow;
 use std::cmp::PartialEq;
 use std::collections::hash_map::Entry;
 use std::hash::{Hash, Hasher};
-use std::sync::atomic::{AtomicUsize, Ordering};
 
 use rspack_error::{internal_error, Result};
 use rspack_identifier::IdentifierMap;
@@ -12,8 +11,7 @@ use crate::{
   BoxModule, BoxModuleDependency, DependencyId, Module, ModuleGraphModule, ModuleIdentifier,
 };
 
-// FIXME: placing this as global id is not acceptable, move it to somewhere else later
-static NEXT_MODULE_GRAPH_CONNECTION_ID: AtomicUsize = AtomicUsize::new(1);
+type ModuleGraphConnectionId = usize;
 
 #[derive(Debug, Clone, Eq)]
 pub struct ModuleGraphConnection {
@@ -23,9 +21,8 @@ pub struct ModuleGraphConnection {
   pub module_identifier: ModuleIdentifier,
   /// The referencing dependency id
   pub dependency_id: usize,
-
   /// The unique id of this connection
-  pub id: usize,
+  id: ModuleGraphConnectionId,
 }
 
 impl Hash for ModuleGraphConnection {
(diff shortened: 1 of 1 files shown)


crates/rspack_core/src/module_graph.rs
```diff
@@ -23,9 +21,8 @@ pub struct ModuleGraphConnection {
   pub module_identifier: ModuleIdentifier,
   /// The referencing dependency id
   pub dependency_id: usize,
-
   /// The unique id of this connection
-  pub id: usize,
+  id: ModuleGraphConnectionId,
```
Boshen, line 25: Removing the `pub` makes this struct's constructor private.

crates/rspack_core/src/module_graph.rs
```diff
@@ -44,37 +41,24 @@ impl PartialEq for ModuleGraphConnection {
...
-
   /// Module identifier to its module
   pub(crate) module_identifier_to_module: IdentifierMap<BoxModule>,
   /// Module identifier to its module graph module
   pub module_identifier_to_module_graph_module: IdentifierMap<ModuleGraphModule>,
 
-  dependency_id_to_connection_id: HashMap<usize, usize>,
-  connection_id_to_dependency_id: HashMap<usize, usize>,
+  dependency_id_to_connection_id: HashMap<DependencyId, ModuleGraphConnectionId>,
+  connection_id_to_dependency_id: HashMap<ModuleGraphConnectionId, DependencyId>,
```
Boshen, line 52: These two lookups are not in hot paths right now so I didn't change it.

crates/rspack_core/src/module_graph.rs
```diff
@@ -12,8 +11,7 @@ use crate::{
   BoxModule, BoxModuleDependency, DependencyId, Module, ModuleGraphModule, ModuleIdentifier,
 };
 
-// FIXME: placing this as global id is not acceptable, move it to somewhere else later
-static NEXT_MODULE_GRAPH_CONNECTION_ID: AtomicUsize = AtomicUsize::new(1);
+type ModuleGraphConnectionId = usize;
```
Boshen, line 14: These ids should be wrapped inside as newtype, but I'll leave it to the next person ;-)

crates/rspack_core/src/module_graph.rs
```diff
@@ -125,7 +124,7 @@ impl ModuleGraph {
   }
 
   pub fn module_identifier_by_dependency_id(&self, id: &DependencyId) -> Option<&ModuleIdentifier> {
-    self.dependency_id_to_module_identifier.get(id)
+    self.dependency_id_to_module_identifier[*id].as_ref()
```
Boshen, line 127: All `[]` index access are safe because we do append only operations. We can refactor the backing `Vec` to a newtype if we want to enforce this. But I'll leave this to the next person.

Boshen:
The IDs should be global, closing for now.

PR #2259, opened 2023-03-15
Chore/logger overwrite

PengBoUESTC:
## Summary

dev-server log info with `webpack-dev-server`
## Types of changes

- [ ] Bug fix
- [x] Refactoring
- [ ] Breaking change

## Checklist


PengBoUESTC:
maybe proxy the console's  functions can deal with these problems (log messages include `webpack` info)

packages/rspack-dev-client/src/index.ts
```diff
@@ -13,6 +13,25 @@ import {
 	setLogLevel,
 	logEnabledFeatures
 } from "webpack-dev-server/client/utils/log";
+import logger from 'webpack-dev-server/client/modules/logger/index'
```
bvanjoi, line 16: Personally, I don't think this piece of code should be changed, as its ultimate goal is to use `webpack-dev-server/clinet` directly

PengBoUESTC: but you didn't use `webpack-dev-server` directly,  `rspack` provides `rspack-dev-server` package,  developer use `rspack` to develop project, but the logs show `webpack`,  it's a little weird.

PR #2272, opened 2023-03-15
fix: add compile hook

zscumt123:
## Summary
I noticed that the `compile hook` functions seem to not be called in source  code, but it in [documents](https://www.rspack.dev/api/plugin-api.html#compile) ,and I'm not sure if this is intentional,and I'm not sure if my implementation is correct.

## Types of changes

- [x] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/rspack/src/compiler.ts
```diff
@@ -374,6 +374,7 @@ class Compiler {
 	}
 
 	#newCompilation(native: binding.JsCompilation) {
+		this.hooks.compile.call();
```
IWANABETHATGUY, line 377: The `call` timing is not right, you could refer https://github.com/webpack/webpack/blob/d96ce76f4d3996f32367b4b4231039451550c1f7/lib/Compiler.js#L1161-L1203 and https://github.com/webpack/webpack/blob/d96ce76f4d3996f32367b4b4231039451550c1f7/lib/Compiler.js#L1117-L1124, 
the correct timing is 
```mermaid
flowchart LR
   compile-. newCompilation .->thisCompilation-->compilation
```

IWANABETHATGUY (changes requested):

hardfist:
this mr is outdated, feel free to open a new one if you're still interested

PR #2279, opened 2023-03-15
refactor module graph lookups

Boshen:
## Summary

There are two bottlenecks around `ModuleIdentifer`:

* it is a long path, which takes a lot of time to compute the hash
* it is a `Ustr`, accessing the contents of the `Ustr` is a roundtrip to the heap, which is slow

When we use `ModuleIdentifier` as the key for `HashMap`s, the profiler shows performance hot spots in hot paths.

This PR removes the need to read `ModuleIdentifier` in hot paths, and uses a vector as the backing storage. `ConnectionId` and `Dependency` Id are the indexes to their respective vectors.

For a large project with 20k modules, code splitting time is now reduced from 7s to 6s.

As a bonus, the `AtomicUsize`s are now gone :-)

### DIFF
--- a/crates/rspack_core/src/module_graph/connection.rs
+++ b/crates/rspack_core/src/module_graph/connection.rs
@@ -1,11 +1,4 @@
-use std::cmp::PartialEq;
-use std::hash::{Hash, Hasher};
-use std::sync::atomic::{AtomicUsize, Ordering};
-
 use crate::{DependencyId, ModuleIdentifier};
-
-// FIXME: placing this as global id is not acceptable, move it to somewhere else later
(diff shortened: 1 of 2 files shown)


hyf0 (dismissed):

IWANABETHATGUY (dismissed):

hyf0:
LGTM. You may need the final review of @h-a-n-a.

h-a-n-a (dismissed):

h-a-n-a (approved):

IWANABETHATGUY (approved):

hyf0 (approved):

PR #2294, opened 2023-03-16
perf(core): improve cold startup performance by caching `depended_modules` function in code splitter

Boshen:
## Summary

Code splitting has a hot path on `depended_modules`, this PR caches the result.
In a project a with 20k modules, code splitting time is reduced from 6s to 2s.

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/crates/rspack_core/src/build_chunk_graph/code_splitter.rs
+++ b/crates/rspack_core/src/build_chunk_graph/code_splitter.rs
@@ -1,6 +1,6 @@
 use anyhow::anyhow;
 use rspack_error::Result;
-use rspack_identifier::IdentifierSet;
+use rspack_identifier::{IdentifierMap, IdentifierSet};
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 use super::remove_parent_modules::RemoveParentModulesContext;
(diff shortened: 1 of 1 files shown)


Boshen:
@hyf0 this has a better perf improvement, please take a look.

crates/rspack_core/src/build_chunk_graph/code_splitter.rs
```diff
@@ -322,24 +324,30 @@ impl<'me> CodeSplitter<'me> {
...
+      .entry(item.module_identifier)
+      .or_insert_with(|| {
+        mgm
+          .depended_modules(&self.compilation.module_graph)
+          .into_iter()
+          .rev()
+          .map(|module_identifier| QueueItem {
+            action: QueueAction::AddAndEnter,
+            chunk: item.chunk,
+            chunk_group: item.chunk_group,
```
hyf0, line 345: Good works!

https://github.com/web-infra-dev/rspack/blob/70519c86eda8a802af7e0602afb3777ffad3c33c/crates/rspack_core/src/build_chunk_graph/code_splitter.rs#L344-L345

But the captured values in the above lines are incorrect. It would add `Module` to the unexpected `Chunk` while hitting the cache.

I suggest you could only cache the modules, such like 
 `depended_modules_cache: IdentifierMap<Vec<ModuleIdentifer>>`

hyf0 (changes requested):

crates/rspack_core/src/build_chunk_graph/code_splitter.rs
Boshen: @hyf0 Done. The performance gain is still significant.

hyf0 (dismissed):

hyf0 (approved):
Sorry for the latency.

PR #2300, opened 2023-03-16
fix(ci): clean up check-rs, attempt to fix ci cache taking 1G

Boshen:
CI cache on check-rs is taking about 1G, I think it's due to double caching here.

### DIFF
--- a/.github/workflows/check-rs.yaml
+++ b/.github/workflows/check-rs.yaml
@@ -26,36 +26,28 @@ jobs:
     name: lint, format and spell check
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout PR Branch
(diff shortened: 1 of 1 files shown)


.github/workflows/check-rs.yaml
```diff
@@ -26,36 +26,28 @@ jobs:
...
-      - name: Checkout PR Branch
+      - name: Checkout repository
         uses: actions/checkout@v3
+
+      - name: Install protoc
+        uses: arduino/setup-protoc@v1
         with:
-          submodules: false
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Support longpaths
```
IWANABETHATGUY, line 34: Why do you remove this line

PR #2325, opened 2023-03-17
fix(cli): set stats default to errors-warning to avoid server info flushhed by module stats

hardfist:
## Summary

none of react cli prints stats.modules info expect webpack-cli, which is not friendly to developer since it will flush server info out of screen
## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/rspack-cli/src/rspack-cli.ts
+++ b/packages/rspack-cli/src/rspack-cli.ts
@@ -155,7 +155,7 @@ export class RspackCLI {
 			}
 
 			if (typeof item.stats === "undefined") {
-				item.stats = { preset: "normal" };
+				item.stats = { preset: "errors-warnings" };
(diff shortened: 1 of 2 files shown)


packages/rspack-cli/src/rspack-cli.ts
```diff
@@ -155,7 +155,7 @@ export class RspackCLI {
 			}
 
 			if (typeof item.stats === "undefined") {
-				item.stats = { preset: "normal" };
+				item.stats = { preset: "errors-warnings" };
```
chenjiahan, line 158: How about add a changeset

h-a-n-a (approved):

ulivz (approved):

chenjiahan (approved):

PR #2404, opened 2023-03-21
refactor: css dependencies

ahabhgk:
## Summary

also fixes `url(https://xxx.xx/~img.png)`
------------------------------- ^ here

## Types of changes

- [x] Bug fix
- [x] Refactoring
- [ ] Breaking change

## Checklist


crates/rspack_plugin_css/src/dependency/code_generation.rs
```diff
@@ -0,0 +1,258 @@
+/**
+ * Some code is modified based on
+ * https://github.com/vercel/turbo/blob/a1947f64443fb98e5c3e10bca6ef9eafd278bd21/crates/turbopack-css/src/path_visitor.rs
```
hardfist, line 3: why more code from turbo，this is unacceptable 😢

ahabhgk: Because css dependencies didn't refactor to the new dependencies architecture before, so there are only JavaScript dependencies helpers which reference from turbo exist.

ahabhgk: Honestly I don't know what the advantage of AST-based dependencies for now, but I know there are advantages of source-based dependencies.
- more friendly for persist cache 
- AST less so swc won't be the bottleneck
- architecture more close to webpack, different from turbo at the root 
But need to rewrite lots of code 🤷

h-a-n-a (approved):

hardfist:
This is a temporarily workaround and will refactor to string replacement in the future
cc @ahabhgk @h-a-n-a

Issue #2464, opened 2023-03-25
[Feature Request]: Add webpack's `ProvidePlugin` to rspack's built-ins


SukkaW:
### What problem does this feature solve?

Currently, I am using `webpack.ProvidenPlugin` to polyfill Node.js built-in objects for some libraries used in my web project:

```js
new ProvidePlugin({
  // Polyfill for Node global "Buffer" variable
  Buffer: [require.resolve('buffer'), 'Buffer']
})
```

https://webpack.js.org/plugins/provide-plugin/

I don't see any alternative option in rspack.

### What does the proposed API of configuration look like?

```js
{
  builtins: {
    provide: {
      Buffer: [require.resolve('buffer'), 'Buffer']
    }
  }
}
```

Issue #1358, opened 2022-12-06
support shim global with modules 


hardfist:
It's very common that some third party library use some non standard global variable like `global`、`process`
```ts
module.exports = {
   builtins: {
      provide: { 
         'global': './shims/global.js'
     }
   }
}
```
like https://webpack.js.org/plugins/provide-plugin/

dhruvkelawala:
> It's very common that some third party library use some non standard global variable like `global`、`process`
> 
> ```ts
> 
> module.exports = {
> 
>    builtins: {
> 
>       provide: { 
> 
>          'global': './shims/global.js'
> 
>      }
> 
>    }
> 
> }
> 
> ```
> 
> like https://webpack.js.org/plugins/provide-plugin/

Is someone working on this? We want to use rspack in our project but we need to "provide" things like process, buffer, etc. I can create a PR if no one is currently working on this.

hyf0:
> I can create a PR if no one is currently working on this.

No one is currently working on this at the moment. But if you're interested in contributing, don't hesitate to open a pull request! Once it's submitted, I'll be happy to review it.

And if you need any guidance, feel free to reach out to me on Discord or GitHub.

Thank you in advance.

hardfist:
there's a workaround with it right now and we also use it in internal projects
* rspack.config.js
```ts
module.exports = {
  entry: { 
     main: ['./shims.js', './src/index.js']
  }
}
```
* shims.js
```ts
var global = globalThis;
```

dhruvkelawala:
Thanks for replying @hyf0 @hardfist. I am almost half-way through the PR adding `provide` as a builtin. Hopefully will get some more work done during the weekend

hyf0:
Closed by #2492

PR #2492, opened 2023-03-27
feat: built-in support for ProvidePlugin

dhruvkelawala:
## Summary

Support the `ProvidePlugin` as a `provide` builtin in rspack

## Related issue (if exists)

#1358 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
@@ -1,0 +1,156 @@
+use std::sync::Arc;
+
+use rspack_core::Provide;
+use swc_core::common::Span;
+use swc_core::common::{errors::Handler, SourceMap, DUMMY_SP};
+use swc_core::ecma::ast::{
+  CallExpr, Callee, ComputedPropName, Expr, ExprOrSpread, Ident, ImportDecl, Lit, MemberExpr,
+  MemberProp, ModuleDecl, ModuleItem, Str,
+};
+use swc_core::ecma::visit::{Fold, FoldWith, VisitMut};
+
+pub fn provide_builtin<'a>(
+  opts: &'a Provide,
+  handler: &'a Handler,
+  cm: &'a Arc<SourceMap>,
+) -> ProvideBuiltin<'a> {
+  ProvideBuiltin::new(opts, handler, cm)
+}
+
+pub struct ProvideBuiltin<'a> {
+  opts: &'a Provide,
+  handler: &'a Handler,
+  cm: &'a Arc<SourceMap>,
+}
+
+impl<'a> ProvideBuiltin<'a> {
+  pub fn new(opts: &'a Provide, handler: &'a Handler, cm: &'a Arc<SourceMap>) -> Self {
+    ProvideBuiltin { opts, handler, cm }
+  }
+
+  fn handle_ident(&self, ident: Ident) -> Expr {
+    if let Some(module_path) = self.opts.get(&ident.sym.to_string()) {
+      self.create_obj_expr(ident.span, module_path)
+    } else {
+      Expr::Ident(ident)
+    }
+  }
+
+  fn handle_member_expr(&self, member_expr: MemberExpr) -> Expr {
+    if let Expr::Ident(ident) = &*member_expr.obj {
+      if let Some(member_prop) = member_expr.prop.as_ident() {
+        let identifier_name = format!("{}.{}", ident.sym.to_string(), member_prop.sym.to_string());
+        if let Some(module_path) = self.opts.get(&identifier_name) {
+          self.create_obj_expr_with_prop(ident.span, module_path)
+        } else {
+          Expr::Member(member_expr)
+        }
+      } else {
+        Expr::Member(member_expr)
+      }
+    } else {
+      Expr::Member(member_expr)
+    }
+  }
+
+  fn create_obj_expr(&self, span: Span, module_path: &[String]) -> Expr {
+    let call_expr = self.create_call_expr(span, &module_path[0]);
+    let mut obj_expr = Expr::Call(call_expr);
+
+    for module_name in module_path.iter().skip(1) {
+      let member_expr = MemberExpr {
+        span,
+        obj: Box::new(obj_expr),
+        prop: MemberProp::Computed(ComputedPropName {
+          span,
+          expr: Box::new(Expr::Lit(Lit::Str(Str {
+            span,
+            value: module_name.to_string().into(),
+            raw: None,
+          }))),
+        }),
+      };
+
+      obj_expr = Expr::Member(member_expr);
+    }
+
+    obj_expr
+  }
+
+  fn create_obj_expr_with_prop(&self, span: Span, module_path: &[String]) -> Expr {
+    let call_expr = self.create_call_expr(span, &module_path[0]);
+
+    if module_path.len() > 1 {
+      let prop_sym = module_path[1..].join(".").to_string();
+
+      let member_expr = MemberExpr {
+        span,
+        obj: Box::new(Expr::Call(call_expr)),
+        prop: MemberProp::Ident(Ident::new(prop_sym.into(), span)),
+      };
+
+      Expr::Member(member_expr)
+    } else {
+      Expr::Call(call_expr)
+    }
+  }
+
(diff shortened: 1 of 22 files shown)


SukkaW:
This also closes https://github.com/web-infra-dev/rspack/issues/2464

crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
```diff
@@ -0,0 +1,156 @@
...
+  fn fold_expr(&mut self, expr: Expr) -> Expr {
+    let new_expr = match expr {
+      Expr::Ident(ident) => self.handle_ident(ident),
+      Expr::Member(member_expr) => self.handle_member_expr(member_expr),
+      _ => expr,
+    };
+    new_expr.fold_children_with(self)
+  }
+
+  fn fold_module_items(&mut self, items: Vec<ModuleItem>) -> Vec<ModuleItem> {
```
hyf0, line 131: Take this as an example. You should be able to do something like
```rust
impl VisitMut for ProvideBuiltin {

  fn visit_mut_module_items(&mut self, node: &mut Vec<ModuleItems>) {
    for (_key, module_paths) in self.opts {
        // println!("key: {:?}", key);
        // println!("module_paths: {:?}", module_paths);

        let import_decl = ImportDecl {
          span: DUMMY_SP,
          specifiers: vec![],
          src: Box::new(Str {
            span: DUMMY_SP,
            value: module_paths[0].to_string().into(),
            raw: Option::None,
          }),
          type_only: false,
          asserts: None,
        };
      node.push(ModuleItem::ModuleDecl(ModuleDecl::Import(import_decl)));
    }

   node.visit_children_with(self)
  }

}

```

hyf0 (commented):
You could also take a look at the code of SWC.

https://github.com/hyf0/swc/blob/6ea29c5cd733ba87bf0700b70cc7d126bf7228a3/crates/swc_ecma_minifier/src/pass/global_defs.rs#L49

You might get some inspiration on how to replace `Expr`/`Ident`.

The code may be complicated due to handling many edge cases. I think handling `Ident` and `MemberExpr` would be enough for the current usage.

hyf0:
Feel free to ping me anytime on GitHub or Discord, But I might not be able to respond immediately if I'm asleep. I'll get back to you as soon as I can.

crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
dhruvkelawala: Sure. I was going to do it with `VisitMut` as it's more performant than`Fold` but inspired from `define` builitin, I thought we should use fold

hyf0: You could use `VisitMut` to implement the feature and use `as_folder` to turn a `VisitMut` into a `Fold`.

For example:

https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/visitors/clear_mark.rs

And yeah, `VisitMut` has better performance than `Fold`.

dhruvkelawala: Thanks for the help!

packages/rspack/tests/configCases/builtins/provide/index.js
```diff
@@ -0,0 +1,68 @@
...
+});
+
+// it("should provide a module for a nested var within a IIFE's argument", function () {
+// 	(function (process) {
+// 		expect(process.env.NODE_ENV).toBe("development");
+// 		var x = process.env.NODE_ENV;
+// 		expect(x).toBe("development");
+// 	})(process);
+// });
+
```
dhruvkelawala, line 18: The commented out tests are failing as of now. Those cases needs to be handled / transformed properly.

crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
```diff
@@ -0,0 +1,168 @@
...
+      Expr::Member(member_expr.clone())
+    }
+  }
+
+  fn create_obj_expr(&self, span: Span, module_path: &[String]) -> Expr {
+    let call_expr = self.create_call_expr(
+      Span {
+        lo: span.lo,
+        hi: span.hi,
+        ctxt: SyntaxContext::from_u32(2),
```
hyf0, line 51: You could get the unresolved `SyntaxContext` by using `SyntaxContext::empty().apply_mark(unresolved_mark)`. 

You could get the unresolved `Mark`  here https://github.com/web-infra-dev/rspack/blob/93975ed98266be67475a0d14bf9501d89cd9134d/crates/rspack_plugin_javascript/src/visitors/mod.rs#L68

crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+  fn visit_mut_module_items(&mut self, node: &mut Vec<ModuleItem>) {
+    for (_key, module_paths) in self.opts {
+      // println!("key: {:?}", key);
+      // println!("module_paths: {:?}", module_paths);
+
+      let import_decl = ImportDecl {
+        span: DUMMY_SP,
+        specifiers: vec![],
```
hyf0, line 153: What's the usage of the `ImportDecl`? I saw it has no specifiers.

dhruvkelawala: Initially I thought we needed to mutate the Import Declarations too. However, after the current implementation, I don't think we need it anymore

hyf0: I see. I'm not sure I understand correctly.

For config like
```js
{
  $: ["jquery", "default"]
}
```

Instead of transforming like
```js
$('div')

// => 

import jquery from 'jquery'
jquery.default('div')
```

the current PR is doing something like

```js
$('div')

// => 

require('jquery').default('div')
```

hyf0: @dhruvkelawala Could you tell me that is my understanding correct? Wonderful work by the way.

dhruvkelawala: Yes, it's working as you described. I think webpack does it in the same way, doesn't it? Am I missing something?

dhruvkelawala: You are right. Maybe we should try to achieve the results similar to webpack with ES module imports. Do you have any pointers how we can achieve that?

dhruvkelawala: Okay 👍

crates/rspack_plugin_javascript/src/visitors/swc_visitor/provide.rs
```diff
@@ -0,0 +1,128 @@
+use rspack_core::Provide;
+use swc_core::common::Span;
+use swc_core::common::{Mark, DUMMY_SP};
+use swc_core::ecma::ast::{
+  CallExpr, Callee, ComputedPropName, Expr, ExprOrSpread, Ident, Lit, MemberExpr, MemberProp, Str,
+};
+use swc_core::ecma::visit::{as_folder, Fold, VisitMut, VisitMutWith};
+
+pub fn provide_builtin<'a>(opts: &'a Provide, unresolved_mark: Mark) -> impl Fold + 'a {
```
hyf0, line 9: The lifetime `'a here seems redundant. You could remove it by following the suggestions of Clippy.

hyf0 (changes requested):

hyf0:
I saw the snapshot tests in js failed, mostly because of adding new options. I will take it from here and finish the rest tedious work.

Hugh thanks for your contribution!

hyf0 (approved):

dhruvkelawala:
> I saw the snapshot tests in js failed, mostly because of adding new options. I will take it from here and finish the rest tedious work.
> 
> Huge thanks for your contribution!

Thanks for you help. Looking forward to more contributions!

PR #2541, opened 2023-03-29
chore: unify line endings for all files for any new clone

ShuiRuTian:
## Summary

Motivation: To make .x file work under Linux.

What is changed: The files will be created with "LF" for any new clone, existing files will not be changed. About the details, please reference this great article about CRLF and LF: https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

Have to say I am kind of confused, the CONTRIBUTING.md seems to be prepared for Linux, but ".x" is using CRLF as line endings, which make it complains "env: node\r: No such file". Note the strange "\r", which is the problem.

We cloud also add such content to .editorconfig to express intension better, I do not add it for now:
``` .editorconfig
[*]
end_of_line = lf
```

This should not cause any problem in Windows, there is no "bat" or "cmd" file in the repo.

## Types of changes

Not sure, I guess it's "refactoring"?
- [ ] Bug fix
- [x] Refactoring
- [ ] Breaking change

## Checklist

Does this change need to fill these checkbox?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 benchcases/** linguist-detectable=false
 fixtures/** linguist-detectable=false
 webpack-test/** linguist-detectable=false
+* text=auto
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,8 @@
 root = true
 
+[*]
+end_of_line = lf
+
 [*.{js,ts}]
 indent_style = tab
 indent_size = 2


hyf0:
IMO, we should enforce `end_of_line = lf` rather than `auto`.

hyf0:
Could you sign the CLA?

And after discussing with the @web-infra-dev/rspack, we think it's better to enforce the end of line with `LF`.

ShuiRuTian:
CLA Signed.

I still keep the change in `.gitattributes`, because in an editor like VSCode, you need an extension to make the file ".editorconfig" work. Then the change in `.gitattributes` provide a good default behavior.

.gitattributes
```diff
@@ -1,3 +1,4 @@
 benchcases/** linguist-detectable=false
 fixtures/** linguist-detectable=false
 webpack-test/** linguist-detectable=false
+* text=auto
```
hyf0, line 4: I'm not familiar with this. Does it achieve the goal of enforcing `line-style: LF`?

ShuiRuTian: The main purpose is to not change every lines of every files.

It will make the codes in remote repo(.i.e, git) ending with LF, and any new clone will be LF, because the files in the remote is LF.

But for your local files, this line does not work. It just tells git to ignore the ending of line as diff. 

You could run `git add --renormalize .`(as suggested in the article link), and install extension "EditorConfig.EditorConfig" to make VSCode to respect ".editorconfig" file.

You could reference this great article https://www.aleksandrhovhannisyan.com/blog/crlf-vs-lf-normalizing-line-endings-in-git/

Issue #2486, opened 2023-03-27
[Feature Request]:  Rspack should validate the version of nodejs


hyf0:
### What problem does this feature solve?

Rspack doesn’t support nodejs lower than 12(include 12).

To provide a better experience, Rspack should validate the version of nodejs where Rspack is running.

### What does the proposed API of configuration look like?

nope.

hyf0:
This is a good first issue!

You could add the following code

```jsx
const semver = require('semver');
const requiredVersion = '>=14.0.0';

if (!semver.satisfies(process.version, requiredVersion)) {
  throw new Error(`Node.js version must be ${requiredVersion}, but found ${process.version}.`)
}
```

 before this line

https://github.com/web-infra-dev/rspack/blob/28b9757460a57998e78474f4c448dadc7e1faf21/packages/rspack-cli/src/bootstrap.ts#L4

to solve this issue.

hyf0:
> also need add notice to website

I think it could be done in a separate PR. I would also create an issue in https://github.com/web-infra-dev/rspack-website.

ycjcl868:
```js
if (semver.lt(semver.clean(process.version), '14.0.0')) {
   // throw error
}
```

amanx123:
Please assign me I would like to make this contribution real quick.

hyf0:
> Please assign me I would like to make this contribution real quick.
> 
> 

Done. Thank you in advance.

hyf0:
@amanx123 Hi, a friendly ping. Do you still work on this?

PR #2575, opened 2023-03-31
feat(rspack-cli): emit warning if the using nodejs version is not recommended

dhruvkelawala:
## Summary

Rspack doesn’t support nodejs lower than 12(include 12).

To provide a better experience, Rspack should validate the version of nodejs where Rspack is running.

## Related issue (if exists)

#2486 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/rspack-cli/src/rspack-cli.ts
```diff
@@ -76,6 +77,13 @@ export class RspackCLI {
 		};
 	}
 	async run(argv: string[]) {
+		if (semver.lt(semver.clean(process.version), "14.0.0")) {
```
dhruvkelawala, line 80: Should the version be 12 or 14?

hyf0: The version should be 14.

packages/rspack-cli/src/rspack-cli.ts
```diff
@@ -76,6 +77,13 @@ export class RspackCLI {
 		};
 	}
 	async run(argv: string[]) {
+		if (semver.lte(semver.clean(process.version), "12.0.0")) {
+			this.getLogger().error(
+				`Node.js version must be greater than 12.0.0, current version is ${process.version}`
+			);
+			process.exit(1);
```
hyf0, line 84: Maybe it's better not to force an exit. I had the same thought as you in the first place, But I think a warning would be enough now. We don't promise Rspack could be running in node12 but if it could, there's no reason to forbid it.

hyf0 (approved):
Thank you!

PR #2576, opened 2023-04-01
fix(dev-client): use webpack-dev-server/client directly

bvanjoi:
This PR brings two changes:

1. Thanks to the implementation of provide, we can now use `webpack-dev-server/client` directly
2. fix pnpm version to 7.25.0


packages/rspack-dev-server/tests/__snapshots__/normalizeOptions.test.ts.snap
```diff
@@ -128,9 +127,8 @@ exports[`normalize options snapshot react-refresh client added when react/refres
 {
   "main": [
     "<prefix>/rspack-dev-client/dist/reactRefresh.js",
-    "<prefix>/rspack-dev-client/dist/devServer.js",
-    "<prefix>/dist/clients/WebSocketClient.js",
```
ahabhgk, line 132: Is this expected?

bvanjoi: yes, because it was added to the `provider`

underfin (approved):

ahabhgk (approved):

hardfist (approved):

PR #2578, opened 2023-04-02
feat(rspack_plugin_dev_friendly_split_chunks): reduce `MAX_MODULES_PER_CHUNK` and add `MAX_SIZE_PER_CHUNK` condition

### LINKED ISSUE #2529: [Feature Request]: `DevFriendlySplitChunksPlugin` should split chunks considering the size of modules and chunks

Some modules are really big, thus, the current implementation only considers the number of modules. It doesn't scale in some cases.

nope.

hyf0:
## Summary

The FS project emits a 30MB js asset due to the number `MAX_MODULES_PER_CHUNK`.

This PR reduced the number of `MAX_MODULES_PER_CHUNK` and add a `MAX_SIZE_PER_CHUNK` condition to split chunks that are bigger than `MAX_SIZE_PER_CHUNK`.

## Related issue (if exists)

Fixes #2529.

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/crates/rspack_plugin_dev_friendly_split_chunks/src/plugin.rs
+++ b/crates/rspack_plugin_dev_friendly_split_chunks/src/plugin.rs
@@ -70,7 +70,7 @@ impl Plugin for DevFriendlySplitChunksPlugin {
 
     // First we group modules by MAX_MODULES_PER_CHUNK
 
-    let splitted_modules = shared_modules
+    let split_modules = shared_modules
       .par_chunks(MAX_MODULES_PER_CHUNK)
       .flat_map(|modules| {
         let chunk_size: f64 = modules
@@ -132,7 +132,7 @@ impl Plugin for DevFriendlySplitChunksPlugin {
       .collect::<DashMap<_, _>>();
 
     // Yeah. Leaky abstraction, but fast.
-    let mut chunk_and_cgc = splitted_modules
+    let mut chunk_and_cgc = split_modules
       .map(|modules| {
         let mut chunk = Chunk::new(None, None, rspack_core::ChunkKind::Normal);
         chunk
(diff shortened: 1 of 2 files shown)


crates/rspack_plugin_dev_friendly_split_chunks/src/plugin.rs
```diff
@@ -64,7 +64,65 @@ impl Plugin for DevFriendlySplitChunksPlugin {
...
+              rspack_core::ModuleType::JsxEsm => 5.0,
+              rspack_core::ModuleType::Tsx => 5.0,
+              _ => 1.5,
+            };
+
+            module.size(&rspack_core::SourceType::JavaScript) * coefficient
+          })
+          .sum();
+
+        if chunk_size > MAX_SIZE_PER_CHUNK {
```
hyf0, line 98: The code doesn't work as expected in the FS project. I suspect that `Module::size` has some problems, but I didn't dig deeper for lacking time. @h-a-n-a cc

In the end, I managed to reduce the emitted js asset sizes by reducing the number of `MAX_MODULES_PER_CHUNK`.

h-a-n-a (approved):
Looks good

crates/rspack_plugin_dev_friendly_split_chunks/src/plugin.rs
```diff
@@ -64,7 +64,65 @@ impl Plugin for DevFriendlySplitChunksPlugin {
...
+          .map(|m| {
+            let module = compilation
+              .module_graph
+              .module_by_identifier(&m.module)
+              .expect("Should have a module here");
+
+            // Some code after transpiling will increase it's size a lot.
+            let coefficient = match module.module_type() {
+              // 5.0 is a number in practice
+              rspack_core::ModuleType::Jsx => 5.0,
```
h-a-n-a, line 87: Maybe we could figure out a better way for this. I will take this as a workaround.

hardfist (approved):

PR #2617, opened 2023-04-04
chore: Update all package.json#repository fields to include the monorepo-relative directory

vjpr:
## Summary

From https://docs.npmjs.com/cli/v9/configuring-npm/package-json#repository:

> If the package.json for your package is not in the root directory (for example if it is part of a monorepo), you can specify the directory in which it lives:

---

1. It allows a more accurate direct link to the source code from www.npmjs.com/package
1. When `rspack` is installed in your project's `node_modules`, if you encounter an error with a stack trace, it makes it easy for the IDE/Terminal to open the Github source code directly by examining the closest `package.json` file.

## Related issue (if exists)

Nil.

## Types of changes

- [ ] Bug fix
- [x] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,10 @@
   },
   "homepage": "https://rspack.dev",
   "bugs": "https://github.com/web-infra-dev/rspack/issues",
-  "repository": "web-infra-dev/rspack",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/web-infra-dev/rspack"
+  },
   "devDependencies": {
     "@changesets/cli": "2.24.4",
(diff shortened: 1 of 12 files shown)


vjpr:
> If these changes should result in a version bump, you need to add a changeset.

No version bump is needed.

No tests are needed either.

hyf0:
I think this is very important. I couldn't remember how many times I met the situation you mentioned. It's minor but annoying.

Thank you!

hyf0 (approved):

Issue #2593, opened 2023-04-03
[Bug Report]: Stats module reasons compared to webpack (userRequest and other fields missing)


lalexdotcom:
### System Info

devcontainer
```
  System:
    OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (5) arm64 unknown
    Memory: 15.93 GB / 23.45 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 18.12.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.2 - /usr/local/bin/npm
  npmPackages:
    @rspack/cli: ^0.1.5 => 0.1.5 
```

### Details

Thanks to #2573 workaround, I could have the reason property on modules' compilation stats.
But some fields are missing in RSPack reasons (JsStatsModuleReason) compared to webpack (StatsModuleReason type)
(I'm showing TS types which should be clearer for reporting)

RSPack
```typescript
export interface JsStatsModuleReason {
  moduleIdentifier?: string
  moduleName?: string
  moduleId?: string
}
```
Webpack
```typescript
declare interface KnownStatsModuleReason {
	moduleIdentifier?: string;
	module?: string;
	moduleName?: string;
	resolvedModuleIdentifier?: string;
	resolvedModule?: string;
	type?: string;
	active: boolean;
	explanation?: string;
	userRequest?: string;
	loc?: string;
	moduleId?: string | number;
	resolvedModuleId?: string | number;
}
```
I'm mainly after the `userRequest` property, but others may be used...

~~By the way, the `module` field is now `moduleId`, shouldn't it be ISO with webpack for better interoperability?~~ (my bad, moduleId is there 😅)

### Reproduce link

https://github.com/lalexdotcom/rspack-bug-report rspack-module-reason-type

### Reproduce Steps

**!! Be sure to be on rspack-module-reason-type branch !!**

1. run `npm install`
2. RSPack output: run `npm run rspack-stats-output`
```
Reasons found fields [ 'moduleIdentifier', 'moduleName', 'moduleId' ]
```
3. Webpack output: run `npm run webpack-stats-output`
```
Reasons found fields [
  'moduleIdentifier',
  'module',
  'moduleName',
  'resolvedModuleIdentifier',
  'resolvedModule',
  'type',
  'active',
  'explanation',
  'userRequest',
  'loc',
[...]

PR #2671, opened 2023-04-08
feat: stats module reason type and userRequest

faga295:
## Summary
This PR includes the `type` and `user_request` fields, which can be obtained directly through `dependency`. The remaining fields, such as `loc` and `resolvedModuleIdentifier`, any of which I think can be used as a separate PR.

## Related issue (if exists)
#2593 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


packages/rspack/tests/Stats.test.ts
```diff
@@ -177,4 +177,58 @@ describe("Stats", () => {
...
 	`);
 	});
+
+	it("should have reasons", async () => {
+		const stats = await compile({
+			context: __dirname,
+			entry: "./fixtures/main5.js",
+			stats: { reasons: true }
+		});
+		expect(stats?.toJson({}, true).modules).toEqual([
```
ahabhgk, line 187:
```suggestion
		expect(stats?.toJson({ all: false, reasons: true }).modules).toMatchInlineSnapshot([
```
Prefer `toMatchInlineSnapshot` instead of `toEqual`, `identifier: "javascript/dynamic|/Users/faga/f/rspack/packages/rspack/tests/fixtures/a.js",` will cause CI failed.

faga295: thx, I use `toMatchInlineSnapshot` now, but it seems has indent problem(`jest --updateSnapshot` result), does it matter?

ahabhgk: Doesn't matter if it only update this test, I will review the snapshot.

packages/rspack/tests/Stats.test.ts
```diff
@@ -177,4 +177,64 @@ describe("Stats", () => {
...
 	`);
 	});
+
+	it("should have reasons", async () => {
+		const stats = await compile({
+			context: __dirname,
+			entry: "./fixtures/main5.js",
+			stats: { reasons: true }
+		});
+		expect(stats?.toJson({}).modules).toMatchInlineSnapshot(`
```
ahabhgk, line 187:
```suggestion
		expect(stats?.toJson({ all: false, reasons: true }).modules).toMatchInlineSnapshot(`
```

add `{ all: false, reasons: true }` can make the snapshot less noisy.

packages/rspack/tests/Stats.test.ts
```diff
@@ -177,4 +177,64 @@ describe("Stats", () => {
 		"
 	`);
 	});
+
+	it("should have reasons", async () => {
+		const stats = await compile({
+			context: __dirname,
+			entry: "./fixtures/main5.js",
+			stats: { reasons: true }
```
ahabhgk, line 185:
```suggestion

```
this line is not necessary if using `stats.toJson({ all: false, reasons: true })`

packages/rspack/tests/Stats.test.ts
```diff
@@ -177,4 +177,64 @@ describe("Stats", () => {
...
+    ],
+    "size": 55,
+    "type": "module",
+  },
+  {
+    "chunks": [
+      "main",
+    ],
+    "id": "597",
+    "identifier": "javascript/dynamic|<PROJECT_ROOT>/tests/fixtures/main5.js",
```
ahabhgk, line 224: CI failed caused by this line, we have already fixed it in the latest main branch.

packages/rspack/tests/Stats.test.ts
faga295: Seems `reasons` option is not work, and I can't find any operation for `options.reasons` in `toJson` function.

faga295:
test failed under node16, looks a bit strange.

packages/rspack/tests/Stats.test.ts
ahabhgk: > Seems `reasons` option is not work

Just found it's a bug, we may fix it in the following PR.

ahabhgk (approved):
Thanks!

PR #2677, opened 2023-04-09
fix: use codegen hash to calculate chunk hashes & fix runtime chunk hash

underfin:
## Summary

- use module codegen cache as workaround 

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


ahabhgk (approved):

underfin:
https://github.com/webpack/webpack/pull/11339 use statscase can improve test case, but we don't port the whole statscase features...

PR #2678, opened 2023-04-10
chore: upgrade to pnpm@8.2.0

zakuru:
## Summary

Chore: 

1. Upgrade to pnpm@8.2.0
2. Setup corepack to use pnpm version 8.2.0
3. Update yaml files node matrix to target node 16 and exclude node 14
4. Update yaml files to use pnpm version 8.2.0 via corepack command

## Types of changes

- [ X] Docs change / Dependency upgrade
- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist


package.json
```diff
@@ -36,7 +36,8 @@
     "*.toml": "pnpm run format:toml"
   },
   "engines": {
-    "pnpm": "7.25.0"
+    "pnpm": "8.2.0",
```
hyf0, line 39: Is there any particular reason you specify this version?

zakuru: @hyf0 As you tagged the PR, it would be a nice to have to be up to date with latest pnpm version and benefits from its latest features / improvements. 
Let me know if it causes any issue 
Thanks

hyf0: I wonder if we could specify a range of versions, such as `>=8.0.0`.

A specific version makes one pretty frustrated once you aren't using the version. You have to learn something like corepack to manipulate the version.

I'm going to wait for the decision of @hardfist. He is a very experienced developer in this aspect.

zakuru: Sure, in the mean time I updated the yaml files to target node 16 and updated the contribution doc as well
The CI passes now

hardfist: I don't think change engines.pnpm too often is a good practice,  it will require every developer to reinstall their global pnpm version, which is not necessary, and we only use very limited features of pnpm, so except pnpm upgrade bring feature| fix we really need, upgrade is not necessary.

zakuru: @hardfist thanks for the feedback 
I am not sure how often it was modified before - but in this case I completely agree apart from potential faster installation of packages thanks to https://github.com/pnpm/pnpm/pull/6281 which was released in `v7.30.4`  there is no critical reason to have this update. It is really just an nice to have as long as it does not disturb other contributors.
However in this same current PR, I also modified some `.yaml` files which were targeting node version 14 when the contribution documentation recommend node 16. Still nothing major, but good to have alignment.
If jugged non necessary / too risky I will close it no worries.
Thanks

zakuru: Hi @hyf0 
Looking at this change https://github.com/web-infra-dev/rspack/commit/99bec3fec54dbea969c157a8236938235ad1b8c2#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R39
In case contributors use different version of `pnpm`, the lock file might change unnecessary as you can see for instance in this PR on another repo https://github.com/TooTallNate/node-degenerator/pull/19 
Not sure its the best way for dev experience, what do you think ?

hyf0: I deeply agree with you we should use the latest pnpm version and benefits from its latest features / improvements.

We are considering switching to `corepack`, which would use a local package manager to override the global one while you under the `rspack` folder.

`corepack` would automatically install the correct package manager with the correct version.

That's how I think the process is going to be done:

1. introduce `corepack` in the rspack repo with the current pnpm version in another PR
2. update pnpm to 8.2.0, which is this PR does

We have a meeting today to discuss the process. I'll post the summary after the meeting.

zakuru: Noted thanks

hyf0: Sync up:

We have discussed this and going to introduce `corepack` in the Rspack repo. After this, The affection of updating the version of Pnpm would be reduced a lot.

We are also going to write a `setup` script to help contributors finish all setup work before running `pnpm install`.

I would mark this PR on hold and merge it after introducing `corepack`. Thank you in advance.

zakuru: thanks a lot for the update @hyf0

package.json
```diff
@@ -40,7 +40,8 @@
     "*.toml": "pnpm run format:toml"
   },
   "engines": {
-    "pnpm": ">=7.25.0"
+    "pnpm": "8.2.0",
```
hyf0, line 43: I'd like to suggest using `>= 8.0.0 <= 8.2.0`(The syntax might be wrong, but you get the idea).

As long as we are using pnpm with the same lockfile-version, we don't need a fixed version here.

If a developer happens to install pnpm@8.1.0 already, there's no need to forbid him from executing `pnpm install` and force him to reinstall pnpm@8.2.0.

package.json
```diff
@@ -40,7 +40,8 @@
...
+    "node" : ">=16.14"
```
hyf0, line 44: Could add some context about why we need this requirement for `node`?

I'm not sure about this, you might want to split this change into a separate PR.

scripts/meta/setup.mjs
```diff
@@ -45,3 +45,14 @@ runInContext(
 			encoding: "utf-8"
 		}).status
 );
+
+runInContext(
+	"corepack prepare pnpm@8.2.0",
```
hyf0, line 50: I thought `corepack` would install a proper version after `corepack enable`, do we need `corepack prepare` explicitly?

.github/workflows/reusable-build.yml
```diff
@@ -188,7 +188,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        node: [14, 16]
+        node: [16]
```
hyf0, line 191: Why remove 14?

package.json
zakuru: Hi @hyf0 looking into it does not seem to be possible right now please check this issue : https://github.com/nodejs/corepack/issues/95
and the feature was actually removed with this PR: https://github.com/nodejs/corepack/pull/18 please check the reasons behind explaining the potential side effects not specifying an exact version
Currently allowed syntax is:
```
String does not match the pattern of "(npm|pnpm|yarn)@\d+\.\d+\.\d+(-.+)?".
```
So i would guess having a strict version there would be the way to go.
To make it smooth for dev experience I would suggest to add a preinstall step as the following:
```
 "preinstall": "node scripts/meta/setup.mjs",
 ```
 This would automatically set the right version of the package manager on `pnpm install command`
 ```
root@codespaces-569785:/workspaces/rspack# pnpm install
npm/darwin-arm64                         |  WARN  Unsupported platform: wanted: {"cpu":["arm64"],"os":["darwin"],"libc":["any"]} (current: {"os":"linux","cpu":"x64","libc":"glibc"})
npm/darwin-x64                           |  WARN  Unsupported platform: wanted: {"cpu":["x64"],"os":["darwin"],"libc":["any"]} (current: {"os":"linux","cpu":"x64","libc":"glibc"})
npm/win32-x64-msvc                       |  WARN  Unsupported platform: wanted: {"cpu":["x64"],"os":["win32"],"libc":["any"]} (current: {"os":"linux","cpu":"x64","libc":"glibc"})
Scope: all 51 workspace projects
. pnpm:devPreinstall$ is-ci || node ./scripts/doctor.mjs
│ Yeah. Doctor here.
└─ Done in 89ms
Lockfile is up to date, resolution step is skipped
Packages: +2160
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
node_modules/@nestjs/core: Running postinstall script, done in 236ms
node_modules/@swc/core: Running postinstall script, done in 89ms
node_modules/core-js: Running postinstall script, done in 116ms
node_modules/core-js-pure: Running postinstall script, done in 136ms
node_modules/esbuild: Running postinstall script, done in 149ms
node_modules/playwright-chromium: Running install script, done in 304ms
node_modules/puppeteer: Running postinstall script, done in 225ms
node_modules/svelte-check/node_modules/svelte-preprocess: Running postinstall script, done in 14ms
node_modules/svelte-preprocess: Running postinstall script, done in 12ms
packages/rspack-plugin-minify/node_modules/esbuild: Running postinstall script, done in 62ms
node_modules/styled-components: Running postinstall script, done in 57ms
packages/rspack prepare$ pnpm precompile-schema
│ > @rspack/core@0.1.7 precompile-schema /workspaces/rspack/packages/rspack
│ > node ./scripts/precompile-schema.js
└─ Done in 722ms
. preinstall$ node scripts/meta/setup.mjs
│ ⏺️ Running `corepack enable`
│ ⏹️ Finish  `corepack enable` with 0
│ (node:21103) ExperimentalWarning: Importing JSON modules is an experimental feature. This feature could change at any time
│ (Use `node --trace-warnings ...` to show where the warning was created)
└─ Done in 134ms
. prepare$ is-ci || husky install
│ husky - Git hooks installed
└─ Done in 91ms
The integrity of 3986 files was checked. This might have caused installation to take longer.
Done in 5.8s
root@codespaces-569785:/workspaces/rspack# 
```

package.json
zakuru: > Could you add some context about why we need this requirement for `node`?
> 
> I'm not sure about this, you might want to split this change into a separate PR.

About this part its for 2 reasons:

1.  It is recommend using node 16 in the contribution doc so I though it would make sense to align the version in the CI as well
2. pnpm version 8 requires node version 16 ++

scripts/meta/setup.mjs
zakuru: > I thought `corepack` would install a proper version after `corepack enable`, do we need `corepack prepare` explicitly?

you are correct, it is not needed and now reverted

.github/workflows/reusable-build.yml
zakuru: > Why remove 14?

About this part its for 2 reasons:

It is recommend using node 16 in the contribution https://github.com/web-infra-dev/rspack/blob/main/CONTRIBUTING.md#install-nodejs  doc so I though it would make sense to align the version in the CI as well
pnpm version 8 requires node version 16 ++

hyf0: > pnpm version 8 requires node version 16 ++

I miss this. Rspack is guaranteed to be running in node14 and node16. If that's the case, I'm not sure we should update to pnpm@8.

package.json
```diff
@@ -5,6 +5,7 @@
   "description": "A Fast Rust-based web bundler",
   "private": true,
   "scripts": {
+    "preinstall": "node scripts/meta/setup.mjs",
```
hyf0, line 8: We already have `pnpm:devPreinstall` due to `preinstall` is actually running after `pnpm install.`

.github/workflows/reusable-build.yml
zakuru: Sure no worries. I was just following the docs, let me know when you have updates 
Thanks

zakuru: > > pnpm version 8 requires node version 16 ++
> 
> I miss this. Rspack is guaranteed to be running in node14 and node16. If that's the case, I'm not sure we should update to pnpm@8.

Seems that the change brings some complications
 I can close it with no concern if its the case

hyf0: We would have a meeting tomorrow to discuss this PR and related topics. You could keep the PR open until the team come up with some conclusions.

zakuru: @hyf0 sure here it is https://github.com/web-infra-dev/rspack/pull/2780

PR #2685, opened 2023-04-10
feat(cli): support mjs&cjs as default config file

hardfist:
## Summary
since we support mjs as config file now, we should support `rspack.config.mjs` as default config file
## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/packages/rspack-cli/src/utils/loadConfig.ts
+++ b/packages/rspack-cli/src/utils/loadConfig.ts
@@ -4,7 +4,7 @@ import fs from "fs";
 import { RspackCLIOptions } from "../types";
 import { RspackOptions, MultiRspackOptions } from "@rspack/core";
 
-const supportedExtensions = [".js", ".ts"];
+const supportedExtensions = [".js", ".ts", ".mjs"];
 const defaultConfig = "rspack.config";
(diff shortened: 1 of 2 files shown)


packages/rspack-cli/src/utils/loadConfig.ts
```diff
@@ -4,7 +4,7 @@ import fs from "fs";
 import { RspackCLIOptions } from "../types";
 import { RspackOptions, MultiRspackOptions } from "@rspack/core";
 
-const supportedExtensions = [".js", ".ts"];
```
IWANABETHATGUY, line 7: what about cjs?

chenjiahan (dismissed):

packages/rspack-cli/src/utils/loadConfig.ts
hardfist: unless people ask

hardfist: seems webpack supports cjs and i will add it

IWANABETHATGUY (approved):

PR #2707, opened 2023-04-11
fix(html): html inject public path incorrect

jerrykingxyz:
## Summary

The builtins html will inject the assets into the html, but injection path is wrong when public path is auto.

https://github.com/web-infra-dev/rspack/issues/2710

## Types of changes

- [ ] Bug fix
- [ ] Refactoring
- [ ] Breaking change

## Checklist

### DIFF
--- a/.changeset/dry-camels-tie.md
+++ b/.changeset/dry-camels-tie.md
@@ -1,0 +1,5 @@
+---
+"@rspack/binding": patch
+---
+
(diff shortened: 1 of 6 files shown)


ahabhgk (approved):

h-a-n-a (approved):
Looks good after digging into the implementation. Could you also provide some context in the PR?

PR #2763, opened 2023-04-15
test: update snapshot

siyou:
## Summary

copilot:summary

## Related issue (if exists)

<details open=true>

copilot:walkthrough

</details>

### DIFF
--- a/crates/rspack_error/tests/fixtures/should_not_warn_export_destructring/fixtures__should_not_warn_export_destructring.snap
+++ b/crates/rspack_error/tests/fixtures/should_not_warn_export_destructring/fixtures__should_not_warn_export_destructring.snap
@@ -1,9 +1,10 @@
 ---
 source: crates/rspack_error/tests/fixtures.rs
+assertion_line: 41
 expression: should_not_warn_export_destructring
 ---
-warning[internal]: ./a.js did not export `notExist`, imported by ./index.js
-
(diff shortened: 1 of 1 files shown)


IWANABETHATGUY:
It seems that the snapshot is incorrect, which makes the ci fail.

siyou:
The test case runs successfully on macOS. Based on the tested JavaScript file, the result on macOS should be correct.

siyou:
> It seems that the snapshot is incorrect, which makes the ci fail.

The test case runs successfully on macOS. Based on the tested JavaScript file, the result on macOS should be correct.

IWANABETHATGUY:
I am sorry we can't accept this pr if Ci is failed.

siyou:
Okay, I will investigate the reason for the difference on different platforms.

IWANABETHATGUY:
> Okay, I will investigate the reason for the difference on different platforms.

Any good news here?

siyou:
> > Okay, I will investigate the reason for the difference on different platforms.
> 
> Any good news here?

Oh my god, has it been that long? I've been so busy that I forgot. I'll take a look this weekend.

siyou:
> > Okay, I will investigate the reason for the difference on different platforms.
> 
> Any good news here?

It seems that the issue is caused by the use of HashMap/HashSet , which leads to unstable iterator ordering.

Issue #2868, opened 2023-04-22
module code generate doesn't include comments


faga295:
### System Info

 System:
    OS: macOS 13.0
    CPU: (8) arm64 Apple M1 Pro
    Memory: 56.02 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    Yarn: 1.22.15 - ~/.nvm/versions/node/v16.14.2/bin/yarn
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Safari: 16.1
  npmPackages:
    @rspack/cli: 0.1.7 => 0.1.7

### Details

webpack:
![image](https://user-images.githubusercontent.com/87003751/233772696-e0928e62-d5e4-42cb-a926-cc3583b4e017.png)
rspack:
![image](https://user-images.githubusercontent.com/87003751/233772712-f99021c5-43de-4cfb-8966-99633cb8f13e.png)
when `optimization.minimize` is false, webpack retains all comments from the original files, but rspack removes them 

### Reproduce link

https://github.com/faga295/rspack-reproduce/tree/module_generate_include_comments

### Reproduce Steps

1. pnpm i
2. pnpm build:rspack
3. pnpm build:webpack
4. check the difference between webpack-dist and rspack-dist

faga295:
Can I fix it by the way in #2882?

IWANABETHATGUY:
We prefer one pr focus on one thing.

PR #2893, opened 2023-04-24
feat: add comments in module code generation

faga295:
## Related issue (if exists)

#2868 
## Summary

copilot:summary

<details open=true>

copilot:walkthrough

</details>


IWANABETHATGUY:
Would you mind adding an option to control this behavior? Not everyone wants comments in their outputs, also this would slow the codegen phase. You could add some tests in another dir, comments are kinds of noisy in tree shaking test.

IWANABETHATGUY:
I don't think it is a good option to mix this behavior with minifyOption, because according to your implementation, this option should control development mode code generation behavior as well. Would you mind waiting about one day or so? Maybe we could discuss where should we put the option, or we don't need a option just codegenerate the comments anyway.

faga295:
Hey @IWANABETHATGUY , after discussing how this option should be designed🤔️

IWANABETHATGUY:
We don't want this feature to introduce too much overhead, so let's bench it first.
!bench

IWANABETHATGUY:
!bench

IWANABETHATGUY:
please make sure to pass the CI, you could run `UPDATE=1 cargo test -p rspack`

IWANABETHATGUY:
!bench

faga295:
> this pr also includes some expected file change, and the `tree-shaking/ts-target-es5`'s test seems has some diff problem, it's failed even though the expected file is same as actual file, and I'm very confused about this.

the `tree-shaking/ts-target-es5` is strange, after run `UPDATE=1 cargo test -p rspack`, test can pass locally, but no files changed

IWANABETHATGUY:
Do you use Windows?

faga295:
> Do you use Windows
 OS: macOS 13.0
 CPU: (8) arm64 Apple M1 Pro

image
the expected file is indeed no difference to actual file
https://github.com/web-infra-dev/rspack/blob/4e25fafa1cf76492c89bacedb3efebbc79001cf8/crates/cargo-rst/src/rst.rs#L347-L363
Is this problem due to diff?

IWANABETHATGUY:
I don't think so, 
image
image
It has some diff but will be replaced by git when you commit to GitHub.

IWANABETHATGUY:
I think you could add an option to control this behavior, like `builtins.codeGeneration.keepComments`

IWANABETHATGUY:
By default this option should be disabled, you could add extra test case to test this behavior, this will not conflict with current test case.

faga295:
I think add this option in minifyOption is better, because webpack will keep comments by default, and remove comments is more like `minify`, and [terser-webpack-plugin](https://github.com/webpack-contrib/terser-webpack-plugin#remove-comments) has similar option

crates/rspack_core/src/options/builtins.rs
```diff
@@ -53,6 +54,11 @@ pub struct Minification {
   pub pure_funcs: Vec<String>,
 }
 
+#[derive(Debug, Clone, Default)]
```
IWANABETHATGUY, line 57: Copy and Clone

crates/rspack_plugin_javascript/src/ast/stringify.rs
```diff
@@ -32,7 +36,15 @@ pub fn stringify(ast: &Ast, devtool: &Devtool) -> Result<TransformOutput> {
...
-      None,
+      if let Some(keep) = keep_comments {
+        if keep {
+          program.comments.as_ref().map(|c| c as &dyn Comments)
+        } else {
+          None
+        }
+      } else {
+        None
+      },
```
IWANABETHATGUY, line 47:
```rs
      if let Some(true) = keep_comments {
        program.comments.as_ref().map(|c| c as &dyn Comments)
      } else {
        None
      },
```

packages/rspack/src/config/builtins.ts
```diff
@@ -373,7 +375,10 @@ export function resolveBuiltinsOptions(
 		pluginImport: resolvePluginImport(builtins.pluginImport),
 		relay: builtins.relay
 			? resolveRelay(builtins.relay, contextPath)
-			: undefined
+			: undefined,
+		codeGeneration: builtins.codeGeneration
```
IWANABETHATGUY, line 379: I think you could put the undefined logic inside `resolveCodeGeneration` function.

packages/rspack/tests/Defaults.unittest.ts
```diff
@@ -73,347 +73,350 @@ describe("snapshots", () => {
...
-		    "css": {
-		      "modules": {
-		        "exportsOnly": false,
-		        "localIdentName": "[path][name][ext]__[local]",
-		        "localsConvention": "asIs",
-		      },
-		    },
-		    "decorator": {
-		      "emitMetadata": true,
-		      "legacy": true,
```
faga295, line 89: these are due to `jest --updateSnapshot`,  I can't restore the original indent

IWANABETHATGUY: I think that's alright if the CI is passed.

IWANABETHATGUY (approved):
Thanks!

PR #2983, opened 2023-04-28
feat(terser): expose terser options

dgmachado:
feat(terser): expose terser options: mangle, keep_classnames and keep_fnames

Exposing these options allow us to customize the terser properties. 

```
config.optimization = {
    minimize: true,
    minimizer: [
      new minifyPlugin({
        minifier: 'terser',
        keep_fnames: true,
      }),
    ]
};
```


hardfist (approved):

hardfist:
just FYI this is a temporarily package we will remove this package once we are compatible with `terser-webpack-plugin`

Issue #2341, opened 2023-03-18
[Bug Report]: devServer.invalidate() 调用报错


crazyxhz:
### System Info

  System:
    OS: macOS 12.6
    CPU: (8) arm64 Apple M1 Pro
    Memory: 91.02 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.13.0 - ~/.nvm/versions/node/v18.13.0/bin/node
    Yarn: 1.22.17 - /usr/local/bin/yarn
    npm: 8.19.3 - ~/.nvm/versions/node/v18.13.0/bin/npm
  Browsers:
    Chrome: 111.0.5563.64
    Safari: 16.0

### Details

devServer.invalidate() 调用报错
<img width="1431" alt="image" src="https://user-images.githubusercontent.com/6551888/226145033-12cf37ba-6d18-421b-a28b-cd120c02fee6.png">

### Reproduce Steps

1. pnpm create rspack@latest
2. `rspack.config.js` 添加配置 
```
devServer:{
    setupMiddlewares: (mid, server) => {
      server.app.get('/reload', (_, response) => {
        server.invalidate();
        response.send('reloaded');
      });
      return mid
    }
  },
```
3. pm dev 后， curl localhost:8080/reload 报错

需求，配置变更之后想调用invalid 重读配置

jerrykingxyz:
The compiler rebuild will update module graph with `modifiedFiles` and `deleteFiles`. Empty for both means that the module graph has not changed, and naturally the final assets has not changed. Maybe we should skip incremental rebuild in make phase when both `modifiedFiles` and `deleteFiles` are empty.(Of course this will be slower). cc @hardfist @hyf0 @Boshen

PR #3105, opened 2023-05-10
fix: call devServer.invalidate() panic

jerrykingxyz:
## Related issue (if exists)

Fixes: #2341 

## Summary

Rspack rebuild need both `modifiedFiles` and `deleteFiles` are exists, but will panic when triggered manually by the user.
This bug doesn't exist in webpack because it only has a `compile` method and doesn't care about `modifiedFiles`

Currently, rspack rebuild performs two more steps of `incremental rebuild` and `calc hmr chunk` than build. And there doesn't seem to be a need for `calc hmr chunk` when manually triggered by the user, so just use the build is enough.

### <samp>🤖 Generated by Copilot at 9bfba87</samp>

Improve file watching and rebuilding logic for `rspack`. Use a new variable `canRebuild` to avoid unnecessary rebuilds and enable rebuilds without `devServer`.

<details open=true>

### <samp>🤖 Generated by Copilot at 9bfba87</samp>

*  Introduce a new variable `canRebuild` to check if there are any file changes that trigger a rebuild ([link](https://github.com/web-infra-dev/rspack/pull/3105/files?diff=unified&w=0#diff-4d50b60f1736e09ee8e9cd235bba1d9bab8e99412926b998e107443fbc7bc8f4L241-R242))
*  Update the condition for calling the `rebuild` method of the compiler to use `canRebuild` instead of `isRebuild` ([link](https://github.com/web-infra-dev/rspack/pull/3105/files?diff=unified&w=0#diff-4d50b60f1736e09ee8e9cd235bba1d9bab8e99412926b998e107443fbc7bc8f4L249-R250))

</details>

### DIFF
--- a/packages/rspack/src/watching.ts
+++ b/packages/rspack/src/watching.ts
@@ -238,7 +238,10 @@ class Watching {
 		this.compiler.hooks.watchRun.callAsync(this.compiler, err => {
 			if (err) return this._done(err, null);
 
-			const isRebuild = this.compiler.options.devServer && !this.#initial;
+			const canRebuild =
+				this.compiler.options.devServer &&
+				!this.#initial &&
+				(modifiedFiles?.size || deleteFiles?.size);
 
 			const onBuild = (err?: Error) => {
 				if (err) return this._done(err, null);
@@ -246,7 +249,7 @@ class Watching {
 				this._done(null, this.compiler.compilation);
 			};
 
-			if (isRebuild) {
+			if (canRebuild) {
 				this.compiler.rebuild(modifiedFiles, deleteFiles, onBuild as any);
 			} else {
 				this.compiler.build(onBuild);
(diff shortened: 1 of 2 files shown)


packages/rspack/src/watching.ts
```diff
@@ -238,15 +238,18 @@ class Watching {
 		this.compiler.hooks.watchRun.callAsync(this.compiler, err => {
 			if (err) return this._done(err, null);
 
-			const isRebuild = this.compiler.options.devServer && !this.#initial;
+			const canRebuild =
+				this.compiler.options.devServer &&
```
hyf0, line 242: What happens when people use Rspack via nodejs instead of `@rspack/cli`?

jerrykingxyz: No `options.devserver` won't run rebuild, I'll fix that in another PR.

hyf0 (approved):

Issue #1909, opened 2023-02-20
perf: matching module rule using regex is slow


Boshen:
### Details

![image](https://user-images.githubusercontent.com/1430279/220087242-3ee63bd1-cac6-4e53-8aa2-0c185c1d6d03.png)

With lots of modules and lots of rules, this quickly becomes a performance bottleneck.

PR #3113, opened 2023-05-11
perf: 🤖 optimize regex

IWANABETHATGUY:
## Related issue (if exists)
Closed https://github.com/web-infra-dev/rspack/issues/1909

## Summary
Using `examples/arco-pro` for  ad-hoc benchmark
construct 14 regex, for each regex test all module identifiers.

<details>
<summary>bench code</summary>

```rs
use std::time::Instant;
use aho_corasick::{AhoCorasick, PatternID};
use regex_syntax::ast::Concat;
use regex_syntax::hir::literal::ExtractKind;
use regex_syntax::hir::{Hir, HirKind, Look};
use rspack_error::Result;

fn main() {
  let string = include_str!("../result.log");
  let string_list = string.split("\n").collect::<Vec<_>>();
  let count = 6;
  let mut reg_list = vec![
    "\\.json$",
    "\\.mjs$",
    "\\.js$",
    "\\.cjs$",
    "\\.js$",
    "\\.jsx$",
    "\\.ts$",
    "\\.tsx$",
    "\\.module\\.css$",
    "\\.css$",
    "\\.less$",
    "\\.module\\.less$",
    "\\.svg$",
    "\\.png$",
  ];

  let test_str = "";
  let total_count = count * string_list.len();
  let start = Instant::now();
  let re_list = reg_list
    .iter()
    .map(|item| Regex2::new(item).unwrap())
    .collect::<Vec<_>>();
  for re in re_list.iter() {
    for i in 0..count {
      for test_str in string_list.iter() {
        re.test(test_str);
      }
    }
  }
  println!(
    "{}  {} module:\n{:?}",
    "custom search",
    total_count,
    start.elapsed()
  );
  let start = Instant::now();
  let re_list = reg_list
    .iter()
    .map(|item| Regex1::new(item).unwrap())
    .collect::<Vec<_>>();
  for re in re_list.iter() {
    for i in 0..count {
      for test_str in string_list.iter() {
        re.test(test_str);
      }
    }
  }
  println!(
    "{}  {} module:\n{:?}",
    "regex",
    total_count,
    start.elapsed()
  );
  let start = Instant::now();

  let re_list = reg_list
    .iter()
    .map(|item| regress::Regex::new(item).unwrap())
    .collect::<Vec<_>>();
  for re in re_list.iter() {
    for i in 0..count {
      for test_str in string_list.iter() {
        _ = re.find(test_str).is_some();
      }
    }
  }

  println!(
    "{}  {} module:\n{:?}",
    "regress",
    total_count,
    start.elapsed()
  );
}

enum Regex1 {
  FastReg(regex::Regex),
  SlowReg(regress::Regex),
}

impl Regex1 {
  pub fn new(str: &str) -> Result<Regex1> {
    match regex::Regex::new(str) {
      Ok(reg) => Ok(Regex1::FastReg(reg)),
      Err(_) => Ok(regress::Regex::new(str).map(Regex1::SlowReg).unwrap()),
    }
  }

  pub fn test(&self, str: &str) -> bool {
    match self {
      Regex1::FastReg(reg) => reg.is_match(str),
      Regex1::SlowReg(reg) => reg.find(str).is_some(),
    }
  }
}

enum Regex2 {
  FastCustom(Box<dyn Fn(&str) -> bool>),
  SlowReg(regress::Regex),
}

impl Regex2 {
  pub fn new(str: &str) -> Result<Regex2> {
    match regex_syntax::parse(str) {
      Ok(ast) => {
        let hir = regex_syntax::parse(str).unwrap();
        let res = regex_syntax::hir::literal::Extractor::new()
          .kind(ExtractKind::Suffix)
          .extract(&hir);
        if res.is_exact() {
          let string_list = res
            .literals()
            .unwrap()
            .iter()
            .map(|item| {
              let s = std::str::from_utf8(item.as_bytes()).unwrap();
              s.to_string()
            })
            .collect::<Vec<_>>();

          Ok(Regex2::FastCustom(Box::new(move |str| {
            string_list.iter().any(|item| str.ends_with(item))
          })))
        } else {
          Ok(regress::Regex::new(str).map(Regex2::SlowReg).unwrap())
        }
      }
      Err(_) => Ok(regress::Regex::new(str).map(Regex2::SlowReg).unwrap()),
    }
  }

  pub fn test(&self, str: &str) -> bool {
    match self {
      Regex2::FastCustom(fast) => fast(str),
      Regex2::SlowReg(reg) => reg.find(str).is_some(),
    }
  }
}
```

</details>

### result
**4251 * 14 regex** for small project
```bash
custom search  4251 module:
361.259µs
regex  4251 module:
1.123925ms
regress  4251 module:
3.834253ms
```
**8502 module  * 14 regex** for medium project
```bash
custom search  8502 module:
640.935µs
regex  8502 module:
2.060185ms
regress  8502 module:
7.55814ms
```

**21255 * 14 regex** for large project
```bash
custom search  21255 module:
1.557561ms
regex  21255 module:
5.104133ms
regress  21255 module:
19.617296ms

```
### Conclusion
1. **custom search** is 10~12 times faster than `regress`  in file extension match.
2. **regex** is about 3~4 times faster than `regress` in file extension match


crates/rspack_regex/Cargo.toml
```diff
@@ -8,6 +8,7 @@ version    = "0.1.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+regex-syntax = "0.7.1"
```
Boshen, line 11: You may want to turn off the `unicode` feature

crates/rspack_regex/Cargo.toml
```diff
@@ -20,4 +21,5 @@ swc_core = { workspace = true, features = [
   "__parser",
   "__utils",
   "ecma_preset_env",
+  "base",
```
Boshen, line 24: can you import the crate instead of base? I think I removed based everywhere ...

Boshen (commented):
I'm unsure about adding another regex parser for this, do we wanna try caching as well?

Boshen:
Maybe the regex algorithm is badly implemented for trailing `$`? Have you looked into this as well?

IWANABETHATGUY:
> I'm unsure about adding another regex parser for this, do we wanna try caching as well?

I don't think caching is helpful according to your profile, the slow part is `match` not the constructor.

IWANABETHATGUY:
> Maybe the regex algorithm is badly implemented for trailing `$`? Have you looked into this as well?

No, digging into `regex` crate under the hood may need too much time, I don't have too much time to investigate.

crates/rspack_regex/Cargo.toml
IWANABETHATGUY: Done

crates/rspack_regex/Cargo.toml
IWANABETHATGUY: Done

crates/rspack_regex/src/algo.rs
```diff
@@ -0,0 +1,99 @@
...
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    match self {
+      Self::FastCustom(_) => write!(f, "FastCustom(...)"),
+      Self::Regress(arg0) => f.debug_tuple("Regress").field(arg0).finish(),
+    }
+  }
+}
+
+impl Algo {
+  pub fn new(expr: &str, flags: &str) -> Result<Algo, Error> {
```
Boshen, line 23: It would be great if we provide some documentation for why we are doing this.

crates/rspack_regex/src/algo.rs
```diff
@@ -0,0 +1,99 @@
...
+          internal_error!("Can't construct regex `/{expr}/{flags}`, original error message: {err}")
+        })
+    }
+  }
+
+  pub fn try_fast_path(expr: &str) -> Option<Algo> {
+    let hir = regex_syntax::parse(expr).ok()?;
+    let seq = regex_syntax::hir::literal::Extractor::new()
+      .kind(ExtractKind::Suffix)
+      .extract(&hir);
```
Boshen, line 39: regress is also a parser right? It can't be used to detect this? 

Also, it's this just matching for an ending `$` inside the string?

crates/rspack_regex/src/algo.rs
```diff
@@ -0,0 +1,99 @@
...
+    matches!(self, Self::FastCustom(..))
+  }
+
+  #[cfg(test)]
+  fn is_regress(&self) -> bool {
+    matches!(self, Self::Regress(..))
+  }
+}
+
+fn is_ends_with_regex(hir: &Hir) -> bool {
```
Boshen, line 73: Can you briefly document what this does?

Boshen (approved):
Go ahead if the perf numbers are good

crates/rspack_regex/src/algo.rs
IWANABETHATGUY: NO, `regress` is a regex engine, it can not provide `ast` or `hir` for regex

crates/rspack_regex/src/algo.rs
IWANABETHATGUY: if it is a `regex` something like ends with `$`

crates/rspack_regex/src/algo.rs
IWANABETHATGUY: `regex-syntax` parsed the original regular expression, and convert the pattern `****$` to a fast string search routine, this is what `regress` can't do.

crates/rspack_regex/src/algo.rs
IWANABETHATGUY: Algo is the abbreviation of `Algorithm` , this layer control how did `RspackRegex` match a string literal, fast  string search, or fallback to a normal regress regex matching.

hyf0:
I must say the idea is great!

crates/rspack_regex/src/algo.rs
edvardchen: It matches any regex ending with `$` but not starting with `^` so that it can safely to use `str.ends_with` matching  rather than normal regex matching. 

If the regex starts with `^`, it should use full text matching instead. Am i right? And it seems trivial to implement.

IWANABETHATGUY: Yeah, but this optimization is now only used in `module.rule.test` matching. 
Also, people could use a string literal as `test` expr when they want a full-text matching. https://www.rspack.dev/config/module#condition

Issue #2237, opened 2023-03-14
[Feature Request]: prebundle rspack config file by rspack itself for better config extensions support


fi3ework:
### What problem does this feature solve?

Support multiple config file extensions, such as `mjs` `ts` `cjs` `mts` `cts`, especially `ts` for better intellisense and type check.

### What does the proposed API of configuration look like?

Utilize rspack itself to pre-bundle the config file. The bundling should have the following features:

1. A default unmodifiable sane nodejs lib config
2. Externalize dependencies in node_modules
3. Support both ESM and CJS output
4. Hijacking `require.extensions` to load CJS and use temp file (before `--loader` landed) for ESM.

wernerbihl:
I'm particularly interested in support for .cjs config because all my projects have "type": "module" in package.json. But when changing the sample provided "rspack.config.js" (with no requires or imports) to "rspack.config.cjs", then I get the following:

`TypeError: Cannot read properties of undefined (reading 'import')
    at getNormalizedEntryStatic (/var/www/rspack-project/node_modules/@rspack/core/src/config/normalization.ts:202:33)
    at getNormalizedRspackOptions (/var/www/rspack-project/node_modules/@rspack/core/src/config/normalization.ts:32:7)
    at createCompiler (/var/www/rspack-project/node_modules/@rspack/core/src/rspack.ts:55:44)
    at create (/var/www/rspack-project/node_modules/@rspack/core/src/rspack.ts:141:20)
    at rspack (/var/www/rspack-project/node_modules/@rspack/core/src/rspack.ts:165:31)
    at RspackCLI.createCompiler (/var/www/rspack-project/node_modules/@rspack/cli/src/rspack-cli.ts:49:26)
    at Object.handler (/var/www/rspack-project/node_modules/@rspack/cli/src/commands/build.ts:94:22)
`

hardfist:
@wernerbihl .cjs is supported

hardfist:
@fi3ework  prebundle may meet lots of issue like https://github.com/web-infra-dev/modern.js/issues/3338,especialy when used in monorepo, if we bundle code in other package it will cause the bundled config can't find the its dependencies, so we think transformer is less prone

fi3ework:
Prebundling could externalize bare imports, which would skip libraries in a monorepo, but still will meet ﻿`type: module/commonjs` issues.

hardfist:
[...]

PR #3222, opened 2023-05-18
refactor(cli): support for `.cts`、`.mts` and follows `type: module`

molvqingtai:
## Related issue

Enhanced configuration files, added support for `.cts` and `.mts`, and fixed the problem of not following the user `type: module`.

## summary

`cli` will automatically register `loader` for `.ts | .cts` files, for `.ts` files marked with `type: nodule` or `.mts` will be considered as ESM and need to use `cli` to specify the loader.

<details open=true>

rspack.config.ts:
```ts
import path from 'node:path'
import { fileURLToPath } from 'node:url'

const __dirname = path.dirname(fileURLToPath(import.meta.url))

export default {
  mode: 'production',
  entry: path.resolve(__dirname, 'src/index.ts'),
  output: {
    path: path.resolve(__dirname, 'dist')
  }
}

```
package.json
```json
{
  "type": "module",
  "scripts": {
    "build": "node --loader ts-node/esm rspack build -c rspack.config.ts",
  },
  "devDependencies": {
    "@types/node": "^20.3.1",
    "ts-node": "^10.9.1",
    "typescript": "^5.1.3"
  }
}

```
tsconfig.json
```json
{
  "compilerOptions": {
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "allowImportingTsExtensions": true,
    "noEmit": true
  }
}
```
It should work. 🎉

</details>


hardfist:
we discuss how to support .ts in https://github.com/web-infra-dev/rspack/issues/2237 and decide to use transform strategy other than bundle strategy, so use esbuild to prebundle config is not ideal

molvqingtai:
> we discussed how to support .ts in #2237 and decide to use transform strategy other than bundle strategy, so use esbuild to prebundle config is not ideal

I tried to use rspack self-compilation in the code, but the output file seems to have a problem and cannot be imported or required correctly.

https://github.com/molvqingtai/rspack/blob/main/packages/rspack-cli/src/utils/loadConfig.ts#L32

hardfist:
> I tried to use rspack self-compilation in the code, but the output file seems to have a problem and cannot be imported or required correctly.
> 
use rspack to bundle config is not ideal which is disscussed in #2237

hardfist:
it's a BUG that rspack not supports "type" property of packageJson, but it should **not** be fixed by bundle config file

hardfist:
i think this should break into two parts type field support and mts|cts support，and we dont plan to support mts|cts

molvqingtai:
> we discussed how to support .ts in #2237 and decide to use transform strategy other than bundle strategy, so use esbuild to prebundle config is not ideal

Hey, I have refactored the code and updated the documentation. I used [jiti](https://github.com/unjs/jiti) without bundling packages, and the dependency issue mentioned in #2237 has been resolved.

hardfist:
> I tested jiti because it only runs the config file, so it's very fast, like node VM, unlike ts-node which pollutes the global scope.

sorry it seems webpack doesn't support `mts | cts | ts with type:module` either, and this implementation is too complicated and introduces tons of dependency(jiti uses babel) and edge case.
if you need to use `mts | cts | ts with type:module` i think you could use jiti on userland

alexander-akait:
@hardfist 

> @alexander-akait does webpack have any plan to support mts | cts | ts with type:module config?

I thought about it, but I still think that it should be on the user side, there are a lot of different tools and approaches, but I opened to discussion

hardfist:
https://github.com/webpack/webpack-cli/pull/3831 since webpack has supported mts|cts now, we may support it as well, @molvqingtai are you still willing to work on this?

alexander-akait:
@hardfist note - we don't use `jiti`, we allow to use any tools i.e. `ts-node`, `babel` and etc

molvqingtai:
> @hardfist note - we don't use `jiti`, we allow to use any tools i.e. `ts-node`, `babel` and etc

Let the user specify node loader like webpack [L11](https://github.com/webpack/webpack-cli/blob/bfddce332f91214b55cde5389c7920a2b7f6f576/test/build/config-format/typescript-esnext-mjs/typescript.test.js#L11), I don't think this is a good idea.

Use babel or other transform? I don't have a good idea at present.

alexander-akait:
@molvqingtai 

> Let the user specify node loader like webpack [L11](https://github.com/webpack/webpack-cli/blob/bfddce332f91214b55cde5389c7920a2b7f6f576/test/build/config-format/typescript-esnext-mjs/typescript.test.js#L11), I don't think this is a good idea.

Why? You can't handle all possible tools and their output

molvqingtai:
> @molvqingtai
> 
> > Let the user specify node loader like webpack [L11](https://github.com/webpack/webpack-cli/blob/bfddce332f91214b55cde5389c7920a2b7f6f576/test/build/config-format/typescript-esnext-mjs/typescript.test.js#L11), I don't think this is a good idea.
> 
> Why? You can't handle all possible tools and their output

ESM/CJS compatibility issues should not be thrown to the user. As you can imagine, when the user does not specify a loader and causes an error, it will be a bad experience to search the document for a solution.

molvqingtai:
> > @molvqingtai
> > > Let the user specify node loader like webpack [L11](https://github.com/webpack/webpack-cli/blob/bfddce332f91214b55cde5389c7920a2b7f6f576/test/build/config-format/typescript-esnext-mjs/typescript.test.js#L11), I don't think this is a good idea.
> > 
> > 
> > Why? You can't handle all possible tools and their output
> 
> ESM/CJS compatibility issues should not be thrown to the user. As you can imagine, when the user does not specify a loader and causes an error, it will be a bad experience to search the document for a solution.



> 
Using `jiit` is currently the simplest solution while maintaining a good DX. As for edge cases, both nuxt (https://github.com/nuxt/vite/pull/201) and tailwindcss(https://github.com/tailwindlabs/tailwindcss/pull/10785) are `jiti` to load configuration files. I believe this solution is stable enough.

alexander-akait:
@molvqingtai I strongly disagree, but I will not argue, we already tried to do it and faced with a lot of problems when new things appear (syntax/different outputs/just features like swc for `ts-node`/ES modules), why I should use babel (and dependecies) when I want to use swc. Setting up the runtime environment is just the task of the developer and no one else, you can use `jiti` under the hood if **nothing** was configured (or no options/setting provided), but always using and not allowing customization is a pretty bad idea, you can of course not take someone else's experience and go through the same problems again, your choice

alexander-akait:
And yes, don't forget about performance, babel is not fast

molvqingtai:
> And yes, don't forget about performance, babel is not fast


Of course, you can use swc. https://github.com/tailwindlabs/tailwindcss/pull/10785/commits/f1614c8d9ef50d507e271a30de6213cbaa932f0f

hardfist:
> > @hardfist note - we don't use `jiti`, we allow to use any tools i.e. `ts-node`, `babel` and etc
> 
> Let the user specify node loader like webpack [L11](https://github.com/webpack/webpack-cli/blob/bfddce332f91214b55cde5389c7920a2b7f6f576/test/build/config-format/typescript-esnext-mjs/typescript.test.js#L11), I don't think this is a good idea.
> 
> Use babel or other transform? I don't have a good idea at present.

since esm is very tricky and contains tons of edge case，i dont think let advanced user to handle esm-loader itself is a bad idea，@xiaoxiangmoe mentions jiti has some tricky design and may contain some edge case

hardfist:
i do agree jiti is simple and can handle 80 percent cases，but jiti has  many options which we cant expose to users and i do notice it has some lockin design which will break when javascript has new syntax or nodes invents new esm only api

molvqingtai:
> i do agree jiti is simple and can handle 80 percent cases，but jiti has many options which we cant expose to users and i do notice it has some lockin design which will break when javascript has new syntax or nodejs invents new esm only api

I agree with you. Can we retain the ability to customize loader and use jiit to load when no loader is specified? This ensures that the ESM configuration file can be used out of the box, and when jiti does not follow the new syntax, or some other edge cases, users can customize the loader processing.

xiaoxiangmoe:
I don't think using jiti is a good idea because it has some edge case. For example:

### TypeScript commonjs syntax

```ts
// file: rspack.config.ts
import * as path from "path";

const config = {
	mode: "production",
	entry: "./main.ts",
	output: {
		path: path.resolve(__dirname, "dist"),
		filename: "foo.bundle.js"
	}
};

export = config;// this stand for module.exports = config in commonjs
```

This is the usage that rspack provided now. But jiti can't handle it correctly. This will break users.

### Abusing cjs and esm

import.meta.url exist in cts
```ts
console.log(import.meta.url) // file:///Users/me/ts-node-playground/a.cts
```
__filename exist in mts

```ts
console.log(__filename) // /Users/me/ts-node-playground/a.mts
```

If we provide these feature. Should we also provide this in `.js`, `.cjs` and `.mjs`?

hardfist:
> To summarize the issues listed, in order to be consistent with webpack (important), if there is no better suggestion, I will delete jiit and adopt webpack's solution.

thank you for your understanding, the current wepback's way is a right(contains less edge case) but not convenient way and we may iterate to improve the DX of this in the future

hardfist:
Sometimes`` it seems to me that the most ideal for configurations was pure JS with jsdocs comments for types - no extra loaders, no extra transpilation, no extra overhead.

i prefer this way too，and that is why all of our examples written in this way

molvqingtai:
@hardfist @alexander-akait 
The refactoring is complete and the description has been updated

packages/rspack-cli/package.json
```diff
@@ -26,27 +26,23 @@
...
-  },
   "dependencies": {
     "@discoveryjs/json-ext": "^0.5.7",
     "@rspack/core": "workspace:*",
     "@rspack/dev-server": "workspace:*",
     "colorette": "2.0.19",
+    "interpret": "^3.1.1",
+    "rechoir": "^0.8.0",
     "semver": "6.3.0",
+    "type-fest": "^2.14.0",
```
hardfist, line 45: this dependency is unnecessary we can write our own type

packages/rspack-cli/src/utils/crossImport.ts
```diff
@@ -0,0 +1,17 @@
+import { pathToFileURL } from "url";
+import isEsmFile from "./isEsmFile";
+
+const crossImport = async <T = any>(
+	path: string,
+	cwd = process.cwd()
+): Promise<T> => {
+	if (isEsmFile(path, cwd)) {
+		const url = pathToFileURL(path).href;
+		const { default: config } = await import(url);
```
hardfist, line 10: this will crash node in jest test due to https://github.com/nodejs/node/issues/35889, and I'm not sure how webpack-cli workaround it, @alexander-akait can help me here?

hardfist:
@molvqingtai thank you for your hard work and detailed test, the overall work LGTM

packages/rspack-cli/package.json
molvqingtai: > this dependency is unnecessary we can write our own type

Looking at the type-fest source code, packageJson actually has a large number of type definitions. Is it necessary to copy the complete type definition, or simply write it as {"type": "module"}

packages/rspack-cli/src/utils/crossImport.ts
alexander-akait: We spawn a new process for the each test in CLI integration tests, so we don't have `jest` in scope, to collect coverage we use `nyc` (yeah, it is deprecated, but works).

hardfist: i dont get it why nyc can workaround this problem here since it still runs jest  and it seems it doesnt enable vm-modules，how can jest run dynamic import without vm-moduels enabled？

alexander-akait: `nyc` is not a workaround, the main workaround - we spawn a new process inside `it`, i.e. (pseudo code):
```
it("serve command", () => { 
  spawnNewProcess(__dirname, ["serve"]); 
});
```

So we don't have `jest` inside a new spawed process, it also allows you to test as if the user has run it - no polluting, just a fresh process.

`nyc` is a workaround to get coverage after testing, because we don't have `jest` inside a new spawned process we can't collect lines, so here we use `nyc` inside a new spawned process and collect coverage (but we collect it only for CI)

packages/rspack-cli/tests/build/config/moonrepo/rspack.config.ts
```diff
@@ -1,6 +1,6 @@
-// @ts-ignore
+// @ts-ignore: Because dynamically create {"type": "module"}
```
hardfist, line 1: you can commit node_modules to git for test

packages/rspack-cli/package.json
hardfist: write type: module is enough for our case

hardfist:
@molvqingtai  I'll keep fixing the left issue to get this pr merged ASAP if you don't mind

hardfist:
closes #3609

molvqingtai:
> @molvqingtai I'll keep fixing the left issue to get this pr merged ASAP if you don't mind

Of course

hardfist:
!canary

hyf0 (approved):

hardfist:
@molvqingtai thank you for your great work

PR #3264, opened 2023-05-23
chore: fix debug shell

hardfist:
## Summary

copilot:summary

<details open=true>

copilot:walkthrough

</details>


Boshen:
You don't need the input, just use         

`if: ${{ github.event_name == 'workflow_dispatch'  }}` is fine?

hardfist:
> You don't need the input, just use
> 
> `if: ${{ github.event_name == 'workflow_dispatch' }}` is fine?

debug in shell may have other side effects, so it would be better to trigger by option

Boshen (changes requested):
The input doesn't get implictly passed into reusable workflow,

you can add the action-tmate step in test.yml before calling the reusable build

Boshen (approved):

Issue #3254, opened 2023-05-22
Dynamic imports make rspack crash when cache is enabled in production mode


janmeier:
### System Info

   System:
    OS: Linux 5.19 Ubuntu 22.04.2 LTS 22.04.2 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
    Memory: 10.86 GB / 31.08 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v18.12.1/bin/yarn
    npm: 8.19.2 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Browsers:
    Chrome: 113.0.5672.126
    Firefox: 113.0.1
  npmPackages:
    @rspack/cli: nightly => 0.0.0-canary-1c6a3dd-20230521164529 

### Details

* Set mode to production
* Enable cache
* Have a dynamic import in your code, such as 
```ts
await import(
    /* webpackChunkName: "iso-countries" */
    `i18n-iso-countries/langs/${language}.json`
  )
```

### Reproduce link

https://github.com/janmeier/webpack-dynamic-import

### Reproduce Steps

1. run `pnpm run build`
2. Crashes with `Is a directory (os error 21)`

PR #3271, opened 2023-05-23
fix(context_module_cache): fix cache directory bug

suxin2017:
## Related issue (if exists)

https://github.com/web-infra-dev/rspack/issues/3254
### <samp>🤖 Generated by Copilot at 538a002</samp>

This pull request improves the resource path handling and module resolution for the `rspack` tool on Windows and Node.js v16. It fixes a bug in the `rspack_plugin_devtool` crate, adds a validation step in the `rspack_core` crate, and removes an unnecessary flag from the `rspack` package.

<details open=true>

### <samp>🤖 Generated by Copilot at 538a002</samp>

* Ensure resource paths are files before resolving modules ([link](https://github.com/web-infra-dev/rspack/pull/3271/files?diff=unified&w=0#diff-70e6b130ff83b83409cc2eccd938b2ccfd9077941bda12af4ae116398dff0ef4L71-R73))
* Replace backslashes with forward slashes in resource paths on Windows for module filename template and source map generation ([link](https://github.com/web-infra-dev/rspack/pull/3271/files?diff=unified&w=0#diff-a3ed34d20f0b15a073bdeaac181d27e0fe74d19b63f7a0eb6cc1868192ef6dd9L115-R122), [link](https://github.com/web-infra-dev/rspack/pull/3271/files?diff=unified&w=0#diff-a3ed34d20f0b15a073bdeaac181d27e0fe74d19b63f7a0eb6cc1868192ef6dd9L239-R251))
* Remove unnecessary flag from test script in `package.json` for Node.js v16.0.0 compatibility ([link](https://github.com/web-infra-dev/rspack/pull/3271/files?diff=unified&w=0#diff-142a5c1a681bf6859bd9e53bbfe87fc9dd2d05e8350b0fb50bfcd652a9e36cbbL13-R13))

</details>


crates/rspack_core/src/cache/occasion/resolve_module.rs
```diff
@@ -68,7 +68,9 @@ impl ResolveModuleOccasion {
     let data = generator(args).await?;
     let mut paths = Vec::new();
     if let ResolveResult::Resource(resource) = &data {
-      paths.push(resource.path.as_path());
+      if resource.path.is_file() {
+        paths.push(resource.path.as_path());
```
underfin, line 72: I don't think it is a correct fix, because the error was caused by missed calculating the hash for directory at snapshot. 
The ` fn create_snapshot<F>(&self, paths: &[&Path], f: F)` should be ` fn create_snapshot<F>(&self, paths: &[&Path], dirs: &[&Dir] f: F)`, can you fix the function and update dirs hash logic?

suxin2017: I use the path in the directory for hash computation, do you think this is OK?

jerrykingxyz: Seems like just setting the hash to the `"directory"` string like webpack is enough
https://github.com/webpack/webpack/blob/5b1fd3ebbaf0fd5b7706c29f0be12b7b0150e743/lib/FileSystemInfo.js#L2907

suxin2017: Thank you. I have modified the code according to this

underfin (approved):

jerrykingxyz:
is there anything problem about use "directory" as hash?

suxin2017:
> is there anything problem about use "directory" as hash?

There is no problem, both ways are OK, but you two have different opinions, I don't know how to modify

jerrykingxyz:
I think computing the hash with subfiles may takes more time cost. cc @underfin

jerrykingxyz (approved):

Issue #3306, opened 2023-05-25
bug #2988 persists for more complicated packages.


aliveornot:
### System Info

same as #2988 

### Details

in #2988, I reported a bug about importing a very simple package that does not depend on other packages. And for such a simple package, it has been resolved. But same bug persists for a more complicated package that has dependencies.

I changed original "simple-demo-package" to "inline-style-prefixer" and the same bug appears. Here you can see a screenshot with single reverse-slashes.

![image](https://github.com/web-infra-dev/rspack/assets/60848153/c6b2922c-844c-4f5b-8941-dfb8b00099d4)

### Reproduce link

https://github.com/aliveornot/rspack-module-id-bug

### Reproduce Steps

same as #2988. the repo is updated to import and export inline-style-prefixer instead of simple-demo-package.

n3rds-inc:
stuck with the same problem. any progress?

aliveornot:
Of course, here are the simplest repos that I can think of:
https://github.com/aliveornot/rspack-bug-d-drive
https://github.com/aliveornot/rspack-bug-c-drive

The first one(d-drive) is the main repo, which imports a package from the other c-drive repo.  Just clone them into  d:\\src and c:\\src respectively.

Your example seems pretty much like the original issue #2988. The problem happens when importing a "package that imports other packages" from cross-drive. And I used `inline-style-prefixer` as an example of that in c-drive repo.

- I used the latest ver of rspack
![image](https://github.com/web-infra-dev/rspack/assets/60848153/3f925920-a456-4639-9df0-d997ea7b4f60)

- screenshot from the result bundle. (You can check the single slash in the second highlight, while the first highlight shows double backslashes.)
![image](https://github.com/web-infra-dev/rspack/assets/60848153/b9a9660e-4865-4660-b266-4945957a37da)

aliveornot:
Just checked v0.2.5 out, and it works! Thank you so much!

PR #3354, opened 2023-05-30
fix(window_path): serialized window path

suxin2017:
## Related issue (if exists)

Fixes https://github.com/web-infra-dev/rspack/issues/3306
### <samp>🤖 Generated by Copilot at 24bab3e</samp>

Escape module ids in require calls with serde_json. This fixes a bug in the `rspack_plugin_javascript` crate where string literals or template literals with special characters or quotes could cause syntax errors or unexpected behavior. The change affects the `require` function in `crates/rspack_plugin_javascript/src/dependency/commonjs/require.rs`.

<details open=true>

### <samp>🤖 Generated by Copilot at 24bab3e</samp>

*  Escape the module id as a JSON string when replacing a string or template literal argument of a require call ([link](https://github.com/web-infra-dev/rspack/pull/3354/files?diff=unified&w=0#diff-e30cf4d1f2a5adaf4b5efbc50db2897a0fb69775de346910bd3dddd218a0eedaL94-R101))

</details>

### DIFF
--- a/crates/rspack_plugin_wasm/tests/fixtures/imports-multiple/expected/main.js
+++ b/crates/rspack_plugin_wasm/tests/fixtures/imports-multiple/expected/main.js
@@ -9,7 +9,7 @@
 
 },function(__webpack_require__) {
 var __webpack_exec__ = function(moduleId) { return __webpack_require__(__webpack_require__.s = moduleId) }
-var __webpack_exports__ = (__webpack_exec__('./index.js'));
+var __webpack_exports__ = (__webpack_exec__("./index.js"));
 
 }
 ]);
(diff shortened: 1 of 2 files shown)


crates/rspack_testing/src/test_config.rs
```diff
@@ -584,6 +584,8 @@ impl TestConfig {
     );
     if self.optimization.module_ids == "named" {
       plugins.push(rspack_ids::NamedModuleIdsPlugin::default().boxed());
+    } else if self.optimization.module_ids == "window" {
```
hyf0, line 587: This is not ideal. It's better to put the os-specific logic in the same `named` Plugin. It's probably not a good idea to have a special value for Windows.

PR #3369, opened 2023-05-31
feat: fix: use custom tokio runtime to customize number threads 

hyf0:
## Summary

copilot:summary

<details open=true>

copilot:walkthrough

</details>


crates/node_binding/src/tokio_runtime.rs
```diff
@@ -0,0 +1,15 @@
+use once_cell::sync::Lazy;
+
+/// TODO(hyf0): we might need a way to cleanup this resource
```
Boshen, line 3: Nit: remove the comment if you are unsure about the problem, it's not usable for readers.

Boshen (commented):
Can you document or comment somewhere on how to enable single threading? 

If we want true single threading, rayon needs to be changed as well.

Boshen:
Interesting, this is breaking Windows?

hyf0:
> Interesting, this is breaking Windows?

Yes. I need to investigate it.

hardfist:
@hyf0 is this still needed?

hyf0:
Nope

PR #3403, opened 2023-06-02
[WIP] feat: assetPath hook

dmitry-stepanenko:


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1394,8 +1394,19 @@ impl Compilation {
...
 
-  pub fn get_asset_path(&self, filename: &Filename, data: PathData) -> String {
-    filename.render(data, None)
+  pub fn get_asset_path(&self, filename: &Filename, data: PathData<'_>) -> String {
+    let rendered_filename = filename.render(data, None);
+    return tokio::task::block_in_place(move || {
+      tokio::runtime::Handle::current().block_on(async move {
+        self
+          .plugin_driver
+          .write()
```
h-a-n-a, line 1403: This will be stuck if you would like to call this method from a JS Hook. As `write` was called as Rust called the JS hook, and if you call this method, then, Rust calls the `write` of `plugin_driver` again, which leads to a deadlock. The only way to solve this is to call the `plugin_driver` as `read`.

h-a-n-a: > Are there any other ways to overcome this issue? As I can see we can't really invoke the hook from elsewhere or at another point of time, right?

If you want to acquire the mutable reference as what `write` does, you can only invoke the hook one at the time. But for hooks that only requires `read`, it does not have the limitation, which means you can invoke as much as you wish.

dmitry-stepanenko: @h-a-n-a can you please point me what exactly should be changed in my implementation? I tried both read and write accesses and behavior is the same

h-a-n-a: The deadlock problem has been resolved in https://github.com/web-infra-dev/rspack/pull/3455. You may continue to implement this hook. Thank you in advance!

dmitry-stepanenko: Thanks a lot @h-a-n-a !

dmitry-stepanenko:
@h-a-n-a this PR is pretty much ready, but there's an error `thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime'`. It's caused by this https://github.com/web-infra-dev/rspack/pull/3403/files#diff-5eb759d814281903aa0355174051f134df9444a46e9a1ce332cad27eed896215R1352

Any ideas how this can be avoided?

h-a-n-a:
> @h-a-n-a this PR is pretty much ready, but there's an error `thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime'`. It's caused by this https://github.com/web-infra-dev/rspack/pull/3403/files#diff-5eb759d814281903aa0355174051f134df9444a46e9a1ce332cad27eed896215R1352
> 
> Any ideas how this can be avoided?

This is mainly because of the code that you're trying to run is not running in the context of Tokio runtime. You should create one in this case.

h-a-n-a:
> @h-a-n-a I've tried doing that but looks like I'm not able to come up with functional code. Would you be so kind to provide me a snippet how `get_asset_path` function should be changed in order to properly process the async code inside?

Would you mind if I push some code to this PR?

dmitry-stepanenko:
Absolutely!

h-a-n-a:
After I dig into this PR, I found out there's basically an issue about multi-thread and single-thread interoperation. Details will be added later.

h-a-n-a:
I initiate a dicussion here: https://github.com/web-infra-dev/rspack/discussions/3642

dmitry-stepanenko:
not stale

dmitry-stepanenko:
not stale

dmitry-stepanenko:
not stale

hardfist:
!bench

hardfist:
we will revisit when we find solution

PR #3406, opened 2023-06-03
fix: devtool default value should be false

### LINKED ISSUE #3392: `devtool: false` not work
### System Info

  System:
    OS: macOS 13.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 64.70 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.volta/tools/image/node/16.14.0/bin/node
    npm: 8.3.1 - ~/.volta/tools/image/node/16.14.0/bin/npm
  Browsers:
    Chrome: 114.0.5735.90
    Firefox: 113.0.2
    Safari: 16.4

### Details

By default,field `devtool` is `false` which should not to generator `sourcemap` files. I think the `false` value has been ignored with rust binding options.
Only `Omit<DevTool,false>` support.

https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_binding_options/src/options/mod.rs#L164

### Reproduce link

_No response_

### Reproduce Steps

```shell
pnpm create rspack@latest
pnpm run build
```

faga295:
## Related issue (if exists)
#3392 

## Summary
Remove the handling of the default value for the devtool option in `rspack-cli.ts`.

### <samp>🤖 Generated by Copilot at 810411e</samp>

Removed code that overrode the `devtool` option for webpack in `rspack-cli.ts`. This allows the user to configure source maps according to their needs and preferences.

<details open=true>

### <samp>🤖 Generated by Copilot at 810411e</samp>

*  Removed the code that set the `devtool` option for webpack based on the `mode` argument ([link](https://github.com/web-infra-dev/rspack/pull/3406/files?diff=unified&w=0#diff-44dee9ba83d25009b434358408609d864a6ae2c804c084a8d6cfeec82ab64ab4L166-L169)). This allows the user to configure their own `devtool` option in the `rspack.config.js` file and improves the performance and debugging experience of the application. This change also matches the documentation of the `devtool` option in the `packages/rspack-cli/README.md` file.

</details>

### DIFF
--- a/packages/rspack-cli/src/rspack-cli.ts
+++ b/packages/rspack-cli/src/rspack-cli.ts
@@ -163,10 +163,6 @@ export class RspackCLI {
 				item.mode = options.mode as Mode;
 			}
 
-			// false is also a valid value for sourcemap, so don't override it
-			if (typeof item.devtool === "undefined") {
-				item.devtool = isBuild ? "source-map" : "cheap-module-source-map";
-			}
 			item.builtins = item.builtins || {};
 			if (isServe) {
 				item.builtins.progress = item.builtins.progress ?? true;


packages/rspack-cli/src/rspack-cli.ts
```diff
@@ -163,10 +163,6 @@ export class RspackCLI {
 				item.mode = options.mode as Mode;
 			}
 
-			// false is also a valid value for sourcemap, so don't override it
-			if (typeof item.devtool === "undefined") {
-				item.devtool = isBuild ? "source-map" : "cheap-module-source-map";
```
hardfist, line 168: this will not executed when devtool is set to false

faga295: but it cause the default value of devtool is `source-map` or `cheap-module-source-map` instead of `false`

faga295: and this makes it not working https://github.com/web-infra-dev/rspack/blob/40fef768bfe8d9c495f6a6ea5c7a13c073517ecd/packages/rspack/src/config/defaults.ts#L66

hardfist:
why default value of devtool should be false?

faga295:
image
I refer to this, maybe document should update

PR #3447, opened 2023-06-06
chore: add nestjs-external example & eslint-webpack-plugin example

hardfist:
## Related issue (if exists)
bundle node_modules is much harder than external node_modules, since we needs to deal with lots packages which not support bundle, so we provide another example for NestJS users who don't need bundle node_modules
### <samp>🤖 Generated by Copilot at a4c3c1e</samp>

No summary available (Limit exceeded: required to process 76420 tokens, but only 50000 are allowed per call)

<details open=true>

### <samp>🤖 Generated by Copilot at a4c3c1e</samp>

No walkthrough available (Limit exceeded: required to process 76420 tokens, but only 50000 are allowed per call)

</details>


hardfist:
cc @zackarychapple @antoinepairet might find this useful

Issue #3247, opened 2023-05-22
simplify nodejs profile process


hardfist:
### What problem does this feature solve?

when we met performance  problems on  js thread  we need to use nodejs profile to analyze, but it's not easy to generate nodejs profile right now. so we need a better way to generate nodejs profile

### What does the proposed API of configuration look like?

rspack build --profile to generated a cpu-profile

sladyn98:
@hardfist This is marked as a good first issue, I could give this a shot if there is nothing blocking it

hardfist:
thanks in advance

hardfist:
you can take vites implementation for reference

sladyn98:
@hardfist Is there a link to the PR for that (vites implementation)

hardfist:
> @hardfist Is there a link to the PR for that (vites implementation)

https://github.com/vitejs/vite/blob/e444375d34db1e1902f06ab223e51d2d63cd10de/packages/vite/bin/vite.js#L47

sladyn98:
@hardfist Thanks I understood, just a couple of questions. 
a) Would we want a new script called node.js inside the `bin` directory that would implement the node profiler
b) Or would we just add functionality for profiling nodejs inside `vite.js`
Thanks in advance

hardfist:
@sladyn98  do you still work on this?

hardfist:
bump

hardfist:
we are making a redesign about how rspack profile work

PR #3470, opened 2023-06-07
Add nodejs profiling support

sladyn98:
## Related issue (if exists)

https://github.com/web-infra-dev/rspack/issues/3247

## Summary

This TypeScript script allows you to profile a Node.js-based rspack binary. It checks if the --profile command-line argument is provided. If it is, an Inspector session is initiated and the V8 profiler is started before executing the rspack binary. When the rspack process completes, the profiler is stopped, and the collected CPU profiling data is written to a .cpuprofile file. This script is helpful for identifying performance issues in the rspack binary by providing detailed CPU profiling information.

### <samp>🤖 Generated by Copilot at 01e42eb</samp>

Add a new file `profiler.ts` that enables CPU profiling for rspack. The file uses the inspector API and a child process to run rspack with the `--cpu-prof` flag and save the profile data.

<details open=true>

### <samp>🤖 Generated by Copilot at 01e42eb</samp>

* Add a profiling option to rspack for performance analysis and optimization

</details>


packages/rspack/src/profiler.ts
```diff
@@ -0,0 +1,43 @@
+import { performance } from 'perf_hooks';
+import { spawn, ChildProcess } from 'child_process';
+import { Session } from 'inspector';
+import fs from 'fs';
+
+const profileIndex = process.argv.indexOf('--profile');
+
+function start(): ChildProcess {
+    const rspackProcess = spawn('node', ['--cpu-prof', './path-to-rspack/rspack', '-c', 'rspack.config.js'], {
```
sladyn98, line 9: I would need to get the rspack binary path, do I need to grab this from the config file

sladyn98 (commented):
Do I create a new command in rpsack that would invoke this script

packages/rspack/src/profiler.ts
sladyn98: cc @hardfist

hardfist: you don't need to spawn a process here, profile session can be trigger by node api like https://github.com/vitejs/vite/blob/92b7fbb1c9526d68a05e370cc246c75d9e1a6e5e/packages/vite/bin/vite.js#L53

hardfist:
closed in favor of #3937

PR #3559, opened 2023-06-16
fix: WebpackDevServer typing import

linbudu599:
## Summary

I found this in a downstream package in rspack which [@nx/rspack](https://github.com/nrwl/nx-labs/blob/main/packages/rspack/src/executors/dev-server/dev-server.impl.ts#L32) is using `DevServer` type exported from rspack：

```typescript
import { DevServer } from '@rspack/core/dist/config';

  // If I don't typecast, it throws an error
  // that port does not exist on type DevServer
  // however, it does exist, since DevServer extends
  // WebpackDevServer.Configuration which has port
  let devServerConfig: DevServer = {
    port: options.port ?? 4200,
    hot: true,
  } as DevServer;


// rspack source code
import webpackDevServer from "webpack-dev-server";

// it's extending an any type actually
export interface DevServer extends webpackDevServer.Configuration {
	hot?: boolean;
}
```

Default import item `webpackDevServer` doesnot provide typings exports, you have to use namespace import for 
accessing complete exports.

And fixup can be applied easily:

```typescript
import * as webpackDevServer from "webpack-dev-server";

import type * as webpackDevServer from "webpack-dev-server";
```

</details>


hyf0 (approved):

hyf0:
Thank you!

PR #3586, opened 2023-06-19
fix: rspack serve will automatically rebuild once

jerrykingxyz:
## Context

When using `rspack serve` it may automatically run a rebuild after the first build completes.
By debugging this behavior, I discovered that watchpack had emitted an array containing empty strings as deleted files which triggered the issue. And this empty string is a file_dependency which is generated by `data uri`.

## Summary

1. Modify tests to ensure all items in `compilation.*deps` are absolute paths
2. Only add the resource_path which is a absolute path

### DIFF
--- a/crates/rspack_loader_runner/src/runner.rs
+++ b/crates/rspack_loader_runner/src/runner.rs
@@ -220,7 +220,9 @@ async fn create_loader_context<'c, C: 'c>(
   context: C,
 ) -> Result<LoaderContext<'c, C>> {
   let mut file_dependencies: HashSet<PathBuf> = Default::default();
-  file_dependencies.insert(resource_data.resource_path.clone());
+  if resource_data.resource_path.is_absolute() {
+    file_dependencies.insert(resource_data.resource_path.clone());
+  }
 
   let loader_context = LoaderContext {
     cacheable: true,
--- a/crates/rspack_core/src/module_factory.rs
+++ b/crates/rspack_core/src/module_factory.rs
@@ -33,7 +33,9 @@ impl ModuleFactoryResult {
   }
 
   pub fn file_dependency(mut self, file: PathBuf) -> Self {
-    self.file_dependencies.insert(file);
+    if file.is_absolute() {
+      self.file_dependencies.insert(file);
+    }
     self
   }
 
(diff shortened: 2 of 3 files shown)


crates/rspack_loader_runner/src/runner.rs
```diff
@@ -219,12 +219,9 @@ async fn create_loader_context<'c, C: 'c>(
   plugins: &'c [Box<dyn LoaderRunnerPlugin>],
   context: C,
 ) -> Result<LoaderContext<'c, C>> {
-  let mut file_dependencies: HashSet<PathBuf> = Default::default();
```
jerrykingxyz, line 222: this code is useless, we add `resource_data.resource_path` as file dependencies when resolve success

crates/rspack_core/src/normal_module_factory.rs
```diff
@@ -162,7 +162,7 @@ impl NormalModuleFactory {
       plugin_driver
         .normal_module_factory_resolve_for_scheme(ResourceData::new(
           request_without_match_resource.to_string(),
-          "".into(),
+          request_without_match_resource.to_string().into(),
```
jerrykingxyz, line 165: Same as webpack, when the scheme is data, resourcePath is the same as resource

ahabhgk: Are you sure? If I remember correctly, it should be undefined in webpack

jerrykingxyz: the resourcePath will be recalculated in js via the `resource` field, i have discard this change.

crates/rspack_loader_runner/src/runner.rs
```diff
@@ -220,7 +220,9 @@ async fn create_loader_context<'c, C: 'c>(
   context: C,
 ) -> Result<LoaderContext<'c, C>> {
   let mut file_dependencies: HashSet<PathBuf> = Default::default();
-  file_dependencies.insert(resource_data.resource_path.clone());
+  if resource_data.resource_path.is_absolute() {
```
ahabhgk, line 223: If this is only for filter data url resource, use resource_path != "", we can refactor resource_path to Option<String> in the future PR, which aligns with data url resourcePath is undefined in webpack

ahabhgk (approved):
Leaved a comment, but you can leave it here, since we'll need to refactor resource_path to Option<String> in future PR

PR #3591, opened 2023-06-20
fix(commonjs_scanner): ignore unresolve require span

### LINKED ISSUE #3558: Bundle `jsonpath` library
### System Info

None

### Details

`https://github.com/dchester/jsonpath/issues/50`

image

Webpack is able to bundle this without resolving the two files. 

### Reproduce link

_No response_

### Reproduce Steps

None

suxin2017:
fixed #3558


packages/rspack/tests/statsCases/issuse-3558/module/jsonpath.js
```diff
@@ -0,0 +1,6842 @@
+/*! jsonpath 1.1.1 */
```
h-a-n-a, line 1: Would you move this to the fixtures?

suxin2017: ok

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_scanner.rs
```diff
@@ -106,11 +110,17 @@ impl VisitAstPath for CommonJsScanner<'_> {
     node: &'ast CallExpr,
     ast_path: &mut AstNodePath<AstParentNodeRef<'r>>,
   ) {
-    if is_require_resolve_call(node) {
-      return self.add_require_resolve(node, ast_path, false);
-    }
-    if is_require_resolve_weak_call(node) {
-      return self.add_require_resolve(node, ast_path, true);
+    if let Callee::Expr(expr) = &node.callee {
```
h-a-n-a, line 113: You should change this part https://github.com/web-infra-dev/rspack/blob/59ed8c9dc637f19aedf722fef9873bcd713c5d4d/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs#L145 too. We are migrating dependency code generation from AST to string.

h-a-n-a: cc @underfin

packages/rspack/tests/statsCases/issuse-3558/module/jsonpath.js
ahabhgk: And hand write the minimal issue code instead of copy all jsonpath.js to keep the noise as low as possible

packages/rspack/tests/statsCases/issuse-3558/index.js
```diff
@@ -0,0 +1,3 @@
+const jsonPath = require('./module/jsonpath');
+
+console.log(jsonPath);
```
h-a-n-a, line 3: Can we just leave `require("test-case-you-want")`? It looks not tidy.

h-a-n-a:
There is a typo in the directory name `issuse-3558`, which is intended to be `issue-3358`

packages/rspack/tests/statsCases/issuse-3558/module/jsonpath.js
h-a-n-a: > And hand write the minimal issue code instead of copy all jsonpath.js to keep the noise as low as possible

Agree

suxin2017: We shouldn't need this file, just verify that we can resolve the redefined require

crates/rspack_plugin_javascript/src/visitors/dependency/util.rs
```diff
@@ -206,3 +206,12 @@ fn test() {
     type_args: None
   }));
 }
+
+pub fn is_unresolved_require_member_expr(expr: &Expr, unresolved_ctxt: &SyntaxContext) -> bool {
```
h-a-n-a, line 210: Would you change the name of this function since it can be used for other expressions too?

h-a-n-a: Something that is not related to `require` looks fine to me, maybe something like `is_unresolved_member_object_ident`

h-a-n-a: Would you mind me pushing some code to this PR?

h-a-n-a (approved):

PR #3596, opened 2023-06-20
chore:  env plugin

IWANABETHATGUY:
## Summary
1. configCases/plugins/environmentPlugin
2. Closed https://github.com/web-infra-dev/rspack/issues/3602


webpack-test/checkArrayExpectation.js
```diff
@@ -82,6 +82,10 @@ module.exports = function checkArrayExpectation(
 		}
 	}
 	if (fs.existsSync(path.join(testDirectory, `${filename}.js`))) {
+		if (fs.existsSync(path.join(testDirectory, `${filename}-sort.js`))) {
```
IWANABETHATGUY, line 85: Add an optional sorter file, to sort `stats.errors`,  the order of `stats.errors` is hard to keep the same as webpack because our parallel architecture. And the order of `stats.errors` does not affect the correctness.

packages/rspack/src/lib/BannerPlugin.js
```diff
@@ -124,4 +124,4 @@ class BannerPlugin {
 	}
 }
 
-module.exports = BannerPlugin;
+export { BannerPlugin };
```
h-a-n-a, line 127: In case of users who directly consumes `"webpack/lib/EnvironmentPlugin"`, should we support this?  What do you think? cc@hardfist

hardfist: can this plugin accessed by webpack.EnvironmentPlugin

IWANABETHATGUY: https://github.com/web-infra-dev/rspack/pull/3596/files#diff-0f776ba59eeb95c015bb94540359c4a7131f29fa6001ab35f864e64f2a7dc065R3, as you can see, yes.

IWANABETHATGUY: Do you mean webpack instance?

hardfist: yes，that is enough now

h-a-n-a (approved):

PR #3599, opened 2023-06-21
chore: add back configCases/target in webpack-test

ahabhgk:
## Summary

add back configCases/target, gitignored before


IWANABETHATGUY:
Looks good, can you rename the pull request, so I don't need to reapprove.

IWANABETHATGUY (approved):

Issue #3593, opened 2023-06-20
[Tracking]: HMR performance monitor


jerrykingxyz:
### What subject does this issue tracking?

# [Subject]

To avoid HMR performance degradation, we need
* add some tests before pr merge
* Generate hmr performance report on release

```[tasklist]
### Tasks
- [ ] discuss the implementation
- [ ] hmr performance ci
- [ ] generate hmr performance report
```

PR #3604, opened 2023-06-21
fix: import export check invalidation on make stage

jerrykingxyz:
## Context

The `compilation.has_module_import_export_change` is used to skip code & bundle splitting, it can improve hmr speed when no module import and export changes.
It is incorrectly set to `true` which will cause code & bundle splitting to never be skipped.

## Summary

Fix the incorrect value of `compilation.has_module_import_export_change` and fix hmr performance degradation.

## Test Plan

Currently no testing plan but I will build hmr monitoring system to avoid hmr speed degradation.
For more information see: https://github.com/web-infra-dev/rspack/issues/3593

### DIFF
--- a/crates/rspack_core/src/compiler/hmr.rs
+++ b/crates/rspack_core/src/compiler/hmr.rs
@@ -173,7 +173,7 @@ where
         new_compilation.code_splitting_cache =
           std::mem::take(&mut self.compilation.code_splitting_cache);
 
-        self.compilation.has_module_import_export_change = false;
+        new_compilation.has_module_import_export_change = false;
         // remove prev build ast in modules
         fast_drop(
           new_compilation


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -109,7 +109,7 @@ impl Compilation {
       module_graph,
       make_failed_dependencies: HashSet::default(),
       make_failed_module: HashSet::default(),
-      has_module_import_export_change: true,
+      has_module_import_export_change: false,
```
h-a-n-a, line 112: I think setting to `true` will always make the result as expected, would this interfere the non-incremental build?

jerrykingxyz: https://github.com/web-infra-dev/rspack/blob/ed922e724e027484c565ccbedafaebfc15307291/crates/rspack_core/src/compiler/compilation.rs#L713-L727

The make phase will recalculate `has_module_import_export_change`. It is in a non-incremental build when the `origin_module_deps` is empty. The reason for this problem is that there is a call to `compilation.rebuildModule`, the previous `has_module_import_export_change` needs to be reused.

jerrykingxyz: https://github.com/web-infra-dev/rspack/blob/ed922e724e027484c565ccbedafaebfc15307291/crates/rspack_core/src/compiler/hmr.rs#L144-L197

I could move this setting to incremental checks and it would solve the problem too, but I just think false as the default is more intuitive

jerrykingxyz:
!canary

crates/rspack_core/src/compiler/compilation.rs
jerrykingxyz: Yes. I have modified this into incremental initialization

h-a-n-a (approved):
Looks good!

PR #3605, opened 2023-06-21
chore:  loader options plugin

IWANABETHATGUY:
## Summary
1. Closed https://github.com/web-infra-dev/rspack/issues/3606
2. Finish the test case webpack-test/configCases/plugins/loader-options-plugin/

### DIFF
--- a/packages/rspack/src/schemas/plugins/LoaderOptionsPlugin.check.js
+++ b/packages/rspack/src/schemas/plugins/LoaderOptionsPlugin.check.js
@@ -1,0 +1,51 @@
+/*
+ * This file was automatically generated.
+ * DO NOT MODIFY BY HAND.
+ * Run `yarn special-lint-fix` to update
+ */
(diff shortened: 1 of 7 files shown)


packages/rspack/src/schemas/plugins/LoaderOptionsPlugin.check.js
```diff
@@ -0,0 +1,51 @@
+/*
```
h-a-n-a, line 1: How would we generate this file if `scheme.json` changes?

h-a-n-a (approved):

packages/rspack/src/schemas/plugins/LoaderOptionsPlugin.check.js
IWANABETHATGUY: For the short term, we will not change the plugin. We can port the `scheme generation` in the next quarter.

PR #3621, opened 2023-06-25
feat: support function `Rule.use`

### LINKED ISSUE #3407: Support Function of `Rule.use`

Be consistent with Webpack behavior

Example: https://github.com/unjs/unplugin/blob/856b0736dbb658ffd6cf32d246359683b622b36f/src/webpack/index.ts#LL68C4-L68C4

 Type declaration of Webpack:
```ts
(data: {
	resource: string;
	realResource: string;
	resourceQuery: string;
	issuer: string;
	compiler: string;
}) => RuleSetUseItem[]
```

faga295:
Closes #3407

## Test Plan
The test case is in `packages/rspack` it use less loader as example and return loaders accroding to `resourceQuery`, `resource`,`realResource`,`issuer`


ahabhgk:
LGTM, @h-a-n-a would you also take a look?

h-a-n-a (dismissed):

ahabhgk (approved):

Issue #3515, opened 2023-06-12
remove sourcemap in dist when publish


hardfist:
https://www.npmjs.com/package/@rspack/core?activeTab=code
we shipped sourcemap without shipping source, and it seems unnecessary ship source and sourcemap in the dist(which will cause npm download slow)
so it would be better we just remove sourcemap when publish(but we need sourcemap in development for better debug experience)

hardfist:
we may need two `tsconfig.json`, `tsconfig.build.json` for build and publish(without sourcemap) and  `tsconfig.json` for local development(with sourcemap).

RussellCanfield:
@hardfist - I would be interested in contributing to this issue. I would likely use the existing "prepare" script in rspack/core, and run a clean/build using another tsconfig.build.json (or tsconfig.lib.json) with source maps disabled.

Assuming we wanted the build to be part of prepare, or just enhance the existing build script since there is already a dev script that performs a watch on the default tsconfig.json.

hardfist:
> @hardfist - I would be interested in contributing to this issue. I would likely use the existing "prepare" script in rspack/core, and run a clean/build using another tsconfig.build.json (or tsconfig.lib.json) with source maps disabled.
> 
> Assuming we wanted the build to be part of prepare, or just enhance the existing build script since there is already a dev script that performs a watch on the default tsconfig.json.

yeah. that is what we want

PR #3631, opened 2023-06-26
feat: Disable source maps for rspack core

RussellCanfield:
## Summary

Disable source maps for rspack/core - #3515. Change is limited to rspack/core, the new tsconfig.lib.config file will only be used on builds which is what the github action triggers. I wasn't sure if we wanted to rimraf the dist folder, similar to rspack/cli, it should be harmless if we want to add it.

## Test Plan

Execute pnpm build:js, or build the rspack/core project directly. Ensure the "dist" folder is deleted prior to running build. Ensure the "dist" folder contains no .map files after the build.


hyf0:
Thanks for your contribution. May I ask why we need to disable the source map of `@rspack/core`?

hardfist:
can you help fix all package？

hardfist:
> Thanks for your contribution. May I ask why we need to disable the source map of `@rspack/core`?

see related issue for more context

hyf0 (approved):

RussellCanfield:
> can you help fix all package？

I can do that, let me do them in separate PRs, I will reference the same issue for context.

hardfist (approved):

PR #3643, opened 2023-06-28
fix(examples/vue): refine module resource matching rule.

sumy7:
## Summary

change matching rule in examples/vue to make less-loader process correct resources.


zoolsher:
Nice!

PR #3664, opened 2023-06-29
test: add more E2E test case

jerrykingxyz:
## Summary

* rspack fixture set default value for context and builtins.define.
* crossoriginloading case test __public_path__
* add tailwindcss test case
* add asset-emitted hook test


Boshen:
LGTM, just need to get CI to pass.

Boshen (approved):

PR #3672, opened 2023-06-30
fix: hoist async or generate anonymous function decl at export decl

underfin:
## Summary

Also need hoist async or generate anonymous function decl at export default decl

## Test Plan
Added test


crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_expression_dependency.rs
```diff
@@ -8,11 +8,21 @@ pub struct HarmonyExpressionHeaderDependency {
   pub start: u32,
   pub end: u32,
   pub declaration: bool,
-  pub function: bool,
+  pub function: Option<(
```
hyf0, line 11: It's much better to use a struct rather than a tuple here.

hyf0 (approved):

PR #3673, opened 2023-06-30
feat: Support passing string to splitChunks.{cacheGroup}.test

Hamzakh777:

### DIFF
--- a/crates/rspack_plugin_split_chunks_new/src/common.rs
+++ b/crates/rspack_plugin_split_chunks_new/src/common.rs
@@ -7,7 +7,9 @@ use std::{
 
 use derivative::Derivative;
 use futures_util::FutureExt;
+use napi::{Either, JsString};
 use rspack_core::{Chunk, ChunkGroupByUkey, Module, SourceType};
+use rspack_napi_shared::JsRegExp;
 use rustc_hash::{FxHashMap, FxHashSet};
 
 pub type ChunkFilter = Arc<dyn Fn(&Chunk, &ChunkGroupByUkey) -> bool + Send + Sync>;
@@ -54,7 +56,7 @@ pub fn create_module_filter_from_rspack_regex(re: rspack_regex::RspackRegex) -> 
   })
 }
 
-pub fn create_module_filter(re: Option<String>) -> ModuleFilter {
+pub fn create_module_filter(re: Option<Either<JsString, JsRegExp>>) -> ModuleFilter {
   re.map(|test| {
     let re =
       rspack_regex::RspackRegex::new(&test).unwrap_or_else(|_| panic!("Invalid regex: {}", &test));
--- a/crates/rspack_binding_options/src/options/raw_split_chunks.rs
+++ b/crates/rspack_binding_options/src/options/raw_split_chunks.rs
@@ -117,7 +117,10 @@ pub struct RawCacheGroupOptions {
   pub priority: Option<i32>,
   // pub reuse_existing_chunk: Option<bool>,
   //   pub r#type: SizeType,
-  pub test: Option<String>,
+  #[serde(skip_deserializing)]
+  #[napi(ts_type = "RegExp | 'async' | 'initial' | 'all'")]
+  #[derivative(Debug = "ignore")]
+  pub test: Option<Either<JsString, JsRegExp>>,
   //   pub filename: String,
   //   pub enforce: bool,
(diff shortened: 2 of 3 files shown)


crates/rspack_plugin_split_chunks_new/src/common.rs
```diff
@@ -54,7 +56,7 @@ pub fn create_module_filter_from_rspack_regex(re: rspack_regex::RspackRegex) ->
   })
 }
 
-pub fn create_module_filter(re: Option<String>) -> ModuleFilter {
+pub fn create_module_filter(re: Option<Either<JsString, JsRegExp>>) -> ModuleFilter {
```
hyf0, line 59: Currently, Rspack only supports `JsRegExp`, so `JsRegExp`  is converted to `RspackRegex`, which we used to create a [closure](https://github.com/web-infra-dev/rspack/blob/5b145f69eadd23f3e79df174bf15743509fe1ded/crates/rspack_plugin_split_chunks_new/src/common.rs#L51-L58) for porting this part code: https://github.com/webpack/webpack/blob/f921a5baca9c7d200eee8204fafaaaf36b7c1451/lib/optimize/SplitChunksPlugin.js#L530-L533.

Now, we want Rspack to support `String` and the corresponding code of Webpack is https://github.com/webpack/webpack/blob/f921a5baca9c7d200eee8204fafaaaf36b7c1451/lib/optimize/SplitChunksPlugin.js#L526-L529.

So, you need to rewrite the `create_module_filter` function.

The behavior should be
- If the passing value is a `JsRegExp` return a closure does [this](https://github.com/webpack/webpack/blob/f921a5baca9c7d200eee8204fafaaaf36b7c1451/lib/optimize/SplitChunksPlugin.js#L530-L533).
  - Actually, we already implement it in the current codebase.
- If the passing value is `String`, return a closure does [this](https://github.com/webpack/webpack/blob/f921a5baca9c7d200eee8204fafaaaf36b7c1451/lib/optimize/SplitChunksPlugin.js#L526-L529).

crates/rspack_plugin_split_chunks_new/Cargo.toml
```diff
@@ -8,15 +8,17 @@ version    = "0.1.0"
...
+rspack_identifier  = { path = "../rspack_identifier" }
+rspack_napi_shared = { path = "../rspack_napi_shared" }
+rspack_regex       = { path = "../rspack_regex" }
 
 async-scoped = { workspace = true, features = ["use-tokio"] }
 async-trait  = { workspace = true }
 dashmap      = { workspace = true }
 derivative   = { workspace = true }
 futures-util = { workspace = true }
+napi         = { workspace = true }
```
hyf0, line 21: It's better not to use `napi` in this crate. You could do some transform.

packages/playground/cases/context/delete-file/index.test.ts
```diff
@@ -1,26 +1,33 @@
...
-    // ensure page status changed
-    const statusText = await page.textContent("#root");
-    return statusText === "__HMR_UPDATED__";
-  });
-  // asset new script content
-  const responseAfterDelete = await request.get(`http://localhost:${rspack.devServer.options.port}/main.js`);
-  const bodyResponseAfterDelete = (await responseAfterDelete.body()).toString();
-  expect(bodyResponseAfterDelete).not.toContain('this is mod1');
-  expect(bodyResponseAfterDelete).toContain('this is mod2');
+test("delete file should work", async ({
```
hyf0, line 3: This file should not be touched.

crates/rspack_plugin_split_chunks_new/src/common.rs
```diff
@@ -57,11 +59,24 @@ pub fn create_module_filter_from_rspack_regex(re: rspack_regex::RspackRegex) ->
   })
 }
 
-pub fn create_module_filter(re: Option<String>) -> ModuleFilter {
-  re.map(|test| {
-    let re =
-      rspack_regex::RspackRegex::new(&test).unwrap_or_else(|_| panic!("Invalid regex: {}", &test));
-    create_module_filter_from_rspack_regex(re)
+pub fn create_module_filter_from_rspack_str(st: String) -> ModuleFilter {
```
hyf0, line 62: Good work.

hyf0 (approved):
Thank you!

PR #3681, opened 2023-07-03
chore: add monaco-editor examples

ahabhgk:

### DIFF
--- a/examples/monaco-editor-js/index.js
+++ b/examples/monaco-editor-js/index.js
@@ -1,0 +1,24 @@
+import * as monaco from 'monaco-editor';
+
+self.MonacoEnvironment = {
+	getWorkerUrl: function (moduleId, label) {
+		if (label === 'json') {
+			return './json.worker.bundle.js';
+		}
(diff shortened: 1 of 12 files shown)


examples/monaco-editor-js/index.js
```diff
@@ -0,0 +1,24 @@
...
+			return './css.worker.bundle.js';
+		}
+		if (label === 'html' || label === 'handlebars' || label === 'razor') {
+			return './html.worker.bundle.js';
+		}
+		if (label === 'typescript' || label === 'javascript') {
+			return './ts.worker.bundle.js';
+		}
+		return './editor.worker.bundle.js';
+	}
```
IWANABETHATGUY, line 18: nit: I prefer to use `switch` here, but it's OK to use multiple `if`

IWANABETHATGUY (approved):

examples/monaco-editor-js/index.js
ahabhgk: I just copy from monaco-editor repo XD

PR #3682, opened 2023-07-03
feat: support stats option "source"

### LINKED ISSUE #3665: Feature Request: add a `source` getter on `JsStatsModule`

I am adding a rspack unit test for one of my webpack loader projects, [`forgetti-loader`](https://github.com/SukkaW/forgetti-loader) (see commit https://github.com/SukkaW/forgetti-loader/commit/4620937e47a6cb6b50818a280cc33f5f45b8e4a9), and I want to make a snapshot from the output of one of the transformed modules.

However, rspack's stats API doesn't include the `source` field in the `stats.modules[]`.

Reading the individual output of each module can and will be very expensive, so the `stats.modules[].source` should only be included with a `verbose` stats preset, and it should be implemented as a getter so that it can be lazily evaluated.

suxin2017:
close #3665

## Test Plan

Configure the following configuration

```
	stats: {
		source: true,
                // or
                all :true
	}
```

simple test case path
packages/rspack/tests/statsCases/simple-module-source/webpack.config.js


packages/rspack/src/config/zod/stats.ts
```diff
@@ -28,7 +28,8 @@ export function stats() {
 				chunkRelations: z.boolean().optional(),
 				timings: z.boolean().optional(),
 				builtAt: z.boolean().optional(),
-				nestedModules: z.boolean().optional()
+				nestedModules: z.boolean().optional(),
+				source: z.boolean().optional()
```
hyf0, line 32: What's the default value of it? Is it aligned with Webpack?

hyf0 (approved):

Boshen:
What should be our strategy for webpack tests for these kind of things?

@suxin2017 It is recommended to fill in the test plan so we know the exact things you did for testing.

```
<!-- Can you please describe how you tested the changes you made to the code? -->
```

crates/node_binding/binding.d.ts
```diff
@@ -350,6 +350,7 @@ export interface JsStatsModule {
   issuerPath: Array<JsStatsModuleIssuer>
   reasons?: Array<JsStatsModuleReason>
   assets?: Array<string>
+  source?: string
```
h-a-n-a, line 353: Why is it always a `string` typed? It should be `Buffer` when it comes to binary assets.

h-a-n-a: > I don't know what to do now. Can you help me? There is a `JsCompatSource` object in `node_binding`, which I want to use, but in the `stats` module of`rspack_core`, `StatsModule` can only get an `Option<&dyn Source>` This `Source` In this case, if I want to implement `&dyn Source` I have to add a reference to StatsModule, I don't think this solution is very good, do you have any suggestions?

Sure, I will take a look in these days.

crates/rspack_core/src/stats.rs
```diff
@@ -418,6 +421,13 @@ impl Stats<'_> {
       reasons,
       assets,
       modules,
+      source: source
+        .then(|| {
+          module
```
h-a-n-a, line 426: You can keep this field and use `Option<&dyn Source>` instead of eagerly converting it to a `String`

crates/node_binding/src/js_values/stats.rs
```diff
@@ -91,6 +91,7 @@ pub struct JsStatsModule {
   pub issuer_path: Vec<JsStatsModuleIssuer>,
   pub reasons: Option<Vec<JsStatsModuleReason>>,
   pub assets: Option<Vec<String>>,
+  pub source: Option<String>,
```
h-a-n-a, line 94: Here, should be `Option<Either<String, Buffer>>`

crates/node_binding/src/js_values/stats.rs
```diff
@@ -111,6 +112,7 @@ impl From<rspack_core::StatsModule> for JsStatsModule {
         .reasons
         .map(|i| i.into_iter().map(Into::into).collect()),
       assets: stats.assets,
+      source: stats.source,
```
h-a-n-a, line 115: Due to the fact that the type of `stats.source` has already changed to `Option<&dyn Source>`, you can now leverage `to_js_compat_source` to convert it `JsCompatSource`, however, as what we want is an `Option` with a union type. You should again implement conversion logic in Rust. 

This logic is almost the same as here: https://github.com/web-infra-dev/rspack/blob/02f313b19605503663d3bafb4d52484e4e5bbe96/packages/rspack/src/util/createSource.ts#L7-L16. 

```rust
if source.is_raw && source.is_buffer {
   // convert to buffer
}

// convert to string
```

h-a-n-a: Note: using `JsCompatSource` here is only for making sure that we get the right `Buffer` data. The difference between `rspack_sources` and `webpack-sources` is that we don't return `Buffer` in the `source` of `RawSource`. 

So, originally, I used the type downcasting to make sure the behavior is correct as I implemented here: https://github.com/web-infra-dev/rspack/blob/02f313b19605503663d3bafb4d52484e4e5bbe96/crates/node_binding/src/js_values/source.rs#L96-L198

h-a-n-a: Would you please also test if a binary asset shows the correct original source in the final stats result?

h-a-n-a: I think this is a problem related to https://github.com/web-infra-dev/rspack/blob/d0e5f05f73f776acd657746031835705dcdf09a5/crates/rspack_core/src/normal_module.rs#L561. `Buffer` and `String` should be converted respectfully. Would you fix this?

h-a-n-a: Tracked here: https://github.com/web-infra-dev/rspack/issues/3795

PR #3683, opened 2023-07-03
refactor: set dependency id at dependency init

underfin:
## Summary

Before we consider  thread lock for assign dependency id and set dependency id at after init dependency, but it caused dependencies reference isn't clear(the workaround code is very complex), so the pr set dependency id at init dependency.

## Test Plan

It's a code refactor, our test case should test it.

### DIFF
--- a/crates/rspack_core/src/dependency/mod.rs
+++ b/crates/rspack_core/src/dependency/mod.rs
@@ -155,10 +155,9 @@ impl Display for DependencyCategory {
 }
 
 pub trait Dependency: AsAny + DynClone + Send + Sync + Debug {
-  fn id(&self) -> Option<DependencyId> {
-    None
+  fn id(&self) -> DependencyId {
+    0.into()
   }
-  fn set_id(&mut self, _id: Option<DependencyId>) {}
 
   fn category(&self) -> &DependencyCategory {
     &DependencyCategory::Unknown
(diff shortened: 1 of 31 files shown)


underfin:
!bench

underfin:
!bench

underfin:
!bench

underfin:
!bench

underfin:
!bench

crates/rspack_core/src/dependency/mod.rs
```diff
@@ -155,11 +157,6 @@ impl Display for DependencyCategory {
 }
 
 pub trait Dependency: AsAny + DynClone + Send + Sync + Debug {
-  fn id(&self) -> Option<DependencyId> {
```
h-a-n-a, line 158: I think we still need this field? What's your purpose of removing this field?

crates/rspack_core/src/dependency/mod.rs
```diff
@@ -267,6 +221,10 @@ pub trait ModuleDependency: Dependency {
 }
 
 impl ModuleDependency for Box<dyn ModuleDependency> {
+  fn id(&self) -> &DependencyId {
```
h-a-n-a, line 224: As far as I concerned, `id` should be available to all dependency types. So we'd better to move this to `Dependency`.

underfin: Because we current only need id for module dependency, it is used generate module graph, but other dependency isn't need it and don't need creat id for it.

h-a-n-a (approved):

crates/rspack_core/src/dependency/mod.rs
h-a-n-a: Looks good to me.

PR #3689, opened 2023-07-03
fix: output.clean can not remove *.hot-update.js

jerrykingxyz:
## Summary

* change `compilation.emitted_asset_versions` to HashMap
* Fix the bug that output.clean cannot remove *.hot-update.js.

## Test Plan

Added a e2e test case to ensure results are as expected.


crates/rspack_core/src/compiler/mod.rs
```diff
@@ -198,6 +198,7 @@ where
 
   #[instrument(name = "emit_assets", skip_all)]
   pub async fn emit_assets(&mut self) -> Result<()> {
+    println!("self.options.output.clean {:?}", self.options.output.clean);
```
underfin, line 201: Should be removed

underfin (approved):

Issue #3205, opened 2023-05-18
Analyzer should follow 1-pass visit


h-a-n-a:
### What problem does this feature solve?

Currently, dependency analysis costs n-passes. From the perspective of performance, we should change it to a single pass.

### What does the proposed API of configuration look like?

Not related: Internal API

Boshen:
Can you link some source code for this problem? I want to study the code a bit.

h-a-n-a:
> Can you link some source code for this problem? I want to study the code.

Currently, Rspack does not granular split analysis, which would cause a bunch of overhead. https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/visitors/dependency/scanner.rs

h-a-n-a:
bump

PR #3690, opened 2023-07-03
feat: port JS `ParserHooks` of Webpack

hyf0:
## Summary

Port JS `ParserHooks` of Webpack.

Related issues & discussions:

- https://github.com/web-infra-dev/rspack/discussions/3619
- https://github.com/web-infra-dev/rspack/issues/3205

This PR introduces `JsParserHookDriver` and the reason why we introduce it is we want to finish all scanning work in a single visit pass.

## Test Plan

The correctness of the build artifact could be ensured by current tests.

I also add some unit tests to test the behaviors of `JsParserHookDriver`.


crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
```diff
@@ -0,0 +1,162 @@
...
+// 3. Copy the name and Ast node type of the method and use a tuple to combine them.
+// 4. We will get a `(visit_expr, ast::Expr)`, then add it in the following postion.
+// 5. Don't forget to add the comma.
+register_hooks!(pub trait JsParserHook {
+    (visit_expr, ast::Expr),
+    (visit_ident, ast::Ident),
+    (visit_call_expr, ast::CallExpr),
+});
+
+// #[test]
```
Boshen, line 106: This should be cleaned up.

crates/rspack_core/Cargo.toml
```diff
@@ -29,7 +29,7 @@ rayon = { workspace = true }
 regex = { workspace = true }
 rspack_database = { path = "../rspack_database", features = ["rayon"] }
 rspack_error = { path = "../rspack_error" }
-rspack_fs = { path = "../rspack_fs", features = ["async"] }
+rspack_fs = { path = "../rspack_fs", features = ["async", "rspack-error"] }
```
hyf0, line 32: Added for running `cargo test -p rspack_plugin_javascript_shared`.

crates/rspack_core/Cargo.toml
```diff
@@ -48,6 +48,7 @@ swc_core = { workspace = true, features = [
   "__visit",
   "__utils",
   "__ecma_transforms",
+  "ecma_preset_env",
```
hyf0, line 51: Added for running `cargo test -p rspack_plugin_javascript_shared`.

hyf0:
!bench

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
```diff
@@ -0,0 +1,160 @@
...
+        }
+      )*
+    }
+
+    impl Visit for JsParserHookDriver {
+        $(
+        fn $name(&mut self, node: &$node) {
+          self.hooks.values_mut().for_each(|hooks| {
+            // Bailout if any hook returns `true`
+            hooks.iter_mut().any(|hook| matches!(hook.$name(&mut self.context, node), Control::Skip));
```
h-a-n-a, line 91: This actually looks pretty weird to me. Why would the hook affect other hooks behavior for __analyzation__ itself? Could you give an example for the real-world use case? If this is for performance reason, I think this will make problem hard to discover in the future.

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
```diff
@@ -0,0 +1,160 @@
...
+pub struct JsParserContext {
+  pub presentational_dependencies: Vec<Box<dyn CodeGeneratableDependency>>,
+  pub dependencies: Vec<Box<dyn ModuleDependency>>,
+}
+
+/// The reason why we introduce [JsParserHookDriver] instead of using plain `swc_core::ecma_visit::Visit` is
+/// we want to finish all work in a single visit pass.
+#[derive(Default)]
+pub struct JsParserHookDriver {
+  hooks: HashMap<StaticStr, Vec<Box<dyn JsParserHook>>>,
```
h-a-n-a, line 21: What is this key for?

hyf0: https://github.com/webpack/webpack/blob/f1ef2eded2fda8136535c60813dcb0c55510a5e0/lib/APIPlugin.js#L193

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
hyf0: That's how webpack works. 

For example: 
https://github.com/webpack/webpack/blob/f1ef2eded2fda8136535c60813dcb0c55510a5e0/lib/APIPlugin.js#L190

Return `true` would skip all the subscribers under the `__webpack_layer__` key.

The real-world example would be `Worker support`. If you have interest in it, we could setup a lark meeting.

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
h-a-n-a: But this looks not the same to me. 

For example: https://github.com/web-infra-dev/rspack/pull/3690/files#diff-8c1f9a7201ddef2c36f695634b4584ccfc786751a7c4eb2975625795b87ebe6fR83. Even after you registered the key `require.context`, you still need to write a guard here: https://github.com/web-infra-dev/rspack/pull/3690/files#diff-8c1f9a7201ddef2c36f695634b4584ccfc786751a7c4eb2975625795b87ebe6fR20.

The expected way of this would be how webpack does.

hyf0: It's not used as a guard and it doesn't affect the AST visit. Let's have a meeting for reviewing.

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
h-a-n-a: This looks good to me.

crates/rspack_plugin_javascript_shared/src/js_parser_hook.rs
h-a-n-a: It would be more convenient to use this as a guard or, personally speaking, this key does not seem necessary to me...

hyf0: No. The key has the other purpose. It's not used to express what ast node I want to visit.

Another use case of Webpack:

- https://github.com/webpack/webpack/blob/00769837bcd2bdeb105ed5786373b6c2a0c0b1e1/lib/dependencies/HarmonyImportDependencyParserPlugin.js#L30
- https://github.com/webpack/webpack/blob/00769837bcd2bdeb105ed5786373b6c2a0c0b1e1/lib/dependencies/HarmonyImportDependencyParserPlugin.js#L187

h-a-n-a: Doesn't this mean that I want to visit identifier for `__webpack_layer__`? https://github.com/webpack/webpack/blob/f1ef2eded2fda8136535c60813dcb0c55510a5e0/lib/APIPlugin.js#L193

hyf0: Nope.

```js
parser.hooks.evaluateIdentifier
  .for("__webpack_layer__")
  .tap(PLUGIN_NAME, expr =>
```

means that I want my callback to be called when the AST node `evaluateIdentifier` is visited. The thing that the callback do is related `__webpack_layer__`, which is just a string. It's more of a description. You could write it in this way

```js
parser.hooks.evaluateIdentifier
  .for("something related to __webpack_layer__")
  .tap(PLUGIN_NAME, expr =>
```

hyf0: So you may have many callbacks that care about the same thing. The key is used to express what thing.

h-a-n-a: I wrote a demo locally. After changing `for("__webpack_layer__")` to `for("something related to __webpack_layer__")`, the callback is never called.

h-a-n-a: You can see here for how webpack implements this feature with the key of hook map: https://github.com/webpack/webpack/blob/6be4065ade1e252c1d8dcba4af0f43e32af1bdc1/lib/javascript/JavascriptParser.js#L3556

hyf0: Get it. Let's have a meeting. This might not fit the current design of this PR.

hyf0:
On hold: Need the final decision of @hardfist.

PR #3697, opened 2023-07-04
fix(minifier): fix minifier options difference between minifications

h-a-n-a:
## Summary

Directly deleting options from the original object will cause the `this.options` cannot be used for the next run. 
This also causes failure on the angular example.

Fixed build exit code for angular and the example itself.

### DIFF
--- a/examples/angular/scripts/build.js
+++ b/examples/angular/scripts/build.js
@@ -1,4 +1,5 @@
 const { spawn, spawnSync } = require("node:child_process");
+const path = require("node:path");
 const ls = spawn("node", ["-v"]);
 
 ls.stdout.on("data", data => {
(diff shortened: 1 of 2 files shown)


hyf0 (approved):

hyf0 (dismissed):

examples/angular/scripts/build.js
```diff
@@ -1,31 +1,45 @@
...
 });
 
 function build() {
 	return new Promise((resolve, reject) => {
 		try {
-			spawnSync("npx", ["rspack", "build"], { stdio: "inherit" });
+			const { status, error, stderr } = spawnSync(
+				"node",
+				[
+					path.resolve(__dirname, "../node_modules/@rspack/cli/bin/rspack"),
```
h-a-n-a, line 32: Didn't know why this binary is not linked to the `.bin` directory in windows, and I used this as a workaround.

hyf0: Changes in this PR should not affect this. Does the CI run failed?

h-a-n-a: Yes. `npx` is not working as expected since `bin` was not exist there.

hyf0: > Yes. npx is not working as expected since bin was not exist there.

Does the main branch have the problem too?

h-a-n-a: > > Yes. npx is not working as expected since bin was not exist there.
> 
> Does the main branch have the problem too?

I don't think this will execute the correct one on the fly as `npx` may also pull the package from remote. This might be the reason that `main` is working. Using this technique is to make sure to test locally linked packages.

PR #3699, opened 2023-07-04
ci: compile dependencies with optimization for faster tests

Boshen:
## Summary

CI build uses debug mode and runs really slow, this can be improved by compiling all dependencies with optimization turned on, and our crates without any optimizations. This should give us the best build time (since third party crates are cached) + the best test execution time. The only downside is cold build will be really slow on windows and mac, but the trade off is worth because we don't have many cold builds due to how are cache management is setup. 

## Test Plan

* CI passes
* tests run less than 10 minutes. See successful run: https://github.com/web-infra-dev/rspack/actions/runs/5451668329/usage, the slowest is 10 minutes from "Test Windows / Test Node 14"

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -85,6 +85,15 @@ jobs:
       - name: Setup Rust Target
         run: rustup target add ${{ inputs.target }}
 
+      # Compile dependencies with optimization to make tests run faster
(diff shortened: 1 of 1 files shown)


jerrykingxyz (approved):

Boshen:
image

Nice!

Issue #2795, opened 2023-04-17
HTML builtin favicon behavior inconsistent with html-webpack-plugin


shawnmcknight:
### System Info

System:                                                       
  OS: Windows 10 10.0.19045                                   
  CPU: (16) x64 11th Gen Intel(R) Core(TM) i9-11900H @ 2.50GHz
  Memory: 16.57 GB / 31.73 GB                                 
Binaries:                                                     
  Node: 18.15.0 - C:\Program Files\nodejs\node.EXE            
  npm: 9.5.0 - C:\Program Files\nodejs\npm.CMD                
Browsers:                                                     
  Edge: Spartan (44.19041.1266.0), Chromium (112.0.1722.48)   
  Internet Explorer: 11.0.19041.1566                          

### Details

When using the `html-webpack-plugin` and supplying a path to a `favicon` property, the file at that path is copied into the output folder and the resulting html file has the favicon link pointing to the path of where the file was copied into the output folder.  This works whether providing a relative or an absolute path.

When using `builtins.html`, the behavior is different.  If providing an absolute path, the final html output will reference that full absolute path which is likely invalid since that path might not be available in the final runtime.  If providing a relative path, the folder structure to obtain that relative path will be copied into the output directory prepended with a forward slash.  However, the path separator character remains a backslash (likely due to Windows path separators).

For comparison purposes:

**Webpack Absolute Path**
Input: `favicon: path.resolve(process.cwd(), 'static', 'favicon.ico'),`
Output: `<link rel="icon" href="/favicon.ico">`

**Rspack Absolute Path**
Input: `favicon: path.resolve(process.cwd(), 'static', 'favicon.ico'),`
Output: `<link rel="icon" href="/C:\Users\<hidden>\<hidden>\<hidden>\<hidden>\<hidden>\static\favicon.ico"`

Note: All the path segments marked as `<hidden>` are to avoid leaking my folder structure into this issue log.  Each one replaces a subdirectory that was present in the output path.

**Webpack Relative Path**
Input: `favicon: path.join('static', 'favicon.ico'),`
Output: `<link rel="icon" href="/favicon.ico">`

**Rspack Relative Path**
Input: `favicon: path.join('static', 'favicon.ico'),`
Output: `<link rel="icon" href="/static\favicon.ico" />`

[...]

PR #3700, opened 2023-07-04
fix(rspack_plugin_html): favicon behavior when using absolute file path

LingyuCoder:
## Summary

Fix the favicon behavior when using absolute file path in rspack_plugin_html (#2795 )


crates/rspack_plugin_html/src/config.rs
```diff
@@ -131,4 +132,14 @@ impl HtmlPluginConfig {
         .render(compilation, filename),
     }
   }
+  pub fn get_relative_path(&self, compilation: &Compilation, filename: &str) -> String {
```
hardfist, line 135: sugar_path already support this

h-a-n-a:
Looks good to me. Could you fix the test?

crates/rspack_plugin_html/src/config.rs
LingyuCoder: It is not just to get the relative path between two PathBufs. This function is used to get the relative path to the project root directory, so exists favicon logic can be reused.

LingyuCoder:
> Looks good to me. Could you fix the test?

Test has been fixed

h-a-n-a (approved):
LGTM!

PR #3714, opened 2023-07-06
refactor(util): 1.better string concat performance 2.reduce code complexity 3.add comments

hyf0:
## Summary

1.better string concat performance. See https://github.com/hoodie/concatenation_benchmarks-rs
2.reduce code complexity
3.add comments

## Test Plan

Not needed.


h-a-n-a (approved):
LGTM, but title looks a little bit cumbersome to me tho.

PR #3716, opened 2023-07-06
chore: 🤖 revert file name generate logic

IWANABETHATGUY:


ahabhgk (approved):

Boshen:
It seems like there's some missing context here. And it would be awesome if there is a test plan for preventing similar issues in the future.

IWANABETHATGUY:
> 

Done

Issue #3715, opened 2023-07-06
add version check for binding


hardfist:
### What problem does this feature solve?

since we distribute @rspack/core and @rspack/binding separately, the version may mismatch here cause unexpected behavior here, we should do version check and provide nicely error message for users other than unreadable error message
![image](https://github.com/web-infra-dev/rspack/assets/8898718/8373f218-0ce0-45fa-9ec0-38d9ba89e8c2)

### What does the proposed API of configuration look like?

no api needed, esbuild provides nice error message for version check
![image](https://github.com/web-infra-dev/rspack/assets/8898718/c3fee7b8-5f9f-4903-a591-b97e65a4351b)

lippzhang:
please assign this to me

hardfist:
thanks in advance

PR #3717, opened 2023-07-06
feat(packages/rspack): add version check for binding

lippzhang:
## Summary

Add version consistency check for core, binding and addon.
closes https://github.com/web-infra-dev/rspack/issues/3715
## Test Plan

ecosystem-ci passing

### DIFF
--- a/packages/rspack/src/compiler.ts
+++ b/packages/rspack/src/compiler.ts
@@ -37,7 +37,7 @@ import { Logger } from "./logging/Logger";
 import { NormalModuleFactory } from "./normalModuleFactory";
 import { WatchFileSystem } from "./util/fs";
 import { getScheme } from "./util/scheme";
-import { bindingVersionCheck } from "./util/bindingVersionCheck";
+import { checkVersion } from "./util/bindingVersionCheck";
 import Watching from "./watching";
 import { NormalModule } from "./normalModule";
 import { normalizeJsModule } from "./util/normalization";
@@ -264,7 +264,9 @@ class Compiler {
 	/**
 	 * Lazy initialize instance so it could access the changed options
 	 */
-	#getInstance(cb: (error?: Error | null, instance?: binding.Rspack) => void) {
+	#getInstance(
+		callback: (error: Error | null, instance?: binding.Rspack) => void
+	): void {
 		const processResource = (
 			loaderContext: LoaderContext,
 			resourcePath: string,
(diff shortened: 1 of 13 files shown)


packages/rspack/scripts/install.mjs
```diff
@@ -0,0 +1,17 @@
...
+	const command = `npm show ${packageName} version`;
+	return execSync(command).toString().trim();
+}
+
+const coreVersion = getVersion("@rspack/core");
+const bindingVersion = getVersion("@rspack/binding");
+
+if (coreVersion !== bindingVersion) {
+	throw new Error(
+		`Expected ${JSON.stringify(coreVersion)} but got ${JSON.stringify(
```
underfin, line 13: The msg isn't clear. Can you update to `The rspack core version xx is not match to binding version xx`. cc @hardfist

lippzhang: > The msg isn't clear. Can you update to `The rspack core version xx is not match to binding version xx`. cc @hardfist

wait a minute

hardfist:
i think runtime maybe better than postinstall check，postinstall causes problems in large monorepo

lippzhang:
> 

I think you are right, so where is the right place to put the check? When before  createCompiler() , or before compiler build?

lippzhang:
> I think you are right, so where is the right place to put the check? When before createCompiler() , or before compiler build?



> i think runtime check maybe better than postinstall check，postinstall causes problems in large monorepo，the postinstall costs a lot in monorepo if it has multi versions of rspack which we met when use esbuild in monorepo

I think you are right, so where is the right place to put the check? When before createCompiler() , or before compiler build?

lippzhang:
> i think runtime check maybe better than postinstall check，postinstall causes problems in large monorepo，the postinstall costs a lot in monorepo if it has multi versions of rspack which we met when use esbuild in monorepo

I think you are right, so where is the right place to put the check? When before createCompiler() , or before compiler build?

packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -0,0 +1,18 @@
...
+function getVersion(packageName: string): string {
+	const command = `npm show ${packageName} version`;
+	return execSync(command).toString().trim();
+}
+
+export const bindingVersionCheck = () => {
+	const coreVersion = getVersion("@rspack/core");
+	const bindingVersion = getVersion("@rspack/binding");
+	if (coreVersion !== bindingVersion) {
+		throw new Error(
```
h-a-n-a, line 12: This error should be propagated property instead of throwing here directly.

packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -0,0 +1,18 @@
+import { execSync } from "child_process";
+
+function getVersion(packageName: string): string {
+	const command = `npm show ${packageName} version`;
```
hardfist, line 4: why not require version from package.json directly, npm show is very slow here

lippzhang: > why not require version from package.json directly, npm show is very slow here

I'll change it

packages/rspack/src/util/bindingVersionCheck.ts
lippzhang: > This error should be propagated property instead of throwing here directly.

I don't really understand why， can you explain？

h-a-n-a: Errors thrown this way is what users cannot expect and it would be better to use the node callback instead of current way.

hardfist:
iam not sure we should check mismatch between core and binding or binding with binding-xx-yy，@h-a-n-a can you help check

h-a-n-a:
> I am not sure we should check mismatch between core and binding or binding with binding-xx-yy，@h-a-n-a can you help check

@hardfist This PR mainly focuses on the version consistency between binding and core, also binding and its corresponding addon. So we should put binding and its addon into consideration too. cc @lippzhang

lippzhang:
I can probably understand, but not sure, please help check it @h-a-n-a

packages/rspack/src/compiler.ts
```diff
@@ -284,10 +285,15 @@ class Compiler {
 				});
 		};
 		const options = getRawOptions(this.options, this, processResource);
-
+		const instanceBindingVersion =
+			require("@rspack/binding/package.json").version;
+		bindingVersionCheck(instanceBindingVersion, err => {
+			if (err) throw err;
```
h-a-n-a, line 291: This part should follow the callback rule too. You can change this getter to `#getInstance(callback)`

packages/rspack/src/compiler.ts
```diff
@@ -353,7 +359,7 @@ class Compiler {
 					buildModule: this.#buildModule.bind(this)
 				},
 				createThreadsafeNodeFSFromRaw(this.outputFileSystem),
-				loaderContext => runLoader(loaderContext, this)
+				(loaderContext: any) => runLoader(loaderContext, this)
```
h-a-n-a, line 362: What is this for?

lippzhang: > What is this for?

In my editor, an error will be reported  : 参数“loaderContext”隐式具有“any”类型。
Maybe it can be changed to this?

```js
(loaderContext: binding.JsLoaderContext) => runLoader(loaderContext, this)
```

h-a-n-a: Could you try `import type * as binding from "@rspack/binding";`? Since NAPI-RS generates named exports in `d.ts` file.

packages/rspack/src/stats.ts
```diff
@@ -7,7 +7,7 @@
  * Copyright (c) JS Foundation and other contributors
  * https://github.com/webpack/webpack/blob/main/LICENSE
  */
-import * as binding from "@rspack/binding";
+import type binding from "@rspack/binding";
```
h-a-n-a, line 10: `type * as binding`

h-a-n-a: Everywhere else should be the same like this too.

lippzhang: thank

packages/rspack/src/compiler.ts
lippzhang: I tried it, still getting an error. So I changed it to this
```js
(loaderContext: binding.JsLoaderContext) => runLoader(loaderContext, this)
```

packages/rspack/src/compiler.ts
lippzhang: OK，I have change it

h-a-n-a:
Would you mind me pushing some code to this PR to get this merged as soon as possible?

lippzhang:
> 



> @lippzhang Thank you for the contribution, it looks great. Would you mind me pushing some code as the final touch to this PR to get this merged as soon as possible?

don't mind, i can study

h-a-n-a:
!eco-ci

h-a-n-a:
!canary

packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -0,0 +1,142 @@
...
+		arm64: {
+			musl: "musl",
+			gnu: "gnu"
+		},
+		arm: "gnueabihf"
+	}
+};
+
+function isMusl() {
+	// For Node 10
```
hardfist, line 38: no need for rspack

packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -0,0 +1,142 @@
...
+	try {
+		const BINDING_PKG_DIR = path.dirname(
+			require.resolve("@rspack/binding/package.json")
+		);
+
+		const isLocal = readdirSync(BINDING_PKG_DIR).some(
+			item => item === `rspack.${platformArchAbi}.node`
+		);
+
+		if (isLocal) {
```
hardfist, line 110: do we really need local check, i thought monorepo already make it works like third_party node_modules

h-a-n-a: > do we really need local check, i thought monorepo already make it works like third_party node_modules

This is for local testing, such as running an example in rspack repo.

hardfist: I know this scenario, but don't we already make sure all the version are same in monorepo?

packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -0,0 +1,142 @@
...
+		if (error instanceof Error) {
+			return (result = new Error(
+				`${error.toString()}. Maybe you forget to install the correct addon package ${`@rspack/binding-${platformArchAbi}`} or forget to build binding locally?`
+			));
+		}
+		return (result = new Error(error));
+	}
+
+	const isMatch = [CORE_VERSION, BINDING_VERSION, ADDON_VERSION].every(
+		(v, i, arr) => i === 0 || (i > 0 && v === arr[i - 1])
```
hardfist, line 132: `(v,i,arr) => v === arr[0]` seems more readable

h-a-n-a: Good point!

packages/rspack/src/util/bindingVersionCheck.ts
h-a-n-a: Of course, but then we have to figure out if we are running locally, or `@rspack/binding-xxx` cannot be found.

h-a-n-a:
The main point of lazily checking is to make sure directly importing `@rspack/core` will not have any problem, even if users might not installed the corresponding `@rspack/binding-xxxx`(i.e addon), making sure the consistency is everything for treating `@rspack/core` and its dependencies as if it's a single library.

hardfist (dismissed):

h-a-n-a:
!eco-ci

h-a-n-a:
!eco-ci !canary

h-a-n-a:
!eco-ci

h-a-n-a:
!canary

jerrykingxyz (approved):

PR #3723, opened 2023-07-07
feat:  nested tree shaking

IWANABETHATGUY:

### DIFF
--- a/crates/rspack_core/src/tree_shaking/visitor.rs
+++ b/crates/rspack_core/src/tree_shaking/visitor.rs
@@ -35,10 +35,10 @@ use crate::{
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum SymbolRef {
-  Direct(Symbol),
+  Declaration(Symbol),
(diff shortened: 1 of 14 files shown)


crates/rspack_core/src/tree_shaking/visitor.rs
```diff
@@ -48,27 +48,29 @@ pub enum SymbolRef {
...
+      SymbolRef::Declaration(d) => d.src(),
       SymbolRef::Indirect(i) => i.importer,
       SymbolRef::Star(s) => s.module_ident(),
       SymbolRef::Url { importer, .. } => *importer,
+      SymbolRef::Usage(_, _, src) => *src,
     }
   }
   /// Returns `true` if the symbol ref is [`Direct`].
   ///
   /// [`Direct`]: SymbolRef::Direct
```
zackarychapple, line 70: Documentation should be updated as well correct?

IWANABETHATGUY: see https://github.com/web-infra-dev/rspack/pull/3755

Issue #2273, opened 2023-03-15
[Feature Request]: `builtins.css.modules` support `auto` config


hzura:
### What problem does this feature solve?

Currently, if you need to support both .less and .module.less files in the project, you need to configure their types separately like below. (plz forgive me if I have used them incorrectly)
```
{
  test: /\.less$/,
  use: [{
	  loader: 'less-loader',
  }],
  type: 'css/module',
},
{
  test: /^(.(?!\.module))+\.less$/,
  use: [{
	  loader: 'less-loader',
  },],
  type: 'css',
},
``` 
I hope that `builtins.css.modules` config can support option `auto` like css-loader, allowing automatic differentiation using regular expressions (as the cost of using a function may be high at the moment)
preference: https://webpack.js.org/loaders/css-loader/#modules

### What does the proposed API of configuration look like?

```
type BuiltinsCssModules = {
  localsConvention?:
    | 'asIs'
    | 'camelCase'
    | 'camelCaseOnly'
    | 'dashes'
    | 'dashesOnly';
  localIdentName?: string;
  exportsOnly?: boolean;
  auto?: boolean | RegExp;
};
```

hardfist:
waiting for https://github.com/webpack/webpack/issues/16572 resolved

evertzz:
I wonder how to support module.less, the configuration above does't work

hardfist:
since https://github.com/webpack/webpack/pull/16577 landed, we can continue supporting it now

PR #3727, opened 2023-07-07
feat(rspack_plugin_css): add css/auto same as webpack

LingyuCoder:
## Summary

According to the issue #2273, equivalent implementation with webpack pull request[ #16572](https://github.com/webpack/webpack/issues/16572)


hyf0:
Webpak doesn't think this as a breaking change. Do we need to? @hardfist

crates/rspack_plugin_css/src/parser_and_generator/mod.rs
```diff
@@ -33,6 +35,9 @@ use crate::swc_css_compiler::{SwcCssSourceMapGenConfig, SWC_COMPILER};
 use crate::utils::{css_modules_exports_to_string, ModulesTransformConfig};
 use crate::{pxtorem::px_to_rem::px_to_rem, visitors::analyze_dependencies};
 
+static REGEX_IS_MODULES: Lazy<Regex> =
+  Lazy::new(|| Regex::new(r"\.module(s)?\.[^.]+$").expect("Invalid regex"));
```
hyf0, line 39: So `module.scss` would be considered as CSS module. Does Webpack do the same thing?

ahabhgk: with css/auto, yes

LingyuCoder: > So `module.scss` would be considered as CSS module. Does Webpack do the same thing?

Webpack use RegExp in module.rules.test field to specific css files. By default, webpack does not have any rule for scss files, only css default rules in webpack/lib/config/defaults.js.

crates/rspack_plugin_css/tests/fixtures/custom/css_auto/expected/main.css
```diff
@@ -0,0 +1,6 @@
+.style {
```
ahabhgk, line 1: avoid to add tests on rust side, we lack of human maintenance and js side test is good enough in most of cases

ahabhgk (approved):

hyf0 (approved):

Issue #3544, opened 2023-06-14
builtin provider should use `ProviderDependency` to implment this


underfin:
### System Info

-

### Details

- Current implement

Using `require('xx')` to do it, but it isn't correct for esm module, please see https://github.com/web-infra-dev/rspack/issues/3203

### Reproduce Steps

-

hardfist:
bump

PR #3731, opened 2023-07-07
fix: react refresh in async entrypoint

ahabhgk:
## Summary

fix react refresh error when importing a react component in worker module

before the react refresh runtime is added by prepending additional entry, but worker chunk is a async entrypoint, so in async entrypoint will loose the react refresh runtime because you can't prepend entry for async entrypoint

now we inject the react refresh runtime by `builtins.provide` to make sure every module using react will have the react refresh runtime

## Test Plan

- `packages/playground/cases/react/worker`


underfin (approved):

crates/rspack_plugin_javascript/src/visitors/swc_visitor/react.rs
```diff
@@ -80,24 +82,146 @@ impl Visit for ReactRefreshUsageFinder {
   }
 }
 
-// __webpack_require__.$ReactRefreshRuntime$ is injected by the react-refresh additional entry
-// See https://github.com/web-infra-dev/rspack/pull/2714 why we have a promise here
-static RUNTIME_CODE: &str = r#"
-function $RefreshReg$(type, id) {
-  __webpack_modules__.$ReactRefreshRuntime$.register(type, __webpack_module__.id+ "_" + id);
+// $ReactRefreshRuntime$ is injected by provide
+fn create_react_refresh_runtime_stmts(unresolved_mark: Mark) -> Vec<Stmt> {
```
underfin, line 86: This should be revert after fixed https://github.com/web-infra-dev/rspack/issues/3544

underfin (approved):

ahabhgk:
!canary

hardfist:
@ahabhgk we should add the breaking change notice in description

PR #3733, opened 2023-07-10
chore: 🤖 enable tree shaking in bench

IWANABETHATGUY:
## Summary

1. Our option normalized logic is done by js side, so we need to enable `treeShaking` by hand on rust side benchmark.


IWANABETHATGUY:
!bench

jerrykingxyz (approved):

Issue #3726, opened 2023-07-07
[Tracking]: misaligned Webpack API


h-a-n-a:
### What subject does this issue tracking?

Tracking APIs that are not provided by Webpack but added in Rspack by mistake.

```[tasklist]
### Tasks
- [ ] compilation.normalModuleFactory
- [ ] compilation.contextModuleFactory
- [ ] compilation.currentNormalModuleHooks
- [ ] compiler.getAsset
```

PR #3735, opened 2023-07-10
refactor: clean non webpack public APIs

h-a-n-a:
## Summary

Cleaned non webpack public APIs to avoid misalignment.
Refactor the current filename to constructor style.

Related: https://github.com/web-infra-dev/rspack/issues/3726

## Test Plan

Passing the original tests.


packages/rspack/src/Compilation.ts
```diff
@@ -103,11 +103,22 @@ export class Compilation {
 	logging: Map<string, LogEntry[]>;
 	name?: string;
 	childrenCounters: Record<string, number> = {};
+	/**
+	 * Note: This is not a webpack public API, maybe removed in future.
```
Boshen, line 107: nitpick: comment on why is this needed, and the requirement for removal. Having "maybe" does not communicate well.

h-a-n-a: Leaving a phrase like "do not use" maybe too aggressive, so I used "maybe" to make the tone softer. What else would you recommend?

PR #3747, opened 2023-07-11
perf: code generation use raw source

underfin:
## Summary

The `ReplaceSource` will caused `compliation.getAsset()` performance regression, so avoid add it to code generation result.

## Test Plan

It is a perf, not a feature, the test is not necessary


underfin:
!bench

Boshen:
> It is a perf, not a feature, the test is not necessary

How did you test the performance impact?

underfin:
> > It is a perf, not a feature, the test is not necessary
> 
> How did you test the performance impact?

Look like it need hmr bench mark.

PR #3748, opened 2023-07-11
perf: convert to one RawSource in devtool to reduce JsCompatSource overhead

ahabhgk:

### DIFF
--- a/crates/rspack_plugin_devtool/src/lib.rs
+++ b/crates/rspack_plugin_devtool/src/lib.rs
@@ -14,7 +14,7 @@ use rspack_core::{
   PluginJsChunkHashHookOutput, PluginProcessAssetsOutput, PluginRenderModuleContentOutput,
   ProcessAssetsArgs, RenderModuleContentArgs, SourceType,
 };
-use rspack_error::{internal_error, Result};
+use rspack_error::{internal_error, Error, Result};
 use rspack_util::swc::normalize_custom_filename;
 use rustc_hash::FxHashMap as HashMap;
 use serde_json::json;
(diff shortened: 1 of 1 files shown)


underfin:
!bench

Boshen:
Can you describe on how did you test the perf change in your test plan?

ahabhgk:
updated

h-a-n-a:
> since we use ReplaceSource in code generation, there will store lots of Source, and **transfer all these Source** to js side when compilation.getAsset()...

What I do understand is that we converts the source to `JsCompatSource` on the Rust side. However, we only send the final result of `source` and `map` calculated on the Rust side to JS as follows: https://github.com/web-infra-dev/rspack/blob/1a3784073aa44045a73e14db4071237562fc9d9c/crates/node_binding/src/js_values/source.rs#L169.

Would you elaborate more on the performance gain?

ahabhgk:
I described wrong, this only avoid one time source map calculation when convert to `JsCompatSource`, will also update it in description

crates/rspack_plugin_devtool/src/lib.rs
```diff
@@ -99,40 +99,58 @@ impl Plugin for DevtoolPlugin {
...
       })
       .collect::<Result<_>>()?;
-    for (filename, map_buffer) in maps {
+    for (filename, (code_buffer, map_buffer)) in maps {
+      let mut asset = args
+        .compilation
+        .assets_mut()
+        .remove(&filename)
+        .expect("should have filename in compilation.assets");
+      // convert to one RawSource to reduce the JsCompatSource js <-> rust overhead
```
ahabhgk, line 147:
```suggestion
      // convert to RawSource to reduce one time source map calculation when convert to JsCompatSource
```

h-a-n-a (approved):

PR #3757, opened 2023-07-12
fix: should escape css url content

ahabhgk:
## Summary

escape css url content

`url(data:text/bad-base64;base64, abcd?#iefix)` => `url("data:text/bad-base64;base64, abcd?#iefix")`

## Test Plan

- `packages/rspack/tests/configCases/css/urls`

### DIFF
--- a/crates/rspack_plugin_css/src/dependency/url.rs
+++ b/crates/rspack_plugin_css/src/dependency/url.rs
@@ -1,3 +1,5 @@
+use once_cell::sync::Lazy;
+use regex::{Captures, Regex};
 use rspack_core::{
   CodeGenerationDataFilename, CodeGenerationDataUrl, Compilation, Dependency, DependencyCategory,
   DependencyId, DependencyTemplate, DependencyType, ErrorSpan, ModuleDependency, ModuleIdentifier,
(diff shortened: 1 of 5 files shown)


crates/rspack_plugin_css/src/dependency/url.rs
```diff
@@ -102,7 +104,44 @@ impl DependencyTemplate for CssUrlDependency {
...
 }
+
+static WHITE_OR_BRACKET_REGEX: Lazy<Regex> =
+  Lazy::new(|| Regex::new(r#"[\n\t ()'"\\]"#).expect("Invalid Regexp"));
+static QUOTATION_REGEX: Lazy<Regex> =
+  Lazy::new(|| Regex::new(r#"[\n"\\]"#).expect("Invalid Regexp"));
+static APOSTROPHE_REGEX: Lazy<Regex> =
+  Lazy::new(|| Regex::new(r#"[\n'\\]"#).expect("Invalid Regexp"));
+
+fn css_escape_string(s: &str) -> String {
```
hyf0, line 120: Where does this logic come from?

ahabhgk: https://github.com/webpack/webpack/blob/6be4065ade1e252c1d8dcba4af0f43e32af1bdc1/lib/dependencies/CssUrlDependency.js#L81

hyf0 (approved):

h-a-n-a (approved):

Issue #3557, opened 2023-06-16
Minify Configuration Options


Boshen:
### What problem does this feature solve?

I'm getting on-calls from applications minifying on already minified files. Usually the minified is chunk grouped.

### What does the proposed API of configuration look like?

One option we are missing is opt-out of minification for chunks, https://webpack.js.org/plugins/terser-webpack-plugin/#exclude

Boshen:
@LingyuCoder I updated the API proposal, it was pointing at the wrong link. It should be https://webpack.js.org/plugins/terser-webpack-plugin/#exclude

PR #3775, opened 2023-07-13
feat(core): add test/include/exclude options to builtins.minifyOptions

LingyuCoder:
## Summary

According to issue #3557 , add test/include/exclude options to builtins.minifyOptions, same as [terser-webpack-plugin](https://github.com/web-infra-dev/rspack/issues/3557)


crates/rspack_regex/src/lib.rs
```diff
@@ -10,6 +12,8 @@ mod algo;
 /// Using wrapper type required by [TryFrom] trait
 #[derive(Debug, Clone)]
 pub struct RspackRegex {
+  pub expr: String,
+  pub flags: String,
   pub algo: Algo,
```
ahabhgk, line 17: I thought out another way for hashing `Algo`, you can define a `AlgoHash` trait and implement it for `Algo`, then use the `AlgoHash` when calling `<RspackRegex as Hash>::hash`, this avoid clone caused by `to_string`

crates/rspack_core/src/options/builtins.rs
```diff
@@ -102,6 +106,97 @@ pub struct Minification {
...
+  #[async_recursion]
+  pub async fn try_match(&self, data: &str) -> rspack_error::Result<bool> {
+    match self {
+      Self::String(s) => Ok(data.starts_with(s)),
+      Self::Regexp(r) => Ok(r.test(data)),
+    }
+  }
+}
+
+impl fmt::Debug for MinificationCondition {
```
ahabhgk, line 128: you can use `#[derive(Debug, Hash)]` to avoid manually implement these trait

crates/rspack_core/src/options/builtins.rs
```diff
@@ -102,6 +106,97 @@ pub struct Minification {
...
+  fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+    match self {
+      Self::String(i) => Debug::fmt(&i, f),
+      Self::Regexp(i) => i.fmt(f),
+    }
+  }
+}
+
+impl Hash for MinificationCondition {
```
ahabhgk, line 137: Maybe we don't need Hash for this, compiler is complaining because `Minification` implemented `Hash`, but `Builtins` didn't implement it, maybe you can try to remove the `Hash` on `Minification`, not sure whether it's still useful

LingyuCoder: It is hashed in js_chunk_hash function in rspack_plugin_javascript/plugin/impl_plugin-for_js_plugin.rs, so this cannot be removed

crates/rspack_regex/src/lib.rs
LingyuCoder: Trait can not has its own fields and Algo is an enum. So I add a wrapper struct named HashRegressRegex which implement the hash function and overwrite the default Debug::fmt function.

ahabhgk (approved):

PR #3779, opened 2023-07-14
fix: avoid rspack_plugin_ensure_chunk_conditions loop

underfin:
## Summary

Avoid rspack_plugin_ensure_chunk_conditions loop.


underfin:
!canary

ahabhgk (approved):

PR #3782, opened 2023-07-17
fix: hmr should work for inject css link to body

underfin:
## Summary

The webpack css hmr runtime use `document.head` to update css but maybe the user inject css link to body, the pr is fixed it.  `mini-css-extract-plugin` do it at https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/src/index.js#L908.

## Test Plan

Added e2e

### DIFF
--- a/packages/playground/cases/css/render-to-body/index.test.ts
+++ b/packages/playground/cases/css/render-to-body/index.test.ts
@@ -1,9 +1,16 @@
 import { test, expect } from "@/fixtures";
 
-test("should update body css", async ({ page, fileAction }) => {
+test("should update body css", async ({ page, fileAction, rspack }) => {
 	await expect(page.locator("body")).toHaveCSS("display", "block");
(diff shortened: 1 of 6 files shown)


hyf0 (dismissed):

packages/playground/cases/css/render-to-body/index.test.ts
```diff
@@ -0,0 +1,9 @@
+import { test, expect } from "@/fixtures";
+
+test("should update body css", async ({ page, fileAction }) => {
+	await expect(page.locator("body")).toHaveCSS("display", "block");
+	fileAction.updateFile("src/index.css", content =>
+		content.replace("block", "flex")
+	);
+	await expect(page.locator("body")).toHaveCSS("display", "flex");
```
jerrykingxyz, line 8: should call rspack.waitingForHmr here

hyf0 (approved):

PR #3786, opened 2023-07-17
fix(core): Typo in html builtins types

tobua:
## Summary

Fixes what looks like an accidental typo in `builtins.html` types generated with zod. The name of `publicPath` instead is `publicPathz`. The [official documentation](https://www.rspack.dev/config/builtins.html#builtinshtml) correctly lists `publicPath`.

## Test Plan

I've discovered the issue in a project. `publicPath` is still the only working option, it's just the types that are wrong, no need to test anything.


hyf0 (approved):

hyf0:
Thank you!

PR #3789, opened 2023-07-18
fix: tree shaking with a cyclic reference 

IWANABETHATGUY:
## Summary
1. Fixed symbol reference when the `ExportAll` Symbol used as an intermediate symbol and connected to an `ImportAllAs` at the same time.


h-a-n-a:
'Cyclic' is a more plausible wording.

IWANABETHATGUY:
> 

nice catch!

hyf0 (approved):

PR #3790, opened 2023-07-18
fix: check symbol used

underfin:
## Summary

Before symbol used check by checking every import symbol used, if they are not used, so every symbol is not used, it isn't correct, the pr is fix it. 

## Test Plan

Added.

### DIFF
--- a/crates/rspack/tests/tree-shaking/class-extend/expected/main.js
+++ b/crates/rspack/tests/tree-shaking/class-extend/expected/main.js
@@ -3,18 +3,30 @@
 'use strict';
 __webpack_require__.r(__webpack_exports__);
 __webpack_require__.d(__webpack_exports__, {
-  'b': function() { return b; }
+  'v': function() { return v; }
 });
+/* harmony import */var _lib__WEBPACK_IMPORTED_MODULE_0_ = __webpack_require__(/* ./lib */"./lib.js");
 
-var __WEBPACK_DEFAULT_EXPORT__ = /* "./lib" unused */null;
- const b = 1;
+ class Lib extends /* "./lib" unused */null {
+}
+ const v = _lib__WEBPACK_IMPORTED_MODULE_0_["value"];
 },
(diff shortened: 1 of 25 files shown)


crates/rspack/tests/tree-shaking/class-extend/expected/main.js
```diff
@@ -0,0 +1,37 @@
+(self['webpackChunkwebpack'] = self['webpackChunkwebpack'] || []).push([["main"], {
+"./app.js": function (__unused_webpack_module, __webpack_exports__, __webpack_require__) {
+'use strict';
+__webpack_require__.r(__webpack_exports__);
+__webpack_require__.d(__webpack_exports__, {
+  'v': function() { return v; }
+});
+/* harmony import */var _lib__WEBPACK_IMPORTED_MODULE_0_ = __webpack_require__(/* ./lib */"./lib.js");
+
+ class Lib extends /* "./lib" unused */null {
```
underfin, line 10: The webpack used `null` instead of `undefined`, because it can be class extends.

crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
```diff
@@ -16,17 +17,47 @@ use swc_core::{
...
+    Self {
+      request,
+      specifier,
+      names,
+    }
+  }
+}
+
+pub type ImportMap =
+  FxHashMap<(JsWord /* ident */, SyntaxContext /* ctxt */), ImporterReferenceInfo>;
```
IWANABETHATGUY, line 42: IIRC, swc has a type alias represented this tuple type.

IWANABETHATGUY: https://github.com/web-infra-dev/rspack/blob/927278e367e87012018b9957ef2f233efd970367/crates/rspack_symbol/src/lib.rs#L4-L4

IWANABETHATGUY (approved):

PR #3794, opened 2023-07-18
fix: compilation hash should hash hot index to avoid hmr loop

underfin:
## Summary

Because the hash of `[hash].hot-update.js/json` is previous compilation hash, the diffrent compilation hash maybe at same `.hot-update.js` file, it will cause hmr loop if you cached `.hot-update.js`, because the compilation hash isn't correct. For detail:
Status A(hash: A) -> Status B(hash: B) will generate `A.hot-update.js`
Status A(hash: A) -> Status C(hash: C) will generate `A.hot-update.js`
The status is different, should generate different hash for `.hot-update.js`, you can see webpack at https://github.com/webpack/webpack/blob/main/lib/HotModuleReplacementPlugin.js#L425.
So use compilation hash update `hot_index` to fix it.

## Test Plan

Added.


underfin:
!canary

ahabhgk (approved):

h-a-n-a:
Is this related to https://github.com/web-infra-dev/rspack/issues/3554?

underfin:
> Is this related to #3554?

Maybe it is, let user have a try?

hyf0 (approved):

PR #3797, opened 2023-07-19
fix: shorthand filed symbol unsed

underfin:
## Summary

Before will replace shorthand filed to null, it will throw `Cannot use a reserved word as a shorthand property`.

## Test Plan

Added


underfin:
!canary

ahabhgk (approved):

PR #3800, opened 2023-07-19
feat: nested import tree shaking

IWANABETHATGUY:
## Summary
1. Before we only support namespace `ImportNamespace` nested level tree shaking. e.g.
```
// index.js
import * as a from './a.js'
a.b

// a.js
export const b = 3;
export const c = 4;
```
We know that a represent all export in `a.js`, any `MemberExpress` access means only `property` element in `a.js` module is referenced, otherwise, you use the whole namespace.
We did the transform logic(transform a namespace access into a normal `ImportDeclaration`, transform `a.b` into `import {b} from './a.js'`, just demonstration, we did not transform the AST in implementation ) in the par module analyze phase, so we can't determine if the imported binding represents a namespace-like binding. e.g.
```js
// index.js
import {a} from './a.js'
a.b

// a.js
export * as a from './c.js'
export const a = 'aaaa'
// c.js
export const b = ''
export const c = ""
```
in the module analysis phase, we have no idea what the `a` is in `a.js` when analyze `index.js`, it might be a `ExportNamedDeclaration` or 
`ExportAllDeclaration`.For safety, we treat all export as used in `c.js`.
2. In this pull request, we defer the transform logic until the graph traversal, so we have the information on what the `a` in `a.js` actually is. 
3. Adding a new symbol `Usage` it takes a `MemberExpressionChainArray`, simply demonstration the algorithm using the example above. Here is the 
**simplified graph representation**
 ```
0: Usage(["a", "b"], 'a.js') 
1: ImportNamedDeclaration("a", "a.js")
2:  ExportAllAs("a", "c.js")
3: ExportNamedDeclaration("a", "a.js")
4: ExportNamedDeclaration("b", "c.js")
5: ExportNamedDeclaration("c", "c.js")

0 -> 1
1 -> 2
2 -> 4
2 -> 5
```
start from `0`, referenced an indirect `ImportNamedDeclaration`, continue `DFS` without pop our path until we meet a real `Export`, node `1` only has one edge connected to `2`, it is a `export` and matches the first chain element `a` so we pop the first element and try if we can go further, node `2` has two edges connected `4` and `5` and node `4` matches the `b`, the whole algo is done. Finally, we know that the `MemberExpression` usage actually referenced `b` in `c.js` so the rest export can be eliminated. The artifacts look something like this: 
```js
// index.js
import {a} from './a.js'
a.b

// a.js
export * as a from './c.js'
// c.js
export const b = ''
```


IWANABETHATGUY:
!bench

crates/rspack/tests/tree-shaking/export_star/expected/main.js
```diff
@@ -5,9 +5,6 @@ __webpack_require__.r(__webpack_exports__);
 __webpack_require__.d(__webpack_exports__, {
   'bar': function() { return _foo__WEBPACK_IMPORTED_MODULE_0_; }
 });
-__webpack_require__.d(__webpack_exports__, {
-  'b': function() { return b; }
-});
 /* harmony import */var _foo__WEBPACK_IMPORTED_MODULE_0_ = __webpack_require__(/* ./foo */"./foo.js");
```
IWANABETHATGUY, line 8: These two var is unused, it is nested export before.

crates/rspack/tests/tree-shaking/reexport-all-as/expected/main.js
```diff
@@ -6,6 +6,7 @@ __webpack_require__.d(__webpack_exports__, {
   'a': function() { return a; }
 });
  function a() {}
+ function dddd() {}
```
IWANABETHATGUY, line 9: We added a extra variable, and only used declaration `a` , https://github.com/web-infra-dev/rspack/pull/3800/files#diff-88e9e2f5deab4237bd10fad2b7a74adc310be0e3b4e1a902f29c59e5f0f40687R34, so the `dddd` is not exported as expected.

underfin (approved):

PR #3803, opened 2023-07-19
feat: built in progress plugin support profile option

faga295:
## Summary
Based on of original progress plugin, I integrate the original logic to `progress_handler`, and add a new `default_handler` to print the profile of progress.In the `update_throttled` function, I chose `50ms` as throttle time, in webpack this time is `500ms`, maybe this time should discuss.
image


crates/rspack_plugin_progress/src/lib.rs
```diff
@@ -32,16 +42,101 @@ impl ProgressPlugin {
...
+            let diff = now - last_state_info[i].time;
+            let report_state = if i > 0 {
+              last_state_info[i - 1].value.clone()
+                + " > "
+                + last_state_info[i].value.clone().as_str()
+            } else {
+              last_state_info[i].value.clone()
+            };
+            println!(
+              "{} {}ms {}",
```
ahabhgk, line 115:
```suggestion
              "{} {} ms {}",
```

crates/rspack_binding_options/src/options/raw_builtins/raw_progress.rs
```diff
@@ -7,12 +7,14 @@ use serde::{Deserialize, Serialize};
 #[napi(object)]
 pub struct RawProgressPluginConfig {
   pub prefix: Option<String>,
+  pub profile: Option<bool>,
```
ahabhgk, line 10: Would be better to change its type to `bool`, and apply the default value `false` at js side (`boolean | undefined` -> default `false` -> `boolean` -> send to rust side by -> `bool`)

Actually `prefix` should be the same, its type should be `String`.

This is not a change request so I will still accept this PR if you don't change this, and it would be very appreciate if you could refactor this in a following PR 😃

crates/rspack_plugin_progress/src/lib.rs
```diff
@@ -54,54 +149,40 @@ impl Plugin for ProgressPlugin {
...
+        .write()
+        .expect("TODO:")
+        .replace(module.raw_request().to_string());
     }
     self.modules_count.fetch_add(1, SeqCst);
+    self.update_throttled();
     Ok(())
   }
-
-  async fn succeed_module(&self, _module: &dyn Module) -> Result<()> {
```
ahabhgk, line 84: Why delete these? I know that webpack didn't hook succeedModule in ProgressPlugin, maybe we can change this to finishMake

faga295: build_module and succeed_module has little difference. What do you mean about `finishMake`?

faga295: @ahabhgk Can you resolve my confusion if you are free?

ahabhgk: Any reason why delete these?

faga295: Since build_module used to be used for `set_message` and `succeed_module` used to be used for `set_position`, now that it's been merged into `update`, `succeed_module` seems unnecessary.

hardfist:
@faga295 are you still working on this?

faga295:
Looks like it works now, but  whether keep `succeed_module` still confused me

ahabhgk (approved):
LGTM, thanks!

PR #3804, opened 2023-07-20
fix: fix css semicolon replacement

### LINKED ISSUE #3798: Generate invalid CSS when building Vue 3 app
### System Info

  System:
    OS: macOS 13.4.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 72.09 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 1.22.19 - ~/Library/Application Support/fnm/node-versions/v16.19.1/installation/bin/yarn
    npm: 8.19.3 - ~/Library/Application Support/fnm/node-versions/v16.19.1/installation/bin/npm
    pnpm: 7.18.1 - ~/Library/Application Support/fnm/node-versions/v16.19.1/installation/bin/pnpm
  Browsers:
    Chrome: 114.0.5735.198
    Safari: 16.5.1
  npmPackages:
    @rspack/cli: 0.2.7 => 0.2.7

### Details

The CSS generated by the Rspack build contains redundant `;` characters, causing the following style to not take effect:

image

This issue only appears when I using vue-loader and import css files in the `index.js` file.

### Reproduce link

https://github.com/chenjiahan/rspack-vue3-style-issue

### Reproduce Steps

1. pnpm i
2. pnpm dev
3. open http://localhost:8080/
4. Check the background color of the page.

h-a-n-a:
## Summary

`@import "foo.css";` will currently be replaced with `;`. 
The expected behavior is to replace the whole at-import statement with empty string.

closes #3798

## Test Plan

Checkout the E2E test `packages/playground/cases/css/import-double-css-files-with-empty/index.test.ts`


underfin (approved):

h-a-n-a:
!canary

underfin (approved):

ahabhgk:
I'm also interested because before we introduce string replacement dependency, the sources don't have entry.css, so I debug into it and found it was caused by this SourceMap

Screenshot 2023-07-21 at 12 20 18

it should be an empty SourceMap, it might be upgrading swc made the SourceMap not empty anymore

since this not affect for debugging with source map so the result is also acceptable/correct, this PR is LGTM

h-a-n-a:
> I'm also interested because before we introduce string replacement dependency, the sources don't have entry.css, so I debug into it and found it was caused by this SourceMap
> 
> Screenshot 2023-07-21 at 12 20 18
> it should be an empty SourceMap, it might be upgrading swc made the SourceMap not empty anymore
> 
> since this not affect for debugging with source map so the result is also acceptable/correct, this PR is LGTM

@ahabhgk If I understand correctly, the replacements of the `ReplaceSource` will not directly result in the `map.sources` and `map.sourcesContent` in the final result? Although the generated code does not output any related mappings, the size of the map file might be largely increased. Nevertheless, it does not introduce any semantic difference and I was asking this out of curiosity and client performance.

ahabhgk:
Yes, ReplaceSource only affect names and mappings

In real cases the size only increase a little and source map shouldn't send to client in production, so it's acceptable to me

h-a-n-a:
!canary

h-a-n-a:
> Yes, ReplaceSource only affect names and mappings
> 
> In real cases the size only increase a little and source map shouldn't send to client in production, so it's acceptable to me

Semantic wise, yes. If it's a problem with webpack, then it's ok to me.

PR #3809, opened 2023-07-20
fix: banner shouldn't be injected to asset

hardfist:
## Summary
fix: https://github.com/web-infra-dev/rspack/issues/3810

rspack's wrong behavior of injecting banner into asset is because rspack doesn't distinguish between auxiliaryFiles and files.


9aoy:
image
need distinguish auxiliaryFiles and files.

https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/Compilation.js#L4643

9aoy:
!canary

9aoy:
!bench

PR #3822, opened 2023-07-21
fix: original source should respect to the module type

### LINKED ISSUE #3795: `OriginalSource` of `NormalModule` should respect its original type
> Due to the fact that the type of `stats.source` has already changed to `Option<&dyn Source>`, you can now leverage `to_js_compat_source` to convert it `JsCompatSource`, however, as what we want is an `Option` with a union type. You should again implement conversion logic in Rust.
> 
> This logic is almost the same as here:
> 
> https://github.com/web-infra-dev/rspack/blob/02f313b19605503663d3bafb4d52484e4e5bbe96/packages/rspack/src/util/createSource.ts#L7-L16
> 
> .
> ```rust
> if js_compat_source.is_raw && js_compat_source.is_buffer {
>    // convert to buffer
> }
> // convert to string
> ```

is_raw && is_buffer condition is always true，Why is that
https://github.com/web-infra-dev/rspack/pull/3682/files#diff-24cf9834c453e48f48a49ce682855a67ec4bf939e01ad4af22356b03d6dbdae8R101-R117

_Originally posted by @suxin2017 in https://github.com/web-infra-dev/rspack/pull/3682#discussion_r1262216731_

Currently, the `create_source` does not separate the difference between `Buffer` and `String`. See: https://github.com/web-infra-dev/rspack/pull/3682#discussion_r1266525208

Furthermore, we should add a content type(`String` or `Buffer`) from loader to support this in the first place.

h-a-n-a:
## Summary

 > Due to the fact that the type of `stats.source` has already changed to `Option<&dyn Source>`, you can now leverage `to_js_compat_source` to convert it `JsCompatSource`, however, as what we want is an `Option` with a union type. You should again implement conversion logic in Rust.
> 
> This logic is almost the same as here:
> 
> https://github.com/web-infra-dev/rspack/blob/02f313b19605503663d3bafb4d52484e4e5bbe96/packages/rspack/src/util/createSource.ts#L7-L16
> 
> .
> ```rust
> if js_compat_source.is_raw && js_compat_source.is_buffer {
>    // convert to buffer
> }
> // convert to string
> ```

is_raw && is_buffer condition is always true，Why is that
https://github.com/web-infra-dev/rspack/pull/3682/files#diff-24cf9834c453e48f48a49ce682855a67ec4bf939e01ad4af22356b03d6dbdae8R101-R117

_Originally posted by @suxin2017 in https://github.com/web-infra-dev/rspack/pull/3682#discussion_r1262216731_

closes #3795
This follows up this comment https://github.com/web-infra-dev/rspack/pull/3682#discussion_r1262216731

## Test Plan

See the snapshot change.

Asset and Wasm related modules should output `Buffer`, and others should output `String` in stats.


packages/rspack/tests/configCases/source-map/cheap/index.js
```diff
@@ -5,7 +5,7 @@ require("./index.scss");
 it("should only map transformed lines if cheap options is used", async () => {
 	const source = fs.readFileSync(__dirname + "/main.css.map", "utf-8");
 	const map = JSON.parse(source);
-	expect(map.sources).toContain("./index.scss");
+	expect(map.sources.some(s => s.includes("./index.scss"))).toBeTruthy();
```
h-a-n-a, line 8: The item of `sources` changes from `"./index.scss"` to `"../../../../node_modules/sass-loader/dist/cjs.js!./index.scss"` and I personally think this is working as intended to mark the processed loaders. Any comments? @ahabhgk

ahabhgk: lgtm

ahabhgk (approved):

PR #3831, opened 2023-07-24
fix: modify Configuration type to support MultiRspackOptions

### LINKED ISSUE #3607: defineConfig MultiplyCompiler Type not export.
### System Info

- 

### Details

MultiplyCompiler Type did not export, caused ts error. 

### Reproduce link

_No response_

### Reproduce Steps

You can use a arrray to test type.

``` .ts
// rspack.config.ts
module.exports = [], // expect array shouldn't cause ts error.
```

suica:
## Summary

Fixes #3607.

Rspack config files should support MultiRspackOptions. But for the `Configuration` type, only "single" `RspackOptions` are used.

This PR changes the definition of `Configuration` type in `packages/rspack/src/config/types.ts` to add support to MultiRspackOptions. After this, we expect following `rspack.config.js` will emit no error:

```js
// @ts-check
/** @type {import('@rspack/cli').Configuration} */
const test = [];
module.exports = test;
```

## Test Plan

Added several test cases to ensure such configuration can pass type checking.

### DIFF
--- a/packages/rspack/tests/configCases/type/multiple-options/index.js
+++ b/packages/rspack/tests/configCases/type/multiple-options/index.js
@@ -1,0 +1,1 @@
+it("should pass type checking", function () {});
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -10,7 +10,8 @@
     "build": "tsc -b ./tsconfig.build.json",
     "prepare": "pnpm precompile-schema",
     "dev": "tsc -w",
-    "test": "node --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --runInBand --logHeapUsage",
+    "test": "pnpm test:tc && node --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --runInBand --logHeapUsage",
+    "test:tc": "tsc -p tsconfig.test.json",
     "precompile-schema": "node ./scripts/precompile-schema.js"
   },
   "files": [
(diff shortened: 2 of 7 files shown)


suica:
@IWANABETHATGUY Tried to fix failed CI and passed all tests on my machine. Could you please trigger it again?

packages/rspack/tests/configCases/type/multiple-options/index.js
```diff
@@ -0,0 +1 @@
+it("should pass type checking", function () {});
```
underfin, line 1: The type checking test is not running, you can remove it because rspack not test export type at now.

suica: Thank your for your review! This file is added deliberately as a placeholder. If it had been removed, the test would fail.
And btw, I added `test:tc` command to `package.json` to execute pure type checking for relevent config files (currently, as specified in `tsconfig.test.json`, only one file will be checked). So we are able to test the types exported by rspack now.

underfin (approved):

ahabhgk:
This PR breaks [nx test suite in rspack-ecosystem-ci](https://github.com/web-infra-dev/rspack-ecosystem-ci/actions/runs/5757255254/job/15608043623)

The proper way for applying type for multi options is `/** @type {ReadonlyArray<import('webpack').RspackOptions> & import('webpack').MultiCompilerOptions} */`, and in rspack `MultiRspackOptions` is an alias for `ReadonlyArray<RspackOptions> & MultiCompilerOptions`, so you can use `/** @type {import('@rspack/cli').MultiRspackOptions} */` as a shorter

So I think we should revert this, cc @underfin

Issue #3649, opened 2023-06-29
Move tree shaking ast analyzer to module build


underfin:
### System Info

-

### Details

The tree shaking needs ast analyze after the module build, it can be moved to the module build prase. 

### Reproduce Steps

-

PR #3839, opened 2023-07-25
chore: use dep_id instead of module id

IWANABETHATGUY:
## Summary
1. Close https://github.com/web-infra-dev/rspack/issues/3649
2. In the make phase, the module graph does not have detailed info on each module, we can't get the module id of each dependencies. So we use `""` by default, in analyze phase transform dependency id into module id. 

4. 

## Test Plan
1. Tests should pass.

### DIFF
--- a/crates/rspack_core/src/tree_shaking/visitor.rs
+++ b/crates/rspack_core/src/tree_shaking/visitor.rs
@@ -1,6 +1,4 @@
-use std::{
-  collections::hash_map::Entry, collections::VecDeque, hash::Hash, path::PathBuf, sync::Arc,
-};
+use std::{collections::hash_map::Entry, collections::VecDeque, hash::Hash, path::PathBuf};
 
 use bitflags::bitflags;
 use hashlink::{LinkedHashMap, LinkedHashSet};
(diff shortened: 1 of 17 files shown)


crates/rspack_core/src/tree_shaking/visitor.rs
```diff
@@ -536,24 +578,24 @@ impl<'a> Visit for ModuleRefAnalyze<'a> {
                 _ => object,
               },
               Part::Url(src) => {
-                let (src_module_id, dep_id) = self
+                let dep_id = self
                   .resolve_module_identifier(src, &DependencyType::NewUrl)
                   .unwrap_or_else(|| panic!("Can't resolve {} in {}", src, self.module_identifier));
 
                 return HashSet::from_iter([SymbolRef::Url {
                   importer: self.module_identifier,
```
IWANABETHATGUY, line 586: Here is why we can't use `Dependency` id directly, because we can't get the dependency id of the module id, one module id may correspond to many dependency id.

underfin (approved):

underfin:
!bench

underfin (approved):

PR #3843, opened 2023-07-26
feat: add some exports info api

underfin:
## Summary

- add exports info api
- add `dependency.resource_identifier()` to merge equal requests, and implement https://github.com/webpack/webpack/blob/main/lib/Compilation.js#L1621
- add `HarmonyImportSpecifierDependency` to dependency instead of  presentational_dependencies

## Test Plan

It's unnecessary at now,because exports info  is not enabled


underfin:
!bench

underfin:
!bench

PR #3852, opened 2023-07-27
refactor: css-import span should be calculated with at-rule

### LINKED ISSUE #3813: `CssImportDependency` should follow the span location of `AtRule`
### System Info

Not related

### Details

Currently, we calculate the import span with `ImportPrelude`, which should be `AtRule`

### Reproduce link

_No response_

### Reproduce Steps

Not related

h-a-n-a:
## Summary

Css-import replacement was originally implemented with dependency template. 
The implementation is supposed to remove the css-import statement in order to do module concatenation.
This PR refactors the current span calculation workaround into a more reasonable way which leverages the at-rule span.

closes #3813

## Test Plan

This is a refactor, so tests has been added before(using snapshots).

### DIFF
--- a/crates/rspack_plugin_css/src/visitors/mod.rs
+++ b/crates/rspack_plugin_css/src/visitors/mod.rs
@@ -2,7 +2,10 @@ use once_cell::sync::Lazy;
 use regex::Regex;
 use rspack_core::{ModuleDependency, SpanExt};
 use rspack_error::{Diagnostic, DiagnosticKind};
-use swc_core::css::ast::{ImportHref, ImportPrelude, Stylesheet, Url, UrlValue};
+use swc_core::common::Span;
+use swc_core::css::ast::{
+  AtRule, AtRuleName, ImportHref, ImportPrelude, Stylesheet, Url, UrlValue,
+};
 use swc_core::css::visit::{Visit, VisitWith};
 
 use crate::{
@@ -21,6 +24,7 @@ pub fn analyze_dependencies(
     deps: Vec::new(),
     code_generation_dependencies,
     diagnostics,
+    nearest_at_import_span: None,
     // in_support_contdition: false,
   };
   ss.visit_with(&mut v);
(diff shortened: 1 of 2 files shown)


ahabhgk:
I'm not sure, @underfin told me it's caused by the span of SWC provided is wrong, should we fix it in SWC?

h-a-n-a:
> I'm not sure, @underfin told me it's caused by the span of SWC provided is wrong, should we fix it in SWC?

@ahabhgk  SWC is working as expected. He has already closed that issue.

crates/rspack_plugin_css/src/visitors/mod.rs
```diff
@@ -54,7 +59,18 @@ fn replace_module_request_prefix(specifier: String, diagnostics: &mut Vec<Diagno
 }
 
 impl Visit for Analyzer<'_> {
+  fn visit_at_rule(&mut self, n: &AtRule) {
+    if let AtRuleName::Ident(ident) = &n.name && &*ident.value == "import" {
+      self.nearest_at_import_span = Some(n.span);
```
hyf0, line 64: This feels weird to me. I think it's better to perform an up-bottom looking up. 

```rust
if let AtRule { prelude: Some(box AtRulePrelude::ImportPrelude(import_predude)), .. } = n {
 // ...
}
```

h-a-n-a: Trying to decouple with `import-prelude` here. For the perspective of `import-prelude`, it should be put into another visit function, which is the same as `support` condition.

h-a-n-a: Hoisting the `import-prelude` function will make this visit function a mess because we have to support `support-condition`, etc in the future.

hyf0: I'm not sure. I don't know how the future goes. But for the current implementation, I'm gonna reserve my opinion on this. An up-bottom looking up is better.

h-a-n-a: Any thoughts on this? @ahabhgk

underfin (approved):

PR #3855, opened 2023-07-28
feat: implement dependency resource_identifier

underfin:
## Summary

It used as an identifier to merge equal requests, it can reduce resolve same request for dependency.

## Test Plan

It should be tested for exits test.

### DIFF
--- a/packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
+++ b/packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
@@ -483,32 +483,6 @@ exports[`StatsTestCases should print correct stats for ignore-plugin 1`] = `
       "chunks": [
         "main",
       ],
-      "id": "665",
-      "identifier": "missing|<PROJECT_ROOT>/tests/statsCases/ignore-plugin/locals/./zh",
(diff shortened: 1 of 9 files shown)


hyf0 (dismissed):
LGTM. @h-a-n-a cc

crates/node_binding/binding.d.ts
```diff
@@ -976,11 +969,3 @@ export interface RawTrustedTypes {
   policyName?: string
 }
 
-export interface ThreadsafeNodeFS {
-  writeFile: (...args: any[]) => any
-  removeFile: (...args: any[]) => any
-  mkdir: (...args: any[]) => any
-  mkdirp: (...args: any[]) => any
-  removeDirAll: (...args: any[]) => any
-}
```
Boshen, line 985: Why is this removed again.

packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
```diff
@@ -580,7 +554,6 @@ exports[`StatsTestCases should print correct stats for ignore-plugin 1`] = `
 exports[`StatsTestCases should print correct stats for ignore-plugin 2`] = `
 "./index.js [10] {main}
 ./locals/en.js [208] {main}
-<PROJECT_ROOT>/tests/statsCases/ignore-plugin/locals/./zh (missing) [665] {main}
```
Boshen, line 583: Was this a bug before? Why is this removed?

crates/node_binding/binding.d.ts
hyf0: napi-rs currently could not generate a stable `.d.ts` file.

underfin: It's my bad.

packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
underfin: This is a same request for next line `<PROJECT_ROOT>/tests/statsCases/ignore-plugin/locals/./zh.js`

h-a-n-a (approved):

hyf0 (approved):

PR #3859, opened 2023-07-28
perf: reduce dependency clone at make

underfin:


underfin:
!bench

hyf0 (approved):
LGTM. But please write the summary and test part. This is very important for someone doesn't know the content.

PR #3863, opened 2023-07-30
fix: type error

xiaoxiangmoe:
## Summary

There are many type errors in the current code that prevent the tsc check of js from being enabled. This PR uses a lot of `//@ts-expect-error` to bypass type checking errors.

### DIFF
--- a/packages/rspack-dev-server/src/middleware.ts
+++ b/packages/rspack-dev-server/src/middleware.ts
@@ -27,11 +27,15 @@ export function getRspackMemoryAssets(
 		}
 
 		// css hmr will append query string, so here need to remove query string
+		// @ts-expect-error
 		const path = parse(url).pathname;
 		// asset name is not start with /, so path need to slice 1
+		// @ts-expect-error
 		const filename = path.startsWith(publicPath)
-			? path.slice(publicPath.length)
-			: path.slice(1);
+			? // @ts-expect-error
+			  path.slice(publicPath.length)
+			: // @ts-expect-error
+			  path.slice(1);
 		const buffer =
 			compiler.getAsset(filename) ??
 			(() => {
@@ -50,6 +54,7 @@ export function getRspackMemoryAssets(
 			contentType = "text/html; charset=utf-8";
 		} else {
 			contentType =
+				// @ts-expect-error
 				mime.contentType(extname(path)) || "text/plain; charset=utf-8";
 		}
 
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -8,7 +8,6 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
-      "types": "./dist/index.d.ts",
       "default": "./dist/index.js"
     },
     "./package.json": "./package.json"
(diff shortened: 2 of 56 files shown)


xiaoxiangmoe:
cc @hardfist

hyf0:
@web-infra-dev/rspack-dx please take a look.

xiaoxiangmoe:
？

hardfist:
bump

hardfist:
since lots of conflicts here, i will try to fix it in separate PR

packages/rspack/package.json
```diff
@@ -35,6 +35,7 @@
     "@rspack/plugin-node-polyfill": "workspace:^",
     "@types/watchpack": "^2.4.0",
     "@types/webpack": "5.28.3",
+    "@types/webpack-dev-server": "^4.7.2",
```
bvanjoi, line 38: why `@types/webpack-dev-server` is needed here?

packages/rspack/tsconfig.json
```diff
@@ -1,12 +1,13 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    // TODO: Remove skipLibCheck when fix upstream type bug
+    "skipLibCheck": true,
     "outDir": "dist",
-    "rootDir": "src",
-    "allowJs": true,
-    "strict": true
```
bvanjoi, line 7: I had noticed `strict` had been append to `tsconifig.base.json`, so this delete is expected.

packages/rspack-dev-server/src/middleware.ts
```diff
@@ -27,7 +27,7 @@ export function getRspackMemoryAssets(
 		}
 
 		// css hmr will append query string, so here need to remove query string
-		const path = parse(url).pathname;
+		const path = parse(url!).pathname!;
```
bvanjoi, line 30: Could you use `@ts-expect-error` rather than `!` ?

packages/rspack/package.json
xiaoxiangmoe: see https://github.com/web-infra-dev/rspack/blob/3c082533ac76e1f2c2a0e0c5ed7955545941a87a/packages/rspack/src/config/zod.ts#L5

bvanjoi:
This pr merely introduces some type checkers and does not affect the existing logic. Therefore, I think it can be merged once the ci is happy

xiaoxiangmoe:
@bvanjoi Can you tell me how to fix ci error?

bvanjoi:
The failure occurred at `﻿rspack-plugin-html`. Could you please verify if this issue can be reproduced locally? You can do this by running the following commands: `﻿npm run build:cli:release && cd packages/rspack-plugin-html && npm run test`.

xiaoxiangmoe:
This issue can be reproduced locally. But I don't know how to fix it. Can you give me some idea about this?

bvanjoi:
I suspect that the issue stems from the `﻿rspack-plugin-html/dist`. Could you revert the changes made to its `﻿tsconfig.json`, and rebuild, then retest it?

bvanjoi:
To be more specific, I believe changing the ﻿`module` to `﻿commonjs` and `﻿moduleResolution` to `﻿node` should resolve this issue.

hardfist:
it seems hits the https://github.com/nodejs/node/issues/35889 bug again!

hardfist (dismissed):

tsconfig.base.json
```diff
@@ -1,8 +1,8 @@
 {
   "compilerOptions": {
-    "module": "CommonJS",
+    "module": "NodeNext",
```
bvanjoi, line 3: I recommend not modifying the output here, as it might lead to unpredictable behavior.

hardfist: since CI passed, let's merge it and if we find problem we can revert this change

PR #3877, opened 2023-08-01
perf(css): only re-parse source code while compiling css modules

hyf0:
## Summary

There's no need to do the parse-codegen process, if the processing module is not considered as css modules.

## Test Plan

Not needed.


hyf0:
!bench

ahabhgk (approved):

ahabhgk (approved):

PR #3880, opened 2023-08-01
refactor: parse result remove ast

underfin:
## Summary

Since the tree shaking ast analyze already moved to make prase, so the ast is unnecessary.

## Test Plan

It isn't need add test.


underfin:
!bench

IWANABETHATGUY (approved):

hyf0 (approved):

h-a-n-a (approved):

PR #3884, opened 2023-08-02
fix(core): add resove failing reason for import in esm file

### LINKED ISSUE #3883: provide error hint about resolve failed when resolve from esm file

webpack give hints about why resolve failed when resolve from esm file which is helpful for debugging this resolve issue
eb367302-6d2b-42f9-aa45-3ac37592158c

rspack should provide similar hint for users

No API Needed

hardfist:
## Summary
 This PR fix #3883 and add warning message for user if resolve failed when import in esm file and also add note support for diagnostic for add extra suggestion info about error message
* before
image

* after
image

## Test Plan
 Add a test in Errors.test.ts


crates/rspack_error/src/emitter.rs
```diff
@@ -175,34 +175,32 @@ fn emit_diagnostic<T: Write + WriteColor>(
...
-
-    let config = codespan_reporting::term::Config::default();
-
-    term::emit(writer, &config, files, &diagnostic).expect("TODO:");
-  } else {
-    let color = match diagnostic.severity {
-      crate::Severity::Error => Color::Red,
-      crate::Severity::Warn => Color::Yellow,
-    };
-    writer.set_color(ColorSpec::new().set_fg(Some(color)))?;
```
hardfist, line 203: @IWANABETHATGUY  I don't understand why use separate emitter logic here and think it would be better unify

h-a-n-a:
Looks good cc @ahabhgk

packages/rspack/tests/Errors.test.ts
```diff
@@ -0,0 +1,153 @@
...
+it("should emit warnings for resolve failure in esm", async () => {
+	await expect(
+		compile({
+			entry: "./resolve-fail-esm"
+		})
+	).resolves.toMatchInlineSnapshot(`
+		Object {
+		  "errors": Array [
+		    Object {
+		      "formatted": "error[internal]: error[internal]: Failed to resolve ./answer in javascript/esm|<cwd>/tests/fixtures/errors/resolve-fail-esm/index.js\\n\\n = Did you mean 'answer.js'?\\n   BREAKING CHANGE: The request './answer' failed to resolve only because it was resolved as fully specified\\n   (probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '\\"type\\": \\"module\\"').\\n   The extension in the request is mandatory for it to be fully specified.\\n   Add the extension to the request.\\n\\n",
```
hardfist, line 145: the error[internal] is concatenated multi times, I'm not sure where it happened?

crates/rspack_error/src/emitter.rs
IWANABETHATGUY: The color is lost according to the screenshot.

packages/rspack/tests/Errors.test.ts
ahabhgk: > `"message": "error[internal]: Failed to resolve ./answer in javascript/esm|<cwd>/tests/fixtures/errors/resolve-fail-esm/index.js\\n",`

message shouldn't contain `error[internal]`

hardfist: the problem is that  the message already contains error[internal] when passed to emit_diagnostic

hardfist: the problem is that when we convert error to diagnostic it prepend error[internal] to message and when we convert diagnostic it prepend error[internal] to message again, cc @ahabhgk @IWANABETHATGUY @h-a-n-a

hardfist: it seems we should remove the prepend logic when convert error to diagnostic

crates/rspack_error/src/emitter.rs
hardfist: image
cargo also only render diagnostic.code to red and keep message as white

packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
```diff
@@ -910,7 +910,8 @@ exports[`StatsTestCases should print correct stats for loader-builtin-swc-plugin
   "errorsCount": 0,
   "warnings": [
     {
-      "formatted": "Experimental plugins are not currently supported.
+      "formatted": "warning[internal]: Experimental plugins are not currently supported.
```
hardfist, line 913: formatted contains disagnostic.code seems expected

ahabhgk (approved):

PR #3886, opened 2023-08-02
refactor: dependency request use JsWord

underfin:
## Summary
Because `importVarMap` will clone request as key, so here use JsWord to avoid string clone.

## Test Plan
It's not requried


underfin:
!bench

ahabhgk:
Hard to say which is better since JsWord is hard for migrating to oxc and there is basically no change in benchmark

jerrykingxyz:
!bench

jerrykingxyz:
!bench

jerrykingxyz:
!bench

jerrykingxyz:
!bench

hardfist:
close for stale

PR #3887, opened 2023-08-02
chore: fix example code

daydayhappychao:
## Summary

Currently, `react-with-less` and `react-with-sass` examples cannot be started normally. 

Modify their `rspack.config.js` so that they can be started easily.

## Test Plan

Goto these example dirs and execute `pnpm dev`. Then check that the view has rendered normally.


hardfist:
thanks

hardfist (approved):

PR #3893, opened 2023-08-03
feat: stats logging

ahabhgk:
## Summary

`stats.logging = "verbose"` =>

Screenshot 2023-08-03 at 15 49 50

## Test Plan

added in `Stats.test.ts`


packages/rspack/tests/Stats.test.ts
```diff
@@ -163,7 +190,7 @@ describe("Stats", () => {
 		    entry ./fixtures/a
 		./fixtures/a.js [876] {main}
 		  entry ./fixtures/a
-		rspack compiled successfully (f0a86c7e70b0de037daf)"
+		rspack compiled successfully (1788bb8a658f2310fbf6)"
```
ahabhgk, line 193: hash change is caused by https://github.com/web-infra-dev/rspack/pull/3893/files#diff-a3ed34d20f0b15a073bdeaac181d27e0fe74d19b63f7a0eb6cc1868192ef6dd9L66-R70, the name will be hashed in js_chunk_hash

crates/rspack_core/src/logger.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.label, start.label
+      );
+    }
+  }
+}
+
+pub type CompilationLogging = Arc<DashMap<String, Vec<LogType>, BuildHasherDefault<FxHasher>>>;
+
+pub struct CompilationLogger {
+  logging: CompilationLogging,
```
jerrykingxyz, line 205: it is hard to distinguishing between `CompilationLogging` and `CompilationLogger`, maybe we can change the name.

crates/rspack_plugin_devtool/src/lib.rs
```diff
@@ -100,6 +102,8 @@ impl Plugin for DevtoolPlugin {
     _ctx: PluginContext,
     args: ProcessAssetsArgs<'_>,
   ) -> PluginProcessAssetsOutput {
+    let logger = args.compilation.get_logger(self.name());
```
jerrykingxyz, line 105: Requiring a compilation for `logger` generation doesn not seem like a good idea. maybe we can
```
let logger = Logger::new(self.name());
....
self.compilation.addLogger(logger);
// or
self.compilation.loggers.push(logger);
```

crates/node_binding/src/js_values/stats.rs
```diff
@@ -377,6 +469,16 @@ impl JsStats {
       .collect()
   }
 
+  #[napi]
+  pub fn get_logging(&self) -> Vec<JsStatsLogging> {
```
jerrykingxyz, line 473: Logging is hard to distinguish from "compilationLogging", maybe we could name something more specific.
```
pub fn get_js_stats_logging(&self) -> Vec<JsStatsLogging>;
// or
pub fn get_stats_logging(&self) -> Vec<JsStatsLogging>;
```

crates/rspack_plugin_devtool/src/lib.rs
ahabhgk: `compilation.addLogger` also require a compilation, `compilation.get_logger` API style is aligned with webpack

crates/rspack_core/src/logger.rs
ahabhgk: logging and logger, this is how webpack naming them...🥲

crates/node_binding/src/js_values/stats.rs
ahabhgk: It's easy to distinguish them by the function signature, and logging basically only consumed by stats, so specific `get_logging` to `get_stats_logging` seems not necessary

jerrykingxyz (approved):

PR #3894, opened 2023-08-03
refactor: dependency codegen use exports_info api

underfin:
## Summary

- dependency codegen use exports_info api
- add HarmonyImportSpecifierDependency to dependency instead of presentational_dependencies

## Test Plan

It's not required for the pr.


underfin:
!bench

underfin:
!bench

IWANABETHATGUY (approved):

Issue #3891, opened 2023-08-03
resolve error lose source_info when package.json contains type:module


hardfist:
### System Info

none

### Details

![image](https://github.com/web-infra-dev/rspack/assets/8898718/d0a2415b-2c4a-4648-8978-ed969ee986cc)

### Reproduce Steps

add package.json to example/basic
```json
{
  "type": "module"
}
```
change  "./answer" to "./answers" to make resolve fail and see the error message

PR #3899, opened 2023-08-04
fix(core): fix missing source_info in esm module error

hardfist:
## Summary

fixs #3891, esm error missing due to importer contains "javascript/esm" which cause fs.readfile failed

## Test Plan
No Test Needed

### DIFF
--- a/crates/rspack_core/src/utils/hooks.rs
+++ b/crates/rspack_core/src/utils/hooks.rs
@@ -91,7 +91,12 @@ pub async fn resolve(
         ResolveError(
           runtime_message,
           TraceableError::from_real_file_path(
-            Path::new(importer),
+            Path::new(
+              importer
+                .split_once('|')
+                .map(|(_, path)| path)
+                .unwrap_or(importer),
+            ),
             span.start as usize,
             span.end as usize,
             "Resolve error".to_string(),
--- a/packages/rspack/tests/Errors.test.ts
+++ b/packages/rspack/tests/Errors.test.ts
@@ -142,9 +142,9 @@ it("should emit warnings for resolve failure in esm", async () => {
 		Object {
 		  "errors": Array [
 		    Object {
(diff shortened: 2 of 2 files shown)


IWANABETHATGUY (approved):

h-a-n-a:
I was wondering how other bundlers solve this problem, if the content of a module has been transformed with additional imports and the resolve failed, the original file path will come up with the incorrect error frame.

underfin (dismissed):

hardfist:
> @hardfist I was wondering how other bundlers solve this problem, if the content of a module has been transformed with additional imports and the resolve failed, the original file path will come up with the incorrect error frame.

webpack has the same problem like rspack it generates wrong errorframe when transformed module contains resolve error

image
 
you can verify it on https://github.com/web-infra-dev/rspack-repro/tree/webpack-error-frame

hardfist:
cc @underfin @h-a-n-a @ahabhgk 
after digging into the webpack source I found that
* webpack doesn't even provide error frame for ModuleNotFoundError, so there's no problem with readFile in error
image
* webpack doesn't use sorucemap to decode span, so if the content is transformed then the span in error message is wrong because the span is valid for the intermediate transformed content but not valid for original module
https://github.com/webpack/webpack/blob/main/lib/stats/DefaultStatsFactoryPlugin.js#L400
* webpack use the initial importer as the filename in errorframe, so if a.scss bundle b.scss and c.scss and error happens at b.scss webpack still use a.scss in errorframe

so there're two ways to handle error frame
* we use sourcemap to decode errorframe(which maybe  not align with webpack and has some performance cost)
* we can seralize the intermediate the result to disk so error_frame can show directly in the intermediate result

hardfist:
cc @underfin @IWANABETHATGUY

ahabhgk (approved):

Issue #3903, opened 2023-08-04
Runtime generation failed causing code is eliminated by accident


h-a-n-a:
### System Info

Not related

### Details

Rspack results show that the code will be eliminated after `optimization.minimize` turned on, which causes `React.useEffect` never triggers.

```js
function useInView({ threshold, delay, trackVisibility, rootMargin, root, triggerOnce, skip, initialInView, fallbackInView, onChange } = {}) {
   // ...
    /*#__PURE__*/ (react__WEBPACK_IMPORTED_MODULE_0__namespace_cache || (react__WEBPACK_IMPORTED_MODULE_0__namespace_cache = __webpack_require__.t(react__WEBPACK_IMPORTED_MODULE_0_, 2))).useEffect(()=>{
        if (skip || !ref) return;
        let unobserve;
        unobserve = observe(ref, (inView, entry)=>{
            setState({
                inView,
                entry
            });
            if (callback.current) callback.current(inView, entry);
            if (entry.isIntersecting && triggerOnce && unobserve) {
                unobserve();
                unobserve = void 0;
            }
        }, {
            root,
            rootMargin,
            threshold,
            // @ts-ignore
            trackVisibility,
            // @ts-ignore
            delay
        }, fallbackInView);
        return ()=>{
            if (unobserve) unobserve();
        };
    }, // We break the rule here, because we aren't including the actual `threshold` variable
    // eslint-disable-next-line react-hooks/exhaustive-deps
    [
        // If the threshold is an array, convert it to a string, so it won't change between renders.
        // eslint-disable-next-line react-hooks/exhaustive-deps
        Array.isArray(threshold) ? threshold.toString() : threshold,
        ref,
        root,
        rootMargin,
        triggerOnce,
        skip,
        trackVisibility,
        fallbackInView,
        delay
    ]);
    // ...
}
```

The expected behavior is how webpack handles the code:

```js
function useInView({
  threshold,
  delay,
  trackVisibility,
  rootMargin,
  root,
  triggerOnce,
  skip,
  initialInView,
  fallbackInView,
  onChange
} = {}) {
// ...
  react.useEffect(
    () => {
      if (skip || !ref)
        return;
      let unobserve;
      unobserve = observe(
        ref,
        (inView, entry) => {
          setState({
            inView,
            entry
          });
          if (callback.current)
            callback.current(inView, entry);
          if (entry.isIntersecting && triggerOnce && unobserve) {
            unobserve();
            unobserve = void 0;
[...]

PR #3910, opened 2023-08-07
fix: HarmonyImportSpecifierDependency interop should scan `xxx.xxx`

underfin:
## Summary

close https://github.com/web-infra-dev/rspack/issues/3903.
Before  HarmonyImportSpecifierDependency interop only scan `xxx.default`, it will cased mjs files import cjs is diffrent with webpack. The pr will fixed it.
Other the pr will scan extra `xx.xx` will generate `xx.["xx"]`, so support generate `xx.xx` to align to webpack.

## Test Plan

Added


crates/rspack_core/src/dependency/runtime_template.rs
```diff
@@ -99,11 +102,66 @@ pub fn get_exports_type_with_strict(
     .get_exports_type(strict)
 }
 
+static SAFE_IDENTIFIER_REGEX: Lazy<Regex> =
+  Lazy::new(|| Regex::new(r"^[_a-zA-Z$][_a-zA-Z$0-9]*$").expect("should init regex"));
+static RESERVED_IDENTIFIER: Lazy<[&str; 37]> = Lazy::new(|| {
```
IWANABETHATGUY, line 107: You can use const array here.

hyf0 (approved):

PR #3915, opened 2023-08-07
feat: support minifyOptions.asciiOnly

9aoy:
## Summary

closed: https://github.com/web-infra-dev/rspack/issues/3917

TODO: support [jsc.minify.format.comment (some)](https://swc.rs/docs/configuration/minification#jscminifyformat)

safari10 does not need to be implemented in rspack, because: https://github.com/swc-project/swc/discussions/3373


crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs
```diff
@@ -245,6 +245,11 @@ impl Plugin for JsPlugin {
         ..Default::default()
       };
 
+      let format = JsMinifyFormatOptions {
+        ascii_only: minify_options.ascii_only.is_some(),
```
h-a-n-a, line 249: Should be using `unwrap_or_default`, since there might be `Some(false)` in this case

crates/rspack_binding_options/src/options/raw_builtins/mod.rs
```diff
@@ -65,6 +65,7 @@ pub struct RawMinificationConditions {
 pub struct RawMinification {
   pub passes: u32,
   pub drop_console: bool,
+  pub ascii_only: Option<bool>,
```
ahabhgk, line 68: It would be better if you could change this to `bool` and apply the default value in js side, and it's ok to not change this since I might make a refactor in two or three weeks later

IWANABETHATGUY:
> https://bytedance.feishu.cn/record/1sfvuPC9kVw334lztoz3G3wnX

Is this link public?

9aoy:
@IWANABETHATGUY  maybe not. I removed it and created an issue. 😭

crates/rspack_binding_options/src/options/raw_builtins/mod.rs
9aoy: ok, i'll handle it

h-a-n-a (approved):

PR #3946, opened 2023-08-14
chore: normalize filename

h-a-n-a:
## Summary

Normalize filenames:
ctor/ctor_name -> `Ctor`/`CtorName`
aa-bb-cc -> `aaBbCc`

Note: non-public API changes should not be considered as breaking changes.


LingyuCoder (approved):

pnpm-lock.yaml
Boshen, line 1: Is this expected?

h-a-n-a: My problem, didn't notice this issue as I'm editing this PR on a case-sensitive OS.

PR #3947, opened 2023-08-14
fix: nested import follow a export all symbol

IWANABETHATGUY:
## Summary
1. fix issue when a importAllAs namespace access follows a export all symbol


IWANABETHATGUY:
!canary

underfin (approved):

PR #3949, opened 2023-08-14
feat(plugin-copy): support info option

JSerFeng:
## Summary

Support object-type `info` option for plugin-copy

## Test Plan

The tests describes that after setting `info.minimized` to `true`, the file won't be processed by minimizer


clijiac:
we need this PR, now the rspack will minify code in Public folder we just copy

JSerFeng:
> we need this PR, now the rspack will minify code in Public folder we just copy

Sorry I thought this was merged, I'll make this work in the next release

crates/rspack_plugin_copy/src/lib.rs
```diff
@@ -597,8 +622,64 @@ fn escape_glob_chars(s: &str) -> String {
...
 #[test]
 fn test_escape() {
   assert_eq!(escape_glob_chars("a/b/**/*.js"), r#"a/b/\*\*/\*.js"#);
   assert_eq!(escape_glob_chars("a/b/c"), r#"a/b/c"#);
 }
+
+// If this test fails, you should modify `set_info` function, according to your changes about AssetInfo
+// Make sure every field of AssetInfo is considered
+#[test]
+fn ensure_info_fields() {
```
ahabhgk, line 670: Why can't just use AssetInfo

JSerFeng: User specify the info config, so user can specify any field they want, for the other fields, stay untouched, so users info fields are all optional

ahabhgk (approved):

Issue #3921, opened 2023-08-08
Enhancement : expose RunTimeGlobals in rspack


pushkargupta9807:
### System Info

rspackv :0.2.11

### Details

we have many webpack plugins which import the RunTimeGlobals from webpack. It would be great if we have such exports in rspack as well. 
![image](https://github.com/web-infra-dev/rspack/assets/93791633/4dd5377d-b16d-4f0b-9a92-6d63e73e97de)

### Reproduce Steps

N/A

lippzhang:
I want to contribute this code

PR #3956, opened 2023-08-15
feat(ts):  expose RunTimeGlobals in rspack

lippzhang:
## Summary
https://github.com/web-infra-dev/rspack/issues/3921


packages/rspack/src/global.d.ts
```diff
@@ -109,3 +109,55 @@ declare module "neo-async" {
 	export const forEach: typeof each;
 	export const forEachLimit: typeof eachLimit;
 }
+
+type RuntimeGlobalsType = {
```
h-a-n-a, line 113: Why adding this manually?

packages/rspack/src/RuntimeGlobals.ts
```diff
@@ -0,0 +1,51 @@
+export const RuntimeGlobals: RuntimeGlobalsType = {
```
h-a-n-a, line 1: `as const` is enough in this case.

packages/rspack/src/index.ts
```diff
@@ -7,6 +7,8 @@ export * from "./Stats";
 export * from "./MultiStats";
 export * from "./ChunkGroup";
 export * from "./NormalModuleFactory";
+export * from "./RuntimeGlobals";
```
h-a-n-a, line 10: Please avoid using `export *`, using this may cause export a field by accident in the future. Please use `export { RuntimeGlobals } from "./xxx"`

packages/rspack/src/global.d.ts
lippzhang: I tried exporting via runtime_globals.rs, but it failed. Can you guide me? Thanks

packages/rspack/src/RuntimeGlobals.ts
lippzhang: OK

packages/rspack/src/index.ts
lippzhang: No problem

packages/rspack/src/global.d.ts
h-a-n-a: Exporting from rust will create a huge runtime overhead. Please stick to JS impl.

h-a-n-a: This type can be removed directly.

lippzhang: i understand ，thank

h-a-n-a:
This seems not synced with webpack's `RuntimeGlobals`. Would you please copy this from webpack? Make sure to add a comment for referencing Webpack's code: https://github.com/web-infra-dev/rspack/blob/0987a0d1f2f13b2531206625af0e4d05546e718c/packages/rspack/src/compilation.ts#L1-L9

lippzhang:
> 

ok

PR #3961, opened 2023-08-15
fix(chore): add exports fields to all package 

hardfist:
## Summary
BREAKING CHANGE: This limits user import internal module from rspack

## Test Plan
No extra tests needed


packages/rspack-dev-server/package.json
```diff
@@ -5,6 +5,12 @@
   "description": "Development server for rspack",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
+  "exports": {
```
Boshen, line 8: `"exports": "./dist/index.js"` should just work 🤔 ?

hardfist: we may add more entryPoint, so I use the more general form

hardfist: I want a publint, does oxclint support now?

Boshen: > we may add more entryPoint later, so I use the more general form

You can always change it later?

> I want a publint, does oxclint support now?

no.

hardfist: actually, it also needs add type field or typescript will not work, so `"exports": "./dist/index.js"` seems not enough here

h-a-n-a:
Seems not rebased?

h-a-n-a (approved):

PR #3972, opened 2023-08-16
perf: reduce memory size of const dependency and dependency type

underfin:
## Summary

Const Dependency size 56 -> 40, DependencyType 32 -> 24


underfin:
!bench

jerrykingxyz:
!bench

PR #4005, opened 2023-08-22
fix: disable DevtoolPlugin when options.devtool=false

LingyuCoder:
## Summary

Disable DevtoolPlugin when options.devtool=false

Can not simplely use devtool.enabled() because it works for module loader.
Method add_module() will also works when devtool is an empty string or been some other illegal strings.
So return directly and add a new method plugin_enabled() to disable the DevtoolPlugin.


hardfist:
!bench

hyf0 (dismissed):

ahabhgk:
snapshot needs update, hash changed is caused by https://github.com/web-infra-dev/rspack/blob/eb22d900e18439afe29627e5bb7dcea378f6bcc5/crates/rspack_plugin_devtool/src/lib.rs#L95

crates/rspack_binding_options/src/options/mod.rs
```diff
@@ -222,17 +222,20 @@ impl RawOptionsApply for RawOptions {
...
-        append: !devtool.hidden(),
-        namespace: output.unique_name.clone(),
-        columns: !devtool.cheap(),
-        no_sources: devtool.no_sources(),
-        public_path: None,
-      })
-      .boxed(),
-    );
+
+    if devtool.plugin_enabled() {
```
ahabhgk, line 226: `devtool.source_map()` looks better to me

ahabhgk: I didn't implement `'eval'` so don't need to care about `'eval'`, even if I need to implement it in the future I will add an another plugin to do that

LingyuCoder: > `devtool.source_map()` looks better to me

source_map() has been exists and not works for devtool=eval

ahabhgk (approved):

Issue #3737, opened 2023-07-10
Support `resolveLoader`


h-a-n-a:
### What problem does this feature solve?

See webpack's `resolveLoader` which provides a way for modifying resolution of loaders.

### What does the proposed API of configuration look like?

The same as Webpack's `resolveLoader`

https://webpack.js.org/configuration/resolve/#resolveloader

hhuwc:
Strongly need this feature

PR #4008, opened 2023-08-22
feat: support option `resolveLoader`

h-a-n-a:
## Summary

closes https://github.com/web-infra-dev/rspack/issues/3737

Added support for `resolveLoader`. However, nodejs-resolver does not support 
alias with a custom query or fragment, which results in the failure with the first webpack test case.

## Test Plan

See test changes and add back the first test once `resolve.alias` supports custom query and fragment.

### DIFF
--- a/webpack-test/configCases/loaders/issue-3320/index.js
+++ b/webpack-test/configCases/loaders/issue-3320/index.js
@@ -1,9 +1,9 @@
 // Webpack supports this, but Rspack does not support `resolve.alias` with a custom query and fragment
-it.skip("should resolve aliased loader module with query", function() {
-	var foo = require('./a');
+// it.skip("should resolve aliased loader module with query", function() {
+// 	var foo = require('./a');
 
-	expect(foo).toBe("someMessage");
-});
+// 	expect(foo).toBe("someMessage");
+// });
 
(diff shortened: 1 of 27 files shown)


webpack-test/configCases/loaders/issue-3320/index.js
```diff
@@ -1,4 +1,5 @@
-it("should resolve aliased loader module with query", function() {
+// Webpack supports this, but Rspack does not support `resolve.alias` with a custom query and fragment
+it.skip("should resolve aliased loader module with query", function() {
```
h-a-n-a, line 2: Currently, Due to the fact that nodejs-resolver does not support resolving with alias, Rspack does not support resolve aliased loader. We should fix this.

hyf0 (approved):

webpack-test/configCases/loaders/issue-3320/index.js
h-a-n-a: tracked here https://github.com/web-infra-dev/rspack/issues/4009

h-a-n-a:
Too bad, CI failed on windows: https://github.com/web-infra-dev/rspack/actions/runs/5936765987/job/16098346469?pr=4008#step:8:673

hyf0 (approved):

Issue #3851, opened 2023-07-27
what is the plan to support `asset.info.sourceFilename` 


ClarkXia:
### What problem does this feature solve?

Asset Info returned by `compilation.getAsset(chunkFile)` lost props of `sourceFilename`.
Do we have a plan to support it.

### What does the proposed API of configuration look like?

`compilation.getAsset(chunkFile)` returns `sourceFilename`.

hyf0:
@web-infra-dev/rspack-dx cc

faga295:
Maybe I can help to support this

h-a-n-a:
> Maybe I can help to support this

@faga295 Thanks for the help! Already assigned.

LingyuCoder:
Solved by #4013

PR #4013, opened 2023-08-22
feat: support asset info source filename

faga295:
#3851


underfin:
The `packages/rspack/tests/fixtures/icon.png` is so large, you use an empty png to test.

faga295:
@underfin cc

ahabhgk:
Looks great, but CI need pass, you can run `pnpm run test -- -u` under `packages/rspack` to update the snapshot

ahabhgk (approved):
Thanks!

PR #4018, opened 2023-08-23
ci: reduce binary size

h-a-n-a:
## Summary

Do not generate debuginfo for debug to reduce binary size. (even for debug binary as a temporary workaround)


.github/workflows/reusable-build.yml
```diff
@@ -96,6 +96,7 @@ jobs:
         shell: bash
         run: |
           echo '[profile.dev.package."*"]' >> Cargo.toml
+          echo 'strip = "debuginfo"' >> Cargo.toml
```
h-a-n-a, line 99:
```suggestion
          echo 'debug = false' >> Cargo.toml
```

underfin (approved):

PR #4022, opened 2023-08-24
fix(runtime): should calculate `initial_chunks` correctly

### LINKED ISSUE #4019: the runtime module of multiple entries would override each other when they have the same name
### System Info

.

### Details

The content of https://github.com/hyf0/rspack-css-only-chunk-runtime-error/blob/main/rspack-dist/build-runtime.js is for the entry `index2.js`.

image
image

Missing chunk info for chunk id `vendor`.

See https://github.com/hyf0/rspack-css-only-chunk-runtime-error

### Reproduce link

_No response_

### Reproduce Steps

.

hyf0:
## Summary

Fixes #4019

## Test Plan

Added

### DIFF
--- a/crates/rspack_core/src/chunk.rs
+++ b/crates/rspack_core/src/chunk.rs
@@ -223,15 +223,19 @@ impl Chunk {
   }
 
   pub fn get_all_async_chunks(&self, chunk_group_by_ukey: &ChunkGroupByUkey) -> HashSet<ChunkUkey> {
+    use rustc_hash::FxHashSet;
+
     let mut queue = HashSet::default();
     let mut chunks = HashSet::default();
-    let initial_chunks: HashSet<ChunkUkey> = self
+
+    let initial_chunks = self
       .groups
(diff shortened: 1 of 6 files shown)


crates/rspack_core/src/chunk.rs
```diff
@@ -223,15 +223,19 @@ impl Chunk {
   }
 
   pub fn get_all_async_chunks(&self, chunk_group_by_ukey: &ChunkGroupByUkey) -> HashSet<ChunkUkey> {
+    use rustc_hash::FxHashSet;
```
underfin, line 226: The file already import `FxHashSet` as `HashSet`

hyf0:
!canary

crates/rspack_core/src/chunk.rs
hyf0: I don't want to run CI again. Let's keep it now.

underfin (approved):

PR #4042, opened 2023-08-25
chore: improve code format

SnowingFox:
## Summary

Remove some unused vars & format code

## Test Plan

None


hardfist:
please sign cla

SnowingFox:
> please sign cla

done

packages/rspack-cli/src/rspack-cli.ts
```diff
@@ -62,7 +57,9 @@ export class RspackCLI {
 			? (config as MultiRspackOptions).some(i => i.watch)
 			: (config as RspackOptions).watch;
 
-		return rspack(config, isWatch ? callback : undefined);
```
h-a-n-a, line 65: Why using type casting here? It's a bad practice to me as it hinders the potential type changes.

SnowingFox: > Why using type casting here? It's a bad practice to me as it hinders the potential type changes.

Because if we remove this, the type will inferred be 
image

Just MultiCompiler reserverd

h-a-n-a: It does not seem to be that case for me.

SnowingFox: > It does not seem to be that case for me.

Maybe my IDE is webstorm, the difference may comes from the different ways in which two IDEs infer types, i will remove this.

h-a-n-a:
@Boshen Does oxc support detecting unused values or types?

SnowingFox:
> @Boshen Does oxc support detecting unused values or types?

I think eslint can do this, but there is no eslint package in this project, should i configure eslint?

h-a-n-a:
> > @Boshen Does oxc support detecting unused values or types?
> 
> I think eslint can do this, but there is no eslint package in this project, should i configure eslint?
> 
> And we can update some ci workflow like typecheck & eslint check

We should like to stick to oxclint for now.

h-a-n-a (approved):

SnowingFox:
> > > @Boshen Does oxc support detecting unused values or types?
> > 
> > 
> > I think eslint can do this, but there is no eslint package in this project, should i configure eslint?
> > And we can update some ci workflow like typecheck & eslint check
> 
> We should like to stick to oxclint for now.

can i join this task?

PR #4046, opened 2023-08-28
refactor: add InitFragment trait

underfin:
## Summary

Add add InitFragment trait, because need add `HarmonyExportInitFragment` for `HarmonyExportSpecifierDependency` codegen. The treeshaking need refactor `HarmonyExportSpecifierDependency` codegen.

## Test Plan

Not need.

### DIFF
--- a/crates/rspack_core/src/init_fragment.rs
+++ b/crates/rspack_core/src/init_fragment.rs
@@ -1,15 +1,24 @@
+use rspack_sources::{BoxSource, ConcatSource, RawSource, SourceExt};
+use rspack_util::ext::AsAny;
 use rustc_hash::FxHashMap as HashMap;
+use swc_core::ecma::atoms::JsWord;
 
-#[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub struct InitFragment {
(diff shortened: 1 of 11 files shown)


crates/rspack_core/src/init_fragment.rs
```diff
@@ -29,4 +66,79 @@ pub enum InitFragmentStage {
...
+
+  let mut sources = ConcatSource::default();
+
+  fragments.iter_mut().for_each(|f| {
+    sources.add(RawSource::from(std::mem::take(&mut f.content)));
+  });
+
+  sources.add(source);
+
+  fragments.iter_mut().rev().for_each(|f| {
```
h-a-n-a, line 88: Just out of suggestion, can we implement a `Fromiterator` trait on the `rspack_source` side for making upstream code base a little bit cleaner?

underfin: `InitFragment` is rspack designed struct, it is un-relevant to `rspack_source`, so i think put it into `rspack` is good for me.

h-a-n-a (approved):

IWANABETHATGUY (approved):

PR #4053, opened 2023-08-29
chore: run oxlint on all files during lint-staged (for now)

Boshen:
Currently when a specific file is passed to oxlint (e.g. `/pwd/to/test.js`), it will lint the file despite having it specified in `.eslintignore`, this makes lint-staged useless when an ignored file is committed.

oxlint has the same behavior as eslint.

This PR makes lint-staged to run on all changed files. The overhead is manageable because oxlint completes within a second.

See
* https://github.com/okonet/lint-staged#how-can-i-ignore-files-from-eslintignore
* https://github.com/okonet/lint-staged#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments

## Test Plan

I verified the behaviour we want: when an ignored file is commited, it does not get linted.

### DIFF
--- a/lint-staged.config.mjs
+++ b/lint-staged.config.mjs
@@ -1,0 +1,8 @@
+export default {
+  "*.rs": "rustfmt --edition 2021",
+  "packages/**/*.{ts,js}": "prettier --write",
+  "x.mjs": "prettier --write",
+  "crates/rspack_plugin_runtime/**/*.{ts,js}": "prettier --write",
+  "*.toml": "npx @taplo/cli format",
+  "**/*.{ts,js,mjs}": () => "oxlint ."
+}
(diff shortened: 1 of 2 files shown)


lint-staged.config.mjs
```diff
@@ -0,0 +1,8 @@
+export default {
+  "*.rs": "rustfmt --edition 2021",
+  "packages/**/*.{ts,js}": "prettier --write",
+  "x.mjs": "prettier --write",
+  "crates/rspack_plugin_runtime/**/*.{ts,js}": "prettier --write",
+  "*.toml": "npx @taplo/cli format",
+  "**/*.{ts,js,mjs}": () => "oxlint ."
```
Boshen, line 7: Taken from https://github.com/okonet/lint-staged#example-run-tsc-on-changes-to-typescript-files-but-do-not-pass-any-filename-arguments

h-a-n-a (approved):

PR #4054, opened 2023-08-29
feat: support disabling default transformation

h-a-n-a:
## Summary

You can opt-in this feature by enabling this:

```js
module.exports = {
  experiments: {
    rspackFuture: {
       disableTransformByDefault: true
    }
  }
}
```

See: https://github.com/web-infra-dev/rspack/discussions/4070 for reference.

Basically, this PR supports disabling transform for JavaScript-like module types and this will become the default strategy for 0.4.0, aiming to support only web-standard syntax.
For `javascript/auto`, `javascript/esm` and `javascript/dynamic` and other jsx like 
module types:
- `options.target` will not be applied to user code. Only runtime code downgrading is supported. After this change, Rspack aligns the behavior of `options.target` with webpack. To downgrade your code, please use `builtin:swc-loader`.
- `builtins.decorator` will not be configurable as it is not a web standard and using these will result in a `ModuleParseError`
- `builtins.relay/pluginImport/emotion/react` are still available and not intended to be deprecated in this feature.

For TypeScript-like module types:
- These `Rule.type` are deprecated

## Test Plan

Added.

## Performance

Performance test of using `disableTransformByDefault` with `builtin:swc-loader` will not be included in this PR.


crates/rspack_plugin_javascript/src/visitors/mod.rs
```diff
@@ -75,12 +80,12 @@ pub fn run_before_pass(
       //      ),
       Optional::new(
         swc_visitor::decorator(assumptions, &options.builtins.decorator),
-        syntax.decorators()
+        transform_by_default && syntax.decorators()
```
hardfist, line 83: why not skip the whole transform_with_handler directly

h-a-n-a: Still need to apply `DefinePlugin` etc.

crates/rspack_plugin_javascript/src/utils.rs
```diff
@@ -40,27 +44,34 @@ pub fn syntax_by_module_type(
...
   match module_type {
     ModuleType::Js | ModuleType::Jsx => Syntax::Es(EsConfig {
       jsx: matches!(module_type, ModuleType::Jsx),
       export_default_from: true,
       decorators_before_export: true,
-      decorators: enable_decorators,
+      // If `disableTransformByDefault` is on, then we treat everything passed in as a web standard stuff,
+      // which means everything that is not a web standard would results in a parsing error.
+      // So as the decorator.
+      decorators: should_transform_by_default && enable_decorators,
```
h-a-n-a, line 57: Checkpoint: only parse if the legacy behavior is enabled.

crates/rspack_plugin_javascript/src/visitors/mod.rs
```diff
@@ -55,108 +198,44 @@ pub fn run_before_pass(
...
-      //        &cm
-      //      ),
       Optional::new(
         swc_visitor::decorator(assumptions, &options.builtins.decorator),
-        syntax.decorators()
+        // Decorator transformation varies from `ModuleType`,
+        // - TypeScript-like: decorators will be transformed by default, with legacy settings and will emit meta data.
+        //      Since this is a default behavior with tsc.
+        // - JavaScript-like: decorators will not be transformed(if `disableTransformByDefault` is on), and the parse will fail.
+        (options.should_transform_by_default() || syntax.typescript()) && syntax.decorators()
```
h-a-n-a, line 209: Checkpoint: with `disableTransformByDefault` enabled, this will never visited as it will result in a `ModuleParseError` in the previous parsing stage.

crates/rspack_plugin_javascript/src/visitors/mod.rs
```diff
@@ -55,108 +198,44 @@ pub fn run_before_pass(
...
-        !resource_path.contains("@swc/helpers")
-          && !resource_path.contains("tslib")
-          && !resource_path.contains("core-js")
+        options.should_transform_by_default()
       ),
-      swc_visitor::reserved_words(),
+      builtins_webpack_plugin(options, unresolved_mark),
+      // This will be deprecated in the future when TypeScript support and transform by default is dropped.
+      // These features are coupled with this.
       swc_visitor::inject_helpers(unresolved_mark),
```
h-a-n-a, line 238: `inject_helpers` are working with `compat_transform` and `decorator` so this cannot be removed now.

crates/rspack_plugin_javascript/tests/fixtures/provide/snapshot/output.snap
```diff
@@ -47,21 +47,29 @@ function defaultClearTimeout() {
 }
 (function() {
     try {
-        if (typeof setTimeout === "function") cachedSetTimeout = setTimeout;
-        else cachedSetTimeout = defaultSetTimout;
+        if (typeof setTimeout === "function") {
```
h-a-n-a, line 50: Snapshot changes because of we don't want to apply `expr_simplifier` and `dead_branch_remover` to every module. This will add additional overhead. Removing this will not change semantic behavior. It works more well with `DefinePlugin`.

ahabhgk (dismissed):

h-a-n-a:
!canary

h-a-n-a:
!canary

PR #4072, opened 2023-08-30
fix(runtime): should calculate installed css chunks correctly

hyf0:
## Summary

The pr has fixed some error with css/js loading runtime modules and get chunk filename runtime module.

1. `installedChunks` at CssLoadingRuntimeModule

Before rspack will render it to initial chunks which has css, if the initial chunk is split by bundling split and it is could other entry async chunk, it will skip the chunk loading. See packages/rspack/tests/runtimeCases/runtime/split-css-chunk test.
So here `installedChunks` is meaning to initial chunks which hasn't css.

2. `JS_MATCH`/ `CSS_MATCH`

Before rspack will render it to `chunkId`, but webpack is render it to which like `"css-chunk_dynamic_js" == chunkId`. If the chunkId hasn't not corresponding chunkName will skip load it. Before we doesn't implement this, so meet `http://xxx/undefined` error. 

3. render chunks map at GetChunkFilenameRuntimeModule

Before rspack always render dynamic.css/js chunk names, but maybe the dynamic chunk hasn't css. The pr add source type to check it.

## Test Plan

Rename some test and add packages/rspack/tests/runtimeCases/runtime/split-css-chunk test.


hyf0:
!canary

underfin:
!canary

underfin:
!canary

underfin (dismissed):

underfin (approved):

underfin:
!canary

underfin (approved):

PR #4077, opened 2023-08-31
chore: skip empty test folder

hardfist:
## Summary
currently rspack try run tests in tests folder, but some folder are moved or ignored on remote branch but the folder is not deleted in local branch, so skip empty folder(without webpack.config.js) for testing

## Test Plan
tests are updated

### DIFF
--- a/packages/rspack/tests/ConfigCase.template.ts
+++ b/packages/rspack/tests/ConfigCase.template.ts
@@ -72,8 +72,12 @@ export const describeCases = config => {
 			describe(category.name, () => {
 				for (const testName of category.tests) {
 					// eslint-disable-next-line no-loop-func
+					const testDirectory = path.join(casesPath, category.name, testName);
+					const configFile = path.join(testDirectory, "webpack.config.js");
+					if (!fs.existsSync(configFile)) {
+						continue;
+					}
 					describe(testName, function () {
-						const testDirectory = path.join(casesPath, category.name, testName);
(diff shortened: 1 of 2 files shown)


packages/rspack/tests/ConfigCase.template.ts
```diff
@@ -204,117 +208,13 @@ export const describeCases = config => {
...
-						//                 return;
-						//             }
-						//             done();
-						//         });
-						//     }, 40000);
-						// }
-						it(`${testName} should compile`, done => {
+						it(`${testName} should compile`, _done => {
+							console.info("running:", testName);
+							console.time(testName);
```
hardfist, line 213: add some time info for spot long task test

packages/rspack/tests/configCases/css/urls/webpack.config.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
h-a-n-a, line 1: What is this file for?

hardfist: make sure this test are tested, if it doesn't have webpack.config.js it will be skipped

hardfist: actually this tests should be put in caseTest other than configTest, since it doesn't change any configuration

hyf0 (approved):

PR #4078, opened 2023-08-31
chore(ci): use release build for ci tesets

Boshen:
It seems like the debug build is crashing a lot, let's try release builds.


Boshen:
Let's wait for the CI to build, then re-run test jobs a few times to see if things crashes.

hardfist:
seems release also panic

PR #4082, opened 2023-08-31
fix(loader/swc): builtin swc don't emit error message

happlex:
Problem:
The built-in swc-loader does't emit any messages when transpile errors occur.

Reason:
The loader_context.content is moved before processing, and the TWithDiagnosticArray factory rejects nil content.

Changes:
move -> clone

Before:
image

After:
image


h-a-n-a:
Really thankful for you reaching out to us! However, I might have to close this as Rspack's underlying diagnostic system is quite complicated and have some mis-design. See https://github.com/web-infra-dev/rspack/pull/4111 for the detail. 
If you have any further questions, feel free to ping me or raise another issue!

PR #4094, opened 2023-09-01
feat: export externals plugin and node target plugin, electron target plugin

ahabhgk:
## Summary

Add export

### DIFF
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -46,10 +46,6 @@ import {
 	deprecated_resolveBuiltins
 } from "./builtin-plugin";
 
-class NodeTargetPlugin {
-	apply() {}
-}
-
 class NodeTemplatePlugin {
 	apply() {}
 }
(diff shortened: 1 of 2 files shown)


packages/rspack/src/Compiler.ts
```diff
@@ -163,12 +159,22 @@ class Compiler {
 			get EntryPlugin() {
 				return require("./builtin-plugin").EntryPlugin;
 			},
+			get ExternalsPlugin() {
+				return require("./builtin-plugin").ExternalsPlugin;
```
h-a-n-a, line 163: This lazy valuation seems not working as `builtin-plugin` is just exporting everything.

ahabhgk: Not affect the correctness, we can make it lazy in following PR

h-a-n-a (approved):

h-a-n-a:
I think this is a `feat`.

h-a-n-a:
@ahabhgk 
Seems it does not change the title after we adding this into a merge queue.
image

PR #4107, opened 2023-09-04
feat: initialize future flags

h-a-n-a:
## Summary

Initializes future flags. Rspack learned the technique for future flag from remix team.
Future flagging is fundamentally the same as other terms you may heard of like `feature-gates`, `feature-flags`.

Essentially, in Rspack, when it comes to adopting a deprecation or a newly added feature that differs from 
the behavior of the previous version, a new future flag is added.

For example:

```js
module.exports = {
  experiments: {
    rspackFuture: {
       disableTransformByDefault: true, // opt-in new deprecations
       newResolver: true
    }
  }
}
```

| Version |  Future flag | Deprecation message | Stability of the future behavior |
| - | - |  - | - |
| Current minor | Added (Default: _same as the old behavior_) | Added | ⚠️ (Not guaranteed)  |
| Next minor |  Still available (Default: _new behavior enabled_)  | Still available | ✅ |
| Minor after the next minor | Removed | Removed | ✅ |

When adopting a new breaking behavior in Rspack, three steps are included: 

1. In the current minor version, a breaking change is added with a new future flag. However, 
**for the default behavior, this flag is off**. You can adopt this feature by switching the flag.
At the same time, a deprecation message with regard to the breaking will also be added. 
2. In the next minor, the flag will be **turned on by default**. This is a **breaking change**. 
At this stage, you can still switch to the old implementation. This leaves people some time 
for adopting the new behavior later. For people who turned off the flag (which is enabled in this version), 
**a deprecation message will be used as a heads-up**.
3. In the minor version after that, the flag and its deprecation message will be **completely removed**.

In summary, developers have time of less than two releases to adopt to the new behavior.
This strategy is subject to change when it hits to our first major version.

## Test Plan

Not necessary for now.

### DIFF
--- a/packages/rspack/src/config/schema.js
+++ b/packages/rspack/src/config/schema.js
@@ -399,6 +399,13 @@ module.exports = {
 				css: {
 					description: "Enable native css support.",
 					type: "boolean"
+				},
+				rspackFuture: {
+					description:
+						"Enable default behavior in the future version of Rspack",
+					type: "object",
+					additionalProperties: false,
+					properties: {}
 				}
 			}
 		},
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -606,6 +606,7 @@ export interface RawExperiments {
   asyncWebAssembly: boolean
   newSplitChunks: boolean
   css: boolean
+  rspackFuture: RawRspackFuture
 }
 
 export interface RawExternalItem {
@@ -955,6 +956,10 @@ export interface RawResolveOptions {
   fullySpecified?: boolean
   exportsFields?: Array<string>
   extensionAlias?: Record<string, Array<string>>
+}
+
+export interface RawRspackFuture {
+
 }
 
 export interface RawRuleSetCondition {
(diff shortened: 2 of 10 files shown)


packages/rspack/src/config/schema.js
```diff
@@ -399,6 +399,13 @@ module.exports = {
 				css: {
 					description: "Enable native css support.",
 					type: "boolean"
+				},
+				rspackFuture: {
```
h-a-n-a, line 403: Is this still necessary for config validation? @ahabhgk

hyf0: It's not. The validation is done by https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/config/zod/experiments.ts.

h-a-n-a: > It's not. The validation is done by https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/config/zod/experiments.ts.

This is what exactly I'm thinking of. Let me remove this in the next one.

ahabhgk (approved):

Issue #2914, opened 2023-04-24
Rspack alias not working as expected


dgmachado:
### System Info

System:
OS: macOS 13.1
CPU: (10) arm64 Apple M1 Pro
Memory 32GB
Shell: /bin/zsh

Binaries: 
node: 16.13.2
npm: 9.5.0
yarn: 1.22.19

Browsers:
Chrome: 111.0.5563.146

npmPackages:
@nrwl/rspack: 15.9.5
@rspack/core: ~0.1.0
@rspack/dev-server: ~0.1.0
@rspack/plugin-minify: ~0.1.0

### Details

The Rspack alias isn't working as expected.  As you can see below only scenario 1 works.

```javascript
resolve: {
    alias: {
      // # Scenario 1
      //  It works
      './environments/environment': '/..{absolutePathToTheApp}/nestapp/src/environments/environment.prod.ts',
      //
      // # Scenario 2 
      // it should work also but it doesn't 
      './environments/environment': './environments/environment.prod.ts',
      //
      // # Scenario 3 (relative path)
      // it should work also but it doesn't 
      './environments/environment.ts': './environments/environment.prod.ts',
    }
  }
```

### Reproduce Steps

I've created a [sample app](https://github.com/dgmachado/nestapp-example/tree/build) you can use the branch "build" and to test you can execute the below commands:

```
npm run build-rspack
npm run serve-rspack
```

What is expected:
```
Environment.production:true
```
What is appearing:
```
Environment.production:false
```

hyf0:
I would suggest using https://github.com/web-infra-dev/rspack-repro to confirm whether Webpack works as expected.

dgmachado:
@hyf0 I just created two alias scenarios not working in Rspack as expected. I just invited you as a contributor on the [repo](https://github.com/dgmachado/rspack-webpack-repro/tree/bug/rspack-alias). In summary, the below scenarios are working in Webpack and not in Rspack.

```
const questionAbsolutePath = path.resolve(__dirname, "./src/question.js")

...
  resolve: {
    alias: {
     "./answer": path.resolve(__dirname, "./src/answer.prod.js?raw"),
     questionAbsolutePath: path.resolve(__dirname, "./src/question.prod.js?raw"),
    },
  },
```

hyf0:
Thank you for taking time to create the repo. It's a bug of Rspack and we plan to fix it.

dgmachado:
[...]

PR #4108, opened 2023-09-04
feat(rspack_core): add oxc_resolver (turned off by default)

Boshen:
## Summary

This PR adds oxc_resolver to rspack_core. The resolver can be enabled by `experiments.newResolver`.

The next step is to switch the CI to use the new resolver, then eventually have it turned on by default.

I also plan to add `tracing` capabilities for debugging the resolver.

## Test Plan

* CI should pass because this is turned off by default
* I have tested the resolver on 3 internal projects and verified results against `nodejs_resolver`

### DIFF
--- a/crates/rspack_core/src/options/experiments.rs
+++ b/crates/rspack_core/src/options/experiments.rs
@@ -23,7 +23,9 @@ impl IncrementalRebuildMakeState {
 
 #[derive(Debug, Default)]
 #[allow(clippy::empty_structs_with_brackets)]
-pub struct RspackFuture {}
+pub struct RspackFuture {
+  pub new_resolver: bool,
+}
 
 #[derive(Debug, Default)]
 pub struct Experiments {
--- a/crates/rspack_core/src/options/experiments.rs
+++ b/crates/rspack_core/src/options/experiments.rs
@@ -22,7 +22,6 @@ impl IncrementalRebuildMakeState {
 }
 
 #[derive(Debug, Default)]
-#[allow(clippy::empty_structs_with_brackets)]
 pub struct RspackFuture {
   pub new_resolver: bool,
 }
--- a/crates/rspack_core/src/resolver/resolver_impl.rs
+++ b/crates/rspack_core/src/resolver/resolver_impl.rs
@@ -87,8 +87,8 @@ pub enum Resolver {
 }
 
 impl Resolver {
-  pub fn new(options: Resolve) -> Self {
-    if std::env::var("OXC_RESOLVER").is_ok() {
+  pub fn new(new_resolver: bool, options: Resolve) -> Self {
+    if new_resolver {
       Self::new_oxc_resolver(options)
     } else {
       Self::new_nodejs_resolver(options)
--- a/crates/rspack_core/src/resolver/resolver_impl.rs
+++ b/crates/rspack_core/src/resolver/resolver_impl.rs
@@ -1,4 +1,5 @@
 use std::{
+  fmt,
   path::{Path, PathBuf},
   sync::Arc,
 };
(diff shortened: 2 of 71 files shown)


crates/rspack_core/src/resolver/resolver_impl.rs
```diff
@@ -68,12 +83,16 @@ impl<'a> ResolveInnerOptions<'a> {
...
 pub enum Resolver {
   NodejsResolver(nodejs_resolver::Resolver, Arc<nodejs_resolver::Cache>),
-  // OxcResolver(oxc_resolver::Resolver),
+  OxcResolver(oxc_resolver::Resolver),
 }
 
 impl Resolver {
   pub fn new(options: Resolve) -> Self {
-    Self::new_nodejs_resolver(options)
+    if std::env::var("OXC_RESOLVER").is_ok() {
```
hardfist, line 91: use `experiments.rspackFuture.enableNewResolver` to control, so we can run  tests against new resolver and old resolver

h-a-n-a: @hardfist  I thought we are omitting the `enable` part as we discussed before.

hardfist: sorry forget this name convention

hardfist:
can you help verify whether https://github.com/web-infra-dev/rspack/issues/2914 and https://github.com/web-infra-dev/nodejs_resolver/issues/185 is fixed in new resolver implementation

packages/rspack/src/config/defaults.ts
```diff
@@ -157,6 +157,7 @@ const applyExperimentsDefaults = (
 	D(experiments, "lazyCompilation", false);
 	D(experiments, "asyncWebAssembly", false);
 	D(experiments, "newSplitChunks", true);
+	D(experiments, "newResolver", false);
```
hardfist, line 160: should be `experiments.rspackFuture.newResolver`

Boshen:
> can you help verify whether https://github.com/web-infra-dev/rspack/issues/2914 and https://github.com/web-infra-dev/nodejs_resolver/issues/185 is fixed in new resolver implementation

The problem is fixed in the repro https://github.com/jmikrut/rspack-absolute-alias from the [mentioned comment](https://github.com/web-infra-dev/nodejs_resolver/issues/185#issuecomment-1653472531)

> https://github.com/web-infra-dev/rspack/tree/main/packages/rspack/tests/configCases/resolve these tests should be running against on newResolver and oldResolver

See https://github.com/web-infra-dev/rspack/pull/4108/commits/1e3f3b4848420d5bef6b32c722a305143635d9a3

Boshen:
[1e3f3b4](https://github.com/web-infra-dev/rspack/pull/4108/commits/1e3f3b4848420d5bef6b32c722a305143635d9a3) Doesn't work, the tests will fail. What's t he proper way of doing this?

> https://github.com/web-infra-dev/rspack/tree/main/packages/rspack/tests/configCases/resolve these tests should be running against on newResolver and oldResolver

Boshen:
> it would be helpful if you paste the performance result of newResolver and oldResolver in tested real world projects

I can't get a reliable benchmark from real projects due to too much noise from other parts of the compiler. The benchmark from the oxc should be good enough.

hardfist:
!bench

Boshen:
> !bench

Bench is not going to work because this is turned off by default.

packages/rspack/tests/ConfigCase.template.ts
```diff
@@ -121,6 +121,12 @@ export const describeCases = config => {
 								if (!options.target) options.target = "node";
 								if (!options.output) options.output = {};
 								if (!options.output.path) options.output.path = outputDirectory;
+								if (config.newResolver) {
```
hardfist, line 124: shouldn't change this configuration, you could just copy configCase/resolve to configCase/new_resolve and add experiments.rspackFuture.newResolver to the webpack.config.js in each case https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/tests/configCases/resolve/fallback/webpack.config.js#L10

hardfist:
> > !bench
> 
> Bench is not going to work because this is turned off by default.

we may need to support cross compare performance of different configuration like webpack in the future, cc @jerrykingxyz

jerrykingxyz:
> > > !bench
> > 
> > 
> > Bench is not going to work because this is turned off by default.
> 
> we may need to support cross compare performance of different configuration like webpack in the future, cc @jerrykingxyz

Yes, this is already planned.

crates/rspack_core/src/options/experiments.rs
```diff
@@ -23,7 +23,9 @@ impl IncrementalRebuildMakeState {
 
 #[derive(Debug, Default)]
 #[allow(clippy::empty_structs_with_brackets)]
```
h-a-n-a, line 25:
```suggestion
```

h-a-n-a: Let's remove this, since it's useless now

hardfist (approved):

hardfist:
@Boshen don't forget writing documentation about this feature

hardfist:
@Boshen it seems oxc_resolver still missing missing.test.js and dependencies.test.js, which seems important to track dependency info during resolve process, will you implement it in the future?

PR #4111, opened 2023-09-04
fix: `builtin:swc-loader` does not report error correctly

h-a-n-a:
## Summary

Continue https://github.com/web-infra-dev/rspack/pull/4082

In webpack, `emitWarning` or `emitError` can be used to create some additional error on the compilation. 

In Rspack, we have the `loader_context.diagnostics` field to represent this behavior. But this one has been misused in builtin-loaders. In order to make error report successfully and treat the loader error as the module build error, I changed it to returning a `Err`. 

But still, there's some misalignment in the `ModuleBuildError`, I will create a new issue to indicate this problem.

## Test Plan

Added.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "prepare": "is-ci || husky install",
     "pnpm:devPreinstall": "is-ci || node ./scripts/doctor.js",
     "test:example": "pnpm --filter \"example-*\" build",
-    "test:unit": "pnpm --filter \"@rspack/*\" test",
+    "test:unit": "NO_COLOR=1 pnpm --filter \"@rspack/*\" test",
     "test:e2e": "pnpm --filter \"@rspack-e2e/*\" test",
     "test:ci": "cross-env NODE_OPTIONS=--max_old_space_size=8192 pnpm run build:js && pnpm run test:example && pnpm run test:unit && pnpm test:webpack",
     "test:webpack": "pnpm --filter \"webpack-test\" test:metric"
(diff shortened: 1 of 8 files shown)


ahabhgk (dismissed):

package.json
```diff
@@ -24,7 +24,7 @@
     "prepare": "is-ci || husky install",
     "pnpm:devPreinstall": "is-ci || node ./scripts/doctor.js",
     "test:example": "pnpm --filter \"example-*\" build",
-    "test:unit": "pnpm --filter \"@rspack/*\" test",
+    "test:unit": "NO_COLOR=1 pnpm --filter \"@rspack/*\" test",
```
ahabhgk, line 27: You can use `stats.colors = false` to omit the color

ahabhgk: Can not control swc's internal error emitter

h-a-n-a: Ok, sounds that is going to make this PR not focused, it's move on to the next one for that.

h-a-n-a: Seems https://github.com/web-infra-dev/rspack/issues/4124 they don't have this issue, have to figure out why.

ahabhgk (approved):

package.json
hardfist: https://github.com/swc-project/swc/blob/4623b3e07f9cedb9652785f977f2bd0e509981d9/crates/swc_error_reporters/src/handler.rs#L53 swc's internal error can be controlled here

Issue #4112, opened 2023-09-04
Rspack dist uncertainty, execute multiple times  may produce unexpected dist, may also be the dist of the right.


jeromehan:
### System Info

System:
    OS: macOS 13.4.1
    CPU: (8) x64 Apple M1
    Memory: 32.07 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 14.18.3 - ~/.nvm/versions/node/v14.18.3/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v14.18.3/bin/yarn
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.3/bin/npm
    pnpm: 7.30.0 - ~/Library/pnpm/pnpm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Browsers:
    Chrome: 116.0.5845.140
    Safari: 16.5.1
  npmPackages:
    @rspack/cli: ^0.3.1 => 0.3.1 

### Details

<img width="981" alt="image" src="https://github.com/web-infra-dev/rspack/assets/19773879/550dd569-561c-4b4f-a3d0-48e7ffa3a681">
After several executions of packaging, the result, as shown in the figure, made me afraid to use Rspack again, perhaps because rust was unsure about performing parallel tasks.

### Reproduce link

https://github.com/jeromehan/vue-rspack-demo

### Reproduce Steps
execute multiple times

npm run build

h-a-n-a:
I believe this is caused by the vue-loader's implementation. See this one for details: https://github.com/web-infra-dev/rspack/issues/4003#issuecomment-1689662380

jeromehan:
> I believe this is caused by the vue-loader's implementation. See this one for details: [#4003 (comment)](https://github.com/web-infra-dev/rspack/issues/4003#issuecomment-1689662380). A fix to vue-loader is required to avoid parallel issue when it is called from Rust.

thanks

hardfist:
@jeromehan sorry for causing problems for your, and since this is a bug and we're gonna fix it

h-a-n-a:
@jeromehan Feel free to test it out with version: 0.3.1-canary-efd2bf0-20230906032925.

jeromehan:
> @jeromehan Feel free to test it out with version: 0.3.1-canary-efd2bf0-20230906032925.

it is great

Issue #4003, opened 2023-08-22
Matching jsx syntax in vue2


Aszzo:
### System Info

System:
    OS: macOS 13.5
    CPU: (10) x64 Apple M2 Pro
    Memory: 295.67 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 14.18.3 - ~/.nvm/versions/node/v14.18.3/bin/node
    npm: 6.14.15 - ~/.nvm/versions/node/v14.18.3/bin/npm
  Browsers:
    Chrome: 116.0.5845.96
    Safari: 16.6
  npmPackages:
    @rspack/cli: ^0.2.11 => 0.2.11 

### Details

在 vue2 中处理 jsx 的语法时，对于 vue 文件中的 `<script lang="jsx">` 块会有匹配不到的情况（大概十次中出现2次），我怀疑是 Rspack 异步loader机制的产生的，但是我未找到明确的原因

When dealing with jsx syntax in vue2, there are failures to match `<script lang="jsx">` blocks in vue files (about 2 out of 10 times), which I suspect is due to Rspack's asynchronous loader mechanism, but I haven't found a clear reason for this

```
// rspack.config.js

const { VueLoaderPlugin } = require("vue-loader");

/** @type {import('@rspack/cli').Configuration} */
const config = {
	context: __dirname,
	entry: {
		main: "./src/main.js"
	},
	builtins: {
		html: [
			{
				template: "./index.html"
			}
		]
	},
	devServer: {
		historyApiFallback: true
	},
	devtool: false,
	plugins: [new VueLoaderPlugin()],
	resolve: {
		extensions: ['.js', '.vue', '.json', '.wasm'],
		alias: {
			vue$: 'vue/dist/vue.runtime.esm.js'
		}
	},
	module: {
		rules: [
			{
				test: /\.vue$/,
				loader: 'vue-loader'
			},
			{
				test: /\.(js|jsx)$/,
				use: [{
					loader: 'babel-loader',
					options: {
						presets: ['@vue/babel-preset-jsx']
					}
				}],
				exclude: /node_modules/
			},
			{
				test: /\.less$/,
				use: ["vue-style-loader", "css-loader", "less-loader"],
				type: "javascript/auto"
			},
			{
				test: /\.css$/,
				use: ["vue-style-loader", "css-loader"],
				type: "javascript/auto"
			},
			{
				test: /\.svg$/,
				type: "asset/resource"
			}
		]
[...]

PR #4115, opened 2023-09-05
fix: improve interoperability of rules with side-effects

h-a-n-a:
## Summary

Fixes https://github.com/web-infra-dev/rspack/issues/4112
Fixes https://github.com/web-infra-dev/rspack/issues/4003

This PR fixes the issue on the Rspack side to provide a better interoperability
with JavaScript runtime. However, **leveraging the order of each rule being called
is, to me, not a good choice to make**, but webpack does not provide a good alternative 
for us to match `resourcePath` and `resourceQuery` at the same time,
which is a bummer for implementations in a multi-threaded environment. 
Nevertheless, it's reasonable for vue-loader or some other loaders to choose as 
they do not have another choice to make.

See: https://github.com/web-infra-dev/rspack/issues/4003#issuecomment-1689662380

## Test Plan

Tested locally on the reproduction demos.

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -761,6 +761,12 @@ export interface RawModuleOptions {
 }
 
 export interface RawModuleRule {
+  /**
+   * A conditional match matching an absolute path + query + fragment
+   * This one is reserved as our escape hatch for those who
+   * relies on some single-thread runtime behaviors.
+   */
+  rspackResource?: RawRuleSetCondition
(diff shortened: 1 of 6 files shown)


ahabhgk:
Could you add a test, just to prevent regression

h-a-n-a:
> Could you add a test, just to prevent regression

@ahabhgk I suggest us to figure out an SOP for testing concurrency issues in the first place and then, come back to the test.

ahabhgk (approved):

PR #4122, opened 2023-09-05
chore: increase windows ci test timeout minutes

underfin:
## Summary
The release is always timeout https://github.com/web-infra-dev/rspack/actions/runs/6080313608


hardfist:
I think this is not caused by timeout? it will increase the ci time, we should revert the timeout back

Issue #4047, opened 2023-08-28
rspack stats.modules missing runtimeModule


hardfist:

coder-xiaotian:
Do we still need to address this issue? I'd like to give it a try.

hardfist:
yes，go try it

hardfist:
fixed in #4126

PR #4126, opened 2023-09-05
feat: support stats runtime modules

faga295:
## Summary
#4047


JSerFeng:
Thanks !

crates/rspack_core/src/stats.rs
```diff
@@ -136,13 +137,26 @@ impl Stats<'_> {
...
+      .runtime_modules
+      .iter()
+      .map(|(identifier, module)| {
+        self.get_runtime_module(identifier, module, reasons, module_assets)
+      })
+      .collect::<Result<_>>()?;
+    let mut modules = modules
+      .into_iter()
+      .chain(runtime_modules.into_iter())
+      .collect::<Vec<_>>();
```
ahabhgk, line 159: I think this can be done isomorphically by line 140, if module's module_type is `ModuleType::Runtime`, then its stats.moduleType will also be "runtime", so the correct solution should be adding `ModuleType::Runtime` to runtime module

ahabhgk (changes requested):

crates/rspack_core/src/stats.rs
faga295: We can add `ModuleType::Runtime` to runtime module in ` impl_runtime_module` marco

faga295: >  done isomorphically by line 140
Does this means we should combine `get_module` and `get_runtime_module` into one function?

ahabhgk:
Could you add `options.runtimeModules = optionOrLocalFallback(options.ids, !context.forToString);` in Compilation.createStatsOptions, and snapshot need update

crates/rspack_core/src/stats.rs
ahabhgk: Oh, I see, then current implementation looks good to me

faga295:
Why use `options.runtimeModules = optionOrLocalFallback(options.ids, !context.forToString);` not `options.runtimeModules = optionOrLocalFallback(options.runtimeModules, !context.forToString);`

ahabhgk:
🤦 I copied and pasted this, forgot to change all of.ids, you're right, it should be .runtimeModules

ahabhgk (approved):
Thanks!

PR #4133, opened 2023-09-06
feat: support `builtin:swc-loader` experimental transformers

h-a-n-a:
## Summary

This PR supports experimental rspack transformer in `builtiln:swc-loader`.
[Discussion](https://github.com/web-infra-dev/rspack/discussions/4006#discussioncomment-6903069).

Builtin transformers will be removed in v0.5.0 and switch to `loaderOption.rspackExperiments`:
```js
module.exports = {
  builtins: {
    emotion: {},
    react: {},
    relay: {},
    pluginImport: {}
  }
}
```

⬇️⬇️⬇️

```js
module.exports = {
  module: {
     rules: [
       { 
          test: /\.jsx/,  
          loader: "builtin:swc-loader", 
          options: { 
            rspackExperiments: { 
               emotion: {}, 
               react: {} 
            } 
          } 
        }
    ]
  }
}
```

This PR is not a breaking change. Internally, a new crate-`rspack_swc_visitors` is created, this is used 
to store visitors like `provide`, `define`, `relay`, `react` and so on. And a few refactors is applied to the 
codebase for use to prepare the deprecation of builtin transformer(compilation-wise).

## Test Plan

Migrated from `builtins`

## Require Documentation?

- [ ] No

- [X] Yes, the document will be added once everything for builtin is supported.


examples/emotion/rspack.config.js
```diff
@@ -1,3 +1,9 @@
+const { createSwcLoaderExperiments } = require("@rspack/cli");
+
+const rspackExperiments = createSwcLoaderExperiments()
```
ahabhgk, line 3: Is `createSwcLoaderExperiments` only exists for migrate? Could we avoid add this to core? Maybe add to a separate package?

h-a-n-a: It's for convenience and might be existing for a long run. Creating a new package maybe too much for this case?

ahabhgk: IMO, core shouldn't contain this, and if we want to remove it from core in the future we will have to make a break change

and if swc's raw options for these visitor is not convenience enough, we can introduce a new stage to normalize/default the options before send raw options to rust side, maybe now is a good time since we already have to make some break

h-a-n-a: Remove the normalization part completely from user side.

packages/rspack/src/config/adapterRuleUse.ts
```diff
@@ -220,6 +226,59 @@ export function createRawModuleRuleUses(
...
+		}/dart-sass-embedded/dart-sass-embedded${
+			process.platform === "win32" ? ".bat" : ""
+		}`
+	);
+	return o;
+};
+
+const getSwcLoaderOptions: GetLoaderOptions = (o, options) => {
+	if (o && typeof o === "object" && o.rspackExperiments) {
+		let expr = o.rspackExperiments;
```
h-a-n-a, line 247: Moved the normalization part here.

ahabhgk (approved):

PR #4135, opened 2023-09-06
feat: react refresh plugin

ahabhgk:
## Summary

- add `@rspack/plugin-react-refresh`
- deprecate `@rspack/dev-client`

## Test Plan

- update e2e test (e2e -> dev-server -> plugin-react-refresh)

## Require Documentation?

- [ ] No
- [x] Yes, the corresponding rspack-website PR is https://github.com/web-infra-dev/rspack-website/pull/458


packages/playground/cases/react/basic/rspack.config.js
```diff
@@ -11,7 +11,10 @@ module.exports = {
 			}
 		]
 	},
-	entry: ["@rspack/dev-client/react-refresh-entry", "./src/index.jsx"],
```
h-a-n-a, line 14: Let's keep this one and create a new test instead until the `dev-client` is fully deprecated.

h-a-n-a (dismissed):

underfin (approved):

hardfist:
@ahabhgk can you help add more context about this change, I'm kind of lost here
* why we deprecate @rspack/dev-client in favor of @rspack/plugin-react-refresh
* what's the difference between @rspack/plugin-react-refresh and https://github.com/pmmmwh/react-refresh-webpack-plugin/tree/main

ahabhgk:
- @rspack/dev-client only contains runtime code related to react refresh, we don't have a react-refresh-plugin before, but now we have one, so these code is been move to @rspack/plugin-react-refresh, and @rspack/dev-client is a re-export for these code
- the community version is more complex, it uses additionalRuntimeRequirement hook and afterResolve hook, which causes it hard to be compatible with for now, our version is more simpler

ahabhgk:
This PR only extract the existing logic which been placed in rspack-dev-server before into a new package: @rspack/plugin-react-refresh, so users who don't use rspack-dev-server can more easily enable react refresh

PR #4136, opened 2023-09-06
fix: css loading initial chunks

underfin:
## Summary

Because rspack always generate a js chunk for css chunk, so here fix some bugs.
- `GetChunkFilenameRuntimeModule` generate js async chunk map. 
- Webpack using `loadCssChunkData` and detect css variables to add install chunk, but rspack not support it, so here use js chunk to add install chunk as workaround.

## Test Plan

Added.

## Require Documentation?

[x] No

[ ] Yes, the corresponding rspack-website PR is \_\_


underfin:
!canary

hyf0 (approved):

PR #4137, opened 2023-09-06
feat: re-export core from cli and update templates

ahabhgk:
## Summary

- re-export core from cli
- migrate to builtin plugin in create template

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack-plugin-node-polyfill/src/index.js
+++ b/packages/rspack-plugin-node-polyfill/src/index.js
@@ -32,18 +32,12 @@ const PolyfilledBuiltinModules = {
 };
 module.exports = class PolyfillBuiltinsPlugin {
 	apply(compiler) {
-		const provide = {
-			Buffer: [require.resolve("buffer/"), "Buffer"],
-			process: [require.resolve("process/browser")]
-		};
-
-		compiler.options.builtins = {
-			...compiler.options.builtins,
-			provide: {
-				...provide,
-				...compiler.options.builtins?.provide
-			}
-		};
+		compiler.options.plugins.push(
+			new compiler.webpack.ProvidePlugin({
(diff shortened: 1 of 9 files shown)


packages/rspack-plugin-node-polyfill/src/index.js
```diff
@@ -32,18 +32,12 @@ const PolyfilledBuiltinModules = {
...
-
-		compiler.options.builtins = {
-			...compiler.options.builtins,
-			provide: {
-				...provide,
-				...compiler.options.builtins?.provide
-			}
-		};
+		compiler.options.plugins.push(
+			new compiler.webpack.ProvidePlugin({
```
h-a-n-a, line 36: I was wondering if `compiler.webpack` has the exact same option and outcome  with webpack? Or maybe we should move it to something like `compiler.rspack`

h-a-n-a: Yes, I think we might have to move `HtmlPlugin` and something like that to `compiler.rspack` to avoid conflict.

h-a-n-a (approved):

packages/rspack-plugin-node-polyfill/src/index.js
ahabhgk: Then let's not merge this PR and https://github.com/web-infra-dev/rspack-website/pull/445 first, I'd like to have a discuss about this and maybe make some changes, after that then merge the doc PR and introduce builtin plugin to public, just to avoid break change

h-a-n-a: Sure, let's go!

ahabhgk:
!canary

PR #4151, opened 2023-09-07
feat: support function for `BannerPlugin`

ahabhgk:
## Summary

- support function for `BannerPlugin`
- replace `lib/BannerPlugin.js` with builtin plugin BannerPlugin
- delete unnecessary code in `lib`, hope we can totally remove `lib` someday

## Test Plan

- pass webpack-test banner-plugin

## Require Documentation?

- [ ] No
- [x] Yes, ~~the corresponding rspack-website PR is \_\_~~ I will add docs after docs of builtin plugins re-merged


crates/node_binding/src/js_values/mod.rs
```diff
@@ -1,5 +1,9 @@
+mod chunk {
+  // TODO: should we merge rspack_binding_options and node_binding?
```
ahabhgk, line 2: cc @h-a-n-a

h-a-n-a: > or `rspack_plugin_xxx` export `RawXxxOptions` under feature `node_binding`

This looks good to me, keeping these options in the same crate.

h-a-n-a: Just for heads-up. @Boshen seems treat every undefined symbol to lookup at runtime. For convenience, you can omit the feature. But still, I would love to add one though.

ahabhgk: > or `rspack_plugin_xxx` export `RawXxxOptions` under feature `node_binding`

I'm not familiar with napi or cargo or some build stuff, not sure if it's possible or good for ci build time, so you guys can decide it

h-a-n-a: Let's stick to feature, it's a common practice and do not coupled with linker options.

packages/rspack/src/builtin-plugin/ProgressPlugin.ts
```diff
@@ -1,8 +1,8 @@
 import { RawProgressPluginConfig } from "@rspack/binding";
 import { BuiltinPluginKind, create } from "./base";
 
-export type ProgressPluginOptions = RawProgressPluginConfig | undefined;
+export type ProgressPluginArgument = RawProgressPluginConfig | undefined;
```
h-a-n-a, line 4: Do we have to even align with the type name? For consistency I would love to use `XXXOptions`

h-a-n-a (approved):

packages/rspack/src/builtin-plugin/ProgressPlugin.ts
ahabhgk: I also prefer `XXXOptions`, but types in webpack are not consistent..., let's keep the type alignment for now, at least for facade type

Issue #4152, opened 2023-09-08
can't ignore internal warnings


9aoy:
### System Info

  System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 128.80 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.20.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 8.19.4 - /usr/local/bin/npm
  Browsers:
    Chrome: 116.0.5845.179
    Safari: 15.6.1
  npmPackages:
    @rspack/cli: latest => 0.3.2 

### Details

I wanted to use ignoreWarnings to ignore some warnings, but unfortunately, It does not seems to work.
<img width="746" alt="image" src="https://github.com/web-infra-dev/rspack/assets/22373761/b944fe20-35ba-44e7-8f25-2913a64e0197">

### Reproduce link

https://github.com/9aoy/rspack-issue-reproduce/tree/main/ignoreWarnings

### Reproduce Steps

1. pnpm i
2. pnpm dev

PR #4153, opened 2023-09-08
fix: ignoreWarnings should work when call stats.getJson

9aoy:
## Summary

fix: https://github.com/web-infra-dev/rspack/issues/4152

## Require Documentation?

- [ ] No


h-a-n-a:
Can we implement this on the Rust side(maybe in the future PR)? Implementing this on JS to me will cause a tear between Rust and Node.

9aoy:
> Can we implement this on the Rust side(maybe in the future PR)? Implementing this on JS to me will cause a tear between Rust and Node.

@h-a-n-a  I think we can, but future PR,  I can create a task for myself.

h-a-n-a (approved):
Sure, let's use JS `IgnoreWarningsPlugin` for now.

PR #4169, opened 2023-09-11
feat(config): warn while using `experiments.newSplitChunks`

hyf0:
## Summary

Part of https://github.com/web-infra-dev/rspack/discussions/4168.

## Test Plan

Not needed.

## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #4174, opened 2023-09-12
feat(config): only warn while `experiments.newSplitChunks` being explicitly setted

hyf0:
## Summary

Fix https://github.com/web-infra-dev/rspack/pull/4169#issuecomment-1715039630

## Require Documentation?

- [ ] No

### DIFF
--- a/packages/rspack/src/config/zod.ts
+++ b/packages/rspack/src/config/zod.ts
@@ -945,7 +945,7 @@ const experiments = z.strictObject({
 		.boolean()
 		.optional()
 		.refine(val => {
-			if (val === false) {
+			if (val === false || val === true) {
 				console.warn(
 					"`experiments.newSplitChunks` will be removed at 0.4.0. See details at https://github.com/web-infra-dev/rspack/discussions/4168"
 				);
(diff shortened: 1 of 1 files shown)


ahabhgk (dismissed):

packages/rspack/src/config/zod.ts
```diff
@@ -944,10 +944,12 @@ const experiments = z.strictObject({
 	newSplitChunks: z
 		.boolean()
 		.optional()
-		.refine(_ => {
-			console.warn(
-				"`experiments.newSplitChunks` will be removed at 0.4.0. See details at https://github.com/web-infra-dev/rspack/discussions/4168"
-			);
+		.refine(val => {
+			if (val === false) {
```
h-a-n-a, line 948: Maybe we should warn every user that sets this value explicitly no matter `true`/`false` is used?

h-a-n-a: Is it possible to get the raw user input here?

h-a-n-a: According to the deprecation message "`experiments.newSplitChunks` will be removed at 0.4.0.", looks like users with option `true` needs to be warned too.

hyf0: the `val` should be the raw user input.

h-a-n-a (approved):

h-a-n-a:
LGTM

PR #4185, opened 2023-09-13
fix: perf regression caused by optimizeTree hook

ahabhgk:
## Require Documentation?

- [x] No

### DIFF
--- a/crates/node_binding/src/hook.rs
+++ b/crates/node_binding/src/hook.rs
@@ -32,7 +32,7 @@ pub enum Hook {
   AfterCompile,
   FinishModules,
   OptimizeModules,
-  OptimizaTree,
+  OptimizeTree,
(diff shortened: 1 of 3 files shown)


ahabhgk:
!bench

ahabhgk:
!bench

h-a-n-a (approved):

crates/node_binding/src/hook.rs
```diff
@@ -32,7 +32,7 @@ pub enum Hook {
   AfterCompile,
   FinishModules,
   OptimizeModules,
-  OptimizaTree,
+  OptimizeTree,
```
hardfist, line 35: don't we have typo check now?

ahabhgk: `optimiza tree` can be detected but `OptimizaTree` can't, I guess

PR #4199, opened 2023-09-15
feat: export HMR plugin

ahabhgk:
## Summary

and now we also compatible with webpack-dev-server

## Test Plan

- add webpack-dev-server test in e2e

## Require Documentation?

- [ ] No
- [x] Yes, the corresponding rspack-website PR is https://github.com/web-infra-dev/rspack-website/pull/472


packages/rspack/src/rspackOptionsApply.ts
```diff
@@ -135,6 +135,10 @@ export function optionsApply_compat(
 			compiler.context,
 			options.entry
 		);
+
+		if (options.devServer?.hot) {
```
ahabhgk, line 139: need to move to dev-server, move to here is for avoiding break change, maybe we need a future flag, or just use webpack-dev-server and deprecate rspack-dev-server

packages/playground/fixtures/rspack.ts
```diff
@@ -28,7 +30,8 @@ class Rspack {
 				item();
 			}
 		});
-		this.devServer = new RspackDevServer(
+		const DevServerConstructor = wds ? WebpackDevServer : RspackDevServer;
```
h-a-n-a, line 33: Presumably speaking, this might not be stable unless we apply the wds patch(packages/rspack-dev-server/src/patch.ts).
But here is working as expected, since we have already applied it.

h-a-n-a (approved):

ahabhgk:
!bench

PR #4221, opened 2023-09-20
fix: remove `react` from `rspackExperiments` of `builtin:swc-loader`

h-a-n-a:
## Summary

`rspackExperiments.react.refresh` is quite different from the original `jsc.transform.react.refresh`. So, at the time of implementing this feature, 
this option was migrated to `rspackExperiments`. However, this would add some problem.
Since the `rspackExperiments.react.refresh` does two things.
1. Apply `react-refresh/babel`
2. Apply `react-refresh-webpack-plugin/loader`

The second one should respect the bail-out path provided by user and the path of `react-refresh` and `react-refresh-webpack-plugin/client/ReactUtils` and something like this,
which cannot be decided directly from user when writing the configuration. So we decide to remove this and respect to swc-loader's original `jsc.transform.react` and not including the `react-refresh-webpack-plugin/loader` transformations.

`@rspack/plugin-react-refresh` plugin haven't finished yet. So this is only a part of the `react-refresh` implementation for Rspack under `rspackFuture.disableTransformByDefault`.

## Test Plan

Updated.

## Require Documentation?

- [ ] No

### DIFF
--- a/packages/rspack/tests/configCases/builtin-swc-loader/react-refresh-false/webpack.config.js
+++ b/packages/rspack/tests/configCases/builtin-swc-loader/react-refresh-false/webpack.config.js
@@ -5,11 +5,12 @@ module.exports = {
 				test: /\.js$/,
 				loader: "builtin:swc-loader",
 				options: {
-					rspackExperiments: {
-						react: {
-							refresh: false
-						}
-					}
+					// TODO: add back this until `@rspack/plugin-react-refresh` is finished
+					// rspackExperiments: {
+					// 	react: {
+					// 		refresh: false
+					// 	}
+					// }
 				}
 			}
 		]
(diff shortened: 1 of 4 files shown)


hardfist (approved):

packages/rspack/tests/configCases/builtin-swc-loader/react-refresh-false/webpack.config.js
```diff
@@ -5,11 +5,12 @@ module.exports = {
 				test: /\.js$/,
 				loader: "builtin:swc-loader",
 				options: {
-					rspackExperiments: {
-						react: {
-							refresh: false
-						}
-					}
+					// TODO: add back this until `@rspack/plugin-react-refresh` is finished
```
hardfist, line 8: maybe should remove this comment, since another separate loader seems a better option

h-a-n-a: Yes, I mean we can switch to `@rspack/plugin-react-refresh`

PR #4233, opened 2023-09-22
feat: support `include` and `exclude` of react-refresh-rspack-plugin

h-a-n-a:
## Summary

- Add `react-refresh-entry` to the loader excluding list
- Support passing `include` or `exclude` of plugin

## Test Plan

Added

## Require Documentation?

- [ ] No
- [X] Yes, the corresponding rspack-website PR is , let's create a page for rspack plugins first.

### DIFF
--- a/packages/rspack-plugin-react-refresh/src/options.json
+++ b/packages/rspack-plugin-react-refresh/src/options.json
@@ -1,0 +1,33 @@
+{
+  "additionalProperties": false,
+  "type": "object",
+  "definitions": {
+    "Path": { "type": "string" },
+    "MatchCondition": {
+      "anyOf": [
+        { "instanceof": "RegExp" },
+        { "$ref": "#/definitions/Path" },
+        { "type": "null" }
+      ]
+    },
+    "MatchConditions": {
+      "type": "array",
+      "items": { "$ref": "#/definitions/MatchCondition" },
+      "minItems": 1
+    }
+  },
+  "properties": {
+    "exclude": {
+      "anyOf": [
+        { "$ref": "#/definitions/MatchCondition" },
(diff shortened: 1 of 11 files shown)


packages/rspack-plugin-react-refresh/src/options.json
```diff
@@ -0,0 +1,33 @@
+{
+  "additionalProperties": false,
```
ahabhgk, line 2: Let's use zod

h-a-n-a: I copied it from react-fast-refresh plugin.

ahabhgk: it's slow, and let's avoid bring another schema validation lib into our codebase, and schema validate at here seems not necessary, so I prefer remove it or use zod

h-a-n-a: To be honest, the loader-runner also has`schema-utils` builtin to validate loader options. So I believe it's already there.

> schema validate at here seems not necessary

Any reason for that?

ahabhgk: emm, seems we can't avoid using schema-utils..

ahabhgk (approved):

packages/rspack-plugin-react-refresh/src/options.json
h-a-n-a: I was thinking of converting json-schema to zod in order to remove schema-utils completely.

ahabhgk: That would be great, but I'm not sure whether there is a library in community to converting json-schema to zod at runtime, and including the convert time cost it can still faster than schema-utils

PR #4245, opened 2023-09-25
refactor: export in index.ts (breaking change)

ahabhgk:
## Summary

Tidy up index.ts export, before some should not be exported accidentally exported through re-export, if you use these unintentionally exported export, you may be using Rspack through the incorrect way, please stop using these hack, if there is a real need to use these exports, please raise an issue in the Rspack repository, we will evaluate it and make appropriate adjustments.

## Require Documentation?

- [ ] No


h-a-n-a:
@ahabhgk  Would you fix this?

ahabhgk:
Will rebase this before 0.4

crates/rspack_core/Cargo.toml
```diff
@@ -23,7 +23,7 @@ itertools = { workspace = true }
 mime_guess = { workspace = true }
 nodejs-resolver = { version = "0.1.0" }
 once_cell = { workspace = true }
-oxc_resolver = { version = "0.5.3", features = ["tracing-subscriber"] }
```
ahabhgk, line 26: This will break RSPACK_PROFILE since there are two tracing subscriber in the global, cc @Boshen

ahabhgk: Screenshot 2023-11-16 at 17 29 39

Boshen: oxc_resolver has `tracing` in it, let's try and enable it on the rspack side.

hardfist:
!eco-ci

hardfist:
@ahabhgk please take a look the eco-ci result

PR #4249, opened 2023-09-26
chore: add builtins.react/... only works for transformByDefault tip

9aoy:
image

## Require Documentation?

- [ ] No

### DIFF
--- a/packages/rspack/src/builtin-plugin/index.ts
+++ b/packages/rspack/src/builtin-plugin/index.ts
@@ -294,7 +294,7 @@ export function deprecated_resolveBuiltins(
 				deprecatedWarn(
 					`'builtins.${key} = ${JSON.stringify(
 						builtins[key]
-					)}' has been deprecated, please migrate to ${termlink(
+					)}' only works for transformByDefault, please migrate to ${termlink(
 						"builtin:swc-loader options",
 						"https://www.rspack.dev/guide/loader.html#builtinswc-loader"
 					)}`,
--- a/packages/rspack/src/builtin-plugin/index.ts
+++ b/packages/rspack/src/builtin-plugin/index.ts
@@ -294,7 +294,7 @@ export function deprecated_resolveBuiltins(
 				deprecatedWarn(
 					`'builtins.${key} = ${JSON.stringify(
 						builtins[key]
-					)}' only works for transformByDefault, please migrate to ${termlink(
+					)}' only works for 'rspack.disableTransformByDefault = false', please migrate to ${termlink(
 						"builtin:swc-loader options",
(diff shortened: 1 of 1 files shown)


packages/rspack/src/builtin-plugin/index.ts
```diff
@@ -279,6 +279,31 @@ export function deprecated_resolveBuiltins(
...
+		).forEach(key => {
+			if (builtins[key]) {
+				deprecatedWarn(
+					`'builtins.${key} = ${JSON.stringify(
+						builtins[key]
+					)}' has been deprecated, please migrate to ${termlink(
+						"builtin:swc-loader options",
+						"https://www.rspack.dev/guide/loader.html#builtinswc-loader"
+					)}`,
+					true
```
h-a-n-a, line 301: Would you remove the second parameter? Adding `true` here will result in this cannot be turned off by env.

9aoy: It may be inappropriate to use deprecatedWarn, because it doesn't worked when disableTransformByDefault is turned on, so I don't think it can be turned off by env.

h-a-n-a (dismissed):

packages/rspack/src/builtin-plugin/index.ts
```diff
@@ -279,6 +279,31 @@ export function deprecated_resolveBuiltins(
...
+				"presetEnv",
+				"emotion",
+				"relay"
+			] as const
+		).forEach(key => {
+			if (builtins[key]) {
+				deprecatedWarn(
+					`'builtins.${key} = ${JSON.stringify(
+						builtins[key]
+					)}' only works for transformByDefault, please migrate to ${termlink(
```
ahabhgk, line 297:
```suggestion
					)}' only works for 'rspack.disableTransformByDefault = false', please migrate to ${termlink(
```

packages/rspack/src/builtin-plugin/index.ts
```diff
@@ -279,6 +279,31 @@ export function deprecated_resolveBuiltins(
...
+				"presetEnv",
+				"emotion",
+				"relay"
+			] as const
+		).forEach(key => {
+			if (builtins[key]) {
+				deprecatedWarn(
+					`'builtins.${key} = ${JSON.stringify(
+						builtins[key]
+					)}' only works for 'rspackFuture.disableTransformByDefault = false', please migrate to ${termlink(
```
h-a-n-a, line 297: Let's use the full path.
```suggestion
					)}' only works for 'experiments.rspackFuture.disableTransformByDefault = false', please migrate to ${termlink(
```

h-a-n-a (approved):

Issue #4081, opened 2023-08-31
not allow use swc target and env configuration toghether


hardfist:
we should not allow use jsc.target and env toghether, since they are incompatible and may cause unexpected behavior
![image](https://github.com/web-infra-dev/rspack/assets/8898718/49e14c25-7880-4fcd-9579-db98abcb3ca7)

hardfist:
guidance
you can use emit_diagnostic generate warnings for conflict options for rspack_loader_swc
https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_loader_swc/src/lib.rs#L80

Hamzakh777:
assign it to me @hardfist

hardfist:
thanks

PR #4257, opened 2023-09-26
fix(rspack_loader_swc): Fail the build when jsc.target and env are used together for the SWC loader

Hamzakh777:
## Summary
This PR adds a check for `rspack_loader_swc` config so the user doesn't use both `env` and `jsc.target` at the same time. This is fixed in accordance with the official way its done under SWC here: https://github.com/swc-project/swc/pull/7704/files#diff-05ec577011f6096aac154c266838a96b40e857713f31a3441fc73baf1d90580eR376

Ticket: #4081
Screenshot 2023-09-26 at 22 01 03

## Test Plan

- Go to `examples/builtin-swc-loader/rspack.config.js`
- Under the `builtin:swc-loader` loader options add `jsx.target: "es2016"` and `env.targets: "es2016"`
- Run the build ` pnpm --filter "./examples/builtin-swc-loader" build`
- Build should fail

## Require Documentation?
- [x] No

### DIFF
--- a/crates/rspack_loader_swc/src/lib.rs
+++ b/crates/rspack_loader_swc/src/lib.rs
@@ -85,6 +85,15 @@ impl Loader<LoaderRunnerContext> for SwcLoader {
       ));
     }
 
+    if swc_options.config.jsc.target.is_some() && swc_options.config.env.is_some() {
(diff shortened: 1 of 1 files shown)


h-a-n-a (approved):

h-a-n-a:
Thank you!

PR #4267, opened 2023-09-28
feat: handle `react-refresh` resolving in `@rspack/plugin-react-refresh`

h-a-n-a:
## Test Plan

CI passes. Should be working if `react-refresh` is removed from the package.

## Require Documentation?

- [X] No

### DIFF
--- a/packages/rspack-plugin-react-refresh/src/index.js
+++ b/packages/rspack-plugin-react-refresh/src/index.js
@@ -44,6 +44,10 @@ module.exports = class ReactRefreshRspackPlugin {
 		 */
 		this.options = normalizeOptions(options);
 	}
+
+	/**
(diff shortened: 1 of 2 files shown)


ScriptedAlchemy (approved):

packages/rspack-plugin-react-refresh/src/index.js
```diff
@@ -59,6 +63,12 @@ module.exports = class ReactRefreshRspackPlugin {
 			},
 			use: "builtin:react-refresh-loader"
 		});
+
+		const refreshPath = path.dirname(require.resolve("react-refresh"));
```
hardfist, line 67: we need to tell users to install react-refresh manully on docs, otherwise it will throw error to user, cc @ahabhgk

Issue #4261, opened 2023-09-27
Dynamic import can not generate async chunks as expected


chenjiahan:
### System Info

  System:
    OS: macOS 13.4.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 282.33 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.20.1 - ~/Library/Caches/fnm_multishells/58681_1695821782583/bin/node
    Yarn: 1.22.19 - ~/Library/Caches/fnm_multishells/58681_1695821782583/bin/yarn
    npm: 8.19.4 - ~/Library/Caches/fnm_multishells/58681_1695821782583/bin/npm
    pnpm: 8.6.12 - ~/Library/Caches/fnm_multishells/58681_1695821782583/bin/pnpm
  Browsers:
    Chrome: 117.0.5938.92
    Safari: 16.5.2
  npmPackages:
    @rspack/cli: latest => 0.3.5 

### Details

1. Use `React.lazy` and dynamic import:

```js
import { lazy } from "react";

const LazyComponent = lazy(() => import("./Component"));

console.log(LazyComponent);
```

2. Add `enforceSizeThreshold` config:

```js
{
  optimization: {
    splitChunks: {
      chunks: "all",
      enforceSizeThreshold: 50000,
    },
  },
}
```

Webpack can split dynamic chunks as expected:

![image](https://github.com/web-infra-dev/rspack/assets/7237365/f6668d10-eeca-4cbf-b09c-34df65dd382d)

Rspack can not split the dynamic chunk:

![image](https://github.com/web-infra-dev/rspack/assets/7237365/a4087f74-d4ba-4664-8e63-6b15b1b0e674)

### Reproduce link

https://github.com/chenjiahan/rspack-dynamic-import-issue

### Reproduce Steps

1. pnpm i
2. pnpm build

JSerFeng:
Fixed by #4280, you can try using 0.3.5-canary-10b7fe6-20231008143835

PR #4280, opened 2023-10-03
feat(split chunks): split chunks should consider chunk combinations

JSerFeng:
## Summary

Fix: #4261 

Add combinations for splitChunks.

Before, every cacheGroup will be a splitted chunk
imagine if module `a` exists in entryA and entryB, but `b` exists in entryA, not in entryB, both `a` and `b` will be splitted to a new chunk, so in the end. there are 3 chunks, entryA, entryB, and shared splitted chunk. EntryB ends up including extra module `b`.

We should use module's combinations as key, for splitted chunks, one cacheGroup may end up with multiple chunks.

Before:
```
key                 value
cacheGroup index 1: { chunks: [entryA, entryB], modules: [a, b] }
```

Now:
```
key  value
ab: { chunks: [entryA], modules: [a, b] }
b:  { chunks: [entryB], modules: [b] }
```

## Require Documentation?

- [x] No


crates/rspack_plugin_split_chunks_new/src/plugin/module_group.rs
```diff
@@ -240,4 +308,70 @@ impl SplitChunksPlugin {
...
+        .entry(count)
+        .and_modify(|set| set.push(chunks.clone()))
+        .or_insert(Vec::new());
+    }
+
+    (chunk_sets_in_graph, single_chunk_sets, chunk_sets_by_count)
+  }
+}
+
+fn is_sub_set<T: std::hash::Hash + std::cmp::Eq>(
```
hyf0, line 362: You could use this API https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.is_subset

hyf0:
Well done. This is a wonderful feature.

JSerFeng:
!canary

JSerFeng:
!canary

ahabhgk (dismissed):

JSerFeng:
!bench

hardfist:
it seem benchmark machine is down...

ahabhgk (approved):

f0reclone:
@JSerFeng when using the latest canary build which includes this fix, the bundle size did go up from 36MB to 48MB.

The splitting seems to be correct (every node module in a new chunk).

hardfist:
!canary

PR #4310, opened 2023-10-12
fix: should have null as placeholder in stats chunks before chunkIds

JSerFeng:
1. Chunk id only exists after the call of plugin.chunkIds, before this call, user may access getStats(), it should not panic.

Align with webpack, before the chunkIds call, the chunks fields in stats should be `null` instead of panic

2. Compilation.chunks should have more information for plugin authors

3. New crate rspack_binding_values

## Require Documentation?

- [ ] No


JSerFeng:
!bench

h-a-n-a (dismissed):

crates/rspack_core/src/lib.rs
```diff
@@ -94,6 +94,18 @@ pub enum SourceType {
   Unknown,
 }
 
+impl ToString for SourceType {
```
ahabhgk, line 97: `fmt::Display` instead

ahabhgk (approved):

Issue #3180, opened 2023-05-17
Misleading error message when request is `""`


IWANABETHATGUY:
### System Info

system unrelated

### Details

```js
import {} from '';

```

rspack will throw error: 
```bash
'should have context', crates/rspack_plugin_externals/src/plugin.rs:126:16

```

### Reproduce Steps

1. npm create @rspack/cli@latest
2. change index.js to above
3. pnpm build

hardfist:
bump

Hamzakh777:
assign it to me @hardfist

hardfist:
thanks

Hamzakh777:
@hardfist it seems now the compiler is throwing a different error: 
![Screenshot 2023-09-28 at 16 47 07](https://github.com/web-infra-dev/rspack/assets/40059557/faeb8f42-e2fe-47ca-b34b-2e6efd7ae82a)
This is coming from this line: https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/normal_module_factory.rs#L247
Should we use `ModuleFactoryResult` to return the error instead?

IWANABETHATGUY:
duplicated with #5082

PR #4320, opened 2023-10-13
fix(3180): panicking when importing empty dependency

Hamzakh777:
## Summary
This PR fixes an issue where Rspacks panics when importing an empty dependency `""`. The solution tries to be as close as possible to Webpacks https://github.com/webpack/webpack/blob/main/lib/NormalModuleFactory.js#L310-L312

Screenshot 2023-10-13 at 15 21 56

Github Issue: #3180 
## Require Documentation?

- [x] No


h-a-n-a:
Thanks for contributing would you please add a test case?

Hamzakh777:
@h-a-n-a where would I add that test case?

hardfist:
packages/rspack/tests/cases/resolve here

Hamzakh777:
@hardfist @h-a-n-a is there any example I can refer to for testing against failures?

hardfist:
> @hardfist @h-a-n-a is there any example I can refer to for testing against failures?

actually webpack has a test case for it https://github.com/webpack/webpack/tree/main/test/cases/parsing/issue-2006, so we could just enable this test case in rspack

Hamzakh777:
@hardfist wouldn't it be better to use `TraceableError` instead of `internal_error`? As the error is traceable to the original source file

hardfist:
> @hardfist wouldn't it be better to use `TraceableError` instead of `internal_error`? As the error is traceable to the original source file. If we show the file path, plus where exactly in the file the error is coming from that would be a much better DX

I think traceable error seems better now, FYI our traceableError seems have some limitation now and @h-a-n-a is working on refactor the error handle.

Hamzakh777:
@hardfist okay, I'll work on refactoring the code to use `TraceableError`. 
On another note, it seems the test is failing, how does the error propagate from the Rust side to the nodejs side?

h-a-n-a:
> @hardfist okay, I'll work on refactoring the code to use `TraceableError`. On another note, it seems the test is failing, how does the error propagate from the Rust side to the nodejs side?

@Hamzakh777 You can check out [this](https://github.com/web-infra-dev/rspack/blob/455ed97eab7d0e35e77c93314c975b6fd0b6aade/crates/rspack_napi_shared/src/errors.rs#L1) file as a reference. However, at this time, I will not recommend you changing this file as it's under refactor. :-)

hardfist:
since we have not finished align rspack error to webpack error，pass the webpack test without modifying the test case is hard，we need copy this case to rspack tests folder

hardfist:
@Hamzakh777 i can help move the test if you dont mind

Hamzakh777:
sure go for it @hardfist

hardfist:
@Hamzakh777  the ci failed because it failed to generate assets which cause jest can rust the test, so we need to make sure the artifacts generated successfully even resolve failed, I'll investigate furthrer

h-a-n-a:
Sorry for the delay, I will take a look at this issue ASAP.

h-a-n-a:
Ok, I think this test is not quite related to the thing that you've been working on. It relies on the other feature. Would you please add a new test to `packages/rspack/tests/configCases/errors` with something like `empty-dependency-error`

h-a-n-a:
Let's merge this first and solve the error in the refactor error PR :-).

h-a-n-a:
Thanks!

h-a-n-a (approved):

Issue #4314, opened 2023-10-13
rspack cli should not print help information when config file is not found and other errors.


XGHeaven:
### System Info

In Github CodeSpaces
```
  System:
    OS: Linux 6.2 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (4) x64 AMD EPYC 7763 64-Core Processor
    Memory: 13.30 GB / 15.61 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 16.18.0 - /usr/local/share/nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - /usr/bin/yarn
    npm: 8.19.2 - /usr/local/share/nvm/versions/node/v16.18.0/bin/npm
  npmPackages:
    @rspack/cli: 0.3.5 => 0.3.5 
    @rspack/plugin-html: 0.3.5 => 0.3.5 
```

### Details

When using a not existing config file. The error messages included the help message that was like from `--help`.

<img width="849" alt="image" src="https://github.com/web-infra-dev/rspack/assets/9291212/4d648c41-df8f-492a-8968-3426b422f492">

### Reproduce Steps

1. Install Rspack cli
2. Run `npx rspack build -c a.js` to run a not existing config file.

hardfist:
it's indeed a bug, are you willing to fix it for us?

OceanPresentChao:
> it's indeed a bug, are you willing to fix it for us?

Would you mind assigning it to me?

hardfist:
thanks

hardfist:
landed in #4322

PR #4322, opened 2023-10-15
fix(4314): remove help info print when throw error

OceanPresentChao:
## Summary

GitHub Issue: #4314 

This bug is exactly caused by the internal bug in "yargs". when we use .command method in yargs, if the async handler function we pass in throws an error, it will result in an unexpected "--help" command execution.

I also propose an issue for "yargs", maybe they will fix this bug.
about the bug detail you can see [yargs issue 2368](https://github.com/yargs/yargs/issues/2368)

For rspack, I wrap the handler code with a "try-catch" block to catch the error we throw. In the catch block, we should exit the process by calling "process.exit()" ourselves instead of throwing an error, which prevents the internal bug in "yargs".

Why don't use "try-catch" block wrap `await this.program.parseAsync(hideBin(argv));`?

I have tried this, but it doesn't work, It may be because the handler function is called asynchronously, so the error it throws cannot be captured by the outer "try-catch" block.

## Test Plan

make sure "a.js" is a file not exist, run `rspack build -c a.js`
it is expected that only error info is outputted,without "--help" info.
## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack-cli/src/rspack-cli.ts
+++ b/packages/rspack-cli/src/rspack-cli.ts
@@ -94,6 +94,7 @@ export class RspackCLI {
 			);
 		}
 
+		this.program.showHelpOnFail(false);
 		this.program.usage("[options]");
 		this.program.scriptName("rspack");
 		this.program.strictCommands(true).strict(true);


hardfist:
@OceanPresentChao thank you for your detail explanation, it seems yargs has an option called `showHelpOnFail`, have you tried set it to false and see whether it works?

OceanPresentChao:
> @OceanPresentChao thank you for your detail explanation, it seems yargs has an option called `showHelpOnFail`, have you tried set it to false and see whether it works?

Oh I have tried it, it works! It's not a bug, my bad.
But, if the handler function is not async, it will not show help info even though we do not specify `showHelpOnFail` as false. So I think it is a bug at first, hahaha

hardfist (approved):

hardfist:
thanks

PR #4352, opened 2023-10-17
chore: be more accurate

hardfist:
## Summary

best-effort makes more sense because 100% compatible is non-goal for rspack which may mislead some users
## Test Plan
No Need

## Require Documentation?

- [x] No


Boshen:
Webpack compatible is our design principle. Adding a "best-effort" is a [slippery slope](https://en.wikipedia.org/wiki/Slippery_slope) argument. Let's sleep on this.

Boshen:
I think the current terminology is correct here - Rspack strives for compatibility with the Webpack ecosystem.

After reading the discussion in one of our user discussion group, it seems like someone interpreted it as "Rspack has 1:1 API compatibility with Webpack". I think we never stated such terms. Such interpretation should be explained to the user in the discussion group instead of changing our current statement about compatibility, which will lead to other confusions in the future.

PR #4262, opened 2023-09-28
feat: source order for harmony import dependency

ahabhgk:
## Summary

- sort initFragment by position
- refactor initFragment, and implement merge initFragment
- add source_order for harmony import dependency, and the source_order will be the position of generated initFragment
- render "use strict" by `build_info.strict`

## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #4356, opened 2023-10-18
fix: variable undefined with multiple same import

ahabhgk:
## Summary

checkout added test case

after we introduce [source_order](https://github.com/web-infra-dev/rspack/pull/4262) for new tree shaking implementation, HarmonyImportSideEffectDependency will be duplicated, and merged by InitFragment, this is aligned behavior with webpack, before it won't be duplicated

```js
// before:
import {a} from "./lib" // => Dependency(100 /* dependency_id */, "./lib")
import {b} from "./lib" // => Dependency(100 /* dependency_id */, "./lib")

```

```js
// after:
import {a} from "./lib" // => Dependency(100 /*dependency_id*/, "./lib", 0 /* source_order */)
import {b} from "./lib" // => Dependency(101 /*dependency_id*/, "./lib", 1 /* source_order */)
```

but in the old tree shaking implementation will find dependency by its request and dependency_type, so when we want to find `Dependency(101)`, we actually will find `Dependency(100)`, if `Dependency(100)` is unused, then we won't generate the code, so `Dependency(101)` will be removed mistakenly

## Test Plan

- packages/rspack/tests/configCases/tree-shaking/multiple-same-import-side-effect

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/infer_async_modules_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/infer_async_modules_plugin.rs
@@ -44,8 +44,8 @@ impl Plugin for InferAsyncModulesPlugin {
           .incoming_connections_unordered(module_graph)?
           .filter(|con| {
             if let Some(dep) = module_graph.dependency_by_id(&con.dependency_id) {
-              *dep.dependency_type() == DependencyType::EsmImport
-                || *dep.dependency_type() == DependencyType::EsmExport
+              matches!(dep.dependency_type(), DependencyType::EsmImport(_))
+                || matches!(dep.dependency_type(), DependencyType::EsmExport)
             } else {
               false
             }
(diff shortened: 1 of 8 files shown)


crates/rspack_plugin_javascript/src/plugin/infer_async_modules_plugin.rs
```diff
@@ -44,8 +44,8 @@ impl Plugin for InferAsyncModulesPlugin {
           .incoming_connections_unordered(module_graph)?
           .filter(|con| {
             if let Some(dep) = module_graph.dependency_by_id(&con.dependency_id) {
-              *dep.dependency_type() == DependencyType::EsmImport
-                || *dep.dependency_type() == DependencyType::EsmExport
+              matches!(dep.dependency_type(), DependencyType::EsmImport(_))
```
IWANABETHATGUY, line 47:
```rs
              matches!(
                dep.dependency_type(),
                DependencyType::EsmImport(_) | DependencyType::EsmExport
              )
```

IWANABETHATGUY (approved):

Issue #4335, opened 2023-10-16
Support splitChunks.automatic_name_delimiter


JSerFeng:
should support splitChunks.automatic_name_delimiter

OceanPresentChao:
there are `rspack_plugin_split_chunks_new` and `rspack_plugin_split_chunks`, should we support `splitChunks.automatic_name_delimiter` for both of them?

hardfist:
> there are `rspack_plugin_split_chunks_new` and `rspack_plugin_split_chunks`, should we support `splitChunks.automatic_name_delimiter` for both of them?

just the new

PR #4365, opened 2023-10-19
feat: support `splitChunks.automaticNameDelimiter`

OceanPresentChao:
## Summary

[issue#4335](https://github.com/web-infra-dev/rspack/issues/4335)

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_binding_options/src/options/raw_optimization.rs
+++ b/crates/rspack_binding_options/src/options/raw_optimization.rs
@@ -36,6 +36,10 @@ impl RawOptionsApply for RawOptimizationOptions {
     self,
     plugins: &mut Vec<Box<dyn rspack_core::Plugin>>,
   ) -> Result<Self::Options, rspack_error::Error> {
+    let delimiter: Option<String> = match self.split_chunks.as_ref() {
+      Some(options) => options.automatic_name_delimiter.clone(),
+      _ => None,
+    };
     if let Some(options) = self.split_chunks {
       let split_chunks_plugin = IS_ENABLE_NEW_SPLIT_CHUNKS.with(|is_enable_new_split_chunks| {
         if *is_enable_new_split_chunks {
@@ -48,8 +52,8 @@ impl RawOptionsApply for RawOptimizationOptions {
       plugins.push(split_chunks_plugin);
     }
     let chunk_ids_plugin = match self.chunk_ids.as_ref() {
-      "named" => NamedChunkIdsPlugin::new(None, None).boxed(),
-      "deterministic" => DeterministicChunkIdsPlugin::default().boxed(),
+      "named" => NamedChunkIdsPlugin::new(delimiter, None).boxed(),
+      "deterministic" => DeterministicChunkIdsPlugin::new(delimiter, None).boxed(),
       _ => {
         return Err(internal_error!(
           "'chunk_ids' should be 'named' or 'deterministic'."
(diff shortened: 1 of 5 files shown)


JSerFeng (dismissed):

OceanPresentChao:
An example:
If we have:
image
we can get:
image

crates/rspack_binding_options/src/options/raw_optimization.rs
```diff
@@ -36,7 +36,12 @@ impl RawOptionsApply for RawOptimizationOptions {
     self,
     plugins: &mut Vec<Box<dyn rspack_core::Plugin>>,
   ) -> Result<Self::Options, rspack_error::Error> {
+    let mut delimiter = "~".to_string();
```
ahabhgk, line 39: default logic should move to js side, checkout https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/config/defaults.js#L1368

OceanPresentChao: done

crates/rspack_binding_options/src/options/raw_optimization.rs
```diff
@@ -48,8 +52,8 @@ impl RawOptionsApply for RawOptimizationOptions {
       plugins.push(split_chunks_plugin);
     }
     let chunk_ids_plugin = match self.chunk_ids.as_ref() {
-      "named" => NamedChunkIdsPlugin::new(None, None).boxed(),
-      "deterministic" => DeterministicChunkIdsPlugin::default().boxed(),
+      "named" => NamedChunkIdsPlugin::new(delimiter, None).boxed(),
+      "deterministic" => DeterministicChunkIdsPlugin::new(delimiter, None).boxed(),
```
ahabhgk, line 56: split_chunks.automatic_name_delimiter shouldn't affect how chunk id generated, webpack doesn't, it should affect how split_chunks_plugin works

crates/rspack_binding_options/src/options/raw_split_chunks.rs
```diff
@@ -24,7 +24,7 @@ pub struct RawSplitChunksOptions {
   #[napi(ts_type = "RegExp | 'async' | 'initial' | 'all'")]
   #[derivative(Debug = "ignore")]
   pub chunks: Option<Chunks>,
-  //   pub automatic_name_delimiter: String,
+  pub automatic_name_delimiter: Option<String>,
```
ahabhgk, line 27: since the default value is assigned on js side, this will never be None, this can change to `String`

crates/rspack_binding_options/src/options/raw_optimization.rs
OceanPresentChao: Yes, but It seems that rspack names the output files depending on chunk id partly. I have posted an example above: https://github.com/web-infra-dev/rspack/pull/4365#issuecomment-1770433366. Maybe I have a wrong understanding of how `automatic_name_delimiter` affects the generated files' name. Could you give me some tutorials? Thanks!

OceanPresentChao: > split_chunks.automatic_name_delimiter shouldn't affect how chunk id generated, webpack doesn't, it should affect how split_chunks_plugin works

I think I get it. But in rspack, it involves some troubled arguments passing, I will try it later. =w=

JSerFeng:
The automatic_name_delimiter can be found here https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_split_chunks_new/src/plugin/max_size.rs#L170, I think just pass it a map like chunkUkey -> name_delimiter, and replace the hard coded `~` according to the map

And note that don't need to handle the old split_chunks.

LingyuCoder:
Thanks for your contribution. This feature has been implemented in #4930, so I will close this. Welcome to contribute code to other features.

PR #4412, opened 2023-10-24
fix: use indirect call for TaggedTpl

### LINKED ISSUE #4380: import TaggedTemplateExpression should generate indirect call

Please get context from https://github.com/webpack/webpack/issues/17390.

-

bvanjoi:
Fixes #4380

The ﻿`TaggedTemplate` should be marked with ﻿`enter_call`. Consequently, it will generate an indirect call in the form of `﻿(0, xxxx)()`.

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
@@ -4,18 +4,13 @@ use rspack_core::{
   ConstDependency, DependencyType, SpanExt,
 };
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
-use swc_core::{
-  common::Span,
-  ecma::{
-    ast::{
-      AssignExpr, AssignOp, Callee, ExportAll, ExportSpecifier, Expr, Id, Ident, ImportDecl,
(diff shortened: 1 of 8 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
```diff
@@ -4,18 +4,13 @@ use rspack_core::{
...
-      ImportSpecifier, Lit, MemberExpr, MemberProp, ModuleExportName, NamedExport, Pat, PatOrExpr,
-      Program, Prop,
-    },
-    atoms::JsWord,
-    visit::{noop_visit_type, Visit, VisitWith},
-  },
-};
+use swc_core::atoms::JsWord;
+use swc_core::common::Span;
+use swc_core::ecma::ast::{AssignExpr, AssignOp, MemberExpr, MemberProp};
```
IWANABETHATGUY, line 9: Maybe we could merge these ast imports?

bvanjoi: This is intentional. I think using the split could makes the code more readable.

IWANABETHATGUY: I could understand you split `atoms`, `common`, but does it make any difference split `ast` into four parts?

IWANABETHATGUY (approved):

ahabhgk (approved):

Issue #4396, opened 2023-10-23
support js inline loader query options


ahabhgk:
```js
import a from "./loader?count=1000!./module";
```
loader.js:
```js
module.exports = function () {
  const options = this.getOptions(); // should be `{ count: 1000 }`
};
```

- webpack-test/cases/large/many-replacements

PR #4418, opened 2023-10-25
feat: support passing query to loader

h-a-n-a:
## Summary

resolves https://github.com/web-infra-dev/rspack/issues/4396

## Test Plan

Added

## Require Documentation?

- [X] No


ahabhgk:
I don't suggest modify test under `webpack-test`, even if there are errors/warnings didn't match, this didn't mean we need to align all of these with webpack, this is only for tracking those mismatch

I made some change on this PR (revert changes under `webpack-test` and copy it into our test dir, and enable more `webpack-test`s which blocked by this)

ahabhgk (dismissed):

PR #4448, opened 2023-10-26
fix: nested member chain

IWANABETHATGUY:
## Summary
1. Passing more tree shaking snapshot. (5 got passed)
2. You could ignore those commented `dbg!`, I will clean up them at once when all snapshot got passed.

## Require Documentation?

- [ ] No


crates/rspack/tests/fixtures.rs
```diff
@@ -40,9 +41,9 @@ fn tree_shaking(fixture_path: PathBuf) {
         options.optimization.used_exports = UsedExportsOption::True;
         options.builtins.tree_shaking = TreeShaking::False;
 
-        if options.optimization.side_effects.is_enable() {
```
IWANABETHATGUY, line 43: disable for now,  I will recover when after covering snapshot tests.

IWANABETHATGUY: temporarily disabled to reduce some noise when investigating diff compared with the previous implementation.

crates/rspack_core/src/exports_info.rs
```diff
@@ -1169,9 +1177,14 @@ impl ExportInfo {
     &mut self,
     key: Option<DependencyId>,
     connection: Option<ModuleGraphConnection>,
-    export_name: Option<&Vec<JsWord>>,
+    export_name: Option<&Nullable<Vec<JsWord>>>,
```
IWANABETHATGUY, line 1180: I am sorry to write code like this, but that's what webpack did.
here `undefined` , `null`, `string[]` has difference semantic.
**undefined**
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/ExportsInfo.js#L346-L351
**string[]** 
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/ExportsInfo.js#L313-L318
**null** 
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/FlagDependencyExportsPlugin.js#L199
https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/Dependency.js#L60-L61

`exportName` will be fallbacked to `[name]` only `fromExport` is equal to `undefined`, https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/FlagDependencyExportsPlugin.js#L247

ahabhgk (approved):

PR #4466, opened 2023-10-27
feat: support weak of `import.meta.wabpackContext`

bvanjoi:
Implement  support for ﻿`import.meta.webpackContext` with weak

we still lack `prefetch/preload`, `include/exclude`, `chunkName`...


ahabhgk (approved):
🐐

PR #4470, opened 2023-10-30
chore: disable timing and hash by default for stats.toString()

hardfist:
## Summary
I brings extra burden for outside contributors to disable timing and hash for stat snapshot and which not aligns with webpack behaviors too.
so we align this behavior with webpack(disable hash and timing by default in string snapshot)

## Require Documentation?

- [x] No


packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
```diff
@@ -1339,6 +1340,7 @@ exports[`StatsTestCases should print correct stats for limit-chunk-count-plugin
       "outputPath": "<PROJECT_ROOT>/tests/statsCases/limit-chunk-count-plugin/dist",
       "publicPath": "",
       "rspackVersion": "0.3.8",
+      "time": 64,
```
ahabhgk, line 1343: Remove this

packages/rspack/tests/statsCases/limit-chunk-count-plugin/webpack.config.js
```diff
@@ -15,8 +15,8 @@ module.exports = [1, 2, 3, 4].map(n => ({
 		})
 	],
 	stats: {
-		hash: false,
-		timings: false,
```
ahabhgk, line 19: Add this back to keep snapshot stable

packages/rspack/package.json
```diff
@@ -15,7 +15,7 @@
   "scripts": {
     "build": "tsc -b ./tsconfig.build.json",
     "dev": "tsc -w",
-    "test": "cross-env NO_COLOR=1 RSPACK_DEP_WARNINGS=false node --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --runInBand --logHeapUsage"
+    "test": "cross-env NO_COLOR=1 RSPACK_DEP_WARNINGS=false node --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --runInBand --logHeapUsage --updateSnapshot"
```
ahabhgk, line 18: I suggest adding a new script to update instead of changing this

hardfist: forget to revert it back

packages/rspack/tests/statsCases/limit-chunk-count-plugin/webpack.config.js
hardfist: this is not included in original webpack tests, so we may don't need this

packages/rspack/tests/__snapshots__/StatsTestCases.test.ts.snap
hardfist: I will check out the default behavior of webpack

packages/rspack/package.json
hardfist: reverted

ahabhgk (dismissed):

PR #4497, opened 2023-11-01
feat: replace top-level this in esm

### LINKED ISSUE #4421: should replace top-level this to undefined in esm
https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/dependencies/HarmonyTopLevelThisParserPlugin.js#L22

- webpack-test/cases/parsing/es2020

Austaras:
## Summary

Closes #4421

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_top_level_this.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_top_level_this.rs
@@ -1,0 +1,70 @@
+use rspack_core::{ConstDependency, DependencyTemplate, SpanExt};
+use swc_core::ecma::{
+  ast::{
+    ClassMember, ClassMethod, ClassProp, Expr, Function, GetterProp, MethodProp, Prop, PropName,
+    SetterProp, ThisExpr,
+  },
+  visit::{noop_visit_type, Visit, VisitWith},
+};
+
+pub struct HarmonyTopLevelThis<'a> {
+  pub presentational_dependencies: &'a mut Vec<Box<dyn DependencyTemplate>>,
+}
+
+impl Visit for HarmonyTopLevelThis<'_> {
+  noop_visit_type!();
+
+  fn visit_function(&mut self, _: &Function) {}
+
+  fn visit_class_member(&mut self, n: &ClassMember) {
+    match n {
+      ClassMember::Method(ClassMethod {
+        key: PropName::Computed(computed),
+        ..
+      })
+      | ClassMember::ClassProp(ClassProp {
+        key: PropName::Computed(computed),
+        ..
+      }) => {
+        computed.visit_with(self);
+      }
+      _ => {}
+    }
+  }
+
+  fn visit_prop(&mut self, n: &Prop) {
+    match n {
+      Prop::KeyValue(..) => {
+        n.visit_children_with(self);
+      }
+      Prop::Getter(GetterProp {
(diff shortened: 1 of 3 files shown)


ahabhgk:
I think now `webpack-test/cases/parsing/es2020` can pass, could you enable it, just need to make `test.filter.js` return true

IWANABETHATGUY:
> I think now `webpack-test/cases/parsing/es2020` can pass, could you enable it, just need to make `test.filter.js` return true, and run `pnpm run test -- -t "parsing es2020"` under `webpack-test` folder

string replacing-based codegen required ast to be untouched (some plugins needs span info), if we add an extra transforming pass, it may affect the optimization result. Can we move the transforming work to the loader side?

IWANABETHATGUY:
Or just like the webpack, use a`ConstDependency` instead.

Austaras:
Turns out the origin impl doesn't work at all. Two separate AST is used for codegen and dep analysis, and modify the latter one doesn't have any effect.

Austaras:
> > I think now `webpack-test/cases/parsing/es2020` can pass, could you enable it, just need to make `test.filter.js` return true, and run `pnpm run test -- -t "parsing es2020"` under `webpack-test` folder
> 
> string replacing-based codegen required ast to be untouched (some plugins needs span info), if we add an extra transforming pass, it may affect the optimization result. Can we move the transforming work to the loader side?

Seems like most passes in `transform_before_pass` belong in the loader, but for this PR I believe it's better not to change this logic.

crates/rspack_plugin_javascript/src/visitors/mod.rs
```diff
@@ -240,6 +242,7 @@ pub fn run_before_pass(
       builtins_webpack_plugin(options, unresolved_mark),
       // This will be deprecated in the future when TypeScript support and transform by default is dropped.
       // These features are coupled with this.
+      replace_top_level_this(unresolved_mark),
```
ahabhgk, line 245: And move this under this branch https://github.com/web-infra-dev/rspack/blob/3c0c6ca9d9cc5f99919520288006e33f14867e57/crates/rspack_plugin_javascript/src/visitors/dependency/mod.rs#L131

ahabhgk (commented):
It would be better if you can implement it by string-replacement way (use `ConstDependency` to replace `this` to `undefined`), this is pretty similar with `ReplaceNestWebpackRequireVisitor`, so you can reference it at https://github.com/web-infra-dev/rspack/blob/3c0c6ca9d9cc5f99919520288006e33f14867e57/crates/rspack_plugin_javascript/src/visitors/dependency/compatibility_scanner.rs#L52-L75

Austaras:
Done. But what's the difference between AST plugin and text plugin? What's the necessity of the latter?

ahabhgk:
We used ast-based replacement for all our previous implementations, but we found it challenging to implement caching and future persistent caching using ast compared to string replacement. Therefore, we switched to the implementation of string replacement, which is currently in an intermediate stage. We currently have both string replacement and ast replacement, as there are still some blockers for the remaining ast replacement. In the long term, we will replace all of it with string replacement.

ahabhgk:
LGTM, can you make the CI passes

Austaras:
Interesting.

ahabhgk (approved):

ahabhgk:
Thanks!

PR #4503, opened 2023-11-01
chore(runtime-diff): alignment of get trusted type policy runtime module

LingyuCoder:
## Summary

Alignment of get trusted type policy runtime module.
- Adding logic of generating code by runtime requirements

> output.trustedTypes.onPolicyCreationFailure is not supported in Rspack yet

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_runtime/src/runtime_module/get_trusted_types_policy.rs
+++ b/crates/rspack_plugin_runtime/src/runtime_module/get_trusted_types_policy.rs
@@ -1,6 +1,6 @@
 use rspack_core::{
   rspack_sources::{BoxSource, RawSource, SourceExt},
-  Compilation, RuntimeModule,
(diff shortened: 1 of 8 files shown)


crates/rspack_plugin_runtime/src/runtime_module/get_trusted_types_policy.rs
```diff
@@ -31,11 +35,35 @@ impl RuntimeModule for GetTrustedTypesPolicyRuntimeModule {
...
+      policy_content.push(
+        r#"
+        createScriptURL: function (url) {
+          return url;
+        }
+        "#
+        .to_string(),
+      );
+    }
+    result = result.replace("$policyContent$", policy_content.join(",").as_ref());
```
JSerFeng, line 64: Do we need \n here: policy_content.join(",\n"), like webpack

JSerFeng (approved):

PR #4510, opened 2023-11-01
feat: support `module.parser.javascript.dynamicImportMode: "eager"`

### LINKED ISSUE #1652: module.parser.javascript.dynamicImportMode eager need in node
See: https://webpack.js.org/configuration/module/#moduleparserjavascriptdynamicimportmode

bvanjoi:
Fixes #1652


crates/rspack_core/src/dependency/mod.rs
```diff
@@ -434,6 +445,22 @@ pub trait ModuleDependency: Dependency {
   }
 }
 
+pub trait ImportDependencyTrait: ModuleDependency {
```
ahabhgk, line 448: This seems unnecessary

bvanjoi: It's intentional. we will add `ImportWeakDependency` after now.

ahabhgk (approved):

PR #4555, opened 2023-11-07
fix: succeedModule hook panic when can't convert to jsModule

IWANABETHATGUY:
## Summary
1. fix: succeedModule hook panic when can't convert to jsModule

## Test Plan
1. Add contextModule in `succeedModule` hook test case.

## Require Documentation?

- [ ] No


crates/rspack_binding_values/src/module.rs
```diff
@@ -21,18 +21,41 @@ impl ToJsModule for dyn Module + '_ {
...
       .try_as_normal_module()
       .map(|normal_module| JsModule {
-        original_source,
-
-        resource: normal_module
-          .resource_resolved_data()
-          .resource_path
-          .to_string_lossy()
-          .to_string(),
+        original_source: original_source.clone(),
```
IWANABETHATGUY, line 28: use `Clone` because rustc not happy, but it is ok because once any `try_xxx` succeeds, the rest `or_else` function will not be executed.

IWANABETHATGUY:
!canary

h-a-n-a (approved):

PR #4572, opened 2023-11-08
feat(runtime): alignment of commonjs exports dependency

LingyuCoder:
## Summary

Alignment of commonjs exports dependency, 
- Currently commonjs tree shaking does not implemented in rspack, so used name will always return itself
- This is a pre-task for alignment of module level runtime requirements and commonjs tree shaking
- Runtime requirements will also affected the result of render_module

## Test Plan

- Test cases have been exists under test/cases/mjs

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_exports_dependency.rs
+++ b/crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_exports_dependency.rs
@@ -234,7 +234,7 @@ impl DependencyTemplate for CommonJsExportsDependency {
         panic!("Define property need value range");
       }
     } else {
-      panic!("Unexpect base type");
+      panic!("Unexpected base type");
     }
   }
 }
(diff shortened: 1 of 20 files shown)


crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_exports_dependency.rs
```diff
@@ -0,0 +1,240 @@
...
+          } else {
+            panic!("Unexpected base type");
+          }
+        } else {
+          init_fragments.push(
+            NormalInitFragment::new(
+              "var __webpack_unused_export__;\n".to_string(),
+              InitFragmentStage::StageConstants,
+              0,
+              InitFragmentKey::uniqie(),
```
IWANABETHATGUY, line 220: Do you mean `unique`?

LingyuCoder: Yes, but the previous code was unique. I'll modify it.

IWANABETHATGUY (approved):

Issue #4557, opened 2023-11-07
Panic in require.context regex


glenjamin:
### System Info

  System:
    OS: macOS 12.7
    CPU: (8) x64 Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
    Memory: 404.37 MB / 32.00 GB
    Shell: 5.1.8 - /usr/local/bin/bash
  Binaries:
    Node: 18.8.0 - ~/.asdf/installs/nodejs/18.8.0/bin/node
    npm: 8.18.0 - ~/.asdf/installs/nodejs/18.8.0/bin/npm
  Browsers:
    Chrome: 119.0.6045.105
    Edge: 119.0.2151.46
    Firefox: 119.0
    Safari: 17.0
  npmPackages:
    @rspack/cli: ^0.3.11 => 0.3.11

### Details

The following line in my code
```
const context = require.context('../src', true, /-story\.js$/);
```
causes this panic
```
> RUST_BACKTRACE=full npx -- rspack build
thread 'tokio-runtime-worker' panicked at crates/rspack_regex/src/algo.rs:110:31:
internal error: entered unreachable code
stack backtrace:
   0:        0x11751be7b - _napi_register_module_v1
   1:        0x117540d73 - _napi_register_module_v1
   2:        0x11751891e - _napi_register_module_v1
   3:        0x11751bc59 - _napi_register_module_v1
   4:        0x11751cfa5 - _napi_register_module_v1
   5:        0x11751cd24 - _napi_register_module_v1
   6:        0x11751d403 - _napi_register_module_v1
   7:        0x11751d2ec - _napi_register_module_v1
   8:        0x11751c359 - _napi_register_module_v1
   9:        0x11751d072 - _napi_register_module_v1
  10:        0x1176da0d5 - _napi_register_module_v1
  11:        0x1176da167 - _napi_register_module_v1
  12:        0x116897596 - <unknown>
  13:        0x116651095 - <unknown>
  14:        0x116594fed - <unknown>
  15:        0x1163ca77c - <unknown>
  16:        0x11642522c - <unknown>
  17:        0x1163ca77c - <unknown>
  18:        0x11642fd3c - <unknown>
  19:        0x116596ae7 - <unknown>
  20:        0x11657ad22 - <unknown>
  21:        0x1166dd6ea - <unknown>
  22:        0x1166f7534 - <unknown>
  23:        0x1166f3290 - <unknown>
[...]

PR #4577, opened 2023-11-08
fix: require.context should not fail when using `endWith` algo (#4557)

hverlin:
## Summary
Fix [Panic in require.context regex](https://github.com/web-infra-dev/rspack/issues/4557) (#4557)
Regexes using `endWith` algorithm should return `false` when checking for sticky or global flag.

## Test Plan

Added test case

Verified that without the fix, it fails with the following error:
```
  ● cases › context › issue-4557 › issue-4557 should compile

    error[internal]: internal error: entered unreachable code

      This is not expected, please file an issue at https://github.com/web-infra-dev/rspack/issues.
         0: <unknown>
         1: <unknown>
         2: <unknown>
         3: <unknown>
         4: <unknown>
         5: <unknown>
         6: <unknown>
         7: <unknown>
         8: <unknown>
         9: <unknown>
        10: <unknown>
        11: <unknown>

  ● cases › context › issue-4557 › issue-4557 should load the compiled test
```

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_regex/src/algo.rs
+++ b/crates/rspack_regex/src/algo.rs
@@ -100,14 +100,14 @@ impl Algo {
   pub(crate) fn global(&self) -> bool {
     match self {
       Algo::Regress(reg) => reg.flags.contains('g'),
-      Algo::EndWith { .. } => unreachable!(),
+      Algo::EndWith { .. } => false,
     }
   }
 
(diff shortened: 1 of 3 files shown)


bvanjoi (approved):

bvanjoi:
thanks!

IWANABETHATGUY:
!canary

PR #4595, opened 2023-11-09
refactor: AsyncDependenciesBlock

ahabhgk:
## Summary

Currently, Rspack determines the code splitting point by matching DependencyType, but it can only indicate a one-to-one relationship, one dependency to one code splitting point. However, AsyncDependenciesBlock can indicate a one-to-many relationship, many dependency to one code splitting point, which will be used in scenarios like ContainerEntryModule, or lazy-once ContextModule.

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_core/src/dependencies_block.rs
+++ b/crates/rspack_core/src/dependencies_block.rs
@@ -6,18 +6,18 @@ use rspack_identifier::{Identifiable, Identifier};
 use crate::{BoxDependency, DependencyId, GroupOptions};
 
 pub trait DependenciesBlock {
-  fn add_block(&mut self, block: AsyncDependenciesBlockId);
+  fn add_block_id(&mut self, block: AsyncDependenciesBlockId);
 
   fn get_blocks(&self) -> &[AsyncDependenciesBlockId];
 
-  fn add_dependency(&mut self, dependency: DependencyId);
+  fn add_dependency_id(&mut self, dependency: DependencyId);
 
   fn get_dependencies(&self) -> &[DependencyId];
 }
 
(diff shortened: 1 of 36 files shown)


crates/rspack_core/src/dependencies_block.rs
```diff
@@ -0,0 +1,103 @@
...
+
+fn get_async_dependencies_block_id() -> AsyncDependenciesBlockId {
+  AsyncDependenciesBlockId::from(
+    ASYNC_DEPENDENCIES_BLOCK_ID
+      .fetch_add(1, Ordering::Relaxed)
+      .to_string(),
+  )
+}
+
+pub type AsyncDependenciesBlockId = Identifier;
```
h-a-n-a, line 28: Would be better if this is a new type, moving some runtime evaluation error to compile time:-)

h-a-n-a (approved):

crates/rspack_core/src/dependencies_block.rs
```diff
@@ -0,0 +1,108 @@
...
+  pub fn add_dependency(&mut self, dependency: BoxDependency) {
+    self.dependency_ids.push(*dependency.id());
+    self.dependencies.push(dependency);
+  }
+
+  pub fn take_dependencies(&mut self) -> Vec<BoxDependency> {
+    std::mem::take(&mut self.dependencies)
+  }
+
+  pub fn add_block(&mut self, block: AsyncDependenciesBlock) {
```
h-a-n-a, line 76: Ooops, just forgot one thing. When would the `AsyncDependenciesBlock` have its own block? I'm kinda of curious.

PR #4601, opened 2023-11-09
feat: build rspack to wasm32-wasi-preview1-threads target

Brooooooklyn:
Build rspack to `wasm32-wasi-preview1-threads` target, should be able to run on Browser with wasi polyfill.

~Blocked by `path-dedot` crate, one of the deps of `nodejs_resolver`~


hardfist:
@Boshen  it seems we could remove nodejs_resolver::AliasMap type in to_oxc_resolver_options in 
https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/resolver/resolver_impl.rs#L296
so we can remove nodejs_resolver entirely in near future

hardfist:
bump

PR #4610, opened 2023-11-10
fix: fix infinite while loop when entry name contains '/' like 'routes/index'

hardfist:
## Summary
It seems when we ported get_relative_path we change the logic and cause infinite loop when entry key contains '/' like 'routes/_index'

## Test Plan
No Need

## Require Documentation?

- [x] No

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,6 +27,13 @@
     },
     {
       "type": "lldb",
+      "request": "attach",
+      "sourceLanguages": ["rust"],
+      "pid": "${command:pickMyProcess}",
+      "name": "debug-rust-attach"
+    },
+    {
+      "type": "lldb",
       "request": "launch",
       "sourceLanguages": ["rust"],
       "name": "debug-rust-basic",
--- a/crates/rspack_plugin_runtime/src/helpers.rs
+++ b/crates/rspack_plugin_runtime/src/helpers.rs
@@ -224,15 +224,19 @@ pub fn generate_entry_startup(
   RawSource::from(source).boxed()
 }
 
+/**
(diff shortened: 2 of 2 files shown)


crates/rspack_plugin_runtime/src/helpers.rs
```diff
@@ -224,15 +224,19 @@ pub fn generate_entry_startup(
...
 }
 
+/**
+ * This is ported from https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/esm/ModuleChunkFormatPlugin.js#L98
+ */
 pub fn get_relative_path(base_chunk_output_name: &str, other_chunk_output_name: &str) -> String {
   let mut base_chunk_output_name_arr = base_chunk_output_name.split('/').collect::<Vec<_>>();
   base_chunk_output_name_arr.pop();
   let mut other_chunk_output_name_arr = other_chunk_output_name.split('/').collect::<Vec<_>>();
-  while !base_chunk_output_name_arr.is_empty() && !other_chunk_output_name_arr.is_empty() {
```
hardfist, line 231: https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/esm/ModuleChunkFormatPlugin.js#L117 I don't why we move the if into while loop but id change the semantics of the logic

.vscode/launch.json
```diff
@@ -25,6 +25,13 @@
       ],
       "cwd": "${workspaceFolder}/packages/rspack"
     },
+    {
+      "type": "lldb",
+      "request": "attach",
+      "sourceLanguages": ["rust"],
+      "pid": "${command:pickMyProcess}",
```
hardfist, line 32: used for debugging outside project

underfin (approved):

Issue #4419, opened 2023-10-25
should warning for direct access import.meta


ahabhgk:
https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/dependencies/ImportMetaPlugin.js#L112

- webpack-test/cases/parsing/asi

PR #4621, opened 2023-11-12
fix: warning for direct access import.meta

OceanPresentChao:
## Summary

This pr resolves [issue4419](https://github.com/web-infra-dev/rspack/issues/4419), warning when directly accessing `import.meta`

### Things need discussion
There is no good way to emit warnings in the parsing stage now.
I encounter the dilemma that requires frequent back-and-forth conversion between `rspack_Error` and `Diagnostics`.We use `rspack_error` to emit an error and `Diagnostics` to emit a warning. So I wonder if could we adopt a normal way to emit both warnings and errors

## Test Plan

run test in `webpack-test/cases/parsing/asi`

## Require Documentation?

- [x] No


bvanjoi (dismissed):

packages/rspack/tests/cases/parsing/asi/index.js
```diff
@@ -0,0 +1,52 @@
+import { a as b, callme, getCount } from "./a";
```
bvanjoi, line 1: This case is identical to `﻿webpack-test/case/parsing/asi`. Therefore, I think you could possibly remove ﻿`webpack-test/case/parsing/asi/test.filter.js` directly.

bvanjoi: Furthermore, you could consider deleting the case due to its redundancy :D

bvanjoi:
Thanks!

bvanjoi (approved):

Issue #4587, opened 2023-11-09
[Tracking]: Improve Rspack Developer's Developer Experience 


hardfist:
### What subject does this issue tracking?

# [DX for Rspack Developer]
As we develop Rspack daily, we met some problems which limits our developer's development efficiency, let's track and progressive improve it
```[tasklist]
### Tasks
- [ ] Too much boilerplate code for adding new napi-rs binding especially when adding function api
- [ ] Js side Unit Testing is hard to debug separately
- [ ] Too much flaky test in daily CI jobs( release nightly failed every day)
- [ ] Hard to Mixed Debug a specific Jest case(vscode not supports interactive select debug case)
- [ ] generated binding.d.ts is unstable, which cause unnecessary new changed file
- [ ] Just a thought: there are lots of abstraction heavily depend on dynamic object upcast/downcast, Dependency <-> ModuleDependency <-> HarmonyImportDependency <-> HarmonyImportSpecifierDependency, Module <-> NormalModule (<-> CssModule), InitFragment <-> HarmonyExportInitFragment, it would be great if we can provide an efficient solution for this
```

chenjiahan:
The pnpm version is not limited by `engine`, some pull request cause the lock file to change a lot.

![image](https://github.com/web-infra-dev/rspack/assets/7237365/a68de517-0ff1-460b-9312-d22f4a95ad2a)

For exmaple: https://github.com/web-infra-dev/rspack/pull/4573/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbL76

hardfist:
Normalize Test Runner process and make it extensible for adding new test type

msdlisper:
hi~ @hardfist , I think this problem #4609  can be looked at together, some package versions do not specify the code can not run, which too affect the development experience

msdlisper:
hi~ @hardfist Why don't we try @swc/jest to speed up our testing ? although not much code is currently compiled with babel, which will slow down as more use cases are added :)

hardfist:
> hi~ @hardfist Why don't we try @swc/jest to speed up our testing ? although not much code is currently compiled with babel, which will slow down as more use cases are added :)

No particular reason, @msdlisper can you give it a try, if it works we would like to change to @swc/jest

PR #4623, opened 2023-11-12
chore(test): replace babel-jest with @swc/jest (#4587)

msdlisper:
## Summary

Try To Improve Rspack Developer's Developer Experience #4587 

## Test Plan

cd to `packages/rspack` and  `pnpm run build && pnpm run test`

## Require Documentation?

- [x] No


msdlisper:
@hardfist I tried it and it was OK, but it improved a lot, from around 160s to around 155s on my mac, but if there is a lot of `.tsx` testing behind it, the effect should show

hardfist:
thanks！

JSerFeng:
Thank you !

packages/rspack/tests/configCases/errors/loader-syntax-error/index.js
```diff
@@ -1,10 +1,10 @@
 it("should report syntax error", () => {
-  let errored = false;
+	let errored = false;
 	try {
 		require("./lib");
 	} catch (e) {
-    errored = true;
-		expect(e.message).toContain("SyntaxError");
+		errored = true;
+		expect(e.message).toContain("Syntax Error");
```
hardfist, line 7: this changed because of swc？

pnpm-lock.yaml
```diff
@@ -1,5 +1,9 @@
 lockfileVersion: '6.0'
 
+settings:
```
hardfist, line 3: do you use corepack？seems your pnpm version is not the same with original？

packages/rspack/tests/configCases/errors/loader-syntax-error/index.js
hardfist: @msdlisper can you help explain this change to me?
image

LingyuCoder: This difference is due to the different error messages between babel-jest and @swc/jest. So this change is safe

LingyuCoder:
Thank you. It looks good to me. On my machine, it gives about 20% improvement (92s => 74s). Could you please resolve the conflicts with the main branch, and then I will merge this branch.

hardfist (approved):

PR #4639, opened 2023-11-14
fix(runtime): get export default when new class

LingyuCoder:
## Summary

Fix get export default when used in new expression. It needs to wrap the brackets when is_call is false

## Require Documentation?

- [x] No


JSerFeng:
!canary

JSerFeng:
!canary

JSerFeng (approved):

PR #4654, opened 2023-11-15
chore: add engine check to ensure pnpm version

hardfist:
## Summary

corepack is easily broken so have to use engine to do force check

## Test Plan
No Need

## Require Documentation?

- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
     "type": "git",
     "url": "https://github.com/web-infra-dev/rspack"
   },
+  "engines": {
+    "pnpm": "7.32.0"
+  },
   "devDependencies": {
     "@rspack/cli": "workspace:*",
     "@taplo/cli": "^0.5.2",


package.json
```diff
@@ -35,6 +35,9 @@
     "type": "git",
     "url": "https://github.com/web-infra-dev/rspack"
   },
+  "engines": {
+    "pnpm": "7.32.0"
```
chenjiahan, line 39: Maybe a range is ok? Such as `>=7.0.0 <8`

Timeless0911: pnpm 的稳定性感觉还是锁定吧

chenjiahan: ok

chenjiahan (approved):

package.json
bvanjoi, line 39: Please update to the latest version of pnpm.

hardfist: that's another thing, we should upgrade to 8.x when we merge 0.4.0 back and remove NodeJS 14 support

PR #4670, opened 2023-11-16
fix uniqueName Acquisition method

ityuany:
## Summary

see [webpack-outputuniquename](https://webpack.js.org/configuration/output/#outputuniquename)

Webpack prefers to take `output.library.name` first, and only if the user has not configured it, will it take the `name` from `package.json`.

## Test Plan

copy by [webpack defaults config](https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/config/defaults.js#L804)

## Require Documentation?

- [x] No

see [rspack-outputuniquename-doc](https://www.rspack.dev/config/output.html#outputuniquename) , In fact, this is how he described it.

### DIFF
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -403,9 +403,43 @@ const applyOutputDefaults = (
 		futureDefaults: boolean;
 	}
 ) => {
+
+
+	const getLibraryName = (library:string | string[]| Record<string,any>):string => {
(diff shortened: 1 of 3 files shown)


ahabhgk:
Could you add a test under `packages/rspack/tests/configCases` and make the CI passes

ahabhgk (approved):
Thanks

PR #4679, opened 2023-11-16
fix(runtime): revert alignment of commonjs exports dependency

JSerFeng:
Original PR: #4572 

`this.foo` should not be considered as CommonJS export if not present in the top_level_scope.

Details: https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs#L120-L167

In this PR, if we have

```javascript
class Foo() {
  this.x.y = 1
}
```

will be transformed to:

```javascript
class Foo {
  this.y = 1;
}
```


JSerFeng:
!canary

hardfist:
@JSerFeng  please write the revert reason

LingyuCoder (approved):

PR #4682, opened 2023-11-16
feat: support `ContextDependency`

### LINKED ISSUE #4594: refactor ContextDependency
ImportContextDependency should impl ContextDependency instead of ModuleDependency

img_v3_0251_bcff72ea-162b-4648-9e21-e5d91d0d8c1g

img_v3_0251_f6645d13-9a04-4d08-a55e-32a7bccd484g

then we don't need `fn options` on ModuleDependency

img_v3_0251_6465a89a-bb5c-409b-8f8f-37c8d82f0e3g

bvanjoi:
Fixes #4594 

This PR introduces ﻿a trait called `ContextDependency`, which mocks the ﻿`ContextDependency` in webpack. This can assist in more closely aligning certain elements with webpack.

WARNING: This PR introduces many changes, and I'm **not** confident that it won't result in breaking changes.

### DIFF
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -41,12 +41,12 @@ use crate::{
   Chunk, ChunkByUkey, ChunkContentHash, ChunkGraph, ChunkGroupByUkey, ChunkGroupUkey,
   ChunkHashArgs, ChunkKind, ChunkUkey, CleanQueue, CleanTask, CleanTaskResult,
   CodeGenerationResult, CodeGenerationResults, CompilationLogger, CompilationLogging,
-  CompilerOptions, ContentHashArgs, DependencyId, DependencyParents, Entry, EntryData,
-  EntryOptions, Entrypoint, FactorizeQueue, FactorizeTask, FactorizeTaskResult, Filename, Logger,
-  Module, ModuleGraph, ModuleIdentifier, ModuleProfile, ModuleType, PathData, ProcessAssetsArgs,
-  ProcessDependenciesQueue, ProcessDependenciesResult, ProcessDependenciesTask, RenderManifestArgs,
-  Resolve, ResolverFactory, RuntimeGlobals, RuntimeModule, RuntimeSpec, SharedPluginDriver,
-  SourceType, Stats, TaskResult, WorkerTask,
+  CompilerOptions, ContentHashArgs, ContextDependency, DependencyId, DependencyParents, Entry,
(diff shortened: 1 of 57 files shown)


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -445,11 +445,15 @@ impl Compilation {
           .module_graph
           .dependency_by_id(&id)
           .expect("dependency not found");
+        if dependency.as_module_dependency().is_none()
+          && dependency.as_context_dependency().is_none()
```
h-a-n-a, line 449: Could we also impl `trait ModuleDependency` to context dependencies ?

bvanjoi: why would you think so? All context dependencies in webpack extend from ﻿`ContextDependency`.

ahabhgk: Actually in webpack not only ModuleDependency can resolve a module, Dependency which implement getResourceIdentifier() and return a non-null value can resolve a module

crates/rspack_core/src/dependency/context_dependency.rs
```diff
@@ -0,0 +1,29 @@
+use crate::{ContextOptions, Dependency};
+
+pub trait ContextDependencyTrait: Dependency {
```
ahabhgk, line 3:
```suggestion
pub trait ContextDependency: Dependency {
```

ahabhgk (commented):
Generally looks good to me

crates/rspack_core/src/compiler/queue.rs
```diff
@@ -85,6 +85,7 @@ impl WorkerTask for FactorizeTask {
       | DependencyType::CommonJSRequireContext
       | DependencyType::RequireContext
       | DependencyType::ImportMetaContext => {
+        assert!(dependency.as_module_dependency().is_none());
```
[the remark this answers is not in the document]
bvanjoi: This is intentional. If a dependency here implements ﻿`ModuleDependency`, then we consider it a bug.

ahabhgk (approved):

crates/rspack_core/src/compiler/compilation.rs
h-a-n-a: > Actually in webpack not only ModuleDependency can resolve a module, Dependency which implement getResourceIdentifier() and return a non-null value can resolve a module

I know this and kinda of weird. Just wondering why context dependencies not using Module dependency abstraction.

PR #4693, opened 2023-11-17
fix: 🐛assignDeterministicIds alignment

IWANABETHATGUY:
## Summary
1. align `assignDeterministicIds` you could refer to https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/ids/IdHelpers.js#L364-L376

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/rspack_ids/src/id_helpers.rs
+++ b/crates/rspack_ids/src/id_helpers.rs
@@ -253,7 +253,7 @@ pub fn assign_deterministic_ids<T: Copy>(
   mut items: Vec<T>,
   get_name: impl Fn(T) -> String,
   comparator: impl Fn(&T, &T) -> Ordering,
-  mut assign_id: impl FnMut(T, String) -> bool,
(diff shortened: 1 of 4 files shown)


crates/rspack_ids/src/id_helpers.rs
```diff
@@ -253,7 +253,7 @@ pub fn assign_deterministic_ids<T: Copy>(
   mut items: Vec<T>,
   get_name: impl Fn(T) -> String,
   comparator: impl Fn(&T, &T) -> Ordering,
-  mut assign_id: impl FnMut(T, String) -> bool,
+  mut assign_id: impl FnMut(T, usize) -> bool,
```
IWANABETHATGUY, line 256: I converted `usize` to `String` at the usage, so the rest type should remain unchanged.  this alignment is required by `mangleExportsPlugin`.

ahabhgk (approved):

PR #4698, opened 2023-11-18
feat: support optimization.nodeEnv

SyMind:
## Summary

At present, I'm required to explicitly add `process.env.NODE_ENV` in `builtins.define`. However, webpack implicitly sets a default for `process.env.NODE_ENV` based on the `mode` settings.

In addition to the above, let `@rspack/core` exports `webpack-sources` through `sources`, and exports `util.serialization`.

Ultimately, to prevent string concatenation from being minified and to ensure tests pass, I modified expressions such as `"fo" + "o"` to `["fo" + "o"].join("")`. However, I am uncertain about the precise cause of this issue. It could be a result of the `DefinePlugin`, but that's only a speculation at this point.

## Test Plan

The `Defaults.unittest.ts` was used to generate snapshots.

## Require Documentation?

- [x] No


packages/rspack/src/index.ts
```diff
@@ -63,4 +63,12 @@ export const wasm = { EnableWasmLoadingPlugin };
 import { LimitChunkCountPlugin } from "./builtin-plugin";
 export const optimize = { LimitChunkCountPlugin };
 
+export const sources = require("webpack-sources");
+
+export const util = {
+	get serialization() {
+		return require("./util/serialization");
```
ahabhgk, line 70: Don't export this for now, we haven't implement serialization related stuff in Rspack

ahabhgk (approved):
Thanks!

PR #4701, opened 2023-11-20
fix(harmony-deps-scanner): should only be `dep.call=true` in call_expr or tagged template's tag

JSerFeng:
## Summary

There is a field in harmonyImportSpecifier called `call`, it should be true if the imported specifier in call_expr or tagged templates'tag.

for example:

```javascript
import { foo } from 'bar'

foo() // call=true
foo`` // call=true
new foo() // call=false (undefined in webpack
`${foo}` // call=false (undefined in webpack
```

Before:
The new expression or tagged template expression both are not correct cases for `call` to be `true`, we only checked callee before

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
@@ -6,8 +6,8 @@ use rspack_core::{
 use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 use swc_core::atoms::JsWord;
 use swc_core::common::Span;
-use swc_core::ecma::ast::{AssignExpr, AssignOp, MemberExpr, OptChainExpr};
-use swc_core::ecma::ast::{Callee, ExportAll, ExportSpecifier, Expr, Id, TaggedTpl};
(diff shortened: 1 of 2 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/harmony_import_dependency_scanner.rs
```diff
@@ -533,19 +533,121 @@ impl Visit for HarmonyImportRefDependencyScanner<'_> {
...
+      r#"
+      import { foo } from 'bar';
+      new foo();
+      new foo().bar;
+      new foo.bar();
+      `${foo}`;
+      nested(foo).call(true);
+      nested(foo)``
+      (`${foo}`)``
+    "#
```
JSerFeng, line 635: You can add more tests here

LingyuCoder: `(new foo()).bar()`, foo should also not be call

JSerFeng: Added

LingyuCoder (approved):

PR #4705, opened 2023-11-20
fix: @swc/jest type error

ahabhgk:
## Require Documentation?

- [ ] No


crates/node_binding/binding.d.ts
```diff
@@ -462,6 +462,13 @@ export interface JsStatsWarning {
   formatted: string
 }
 
+export interface NodeFS {
```
hardfist, line 465: @h-a-n-a can you take some time to fix this problem, it's kind of annoying now

PR #4730, opened 2023-11-21
fix(resolver): make sure package.json path is inside the resolved path

Boshen:
## Summary

Previously Package.json path returned the non-symlinked path, which breaks code splitting where it tries to resolve package_json.sideEffects.

https://github.com/oxc-project/oxc/pull/1481/files

## Test Plan

* I ran the code where it went wrong and made sure the path returned the symlinked one.
* I also added a runtime check inside oxc_resolver to make sure all paths are correct.

## Require Documentation?

- [x] No


IWANABETHATGUY (approved):

JSerFeng:
!canary

PR #4750, opened 2023-11-22
chore: fix debugger launcher on window

h-a-n-a:
## Summary

Fix debugger launcher on windows

## Require Documentation?

- [x] No


scripts/tsconfig.json
```diff
@@ -45,7 +45,7 @@
 
     /* JavaScript Support */
     "allowJs": true /* Allow JavaScript files to be a part of your program. Use the 'checkJS' option to get errors from these files. */,
-    "checkJs": true /* Enable error reporting in type-checked JavaScript files. */,
+    "checkJs": false /* Enable error reporting in type-checked JavaScript files. */,
```
hardfist, line 48: why change this?

h-a-n-a: I don't think it's necessary for JS files...

h-a-n-a: Actually, we didn't do type checking on the CI anyway and it make ts server complaining. So I turned it off.

hardfist: does not it provide autocomplete

h-a-n-a: Yes, it does ;--)

hardfist (approved):

PR #4758, opened 2023-11-23
chore(webpack-test): make tests more future proof

Boshen:
## Summary

This PR tries to make webpack-test more future proof by catching more errors where appropriate.

`webpack-test` has the assumption where all configs are valid and everything will compile, but this is not the case with Rspack.

I also found 2 passing tests so I added to this PR.

## Test Plan

* CI passes.

## Require Documentation?

- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "license": "MIT",
   "description": "A Fast Rust-based web bundler",
   "private": true,
+  "main": "./packages/rspack/dist/index.js",
   "scripts": {
     "x": "zx x.mjs",
     "dev": "pnpm --filter @rspack/cli run dev",
--- a/webpack-test/ConfigTestCases.template.js
+++ b/webpack-test/ConfigTestCases.template.js
@@ -59,7 +59,9 @@ const describeCases = config => {
 			stderr = captureStdio(process.stderr, true);
 		});
 		afterEach(() => {
-			stderr.restore();
+			if (stderr) {
+				stderr.restore();
+			}
 		});
 		jest.setTimeout(TIMEOUT);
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
   "license": "MIT",
   "description": "A Fast Rust-based web bundler",
   "private": true,
+  "main": "./packages/rspack/dist/index.js",
```
Boshen, line 7: In some tests it tries to import ...

https://github.com/web-infra-dev/rspack/blob/d3ddbe7d2c3bca2995998a3365c375e19ce8eb7f/webpack-test/configCases/web/nonce/webpack.config.js#L1

This will correctly resolve it.

webpack-test/ConfigTestCases.template.js
```diff
@@ -88,10 +90,17 @@ const describeCases = config => {
...
 						let options, optionsArr, testConfig;
 						beforeAll(() => {
-							options = prepareOptions(
-								require(path.join(testDirectory, "webpack.config.js")),
-								{ testPath: outputDirectory }
-							);
+							expect(() => {
+								try {
+								options = prepareOptions(
+									require(path.join(testDirectory, "webpack.config.js")),
```
Boshen, line 96: The test can crash in some webpack configs, e.g. 

https://github.com/web-infra-dev/rspack/blob/d3ddbe7d2c3bca2995998a3365c375e19ce8eb7f/webpack-test/configCases/plugins/profiling-plugin/webpack.config.js#L7

webpack-test/ConfigTestCases.template.js
```diff
@@ -59,7 +59,9 @@ const describeCases = config => {
 			stderr = captureStdio(process.stderr, true);
 		});
 		afterEach(() => {
-			stderr.restore();
+			if (stderr) {
+				stderr.restore();
```
Boshen, line 63: Sometimes the test will crash and leave `stderr = undefined` , which crashes jest

hardfist (approved):

PR #4764, opened 2023-11-23
feat: support String.raw for context module

### LINKED ISSUE #4420: support String.raw for context module
https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/dependencies/ContextDependencyHelpers.js#L130

- webpack-test/cases/parsing/complex-require

faga295:
## Summary
resolve #4420 
Add `templateStringKind` param for `scan_context_module_tpl`, if there is no `String.raw` tag function, inner_reg will created by `templateElement.cooked`, otherwise `templateElement.raw` will be used

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/context_helper.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/context_helper.rs
@@ -179,7 +179,7 @@ fn scan_context_module_tagged_tpl(tpl: &TaggedTpl) -> (String, String) {
     {
       scan_context_module_tpl(tpl.tpl.as_ref(), TemplateStringKind::Raw)
     }
-    _ => scan_context_module_tpl(tpl.tpl.as_ref(), TemplateStringKind::Cooked),
+    _ => (String::from("."), String::new()),
   }
 }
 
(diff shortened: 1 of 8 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/context_helper.rs
```diff
@@ -145,6 +162,33 @@ fn scan_context_module_concat_call(expr: &CallExpr) -> Option<(String, String)>
   Some((context, reg))
 }
 
+// require(String.raw`./${a}.js`)
+fn scan_context_module_tagged_tpl(tpl: &TaggedTpl) -> Option<(String, String)> {
```
IWANABETHATGUY, line 166: since we return `Some(String, String)` in every branch, so the `Option` looks redundant.

bvanjoi:
Would you mind adding a test case to further ensure the functionality? You can adding this to ﻿`rspack/test/case/`

bvanjoi (approved):

bvanjoi:
Thanks!

PR #4766, opened 2023-11-23
fix: only register `SplitChunkPlugin` once

### LINKED ISSUE #4754: panic: should run render_chunk hook
### System Info

"@rspack/cli": "0.4.0"

### Details

```js
// src/index.js
function add() {
  console.log("hello world")
}
// src/child-entry.js
const value = {};

// rspack.config.js
const path = require("path");

module.exports = {
  entry: path.resolve(__dirname, "./src/index"),
  plugins: [
    {
      apply(compiler) {
        compiler.hooks.make.tap('make', (compilation) => {
          const childEntry = path.resolve(__dirname, './src/child-entry');
          const childCompiler = compilation.createChildCompiler('name', {}, [
            new compiler.webpack.EntryPlugin(compiler.context, childEntry),
          ]);
          childCompiler.compile(() => {});
        });
      },
    }
  ],
  optimization: {
    splitChunks: {
      cacheGroups: { 
        singleVendor: { 
          chunks: "all", 
          enforce: true, 
          name: "vendor", 
        } 
      }
    }
  },
};
```

### Reproduce link

_No response_

### Reproduce Steps

npx rspack build

bvanjoi:
Fixes #4754

#4754 panics because it erroneously generates chunks by registering ﻿`SplitChunksPlugin` twice, while in reality, webpack only registers it once (specifically in the root compiler). As a result, it executes an additional `﻿render_chunk` operation, leading to a panic.

In this PR, I have relocated the registration of `﻿SplitChunksPlugin` to `﻿createCompiler` to ensure that it only registers once.

### DIFF
--- a/crates/rspack_binding_options/src/options/raw_optimization.rs
+++ b/crates/rspack_binding_options/src/options/raw_optimization.rs
@@ -6,7 +6,6 @@ use rspack_ids::{
   DeterministicChunkIdsPlugin, DeterministicModuleIdsPlugin, NamedChunkIdsPlugin,
   NamedModuleIdsPlugin,
 };
-use rspack_plugin_split_chunks::SplitChunksPlugin;
 use serde::Deserialize;
 
 use crate::{RawOptionsApply, RawSplitChunksOptions};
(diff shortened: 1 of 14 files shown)


crates/rspack_binding_options/src/options/raw_optimization.rs
```diff
@@ -36,17 +35,6 @@ impl RawOptionsApply for RawOptimizationOptions {
     self,
     plugins: &mut Vec<Box<dyn rspack_core::Plugin>>,
   ) -> Result<Self::Options, rspack_error::Error> {
-    if let Some(options) = self.split_chunks {
```
bvanjoi, line 39: this implies that we will no longer use ﻿`old_split_chunks_plugin`.

ahabhgk: https://www.rspack.dev/blog/announcing-0.4.html#deprecatingbuiltinsdevfriendlysplitchunks-andexperimentsnewsplitchunks

this is a breaking and should move to 0.5.0, we haven't remove oldSplitChunks currently, so user can `newSplitChunks: false` to enforce to use oldSplitChunks

ahabhgk: it's a mistake we haven't remove it earlier, since that some shit code is needed

packages/rspack/src/rspackOptionsApply.ts
```diff
@@ -189,6 +190,11 @@ export class RspackOptionsApply {
 		if (options.builtins.devFriendlySplitChunks) {
 			options.optimization.splitChunks = undefined;
 		}
+
+		if (options.optimization.splitChunks) {
+			new SplitChunksPlugin(options.optimization.splitChunks).apply(compiler);
```
ahabhgk, line 195: My suggest is add a BuiltinPlugin::OldSplitChunksPlugin and register it based on experiments.newSplitChunks

ahabhgk (approved):

Issue #4767, opened 2023-11-23
Support resource end with `.PNG`


hanhan9449:
### System Info

rspack: 0.4.0
npm: 10.1.0
node: 20.9.0
os: windows 11 23e2

### Details

When trying to use resources ending in .PNG, the system reports an error.

code:

![image](https://github.com/web-infra-dev/rspack/assets/28822945/fa6594f3-b0a9-4914-98c8-9ee8b57b0b91)

config:

![image](https://github.com/web-infra-dev/rspack/assets/28822945/f6375f26-3fdc-465c-85e5-ac99c60a688d)

error:

![image](https://github.com/web-infra-dev/rspack/assets/28822945/dcc0ae15-5233-4c4d-93ae-31d9e6c29b27)

### Reproduce link

https://github.com/hanhan9449/rspack-040-demo

### Reproduce Steps

- npm i
- npm run build

hanhan9449:
ref: web-infra-dev/rspress#395

nomyfan:
It seems it's caused by no copying regex flags from JS side to RS side.

nomyfan:
Hi, I want to have a try to fix this, would you like to assign it to me? @hardfist

PR #4780, opened 2023-11-24
fix: when use regex with flags in module.test property (such as i), the flags don't work nice

hanhan9449:
## Summary

修复了在 rspack.config.js 中，`module.rules.x.test` 属性使用带flags的正则表达式不生效。例如最常见的i(gnore)。

Fixed the issue in rspack.config.js where the `module.rules.x.test` attribute using regular expressions with flags did not take effect. For example, the most common i(gnore).

issue ref：#4767 

## Test Plan

在rspack.test.build目录添加测试(issue-4767)，测试以PNG、SVG等大写字符结尾的文件引用能力。

Add a test in the rspack.test.build directory to test the ability to reference files ending with uppercase characters such as PNG and SVG.

```js
import PNG from "./rspress-icon.PNG";
import SVG from "./react.SVG";

console.log("png", PNG);
console.log("SVG", SVG);
```

## Require Documentation?

- [x] No


ahabhgk:
Thanks for your PR! But more prefer https://github.com/web-infra-dev/rspack/pull/4781, test should be added under `packages/rspack/tests` instead of `packages/rspack-cli/`, contribution welcome in the future!

PR #4828, opened 2023-11-30
fix(rust): s/chars().enumerate()/char_indices()/g

Boshen:
🦀🦶🔫 https://twitter.com/boshen_c/status/1719033308682870891

### DIFF
--- a/crates/rspack_loader_runner/src/scheme.rs
+++ b/crates/rspack_loader_runner/src/scheme.rs
@@ -68,7 +68,7 @@ const HASH: char = '#';
 const QUERY: char = '?';
 
 pub fn get_scheme(specifier: &str) -> Scheme {
-  let mut chars = specifier.chars().enumerate().peekable();
+  let mut chars = specifier.char_indices().peekable();
 
   // First char maybe only a letter
   let start = chars.next();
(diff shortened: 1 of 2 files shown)


crates/swc_plugin_import/src/lib.rs
```diff
@@ -473,7 +473,7 @@ where
     let s: &str = self.as_ref();
     let mut output = String::with_capacity(s.len());
 
-    s.chars().enumerate().for_each(|(idx, c)| {
```
IWANABETHATGUY, line 476: Using `chars().enumerate()` is fine here, cause the function looks just wants to do something different for the first char.

crates/rspack_loader_runner/src/scheme.rs
```diff
@@ -68,7 +68,7 @@ const HASH: char = '#';
 const QUERY: char = '?';
 
 pub fn get_scheme(specifier: &str) -> Scheme {
-  let mut chars = specifier.chars().enumerate().peekable();
```
IWANABETHATGUY, line 71: For this one, you can remove `enumerate` directly, since the index is unused.

Boshen:
It seems like both usages are safe. Closing instead of changing the code for avoiding unintended behaviours.

PR #4841, opened 2023-12-01
ci: speed up diff ci

LingyuCoder:
## Summary

Use github.sha to generate self-hosted runner local cache path, and then the artifacts can be shared between workflows

## Require Documentation?

- [x] No


LingyuCoder:
!diff

jerrykingxyz (approved):

PR #4851, opened 2023-12-02
refactor: key performance in merge_matched_item_into_module_group_map

aweary:


crates/rspack_plugin_split_chunks_new/src/plugin/module_group.rs
```diff
@@ -214,12 +215,12 @@ impl SplitChunksPlugin {
...
-              let selected_chunks_key = Self::get_key(
-                selected_chunks.iter().map(|chunk| &chunk.ukey),
-                chunk_idx_map,
-              );
 
-              merge_matched_item_into_module_group_map(
+
+
+              let selected_chunks_key = selected_chunks.iter().map(|c| c.ukey.as_usize().to_string()).collect::<Vec<_>>().join("_");
+
```
JSerFeng, line 222: Why remove get_key here

crates/rspack_plugin_split_chunks_new/src/plugin/module_group.rs
```diff
@@ -69,7 +69,7 @@ impl SplitChunksPlugin {
       cache_group_index: usize,
       cache_group: &'a CacheGroup,
       selected_chunks: Box<[&'a Chunk]>,
-      selected_chunks_key: ChunksKey,
+      selected_chunks_key: String,
```
JSerFeng, line 72: Why change the ChunksKey here, chunk key is useful to avoid unnecessary modules be merged into chunk

aweary: @JSerFeng `ChunksKey` is the main performance problem here from the traces I took, construction a `BigUint` and then serializing that to a `String` was taking a significant time.

JSerFeng:
!bench

JSerFeng:
!bench

JSerFeng (approved):

JSerFeng:
This PR changes the splitChunks get_chunks_combination algorithm by @aweary,
## Before
The chunksKey is caculated using bitmap, eg. module foo belongs to chunk A, B and C
```
A: 001
B: 010
C: 100
```
So module foo has chunksKey: A | B | C which is 111, if another module also has chunksKey 111, means they are in the same chunk combinations, and we will do something about it.

But this way, the chunksKey used num-bigint crate which has big perf issue when converting to string.

## Now
Now we caculate chunksKey through hash directly, thanks to @aweary, we hash all the chunks for one module, if there hash is equal, they are in the same combinations.

## Other changes

Change the concurrency, make per thread do more jobs by rayon

JSerFeng:
!canary

JSerFeng:
!bench

JSerFeng (dismissed):

JSerFeng (approved):

JSerFeng:
!canary

PR #4880, opened 2023-12-04
refactor: should not always rely on `Result`

h-a-n-a:
## Summary

Adjust some of expectations with the philosophy of https://github.com/web-infra-dev/rspack/pull/4848:

1. NAPI should panic if failed to call
2. Assume every source-map emitted from SWC are valid source-maps.
3. Some minor changes

## Test Plan

as-is.

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs
+++ b/crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs
@@ -11,9 +11,7 @@ use rspack_core::{
   DependencyId, GenerateContext, Module, ParseContext, ParseResult, ParserAndGenerator, SourceType,
   TemplateContext, TemplateReplaceSource,
 };
-use rspack_error::{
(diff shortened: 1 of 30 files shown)


crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs
```diff
@@ -256,7 +256,7 @@ impl ParserAndGenerator for JavaScriptParserAndGenerator {
       SourceMapSource::new(SourceMapSourceOptions {
         value: output.code,
         name: resource_data.resource_path.to_string_lossy().to_string(),
-        source_map: SourceMap::from_json(&map).map_err(|e| internal_error!(e.to_string()))?,
```
h-a-n-a, line 259: Always assume source maps passed back from swc is valid, otherwise, it's an internal bug

ahabhgk (approved):

Issue #4424, opened 2023-10-25
support compile time evaluation


ahabhgk:
https://github.com/webpack/webpack/blob/1f13ff9fe587e094df59d660b4611b1bd19aed4c/lib/ConstPlugin.js#L164-L172

- webpack-test/cases/parsing/evaluate

PR #4918, opened 2023-12-06
feat: eval binary expression

bvanjoi:
Related to #4424 

support evaluate binary expression


crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
```diff
@@ -222,6 +222,11 @@ impl Visit for CommonJsImportDependencyScanner<'_> {
     let value = if self.in_if { "true" } else { "undefined" };
     self.replace_require_resolve(&bin_expr.left, value);
     self.replace_require_resolve(&bin_expr.right, value);
-    bin_expr.visit_children_with(self);
+
+    if let Some(keep_right) = expression_logic_operator(self, bin_expr)
```
IWANABETHATGUY, line 226: `if let Some(true)` is enough here?

IWANABETHATGUY (approved):

PR #4931, opened 2023-12-06
feat: context module support inline loader

faga295:
## Require Documentation?

- [ ] No


webpack-test/cases/loaders/context/test.filter.js
```diff
@@ -1,12 +0,0 @@
-
-/*
-const supportsRequireInModule = require("../../../helpers/supportsRequireInModule");
-
-module.exports = config => {
```
bvanjoi, line 5: In my opinion, this ﻿`module.exports` should not be deleted.

crates/rspack_core/src/context_module_factory.rs
```diff
@@ -86,7 +86,12 @@ impl ContextModuleFactory {
     let mut file_dependencies = Default::default();
     let mut missing_dependencies = Default::default();
     let context_dependencies = Default::default();
-    let specifier = dependency.request();
+    let request = dependency.request().to_string();
```
bvanjoi, line 89: `.to_string` is not neccessary.

packages/rspack/tests/configCases/loader/context/index.js
```diff
@@ -0,0 +1,5 @@
+it("should be able to use a context with a loader", () => {
+	let a = "a";
+	let result = require(`raw-loader!./${a}`).default;
```
bvanjoi, line 3: There's no requirement to utilise ﻿`raw-loader` in this. I suggest creating a `﻿loader.js` file with the following code:

```js
module.exports = (content) => content;
```

and then replacing ﻿`require(raw-loader!./${a}﻿)` with `﻿require(./loader.js!./${a}﻿)`.

bvanjoi (commented):
Once these nits are addressed and the ci pass, I believe it's ready to merge.

webpack-test/cases/loaders/context/test.filter.js
faga295: I delete test.filter.js according to https://github.com/web-infra-dev/rspack/blob/77f57e52180462019274bada5a2b77e90a6ffca4/webpack-test/README.md?plain=1#L17-L24

packages/rspack/tests/configCases/loader/context/index.js
faga295: IMO, using raw-loader can help confirm if the loader is functioning as expected. The `loader.js` file appears to return the original content, making it challenging to determine whether `loader.js` is working.

h-a-n-a: To test if the loader is working, It's not necessary to use an additional loader.
Even a `content.replace("1", "2")` is enough.

faga295: > To test if the loader is working, It's not necessary to use an additional loader. Even a `content.replace("1", "2")` is enough.

get~

PR #4945, opened 2023-12-07
feat!: use native resolver in loader

bvanjoi:
Fixes #4672

PR #4946, opened 2023-12-07
fix: rm dep type for resolve

bvanjoi:
Prerequisite for #4945.

Clean up the unnecessary argument in the ﻿resolve.

### DIFF
--- a/crates/rspack_core/src/resolver/factory.rs
+++ b/crates/rspack_core/src/resolver/factory.rs
@@ -4,14 +4,12 @@ use dashmap::DashMap;
 use rustc_hash::FxHasher;
 
 use super::resolver_impl::Resolver;
-use crate::DependencyType;
 use crate::{DependencyCategory, Resolve};
 
 #[derive(Debug, Hash, Eq, PartialEq, Clone)]
 pub struct ResolveOptionsWithDependencyType {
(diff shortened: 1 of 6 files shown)


crates/rspack_core/src/resolver/factory.rs
```diff
@@ -4,14 +4,12 @@ use dashmap::DashMap;
 use rustc_hash::FxHasher;
 
 use super::resolver_impl::Resolver;
-use crate::DependencyType;
 use crate::{DependencyCategory, Resolve};
 
 #[derive(Debug, Hash, Eq, PartialEq, Clone)]
 pub struct ResolveOptionsWithDependencyType {
```
ahabhgk, line 10:
```suggestion
pub struct ResolveOptionsWithDependencyCategory {
```

bvanjoi: It's align to webpack: https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/ResolverFactory.js#L24

ahabhgk: I know, but ResolveOptionsWithDependencyType don't have dependency_type field feels wired, so add some comments if you want keep the name

ahabhgk (approved):

Issue #4427, opened 2023-10-25
allow parse js hashbang syntax


ahabhgk:
https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/javascript/JavascriptParser.js#L221

PR #4970, opened 2023-12-11
feat(es): allow parse js hashbang syntax

magic-akari:
## Summary

- Closes: #4427

## Test Plan

crates/rspack_loader_swc/tests/fixtures/hashbang added.
If there are any additional locations where test cases need to be added, please inform me.

## Require Documentation?

- [x] No


LingyuCoder:
This modification only works when using builtin:swc-loader, may consider adding it in `crates/rspack_plugin_javascript/src/ast/parse.rs` to prevent compilation failure when not using builtin:swc-loader.

And it would be better if you could align the addition of comments with [webpack](https://github.com/webpack/webpack/blob/v5.89.0/lib/CompatibilityPlugin.js#L165)

magic-akari:
@LingyuCoder Do you have any suggestions about test cases?

LingyuCoder:
> @LingyuCoder Do you have any suggestions about test cases?

It is recommended to add test cases to ensure that the bundled script runs correctly. You can reuse webpack's test cases by removing `webpack-test/cases/parsing/hashbang/test.filter.js`. And also you can add additional test cases in the `packages/rspack/tests` directory.

LingyuCoder:
Thanks

PR #4984, opened 2023-12-12
fix: handle chunk relation

JSerFeng:
## Summary

Chunk relation should be handled at any time processing chunk group connection

## Require Documentation?

- [ ] No


IWANABETHATGUY (commented):
Do we need some test cases?

JSerFeng:
!canary

ahabhgk (approved):

PR #4994, opened 2023-12-13
fix: remove available modules from ancestor

JSerFeng:
## Summary

Giving chunk group relations as following:

```
A
|  \
B  C
|  /
D
```

Remove availableModules can remove modules from parents and ancestor, we have this logic before, but for some reason it didn't work properly, this is a fix for this

## Require Documentation?

- [ ] No


JSerFeng:
!bench

JSerFeng:
!bench

ahabhgk (approved):

PR #5000, opened 2023-12-13
fix: parse left and right when eval failed for logic expr

### LINKED ISSUE #4997: Failed to detect cjs require dependency
### System Info

Not matter

### Details

Failed to detect require in binary expression and unary expression

for example:

```javascript
!require() && !require()
```

### Reproduce link

_No response_

### Reproduce Steps

type above code

bvanjoi:
Fixes #4997 

 Parsing of left and right expressions should continue even if the evaluation of the logical expression fails.


ahabhgk (approved):

JSerFeng:
!canary

Issue #4973, opened 2023-12-11
progressPlugin({profile:true}) result is not accurate


hardfist:
the processAssets actually only takes 2ms but the profile reports 185ms
![image](https://github.com/web-infra-dev/rspack/assets/8898718/ef847047-0115-4e7d-835d-3942323c68fe)
![image](https://github.com/web-infra-dev/rspack/assets/8898718/6059ee7a-3f21-4430-b3d9-ff5b540e8b6c)

hardfist:
@faga295 can you help take a look, I'm also wondering how profile caculates the process_assets duration

faga295:
Actually the start time of each hook's duration is the end time of its previous hook, so the process_assets duration is caculated using `optimize_chunk —— process_assets`.In webpack it intercept many hooks, so each hook's duration will be more precise

https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/ProgressPlugin.js#L468-L508

Maybe I can help to intercept more hooks

hardfist:
@faga295 give it a try

xc2:
link #4890

hardfist:
I'm not sure whether we need this functionality considering it's mechanism it not accurate enough, maybe logging.json is more suitable and accurate than progressPlugin({profile:true}), cc @ahabhgk

ahabhgk:
I prefer stats logging for inspecting the time of every hook, but it is not nevertheless, this is a addon feature on top of progress plugin, it's useful for simple cases, but the premise is that it must be accurate enough.

xc2:
bump

hardfist:
seems fixed in #187

PR #5008, opened 2023-12-13
feat: progress plugin

faga295:
## Summary
#4973 

Last time in #3803, we only supported the activeModule report and progress_handler compatibility. However, it had some issues, such as the default_handler not correctly reflecting the build duration. Due to several missing sealing hooks, the process duration wasn't accurate either.

Comparing to webpack, rspack will build modules with mulitple threads, between `build_module` and `succeed_module` , maybe other module will be build in other thread. Now we have `default_handler` and `progress_handler`, and in the future there may be some custom handler from user.
We don't want the progress bar shows "building module2" after module2 build succeed, so we should correct the progress bar message, when module2 `succeed_module`. In code side, I use `linked_hash_map` which can keep the insertion order to record the active modules, so after module2 build succeed, we can get the last active module accroding to the insertion order(I'm not sure whether there has performance problem)
image
In profile mode, this mode will print module build message in terminal, we should guarantee module's build time instead of module's build order
image

image

## Require Documentation?

- [ ] No

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2956,8 +2956,6 @@ dependencies = [
  "linked-hash-map",
  "rspack_core",
  "rspack_error",
- "rustc-hash",
- "tokio",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2955,6 +2955,7 @@ dependencies = [
  "indicatif",
  "rspack_core",
  "rspack_error",
+ "rustc-hash",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2956,6 +2956,7 @@ dependencies = [
  "rspack_core",
  "rspack_error",
  "rustc-hash",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


faga295:
@ahabhgk  @hardfist  I've encountered an issue, which can be described as follows: Due to rspack using multiple threads for building modules, it means that during the build process of one module, other modules may complete their builds. Since the profile report's time is calculated by subtracting the time of the last handler trigger from the time of the current handler trigger, both succeed_module and build_module will trigger the handler. Through these two hooks, the module's build time can be obtained. However, if other modules trigger succeed_module in the middle, the accuracy of the module's build time may be compromised. I'm not quite sure how to achieve the best compatibility with webpack plugins and ensure the accuracy of module build times. Therefore, I wanted to ask if you have any recommended solutions or best practices for this situation.

ahabhgk:
I suggest update the progress at make start and make end, missing the each module's progress is acceptable.

The progress of seal phase should be more accurate

faga295:
I added a time parameter to `default_handler`. If this parameter `is_some`, the report time will use this time.The `self.active_modules` will store the build start time.The `succeed_module` will use the build start time to get the real build time and finaly pass the time to `handler`.

ahabhgk:
!bench

ahabhgk (approved):
Thanks

PR #5032, opened 2023-12-15
refactor: make sure add all runtime globals when process_runtime_requirements

ahabhgk:
## Summary

https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/Compilation.js#L3509-L3513

`set` can insert when iterating it

## Require Documentation?

- [ ] No


ahabhgk:
!bench

ahabhgk:
!eco-ci

ahabhgk:
!diff

crates/rspack_plugin_runtime/src/runtime_module/css_loading.rs
```diff
@@ -7,21 +7,22 @@ use rspack_identifier::Identifier;
...
+  get_chunk_runtime_requirements,
+  runtime_module::{render_condition_map, stringify_chunks},
+};
 
-#[derive(Debug, Default, Eq)]
+#[derive(Debug, Eq)]
 pub struct CssLoadingRuntimeModule {
   id: Identifier,
   chunk: Option<ChunkUkey>,
-  runtime_requirements: RuntimeGlobals,
```
ahabhgk, line 16: `runtime_requirements` should use `get_chunk_runtime_requirements`, this is a copied value, but actually we need a reference, but using reference will break the borrow checker rules, so using `get_chunk_runtime_requirements` to get it

LingyuCoder (approved):

Issue #5022, opened 2023-12-14
'symlinks' at "resolve" validation error


bostondv:
### System Info

System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Xeon(R) Platinum 8175M CPU @ 2.50GHz
    Memory: 37.57 GB / 62.01 GB
    Container: Yes
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 18.19.0 - ~/.nvm/versions/node/v18.19.0/bin/node
    Yarn: 1.22.4 - ~/carrot/customers/store/node_modules/.bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v18.19.0/bin/npm
    Watchman: 20231210.093346.0 - /usr/local/bin/watchman
  npmPackages:
    @rspack/cli: 0.4.3 => 0.4.3 
    @rspack/core: 0.4.3 => 0.4.3 
    @rspack/plugin-react-refresh: 0.4.3 => 0.4.3

### Details

When specifying the `symlinks` option in `resolve` section a config validation error is thrown `Unrecognized key(s) in object: 'symlinks' at "resolve"`

However this is a valid option in webpack and appears to work in rspack.

The option is missing from the validation types here: https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/config/zod.ts#L353

Can we add it to the config validation? 🙏 

### Reproduce Steps

1. Take the [basic example](https://github.com/web-infra-dev/rspack/blob/main/examples/basic/rspack.config.js) and change the config to

```
{
	entry: "./src/index.js",
	plugins: [
		new rspack.HtmlRspackPlugin({
			template: "./index.html"
		})
	],
        resolve: {
               symlinks: false
        }
}
```

2. Run build or serve
3. Observe config error (or warning)

hardfist:
@bostondv can you help make a pr to fix this

bostondv:
> @bostondv can you help make a pr to fix this

For sure, thanks for looking!

PR #5034, opened 2023-12-15
fix: add "symlinks" option to resolve config validation

bostondv:
## Summary

Fixing https://github.com/web-infra-dev/rspack/issues/5022

When specifying the symlinks option in resolve section a config validation error is thrown Unrecognized key(s) in object: 'symlinks' at "resolve"

## Test Plan

- Added `resolve.symlinks` to basic example and ran the build. No error thrown or warning printed
- Ran the test suite

I don't think any new tests need to be written for this case?

## Require Documentation?

- [x] No


ahabhgk (approved):
thanks

PR #5036, opened 2023-12-16
fix(remove-parent-modules): only filter runtime chunk

JSerFeng:
## Summary

When removing parent modules, filter runtime chunk properly

## Require Documentation?

- [ ] No


JSerFeng:
!canary

JSerFeng:
!bench

jerrykingxyz (approved):

PR #5055, opened 2023-12-18
fix: unnecessary import runtime code from css_loading_with_loading.js

SyMind:
## Summary

Avoid importing runtime code from `css_loading_with_loading.js` when no CSS files are imported.

The relevant code lines in the webpack can be found here: https://github.com/webpack/webpack/blob/main/lib/css/CssLoadingRuntimeModule.js#L87

## Require Documentation?

- [x] No


LingyuCoder (approved):
LGTM, thanks

PR #5103, opened 2023-12-22
chore: extends DevServer types in @rspack/dev-server instead of declare them in @rspack/core, as @rspack/core has no ability of dev server

xc2:
## Require Documentation?

- [ ] No


xc2:
!canary

PR #5108, opened 2023-12-24
feat: improve SourceMapDevToolPlugin

### LINKED ISSUE #4893: Support output.devtoolModuleFilenameTemplate config

Rsbuild / create-react-app use `output.devtoolModuleFilenameTemplate` to support VS Code break points in monorepo.

Provide the same `output.devtoolModuleFilenameTemplate` config as webpack.

Ref: https://webpack.js.org/configuration/output/#outputdevtoolmodulefilenametemplate

### LINKED ISSUE #5060: support rspack.SourceMapDevToolPlugin

support custom sourceMap options 

same as https://webpack.js.org/plugins/source-map-dev-tool-plugin/

SyMind:
## Summary

Fix #5060
Fix #4893

## Test Plan

Ensure the normal execution of current source map test cases and migrate more test cases from webpack.

## Require Documentation?

- [x] No


SyMind:
In this PR, I've attempted to support all options of the `SourceMapDevToolPlugin`, but some options are still not supported so far, such as `fileContext`.

However, due to the extensive changes made, I believe it is now appropriate to proceed with a code review to avoid any potential directional mistakes.

crates/rspack_core/src/options/compiler_options.rs
```diff
@@ -15,7 +17,7 @@ pub struct CompilerOptions {
   pub resolve: Resolve,
   pub resolve_loader: Resolve,
   pub module: ModuleOptions,
-  pub devtool: Devtool,
+  pub devtool: Mutex<Devtool>,
```
ahabhgk, line 20: The Mutex is not necessary at all, this should be readonly, all options are finalized at js side

crates/rspack_core/src/normal_module.rs
```diff
@@ -398,6 +402,7 @@ impl Module for NormalModule {
         code_generation_dependencies: &mut code_generation_dependencies,
         build_info: &mut build_info,
         build_meta: &mut build_meta,
+        devtool: &devtool,
```
ahabhgk, line 405: should have this in ParseResult

crates/rspack_plugin_devtool/src/lib.rs
```diff
@@ -66,88 +200,231 @@ impl Plugin for SourceMapDevToolPlugin {
...
+      devtool.add_hidden();
+    }
+    if !self.columns {
+      devtool.add_cheap();
+    }
+    if self.no_sources {
+      devtool.add_no_sources();
+    }
+    if self.module {
+      devtool.add_module();
```
ahabhgk, line 224: module should setup by SourceMapDevToolModuleOptionsPlugin

crates/rspack_core/src/options/compiler_options.rs
SyMind: The `devtool` setting currently is not yet finalized on the JavaScript side. The `SourceMapDevToolPlugin` will configure it completely.

In Webpack, the [SourceMapDevToolModuleOptionsPlugin](https://github.com/webpack/webpack/blob/main/lib/SourceMapDevToolModuleOptionsPlugin.js) is used to set `useSourceMap` or `useSimpleSourceMap` to true, which signals that each module should generate a source map.

In rspack, I've used the `devtool` setting to achieve the aim .

crates/rspack_plugin_devtool/src/lib.rs
```diff
@@ -274,6 +756,24 @@ impl Plugin for EvalSourceMapDevToolPlugin {
...
+    // TODO: Temporarily use `devtool` to pass source map configuration information
+    let mut devtool = _args.compilation.options.devtool.lock().await;
+    devtool.add_source_map();
+    devtool.add_eval();
+    if !self.columns {
+      devtool.add_cheap();
+    }
+    if self.no_sources {
+      devtool.add_no_sources();
+    }
```
ahabhgk, line 773: also need to handle "module" (eval-cheap-module-source-map)

crates/rspack_plugin_devtool/src/lib.rs
SyMind: I believe it is not the right time to implement `SourceMapDevToolModuleOptionsPlugin`. We should wait for the design to finalize regarding `useSourceMap` and `useSimpleSourceMap` within the module.

crates/rspack_core/src/normal_module.rs
SyMind: This `devtool` setting will be used in the parse function, which is not asynchronous.

crates/rspack_plugin_devtool/src/lib.rs
SyMind: The decision might be made in `packages/rspack/src/rspackOptionsApply.ts`.

crates/rspack_core/src/options/compiler_options.rs
SyMind: It's worth mentioning that a comment in the webpack code suggests the `useSourceMap` and `useSimpleSourceMap`  might be refactored in webpack 6, which is something I'm not very familiar with. You can see the comment here: https://github.com/webpack/webpack/blob/main/lib/Module.js#L169

ahabhgk: it's only a refactor without any logic change, refactor `useSourceMap` and `useSimpleSourceMap` to one enum

SyMind: // TODO refactor this -> options object filled from Factory

ahabhgk: ohh, then we can delay SourceMapDevToolModuleOptionsPlugin, but devtool shouldn't wrapped by Mutex

SyMind: If we don't wrap `devtool` with a Mutex, what method should I use to modify it in `SourceMapDevToolModulePlugin`?

SyMind: @ahabhgk I have used `RwLock` instead of `Mutex`. It's enough for the goal of modifying `devtool` in `compilation` hook and it's more cheap.

SyMind: Do you have any other methods?

SyMind:
@ahabhgk I've introduced a new trait called `SourceMapConf`. The `Module` trait now extends it. This is intended to implement features similar to webpack's `useSourceMap` and `useSimpleSourceMap`.

Additionally, I've removed the `Devtool` struct as the responsibilities are now handled by the `SourceMapDevToolPlugin` and the `SourceMapConf` trait.

crates/rspack_core/src/options/compiler_options.rs
```diff
@@ -15,7 +15,7 @@ pub struct CompilerOptions {
   pub resolve: Resolve,
   pub resolve_loader: Resolve,
   pub module: ModuleOptions,
-  pub devtool: Devtool,
+  pub devtool: String,
```
ahabhgk, line 18: This can be removed

crates/rspack_core/src/parser_and_generator.rs
```diff
@@ -2,6 +2,7 @@ use std::{collections::HashMap, fmt::Debug};
 
 use derivative::Derivative;
 use rkyv::AlignedVec;
+use rspack_common::SourceMapKind;
```
ahabhgk, line 5: Can we move this into rspack_core?

crates/rspack_plugin_css/src/runtime/mod.rs
```diff
@@ -10,6 +11,7 @@ use rspack_plugin_runtime::{
 };
 use rustc_hash::FxHashSet as HashSet;
 
+#[impl_runtime_module]
```
ahabhgk, line 14: We can also improve other code like `DependenciesBlock` by proc macro

ahabhgk (commented):
Generally looks good to me

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
bench!

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

ahabhgk (approved):

SyMind:
!bench

PR #5117, opened 2023-12-25
fix(plugin-runtime): `stringify_chunk` should escape special chars in chunk names as js\'s JSON.stringify do

xc2:
## Summary

Fix artifact problem on case:

https://github.com/web-infra-dev/rspack/blob/a5e489842cdd0f4599757f8ae868a9a6154cfad0/packages/rspack/tests/configCases/chunk-loading/chunk-name/webpack.config.js

image

```markdown
* Feature Related: Chunk Loading
* Case: There's `\\u`, `\\x` or `"` in chunk name
```

The artifact would contain code like `var installedChunks = {"aa\uaaa\xaaa"aaa": 0,};` that is invalid.

### Real world case:

```markdown
* OS: Windows
* Setup: html-webpack-plugin@5.6.0
* Edge situation:  There's any path component starting with `u` or `x`
```

Error sample:

```
start   Compiling...
error   Compile error: 
Failed to compile, check the errors for troubleshooting.
SyntaxError
  × [html-rspack-plugin]: Invalid Unicode escape sequence
  │ C:\userCode\23-2-2\vue-app6\node_modules\@rsbuild\core\static\template.html:74
  │ var installedChunks = {"HtmlWebpackPlugin_0-C:\userCode\23-2-2\vue-app6\node_modules\html-webpack-plugin\lib\loader.js!C:\userCode\23-2-2\vue-
  │ app6\node_modules\@rsbuild\core\static\template.html": 0,};
  │                                               ^^^^^^
```

## Test Plan

Unit tests and configTest case added

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_runtime/src/runtime_module/utils.rs
+++ b/crates/rspack_plugin_runtime/src/runtime_module/utils.rs
@@ -132,7 +132,12 @@ pub fn stringify_chunks(chunks: &HashSet<String>, value: u8) -> String {
   format!(
     r#"{{{}}}"#,
     v.iter().fold(String::new(), |prev, cur| {
-      prev + format!(r#""{cur}": {value},"#).as_str()
+      prev
+        + format!(
+          r#"{}: {value},"#,
+          serde_json::to_string(cur).expect("chunk to_string failed")
+        )
+        .as_str()
     })
   )
 }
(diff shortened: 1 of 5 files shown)


xc2:
!bench

crates/rspack_plugin_runtime/src/runtime_module/utils.rs
```diff
@@ -346,3 +351,21 @@ fn test_get_undo_path() {
     "../../"
   );
 }
+
+#[test]
+fn test_stringify_chunks() {
+  #[rustfmt::skip]
```
ahabhgk, line 357: Actually add test at `packages/rspack/tests` is good enough, it's ok to add unit test at rust side, but since we are iterating fast about code, so unit test usually not keep long. And can you don't skip rustfmt is you wan't to keep this unit test?

ahabhgk (commented):
General looks good to me

crates/rspack_plugin_runtime/src/runtime_module/utils.rs
xc2: > Actually add test at `packages/rspack/tests` is good enough, it's ok to add unit test at rust side, but since we are iterating fast about code, so unit test usually not keep long. And could you make this not skip rustfmt is you still want to keep this unit test?

I’ll drop the unit test. It was useful before when I hadn’t switched to `serde_json`, but it’s not as relevant now that I’ve made the switch.

btw i add `#[rustfmt::skip]` there as i want the case array `pass` clean.

ahabhgk (approved):
Thanks

xc2:
Thank you @ahabhgk

PR #5119, opened 2023-12-26
fix(rspack-dev-server): remove webpack dep

hardfist:
## Summary
It seems we don't use webpack in rspack-dev-server

## Test Plan
No Need

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack-dev-server/package.json
+++ b/packages/rspack-dev-server/package.json
@@ -43,7 +43,6 @@
     "express": "4.18.1",
     "http-proxy-middleware": "2.0.6",
     "mime-types": "2.1.35",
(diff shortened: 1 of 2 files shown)


LingyuCoder (approved):

ahabhgk:
This will cause `Error: Cannot find module 'webpack/hot/dev-server'`, revert in https://github.com/web-infra-dev/rspack/pull/5261/files#diff-fa2af2cbbd416f67bb21aaa181218db15fb78d0b36da8b031ff14bf18d0770dc

hardfist:
@ahabhgk we should reopen this and find alternative solution other than depends on webpack(which increase size)

Issue #5067, opened 2023-12-19
removes all deprecate warning in CI 


hardfist:
https://github.com/web-infra-dev/rspack/actions/runs/7261029436/job/19781998400#step:5:77
some of the test cases are still using deprecated configuration which should be migrated to new configuration
![image](https://github.com/web-infra-dev/rspack/assets/8898718/2dcb2bc1-417a-4707-8d3f-ebbb2f813cf0)

lippzhang:
please assige to me

LingyuCoder:
resolved by #5127

PR #5127, opened 2023-12-27
feat: removes all deprecate warning in CI

lippzhang:
## Summary
removes all deprecate warning in CI

https://github.com/web-infra-dev/rspack/issues/5067

## Test Plan

watch CI info

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack/src/util/index.ts
+++ b/packages/rspack/src/util/index.ts
@@ -1,6 +1,9 @@
-import type { JsAssetInfo, JsStatsError } from "@rspack/binding";
-import { AssetInfo } from "../Compilation";
 import terminalLink from "terminal-link";
+import isCI from "is-ci";
+
+import type { JsAssetInfo, JsStatsError } from "@rspack/binding";
+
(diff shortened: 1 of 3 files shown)


LingyuCoder (approved):

LingyuCoder:
LGTM, thanks

packages/rspack/src/util/index.ts
```diff
@@ -115,7 +118,8 @@ const getDeprecationStatus = () => {
 	const defaultEnableDeprecatedWarning = true;
 	if (
 		process.env.RSPACK_DEP_WARNINGS === "false" ||
-		process.env.RSPACK_DEP_WARNINGS === "0"
+		process.env.RSPACK_DEP_WARNINGS === "0" ||
+		isCI
```
h-a-n-a, line 122: I don't think we should disable deprecation warnings for every user. Only warnings in Rspack repo should be disabled instead.

lippzhang: i will fix it

PR #5129, opened 2023-12-27
fix: user freindly error msg for output.hashFunction

### LINKED ISSUE #4504: Improve error message when unsupported output.hashFunction method is used

I got a cryptic error message:

> thread '<unnamed>' panicked at 'not implemented: sha256', crates/rspack_hash/src/lib.rs:22:12

my config:
```
...
output: {
  ...,
  hashFunction: "sha256"
}
```

and it's completely valid for webpack, but not for rspack.

It would be nice to print a list of supported methods instead.
As a workaround I used `hashFunction: "xxhash64",`

bvanjoi:
Fixes #4504

prev:

image

now:

image

### DIFF
--- a/packages/rspack/tests/ConfigCase.template.ts
+++ b/packages/rspack/tests/ConfigCase.template.ts
@@ -234,7 +234,9 @@ export const describeCases = config => {
 							const deprecationTracker = deprecationTracking.start();
 							const onCompiled = (err, stats) => {
 								const deprecations = deprecationTracker();
-								if (err) return handleFatalError(err, done);
+								if (err) {
+									return handleFatalError(err, done);
+								}
 								const statOptions = {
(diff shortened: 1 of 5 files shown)


packages/rspack/tests/ConfigCase.template.ts
```diff
@@ -632,7 +634,11 @@ export const describeCases = config => {
 									})
 									.catch(done);
 							};
-							rspack(optionsArr, onCompiled);
+							try {
+								rspack(optionsArr, onCompiled);
+							} catch (err) {
+								onCompiled(err, undefined);
```
bvanjoi, line 640: I have updated the test logic, and it should now function properly.

ahabhgk (approved):

PR #5135, opened 2023-12-27
revert: removes all deprecate warning in CI by github CI

lippzhang:
## Summary
Partially reverts: https://github.com/web-infra-dev/rspack/pull/5127

Keep `@types/is-ci` in `@rspack/core`

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack/src/util/index.ts
+++ b/packages/rspack/src/util/index.ts
@@ -1,5 +1,4 @@
 import terminalLink from "terminal-link";
-import isCI from "is-ci";
 
 import type { JsAssetInfo, JsStatsError } from "@rspack/binding";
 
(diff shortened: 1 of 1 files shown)


packages/rspack/src/util/index.ts
```diff
@@ -1,5 +1,4 @@
 import terminalLink from "terminal-link";
-import isCI from "is-ci";
```
h-a-n-a, line 2: Please remove this package as well ;-).

h-a-n-a: Would you also migrate the configuration to the newest version? This is what the issue means https://github.com/web-infra-dev/rspack/issues/5067.

lippzhang: is-ci existed before, in the prepare field in the script in package.json under the root directory。 i just add @type/is-ci

h-a-n-a (approved):

Issue #5138, opened 2023-12-27
No ESM export when compiled to ESM


hardfist:
### System Info

@rspack/cli@0.4.5

### Details

Rspack should generate export { xxx} but generate var xxx
https://github.com/web-infra-dev/rspack-repro/blob/esm-output/rspack-dist/main.mjs#L68 
which webpack is right 
https://github.com/web-infra-dev/rspack-repro/blob/esm-output/webpack-dist/main.mjs#L89

### Reproduce link

https://github.com/web-infra-dev/rspack-repro/tree/esm-output

### Reproduce Steps

```
yarn install && yarn build
```

LingyuCoder:
New tree shaking will resolve this issue, close as not planned

alfaproject:
@LingyuCoder I'm using the new tree-shaking. This was reported under it

LingyuCoder:
> @LingyuCoder I'm using the new tree-shaking. This was reported under it

@alfaproject Try this version `[0.4.5-canary-285c5f3-20231228011645](https://www.npmjs.com/package/@rspack/core/v/0.4.5-canary-285c5f3-20231228011645)`, it has been resolved but not published yet

alfaproject:
I can confirm that version fixes the module output but now I'm getting this panic in another library:
```
Bundling apps/lambda-api/src/handlers/slack-server.ts
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  range end index 18446744073709551615 out of range for slice of length 0
Location: crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_self_reference_dependency.rs:57

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
<empty backtrace>
```

Not sure if it's related to your fix but it looks like something else. That is a regression tho

LingyuCoder:
> I can confirm that version fixes the module output but now I'm getting this panic in another library:
> 
> ```
> Bundling apps/lambda-api/src/handlers/slack-server.ts
> Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
[...]

PR #5150, opened 2023-12-28
fix: libraryType=module when tree shaking is disabled

LingyuCoder:
## Summary

resolve: #5138

- When libraryType=module, ModuleLibraryPlugin needs exportsInfo to generate startup code
- The `providedExports` will always be true even when tree shaking is disabled
- Old tree shaking will not recognize `providedExports`, so this pr fixes it
- But `JsModule::analyze` may cause performance regression, so exportsInfo will be analyzed only when `providedExports=true` and `libraryType=module` 

## Test Plan

- Add `packages/rspack/tests/configCases/library/issue-5138`

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -837,7 +837,15 @@ impl Compilation {
                 }
               }
 
-              if self.options.builtins.tree_shaking.enable() {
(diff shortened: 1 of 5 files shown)


LingyuCoder:
The old version of tree shaking should not be maintained anymore, so this issue will not be fixed unless users report it.

PR #5154, opened 2023-12-28
fix: empty name and complex member expression in cjs dependency

LingyuCoder:
## Summary

- Compatible with empty names of CommonJsSelfReferenceDependency
- Fix commonjs export scanner to compatible complex member expression

## Test Plan

- Add `packages/rspack/tests/cases/cjs-tree-shaking/complex-member`

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_self_reference_dependency.rs
+++ b/crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_self_reference_dependency.rs
@@ -53,9 +53,13 @@ impl ModuleDependency for CommonJsSelfReferenceDependency {
     _runtime: Option<&rspack_core::RuntimeSpec>,
   ) -> Vec<rspack_core::ExtendedReferencedExport> {
     if self.is_call {
-      vec![ExtendedReferencedExport::Array(
-        self.names[0..self.names.len() - 1].to_vec(),
-      )]
+      if self.names.is_empty() {
+        vec![ExtendedReferencedExport::Array(vec![])]
+      } else {
+        vec![ExtendedReferencedExport::Array(
+          self.names[0..self.names.len() - 1].to_vec(),
+        )]
+      }
     } else {
       vec![ExtendedReferencedExport::Array(self.names.clone())]
     }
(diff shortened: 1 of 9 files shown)


LingyuCoder:
!canary

alfaproject:
Something is wrong with this canary. It's removing `exports` properties that it shouldn't. Two examples:

```diff
-    for(let i = 0; i < keys.length; i++)debug.inspectOpts[keys[i]] = exports.inspectOpts[keys[i]];
+    for(let i = 0; i < keys.length; i++)debug.inspectOpts[keys[i]] = exports;
```

```diff
-         */ this.level = exports.ReadConcernLevel[level] ?? level;
+         */ this.level = exports ?? level;
```

alfaproject:
It also added a lot more copies than stable release: 
Screenshot 2023-12-28 at 19 05 32

It looks like a lot of modules now import the cjs versions of themselves when they self reference (in addition to the esm version)

LingyuCoder:
> Something is wrong with this canary. It's removing `exports` properties that it shouldn't. Two examples:
> 
> ```diff
> -    for(let i = 0; i < keys.length; i++)debug.inspectOpts[keys[i]] = exports.inspectOpts[keys[i]];
> +    for(let i = 0; i < keys.length; i++)debug.inspectOpts[keys[i]] = exports;
> ```
> 
> ```diff
> -         */ this.level = exports.ReadConcernLevel[level] ?? level;
> +         */ this.level = exports ?? level;
> ```

This seems to be a bug in extract_member_expression_chain, and I'll try to fix it.

LingyuCoder:
!canary

LingyuCoder:
> 0.4.5-canary-2639105-20231229105328

@alfaproject This is a bug of commonjs export scanner. You can try this version.

LingyuCoder:
> It also added a lot more copies than stable release: Screenshot 2023-12-28 at 19 05 32
> 
> It looks like a lot of modules now import the cjs versions of themselves when they self reference (in addition to the esm version)

This is an known issue. In webpack, SelfModuleFactory was used to handle self dependency and return the reference of parent module, but in rust it may cause a borrowed issue. So rspack use an empty SelfModule to handle self dependency. 

Perhaps SelfModule should be excluded in stats, I created an issue for tracking #5158

IWANABETHATGUY (approved):

PR #5162, opened 2023-12-29
feat: support logical evaluation

LingyuCoder:
## Summary

Fix #5038

## Test Plan

- Add `packages/rspack/tests/cases/parsing/evaluate-logical`

## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #5161, opened 2023-12-29
feat: support evaluate_typeof of module variables

LingyuCoder:
## Summary

Support evaluate_typeof of module variables. 

## Test Plan

- Add `packages/rspack/tests/cases/parsing/typeof-api`

## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #5163, opened 2023-12-29
fix: do not generate dependency after removing

LingyuCoder:
## Summary

Fix code generation error when code removed by compile time evaluation

```js
// source
if (typeof require !== 'function' /* always false */) {
  __webpack_hash__
}

// before
if (false /* always false */ ) __webpack_require__.h(){}

// after
if (false /* always false */ ) {}
```

Due to the lack of parser hooks, scanners will not be stopped even when the statements had been removed. So dependencies will be generated and trigger DependencyTemplate.apply as well. The code will be insert and cause runtime error.

## Test Plan

- Add `packages/rspack/tests/cases/parsing/remove-dependency`

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs
@@ -1,6 +1,7 @@
 use rspack_core::{
   extract_member_expression_chain, BoxDependency, BuildMeta, BuildMetaDefaultObject,
-  BuildMetaExportsType, DependencyTemplate, ModuleType, RuntimeGlobals, SpanExt,
+  BuildMetaExportsType, DependencyLocation, DependencyTemplate, ModuleType, RuntimeGlobals,
+  SpanExt,
 };
 use swc_core::{
   atoms::Atom,
@@ -36,6 +37,7 @@ pub struct CommonJsExportDependencyScanner<'a> {
   stmt_level: u32,
   last_stmt_is_expr_stmt: bool,
   is_top_level: bool,
+  removed: &'a mut Vec<DependencyLocation>,
 }
 
 impl<'a> CommonJsExportDependencyScanner<'a> {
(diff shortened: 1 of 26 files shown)


LingyuCoder:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/5163?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

Current dependencies on/for this PR:
* `main`
  * **PR #5161** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/5161?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
    * **PR #5162** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/5162?utm_source=stack-comment-icon" target="_blank">Graphite</a> 
      * **PR #5163** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/5163?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/web-infra-dev/rspack/5163?utm_source=stack-comment).

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs
```diff
@@ -36,6 +37,7 @@ pub struct CommonJsExportDependencyScanner<'a> {
   stmt_level: u32,
   last_stmt_is_expr_stmt: bool,
   is_top_level: bool,
+  ignored: &'a mut Vec<DependencyLocation>,
```
bvanjoi, line 40: I don't think that ﻿`ignored` is an optimal solution. Could we perhaps align the behavior with that of webpack?

LingyuCoder: If we want to align with webpack, we need full implement of parser hooks. In webpack, JavascriptParser use SyncBailHook, means that if some plugin which tap these hooks and return anthing but not `undefined`, the other taps will not execute, just equivalent to `ignored`. But in rspack, each scanner runs independently.

bvanjoi (approved):

PR #5180, opened 2024-01-02
feat: add common_js_full_require_dependency

LingyuCoder:
## Summary

Add CommonJsFullRequireDependency

## Test Plan

- Add `packages/rspack/tests/cases/cjs-tree-shaking/full-require`

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_core/src/compiler/queue.rs
+++ b/crates/rspack_core/src/compiler/queue.rs
@@ -222,6 +222,26 @@ impl AddTask {
     if let Some(current_profile) = &self.current_profile {
       current_profile.mark_integration_start();
     }
+
+    if self.module.as_self_module().is_some() {
+      let issuer = self
+        .module_graph_module
+        .get_issuer()
+        .identifier()
+        .expect("self module should have issuer");
+
+      set_resolved_module(
+        &mut compilation.module_graph,
+        self.original_module_identifier,
+        self.dependencies,
+        *issuer,
+      )?;
+
+      return Ok(TaskResult::Add(Box::new(AddTaskResult::ModuleReused {
+        module: self.module,
+      })));
+    }
+
     let module_identifier = self.module.identifier();
 
     if compilation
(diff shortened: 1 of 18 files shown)


packages/rspack/tests/__snapshots__/Stats.test.ts.snap
```diff
@@ -122,6 +122,13 @@ exports[`Stats should have stats 1`] = `
               "type": "entry",
               "userRequest": "./fixtures/a",
             },
+            {
+              "moduleId": "585",
+              "moduleIdentifier": "<PROJECT_ROOT>/tests/fixtures/a.js",
+              "moduleName": "./fixtures/a.js",
+              "type": "cjs self exports reference",
```
LingyuCoder, line 129: Webpack also have cjs self exports reference info in stats

crates/rspack_core/src/self_module_factory.rs
```diff
@@ -12,8 +12,6 @@ impl ModuleFactory for SelfModuleFactory {
     let issuer = data
       .issuer_identifier
       .expect("self module must have issuer");
-    Ok(ModuleFactoryResult::new_with_module(Box::new(
-      SelfModule::new(issuer),
-    )))
+    Ok(ModuleFactoryResult::new_with_module(Box::new(SelfModule::new(issuer))).from_cache(true))
```
LingyuCoder, line 15: This is a proxy module which no need to be resolved, so always from_cache=true

crates/rspack_core/src/compiler/queue.rs
```diff
@@ -222,6 +222,26 @@ impl AddTask {
...
+        self.original_module_identifier,
+        self.dependencies,
+        *issuer,
+      )?;
+
+      return Ok(TaskResult::Add(Box::new(AddTaskResult::ModuleReused {
+        module: self.module,
+      })));
+    }
+
```
LingyuCoder, line 244: proxy self module to reuse parent module, then the exports info should be valid

IWANABETHATGUY (approved):

JSerFeng:
!bench

PR #5190, opened 2024-01-03
test: jest serializer path should be working with non-ascii code

h-a-n-a:
## Summary

The old package `slash`  introduced in `jest-serializer-path` which is a long not maintained project is later fixed this issue.

This PR fixed `jest-serializer-path` not working when non-ascii code contains by syncing the implementation from the newest version.

## Require Documentation?

- [ ] No

### DIFF
--- a/scripts/test/slash.cjs
+++ b/scripts/test/slash.cjs
@@ -1,0 +1,15 @@
+/**
+ * The following code is copied from
+ * https://github.com/sindresorhus/slash/blob/98b618f5a3bfcb5dd374b204868818845b87bb2f/index.js
+ *
+ * MIT Licensed
+ * Author Sindre Sorhus @sindresorhus
+ */
+module.exports = function slash(path) {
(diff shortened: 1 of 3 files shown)


scripts/test/slash.cjs
```diff
@@ -0,0 +1,15 @@
+/**
+ * The following code is copied from
+ * https://github.com/sindresorhus/slash/blob/98b618f5a3bfcb5dd374b204868818845b87bb2f/index.js
+ *
+ * MIT Licensed
+ * Author Sindre Sorhus @sindresorhus
+ */
+module.exports = function slash(path) {
```
h-a-n-a, line 8: This one is intended to be used only with path input, however, `jest-serializer-path` use it to everything. It's working in most cases though.

ahabhgk (approved):

jerrykingxyz (approved):

PR #5195, opened 2024-01-03
refactor: store build_info and build_meta in normal_module

JSerFeng:
## Summary

Store build_info and build_meta in normal_module

## Require Documentation?

- [ ] No


IWANABETHATGUY:
Does this mean we can't mutate the `build_info|data` after the build process?

IWANABETHATGUY:
Can you double check if webpack has mutated the build_meta after `build` phase ?

PR #5212, opened 2024-01-04
fix: remove module/exports evaluation 

LingyuCoder:
## Summary

Source Code: 
```js
// lodash-es/isBuffer.js
import root from './_root.js';
import stubFalse from './stubFalse.js';

/** Detect free variable `exports`. */
var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;

// ...
```

Webpack register module&exports evaluateTypeOf only when module type is JAVASCRIPT_MODULE_TYPE_AUTO or JAVASCRIPT_MODULE_TYPE_DYNAMIC. So typeof module and typeof exports will not be evalutated in ESM.

```js
// exports is undefined in esm, so the following code will not execute
var freeExports = typeof exports == 'object' && exports && !exports.nodeType && exports;
```

But currently rspack CommonJsImportScanner will evaluate in all types of javascript modules.
```js
var freeExports = true && exports && !exports.nodeType && exports;
```

According to this, should remove module&exports evaluation and add them back in the future when scanner can be full align with webpack.

## Test Plan

- Fixed

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_export_scanner.rs
@@ -22,37 +22,8 @@ use crate::{
     CommonJsExportRequireDependency, CommonJsExportsDependency, CommonJsSelfReferenceDependency,
     ExportsBase, ModuleDecoratorDependency,
   },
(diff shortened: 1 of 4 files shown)


bvanjoi (approved):

LingyuCoder:
!canary

PR #5214, opened 2024-01-04
refactor: use glob import for prelude

bvanjoi:
clean up


ahabhgk (approved):

JSerFeng:
!bench

bvanjoi:
I can guarantee that it will not impact performance, either during compilation or runtime.

JSerFeng:
I'm just doing bisect searching as I found hmr performance has big regression today

PR #5215, opened 2024-01-04
ci: fix flaky e2e tests

h-a-n-a:
## Summary

NOBODY likes flaky tests ;-(

Use auto-retrying assertions and remove manual pollings.
Ref: https://playwright.dev/docs/api/class-locator#locator-get-attribute
Ref: https://playwright.dev/docs/api/class-locator#locator-text-content

If there's any further tests not working, then It should at least be timeouts instead of something like context destroyed.

## Require Documentation?

- [ ] No

### DIFF
--- a/packages/playground/cases/html/issue-4366/index.test.ts
+++ b/packages/playground/cases/html/issue-4366/index.test.ts
@@ -1,17 +1,12 @@
 import { test, expect } from "@/fixtures";
 
-test("html should refresh after reload", async ({
-	page,
-	fileAction,
-	rspack
-}) => {
-	await expect(page.title()).resolves.toBe("123");
+test("html should refresh after reload", async ({ page, fileAction }) => {
+	await expect(page).toHaveTitle("123");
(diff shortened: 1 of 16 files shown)


packages/playground/cases/html/issue-4366/index.test.ts
```diff
@@ -1,17 +1,12 @@
...
-	rspack
-}) => {
-	await expect(page.title()).resolves.toBe("123");
+test("html should refresh after reload", async ({ page, fileAction }) => {
+	await expect(page).toHaveTitle("123");
 	fileAction.updateFile("./src/index.html", content =>
 		content.replace("123", "456")
 	);
-	await rspack.waitUntil(async () => {
+	await expect(async () => {
```
h-a-n-a, line 8: Use playwright's polling.

bvanjoi: remark: https://playwright.dev/docs/test-assertions#expecttopass

bvanjoi (approved):

PR #5228, opened 2024-01-04
refactor: move swc helpers and react-refresh to peer

ahabhgk:
## Require Documentation?

- [ ] No


packages/rspack/package.json
```diff
@@ -69,14 +69,11 @@
   "dependencies": {
     "@module-federation/runtime-tools": "0.0.0-next-20240102063242",
     "@rspack/binding": "workspace:*",
-    "@swc/helpers": "0.5.1",
```
h-a-n-a, line 72: Should add this to `"peerDependencies"`

ahabhgk: I'm not sure, since this is only necessary when using builtin:swc-loader `externalHelpers: true`

[swc-loader didn't add it as peer](https://github.com/swc-project/pkgs/blob/cbd1c461b128e1d6d350c13409b78153348a84f9/packages/swc-loader/package.json#L14), but swc-loader --peer-> @swc/core --peer-> @swc/helpers

ahabhgk: I'm ok with both, you make the decision

h-a-n-a: I think whether it is optional is to use `"peerDependenciesMeta"`, which indicates '@swc/helper' is not a required one.

h-a-n-a (approved):

PR #5239, opened 2024-01-05
fix(mf1.5): add resolve alias for mf runtime

ahabhgk:
## Require Documentation?

- [ ] No

### DIFF
--- a/packages/rspack/src/container/ModuleFederationPlugin.ts
+++ b/packages/rspack/src/container/ModuleFederationPlugin.ts
@@ -1,3 +1,4 @@
+import path from "node:path";
 import { ExternalsType, externalsType } from "../config";
 import { Compiler } from "../Compiler";
 import { type ModuleFederationPluginV1Options } from "./ModuleFederationPluginV1";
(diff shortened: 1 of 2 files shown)


packages/rspack/src/container/ModuleFederationPlugin.ts
```diff
@@ -147,9 +171,7 @@ function getDefaultEntryRuntime(
 		runtimePluginVars.push(`${runtimePluginVar}()`);
 	}
 	const content = [
-		`import __module_federation_implementation__ from ${JSON.stringify(
-			implementationPath
-		)}`,
+		`import __module_federation_bundler_runtime__ from "@module-federation/runtime-tools/webpack-bundler-runtime"`,
```
h-a-n-a, line 174: Is this intended to be hardcoded?

ahabhgk: No, forgot change

h-a-n-a:
Additionally, I took a look at the `package.json` file. It turns out module federation related dependencies are not added. https://github.com/web-infra-dev/rspack/blob/c24cd5506895547acb274b122449a62abc078308/packages/rspack/package.json#L84

It may cause pnpm resolution issues. Would you please add?

ahabhgk:
I added it to dependencies, should I also add it to peer?

h-a-n-a:
> I added it to dependencies, should I also add it to peer?

Sorry, I misunderstand your intention. If module-federation dependencies are intended to be provided by '@rspack/core' then it's not necessary.

ahabhgk:
it's in the @module-federation/runtime-tools, and resolved from @module-federation/runtime-tools

h-a-n-a:
I see. Then, it's good to go ;-)

h-a-n-a (approved):

PR #5241, opened 2024-01-05
refactor: change return type of `render_manifest` to `TWithDiagnosticArray`

JSerFeng:
## Summary

Change ret type of render_manifest to TWithDiagnostics, so that plugin can emit diagnostics in render_manifest hook

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs
@@ -11,7 +11,7 @@ use rspack_core::{
   PluginCompilationHookOutput, PluginContext, PluginRenderManifestHookOutput, RenderManifestEntry,
   RuntimeGlobals, SelfModuleFactory, SourceType,
 };
-use rspack_error::Result;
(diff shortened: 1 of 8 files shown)


crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs
```diff
@@ -284,13 +284,13 @@ impl Plugin for JsPlugin {
         .runtime(&chunk.runtime),
     );
     asset_info.set_javascript_module(compilation.options.output.module);
-    Ok(vec![RenderManifestEntry::new(
+    Ok(TWithDiagnosticArray::from(vec![RenderManifestEntry::new(
```
h-a-n-a, line 287: Can we just import trait `IntoTWithDiagnosticArray` and use `with_empty_diagnostic`? `From` implementation seems not that necessary to me.

JSerFeng: Wow, didn't notice that before, done

h-a-n-a (approved):

PR #5257, opened 2024-01-07
fix: replace undefined with void in hooks

tmkx:
## Summary

`undefined` requires an explicit `return` statement.

## Require Documentation?

- [x] No


jerrykingxyz (approved):
LGTM, thanks

PR #5261, opened 2024-01-08
chore: fix template version

hardfist:
## Summary
currently template use fixed outdated version which likely breaking in new minor version, we should use latest publish version in template and test against it

## Test Plan
integrated it in eco-ci

## Require Documentation?

- [x] No


ahabhgk:
!canary

ahabhgk (approved):

PR #5264, opened 2024-01-08
refactor: mv some test

bvanjoi:
clean up


packages/rspack/tests/case.template.ts
```diff
@@ -92,6 +92,11 @@ export function describeCases(config: { name: string; casePath: string }) {
 											assert(statsJson.errors!.length > 0);
 										} else if (category.name === "warnings") {
 											assert(statsJson.warnings!.length > 0);
+										} else if (category.name === "no-warnings") {
```
h-a-n-a, line 95: Kindly wondering the reason of this part not  being aligned with webpack. 

From my perspective, these tests can be added into `warning` category with an empty warning array.

Secondly, using `assert` could result tests that failed not being caught by jest, which results in a timeout. Webpack's way is more clear to me: https://github.com/webpack/webpack/blob/6be4065ade1e252c1d8dcba4af0f43e32af1bdc1/test/TestCases.template.js#L354

bvanjoi: > these tests can be added into warning category with an empty warning array

The `﻿warnings` necessitate that the length of ﻿`stats.warnings` be greater than `0`. see above: https://github.com/web-infra-dev/rspack/pull/5264/files#diff-6c0bf465b39653e539e7adf0b2155d9f6f308e0c5eb316ec78059afa491b5348R94

h-a-n-a: Why don't we remove these assertions and use `checkArrayExpectation` to do this job instead?

h-a-n-a: Sorry, maybe I caused some misunderstandings. What I mean is to remove all of this:
https://github.com/web-infra-dev/rspack/pull/5264/commits/f6fe857df8b6dc15c8a2f3b62a9ff3906228ffe2#diff-6c0bf465b39653e539e7adf0b2155d9f6f308e0c5eb316ec78059afa491b5348R90-R105

bvanjoi: The `﻿warnings` and ﻿`errors` can be relocated to the `﻿tests/diagnostics` directory.

h-a-n-a (approved):

packages/rspack/tests/diagnostics/factorize/module-build-error/stats.err
```diff
@@ -0,0 +1,10 @@
+ERROR in ./index.scss NapiError
+
+  × Napi Error: SassError: SassError: Undefined variable.
+  │   ╷
+  │ 2 │   wrong: $wrong;
+  │   │          ^^^^^^
+  │   ╵
+  │   tests/diagnostics/factorize/module-build-error/index.scss 2:10  root stylesheet
+  │     at Object.loader (<HOME_DIR>/rspack/node_modules/sass-loader/dist/index.js:69:14)
```
bvanjoi, line 9: I strongly prefer not to use external libraries in test cases

h-a-n-a (approved):

PR #5272, opened 2024-01-08
feat(cli): support built-in rspack envs

xc2:
### 1. tell the end devs they are using rspack

for sharing configuration file between rspack and webpack.

for instance:

```js
// rspack.config.js
module.exports = function(env) {
  /** align with webpack's WEBPACK_BUNDLE @{link https://github.com/webpack/webpack-cli/blob/eea6adf7d34dfbfd3b5b784ece4a4664834f5a6a/packages/webpack-cli/src/webpack-cli.ts#L2500-L2503} */
  const isRspack = env.RSPACK_BUNDLE;
  const config = {};
  if (isRspack) {
    config.experiments.rspackFuture = {
      newTreeShaking: true
    };
  }
  return config;
}
```

i'm also going to add `WEBPACK_` prefixed builtin env in 0.5.0.

### ~2. tell the plugin devs they are using rspack~

the plugin dev can separately introduce different logics for rspack and webpack.

for instance:

```js
const plugin = {
  apply(compiler) {
    const { isRspack } = compiler;
    if (isRspack) {
      // do something special
    }
  }
}
```

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack-cli/src/commands/build.ts
+++ b/packages/rspack-cli/src/commands/build.ts
@@ -1,7 +1,7 @@
 import * as fs from "fs";
 import type { RspackCLI } from "../rspack-cli";
 import { RspackCommand } from "../types";
-import { commonOptions } from "../utils/options";
+import { commonOptions, setBuiltinEnvArg } from "../utils/options";
 import { MultiStats, Stats } from "@rspack/core";
 
 export class BuildCommand implements RspackCommand {
@@ -21,6 +21,15 @@ export class BuildCommand implements RspackCommand {
 					}
 				}),
 			async options => {
+				if (options.watch) {
+					// @ts-expect-error
+					setBuiltinEnvArg(options.env, "WATCH", true);
+				} else {
+					// @ts-expect-error
+					setBuiltinEnvArg(options.env, "BUNDLE", true);
+					// @ts-expect-error
+					setBuiltinEnvArg(options.env, "BUILD", true);
+				}
 				const logger = cli.getLogger();
 				let createJsonStringifyStream: typeof import("@discoveryjs/json-ext").stringifyStream;
 				if (options.json) {
--- a/packages/rspack-cli/src/commands/build.ts
+++ b/packages/rspack-cli/src/commands/build.ts
@@ -1,7 +1,11 @@
 import * as fs from "fs";
 import type { RspackCLI } from "../rspack-cli";
 import { RspackCommand } from "../types";
-import { commonOptions, setBuiltinEnvArg } from "../utils/options";
+import {
+	commonOptions,
+	ensureEnvObject,
+	setBuiltinEnvArg
+} from "../utils/options";
 import { MultiStats, Stats } from "@rspack/core";
 
 export class BuildCommand implements RspackCommand {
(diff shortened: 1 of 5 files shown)


h-a-n-a:
I personally may discourage this change, as we already have `compiler.webpack.rspackVersion` to determine in plugins.

packages/rspack-cli/src/commands/build.ts
```diff
@@ -21,6 +21,15 @@ export class BuildCommand implements RspackCommand {
 					}
 				}),
 			async options => {
+				if (options.watch) {
+					// @ts-expect-error
+					setBuiltinEnvArg(options.env, "WATCH", true);
```
h-a-n-a, line 26: Does webpack support this?

xc2: > Does webpack support this?

yes webpack-cli support it.

https://github.com/webpack/webpack-cli/blob/eea6adf7d34dfbfd3b5b784ece4a4664834f5a6a/packages/webpack-cli/src/webpack-cli.ts#L2500-L2503

WEBPACK_BUILD / WEBPACK_BUNDLE for non-watch build
WEBPACK_WATCH for watch build
WEBPACK_SERVE for serve

thus i add `RSPACK_*` correspondingly.

another question is whether we should add these `WEBPACK_*` together to align with webpack-cli. I didn't do it as it will break those who use `WEBPACK_*` to determine who (webpack or rspack) is using the function configuration .

xc2:
> I personally may discourage the second change, as we already have `compiler.webpack.rspackVersion` to determine in plugins.

Oh right. i'll drop the change later

xc2:
> > I personally may discourage the second change, as we already have `compiler.webpack.rspackVersion` to determine in plugins.
> 
> Oh right. i'll drop the change later

done

packages/rspack-cli/src/commands/build.ts
h-a-n-a: > another question is whether we should add these WEBPACK_* together to align with webpack-cli. I didn't do it as it will break those who use WEBPACK_* to determine who (webpack or rspack) is using the function configuration .

I would say let's hold the `WEBPACK_*` for now.

packages/rspack-cli/src/commands/build.ts
```diff
@@ -21,6 +21,15 @@ export class BuildCommand implements RspackCommand {
...
+				} else {
+					// @ts-expect-error
```
h-a-n-a, line 28: `@ts-expect-error` should be removed

xc2: done

packages/rspack-cli/src/commands/build.ts
xc2: > > another question is whether we should add these WEBPACK_* together to align with webpack-cli. I didn't do it as it will break those who use WEBPACK_* to determine who (webpack or rspack) is using the function configuration .
> 
> I would say let's hold the `WEBPACK_*` for now.

ok

packages/rspack-cli/tests/build/basic/entry.env.js
```diff
@@ -0,0 +1,11 @@
+module.exports = function (env) {
+	console.log(
```
h-a-n-a, line 2: I don't suggest you to use stdout for testing purpose. Please use other solutions like `DefinePlugin`

h-a-n-a: Let's get it merged. This looks ok to me.

h-a-n-a:
MORE CONTEXT: https://webpack.js.org/api/cli/#env

h-a-n-a (approved):

PR #5124, opened 2023-12-26
fix: should not create additional missing modules

h-a-n-a:
## Summary

Follow up on https://github.com/web-infra-dev/rspack/pull/5084.
closes https://github.com/web-infra-dev/rspack/issues/5085.

This PR removes the old missing module. To replace with the old one, this PR aligns the missing module in code generation stage of each referencing module of the missing modules.

This PR also moves `diagnostics` to `ModuleFactoryCreateData`, as `ModuleFactory` is singleton if `DependencyType` is the same to avoid conflicts between different creations.

## Test Plan

Enabled more compiler tests for `options.bail`.
Added diagnostic test for `empty-dependency`.

## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #5301, opened 2024-01-11
fix: module graph incorrect when delete or create file

jerrykingxyz:
## Summary

This PR fix the module graph incorrect when delete or create file. There are some summary:

* Add parent module to `force_build_module` when module factorize failed instead of add deps to `force_build_deps`. if the module is entry module, we should still add deps to `force_build_deps`, because of it no parent module.
* Deleting the file should not only mark the affected module as `force_build_module` but also add its parent module to `force_build_module`

We can see a simply example:
``` js
// index.js
import a form "./a" // <-- HarmonyImportSideEffectDependency
console.log(a) // <-- HarmonyImportSpecifierDependency
```
the `index.js` have two deps `HarmonyImportSideEffectDependency` and `HarmonyImportSpecifierDependency` . 

If we don't have `a.js` and create it, we should not only create the module for `a.js` in HMR but also recreate `index.js` to make sure all dependencies are correct
if we have `a.js` and delete it, we should not only delete the module for `a.js` in HMR but also recreate `index.js` to make sure all dependencies are correct

## Test Plan

* enable missing file e2e test

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/node_binding/src/plugins/mod.rs
+++ b/crates/node_binding/src/plugins/mod.rs
@@ -148,7 +148,7 @@ impl rspack_core::Plugin for JsHooksAdapter {
     &self,
     _ctx: rspack_core::PluginContext,
     _compilation: &mut rspack_core::Compilation,
-    _param: &mut rspack_core::MakeParam,
+    _params: &mut Vec<rspack_core::MakeParam>,
   ) -> rspack_core::PluginMakeHookOutput {
     if self.is_hook_disabled(&Hook::Make) {
       return Ok(());
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -432,13 +432,13 @@ impl Compilation {
   }
 
   #[instrument(name = "compilation:make", skip_all)]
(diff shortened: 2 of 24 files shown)


h-a-n-a:
Summary looks good to me. 

To be specific, missing dependencies will interfere the code generation of parent modules. 

To demonstrate the issue even further, here's the result if module "./a" is not exist:
```
// index.js
Object(function webpackMissingModule() { var e = new Error("Cannot find module './a'"); e.code = 'MODULE_NOT_FOUND'; throw e; }());
console.log(Object(function webpackMissingModule() { var e = new Error("Cannot find module './a'"); e.code = 'MODULE_NOT_FOUND'; throw e; }());)
```

In order to recover from the missing dependency. './index.js' should be generated again. This is a breaking change introduced by https://github.com/web-infra-dev/rspack/pull/5124.

h-a-n-a:
I will approve this PR, but I would also like to see if there's any performance regression in this case change.
The current benchmark suites are not located in this repo, causing adding benchmark case for current PR is not as much easy as it used to be ;-(

h-a-n-a (approved):

jerrykingxyz:
!bench

PR #5310, opened 2024-01-11
feat(runtime): add rspack version

LingyuCoder:
## Summary

Add rspack version:
- Add `__rspack_version__` api to get current rspack version in runtime
- Manually specific version value by `experiments.rspackFuture.bundlerInfo.version`
- Use `experiments.rspackFuture.bundlerInfo.force=true|string[]` to inject runtime module even when user do not use it, this feature is useful for debugging

## Test Plan

- Add `packages/rspack/tests/configCases/version`

## Require Documentation?

- [ ] No


crates/rspack_core/src/runtime_globals.rs
```diff
@@ -245,6 +245,9 @@ bitflags! {
     const PRELOAD_CHUNK = 1 << 60;
 
     const PRELOAD_CHUNK_HANDLERS = 1 << 61;
+
+    // rspack only
+    const RSPACK_VERSION = 1 << 62;
   }
```
hardfist, line 251: will it conflict with future webpack new runtime_globals?

LingyuCoder: I think not, if webpack use `__webpack_require__.rv`, we should just change to another. It only affects the result of code generation, users will not feel the change.

hardfist:
!canary

KyrieLii:
LGTM 🫡

hardfist (approved):

PR #5319, opened 2024-01-14
fix(parser): reserved binding name in strict mode

### LINKED ISSUE #5120: Strict mode warning not working in entry module
### System Info

Not related

### Details

In webpack, entry module is being translated into an `EntryDependency` and it will be using strict parsing as it's categorized as `Esm`. Rspack does not respect to this option. 

### Reproduce link

_No response_

## Expected
1. Create a demo
2. Add `let let = let` to entry module
3. Run and there should be a parsing error.

## Actual
Rspack does not have a parsing error.

LingyuCoder:
## Summary

Fix #5120 

## Test Plan

- Added `packages/rspack/tests/cases/parsing/issue-5120-xxxx`

## Require Documentation?

- [x] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
@@ -6,8 +6,13 @@ use rspack_core::{
 };
 use rspack_core::{BoxDependency, ConstDependency, ContextMode, ContextNameSpaceObject};
 use rspack_core::{DependencyTemplate, SpanExt};
+use rspack_error::miette::{diagnostic, Diagnostic};
+use rspack_error::DiagnosticExt;
 use swc_core::common::{Spanned, SyntaxContext};
-use swc_core::ecma::ast::{BinExpr, BlockStmt, CallExpr, Callee, Expr, IfStmt, MemberExpr};
+use swc_core::ecma::ast::{
+  BinExpr, BlockStmt, CallExpr, Callee, Expr, Ident, IfStmt, MemberExpr, ObjectPatProp, Pat,
+  Program, Stmt, VarDecl, VarDeclKind,
+};
 use swc_core::ecma::ast::{Lit, TryStmt, UnaryExpr, UnaryOp};
 use swc_core::ecma::visit::{noop_visit_type, Visit, VisitWith};
 
@@ -15,7 +20,7 @@ use super::api_scanner::ApiParserPlugin;
 use super::context_helper::scanner_context_module;
 use super::expr_matcher::{is_module_require, is_require};
 use super::{
-  expr_matcher, extract_require_call_info, is_require_call_start,
+  expr_matcher, extract_require_call_info, is_require_call_start, is_reserved_word_in_strict,
   is_unresolved_member_object_ident, is_unresolved_require,
 };
 use crate::dependency::{
@@ -35,6 +40,7 @@ pub struct CommonJsImportDependencyScanner<'a> {
   pub(crate) is_strict: bool,
   pub(crate) plugin_drive: Rc<JavaScriptParserPluginDrive>,
   pub(crate) ignored: &'a mut Vec<DependencyLocation>,
+  pub(crate) errors: &'a mut Vec<Box<dyn Diagnostic + Send + Sync>>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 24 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
```diff
@@ -232,6 +239,62 @@ impl<'a> CommonJsImportDependencyScanner<'a> {
   pub fn walk_expression(&mut self, expr: &Expr) {
     expr.visit_children_with(self);
   }
+
+  pub fn check_ident(&mut self, ident: &Ident) {
+    if is_reserved_word_in_strict(ident.sym.as_str()) {
+      if self.is_strict {
+        self.errors.push(
+          diagnostic!(
```
h-a-n-a, line 247: Could you please change the diagnostic into a traceable error? This looks more beautiful to user end ;-). You can take this as a reference. https://github.com/web-infra-dev/rspack/blob/bfdc4279f1ae1f1f6332e74db963352d27c012a2/crates/rspack_plugin_javascript/src/utils/mod.rs#L63

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
```diff
@@ -232,6 +239,62 @@ impl<'a> CommonJsImportDependencyScanner<'a> {
...
+            "SyntaxError: The keyword '{}' is reserved in strict mode",
+            ident.sym
+          )
+          .boxed(),
+        );
+      } else {
+        self.errors.push(
+          diagnostic!(
```
h-a-n-a, line 255: This too ;-)

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
h-a-n-a: Also I'm planning to move other diagnostic to traceable errors as much as possible.

LingyuCoder: Resolved, but only these new errors were modified to traceable, I will modify all the left errors of scanners in another pull request.

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
```diff
@@ -232,6 +242,60 @@ impl<'a> CommonJsImportDependencyScanner<'a> {
   pub fn walk_expression(&mut self, expr: &Expr) {
     expr.visit_children_with(self);
   }
+
+  pub fn check_ident(&mut self, ident: &Ident) {
+    if is_reserved_word_in_strict(ident.sym.as_str()) {
+      if self.is_strict {
+        self.errors.push(Box::new(create_traceable_error(
+          "SyntaxError".into(),
```
h-a-n-a, line 250: I think you may use 'Javascript parsing error' here to align with other error messages.

h-a-n-a:
Looks almost good to me. But sorry to bother you again. Would you also add a diagnostic test here to avoid future error span regression since it's a graphical diagnostic? `packages/rspack/tests/diagnostics/module-parse-failed`. You may copy other test cases.

LingyuCoder:
> Looks almost good to me. But sorry to bother you again. Would you also add a diagnostic test here to avoid future error span regression since it's a graphical diagnostic? `packages/rspack/tests/diagnostics/module-parse-failed`. You may copy other test cases. Then I will say this PR is good to go :-)

Done

h-a-n-a (approved):

crates/rspack_plugin_javascript/src/visitors/dependency/common_js_import_dependency_scanner.rs
```diff
@@ -401,13 +465,34 @@ impl<'a> Visit for CommonJsImportDependencyScanner<'a> {
...
+      Program::Script(s) => is_strict(&s.body),
+    };
+    program.visit_children_with(self);
+  }
+
   fn visit_block_stmt(&mut self, n: &BlockStmt) {
     let old_in_strict = self.is_strict;
 
-    self.is_strict = is_strict(n);
+    self.is_strict = is_strict(&n.stmts);
```
bvanjoi, line 479: `blockStmts` should not affect the "use strict" within it, for example this case should work:

```rs
if (true) {
  "use strict"
  let = 1
}
```

PR #5370, opened 2024-01-18
feat: support compilation.hooks.runtimeModule

### LINKED ISSUE #5365: [Feature Request]: runtimeModule hook support
## Background
I want to build a chrome extension using rspack but encountered some problems:

1. In [content-scripts](https://developer.chrome.com/docs/extensions/develop/concepts/content-scripts), the style tags may exist under ShadowRoot, [`document.getElementsByTagName("link")`](https://github.com/web-infra-dev/rspack/blob/1fcafc3fa6499ead35577d9bd6f8bc77ad6254bc/crates/rspack_plugin_runtime/src/runtime_module/runtime/css_loading.js#L9) cannot find the style elements, leading to a non-working HMR.
2. A specified entry runs in the Service Worker environment ([background](https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers)), and it is unable to load JS code dynamically. It is hoped that the `webpack/runtime/load` runtime script code can be customized, so that I can perform a reload action when the entry receives a HMR message.
3. The public path of content-script files is runtime-evaluated.

## Webpack demo
It's a demo that shows how to customize the runtime public path:
```ts
export class ContentScriptPublicPathPlugin extends ContentScriptBasePlugin {
  apply(compiler: webpack.Compiler) {
    compiler.hooks.thisCompilation.tap('ContentScriptPublicPathPlugin', (compilation) => {
      compilation.hooks.runtimeModule.tap('ContentScriptPublicPathPlugin', (module, chunk) => {
        if (chunk.getEntryOptions()?.name !== 'content-script') return;
        if (module.name === 'publicPath') patchPublicPathRuntimeModule(module, compiler);
      });
    });
  }
}

function patchPublicPathRuntimeModule(
  module: webpack.RuntimeModule,
  { webpack, webpack: { RuntimeGlobals } }: webpack.Compiler
) {
  module.name = 'content-script publicPath';
  module.generate = function (this: webpack.RuntimeModule) {
    return webpack.Template.asString([`${RuntimeGlobals.publicPath} = chrome.runtime.getURL('/');`]);
  };
}
```

_Originally posted by @tmkx in https://github.com/web-infra-dev/rspack/discussions/5269_

LingyuCoder:
## Summary

Resolve #5365. Modify sources of runtime modules by `compilation.hooks.runtimeModule`.

```js
// webpack
compilation.hooks.runtimeModule.tap('xxx', (module, chunk) => {
	if (chunk.name === 'xxx' && (module.name === 'xxx' || module.constructor.name === 'xxx')) {
		const originSource = module.getGeneratedCode();
		const appendSource = "blabla";
		module.getGeneratedCode = () => `${originSource}\n${appendSource}`;
	}
});

// rspack
compilation.hooks.runtimeModule.tap("TestFakePlugin", (module, chunk) => {
	if (chunk.name === 'xxx' && (module.name === 'xxx' || module.constructorName === 'xxx')) {
		const originSource = module.source.source.toString('utf-8');
		const appendSource = "blabla";
		module.source.source = Buffer.from(
			`${originSource}\n${appendSource}`,
			"utf-8"
		);
	}
});
```

## Test Plan

- Add `packages/rspack/tests/configCases/hooks/runtime-module`

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/node_binding/src/plugins/mod.rs
+++ b/crates/node_binding/src/plugins/mod.rs
@@ -67,6 +67,7 @@ pub struct JsHooksAdapter {
   pub succeed_module_tsfn: ThreadsafeFunction<JsModule, ()>,
   pub still_valid_module_tsfn: ThreadsafeFunction<JsModule, ()>,
   pub execute_module_tsfn: ThreadsafeFunction<JsExecuteModuleArg, Option<String>>,
+  pub runtime_module_tsfn: ThreadsafeFunction<JsModule, ()>,
 }
 
 impl Debug for JsHooksAdapter {
@@ -850,6 +851,22 @@ impl rspack_core::Plugin for JsHooksAdapter {
       .blocking_recv()
       .unwrap_or_else(|recv_err| panic!("{}", recv_err.to_string()))
   }
+
+  async fn runtime_module(&self, module: &mut dyn rspack_core::Module) -> rspack_error::Result<()> {
+    if self.is_hook_disabled(&Hook::RuntimeModule) {
+      return Ok(());
+    }
+
+    self
+      .runtime_module_tsfn
+      .call(
+        module.to_js_module().expect("Convert to js_module failed."),
+        ThreadsafeFunctionCallMode::NonBlocking,
+      )
+      .into_rspack_result()?
+      .await
+      .unwrap_or_else(|err| panic!("Failed to call runtime module: {err}"))
+  }
 }
 
 impl JsHooksAdapter {
@@ -897,6 +914,7 @@ impl JsHooksAdapter {
       succeed_module,
       still_valid_module,
       execute_module,
+      runtime_module,
     } = js_hooks;
 
     let process_assets_stage_additional_tsfn: ThreadsafeFunction<(), ()> =
@@ -982,6 +1000,8 @@ impl JsHooksAdapter {
       js_fn_into_threadsafe_fn!(still_valid_module, env);
     let execute_module_tsfn: ThreadsafeFunction<JsExecuteModuleArg, Option<String>> =
       js_fn_into_threadsafe_fn!(execute_module, env);
+    let runtime_module_tsfn: ThreadsafeFunction<JsModule, ()> =
+      js_fn_into_threadsafe_fn!(runtime_module, env);
 
     Ok(JsHooksAdapter {
       disabled_hooks,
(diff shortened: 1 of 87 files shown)


tmkx:
awesome!

hardfist:
!bench

crates/node_binding/src/plugins/mod.rs
```diff
@@ -850,6 +856,50 @@ impl rspack_core::Plugin for JsHooksAdapter {
...
+        },
+        ThreadsafeFunctionCallMode::NonBlocking,
+      )
+      .into_rspack_result()?
+      .await
+      .unwrap_or_else(|err| panic!("Failed to call runtime module hook: {err}"))
+      .map(|r| {
+        r.and_then(|s| s.source).map(|s| {
+          std::str::from_utf8(&s.source)
+            .expect("Invalid utf-8 sequence")
```
h-a-n-a, line 898: Should report this error (user's return value)

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1970,10 +1969,15 @@ impl Compilation {
...
 
-    self
+    if let Some(new_source) = self
       .plugin_driver
-      .runtime_module(module.as_mut())
-      .expect("Run runtime_module hook failed");
+      .clone()
+      .runtime_module(module.as_mut(), chunk, self)
+      .await
+      .expect("Run runtime_module hook failed")
```
h-a-n-a, line 1977: should also propagate error from the hook

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1970,10 +1969,15 @@ impl Compilation {
       .chunk_graph
       .connect_chunk_and_runtime_module(*chunk_ukey, runtime_module_identifier);
 
-    self
+    if let Some(new_source) = self
       .plugin_driver
-      .runtime_module(module.as_mut())
-      .expect("Run runtime_module hook failed");
+      .clone()
```
h-a-n-a, line 1974: Can we avoid cloning here? not passing `self` down to the hook

LingyuCoder: Fixed cloning. Currently runtime modules need the full compilation to generate their codes. So `self` here need to be passed.

crates/node_binding/src/plugins/mod.rs
LingyuCoder: Fixed by report the error with panic. But I think user's return value is a buffer and it is meanless to output it.

h-a-n-a:
Looks good. I think the difference with webpack should be well documented

h-a-n-a (approved):

PR #5381, opened 2024-01-22
chore: enable newTreeshaking on arco-pro diff case

LingyuCoder:
## Summary

Modify configuration of diff repo:
1. Enable newTreeshaking
2. Bump webpack to 5.90.0 to support namedExports of experiments.css
3. Add generator and parser config to config file to prevent difference between less files
4. Align runtime module code with webpack 5.90.0 (fetchPritority)

## Require Documentation?

- [x] No

### DIFF
--- a/diffcases/arco-pro/rspack.config.js
+++ b/diffcases/arco-pro/rspack.config.js
@@ -87,7 +87,9 @@ const config = {
 	output: {
 		publicPath: "/",
 		filename: "[name].js",
-		chunkFilename: "[name].js"
+		chunkFilename: "[name].js",
+		cssChunkFilename: "[name].css",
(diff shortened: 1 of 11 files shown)


diffcases/arco-pro/rspack.config.js
```diff
@@ -92,7 +92,10 @@ const config = {
 	optimization: {
 		minimize: false, // Disabling minification because it takes too long on CI
 		realContentHash: true,
-		usedExports: false,
+		providedExports: true,
```
LingyuCoder, line 95: !diff

LingyuCoder: !diff

PR #5390, opened 2024-01-22
perf: cleanup `ApiScanner`

bvanjoi:
Visiting each identifier incurs a significant amount of time. Therefore, optimizing the `﻿ApiScanner` can aid in improving performance.

before 3.3s:

img_v3_027b_d0f441f4-9595-4c5f-a925-0b4d629f790g

after this PR, 2.7s:

img_v3_027b_dd4534ad-56d3-4b7f-b6ff-2605d46a968g

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
@@ -350,6 +350,7 @@ impl<'parser> JavascriptParser<'parser> {
   }
 
   fn walk_variable_declaration(&mut self, decl: &VarDecl) {
+    self.enter_assign = true;
     for declarator in &decl.decls {
       // if let Some(renamed_identifier) = declarator
       //   .init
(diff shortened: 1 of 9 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
```diff
@@ -350,6 +350,7 @@ impl<'parser> JavascriptParser<'parser> {
   }
 
   fn walk_variable_declaration(&mut self, decl: &VarDecl) {
+    self.enter_assign = true;
```
ahabhgk, line 353: Should this in the for loop? since `= false` is in the for loop

jerrykingxyz:
!bench

ahabhgk (approved):

crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
bvanjoi: The logic within `﻿ApiScanner` was simply replicated, and additionally, the ﻿`walk_variable_declaration` will not influence the for-loop body.

PR #5418, opened 2024-01-23
chore: bump ustr to v1.0.0

Boshen:


Boshen:
hmm ... there's a change to ahash https://github.com/anderslanglands/ustr/commit/11c81783a1dd15559f76a0d90190c22affd30ab0#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R22

Boshen:
ustr bumps ahash, from the doc https://github.com/tkaitchuck/aHash

ecause it does not have a fixed standard, different computers or computers on different versions of the code will observe different hash values.

hash values are not consistent :-)

PR #5429, opened 2024-01-23
chore: use `cargo machete` for checking unused dependencies

Boshen:
## Summary

https://github.com/est31/cargo-udeps is outdated and broken.

https://github.com/bnjbvr/cargo-machete makes everything easy by using greping for the dependency, which is correct 99% of the time for a Rust project :-)


Boshen:
@SyMind You mentioned cargo-udeps in another place, this is a better one.

jerrykingxyz (approved):

PR #5431, opened 2024-01-23
fix: evaluate if stmt should skip import specifier in dead branch

### LINKED ISSUE #5430: Invalid syntax after evaluate if_stmt
### System Info

no need

### Details

giving:
```js
import { foo } from './c'

if (typeof require !== "undefined") {
  console.log('hello')
} else {
  console.log(foo)
}
```
result:
```js
if (true) console.log('hello');
else {}_c__WEBPACK_IMPORTED_MODULE_0__.foo
}),
```
The cause I believe is ReplaceSource, replace both the branch and imported identifier, causing one replacement appends at the end of first replacement

### Reproduce link

_No response_

### Reproduce Steps

```js
import { foo } from './c'

if (typeof require !== "undefined") {
  console.log('hello')
} else {
  console.log(foo)
}
```

JSerFeng:
## Summary

fix #5430 

before:
```js
it('should have correct syntax', () => {
        if (true) console.log('hello');
        else _foo__WEBPACK_IMPORTED_MODULE_0__.foo{ }
      });
```
after:
```js
it('should have correct syntax', () => {
        if (true) console.log('hello');
        else { }
      });
```

## Require Documentation?

- [ ] No

### DIFF
--- a/packages/rspack/tests/cases/parsing/evaluate-if/index.js
+++ b/packages/rspack/tests/cases/parsing/evaluate-if/index.js
@@ -1,0 +1,8 @@
+import { foo } from "./foo";
+
+// see issue https://github.com/web-infra-dev/rspack/issues/5430
+it("should compile", () => {
+	if (typeof require === "undefined") {
+		foo;
+	}
+});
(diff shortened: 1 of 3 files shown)


packages/rspack/tests/cases/parsing/evaluate-if/index.js
```diff
@@ -0,0 +1,10 @@
+import { foo } from "./foo";
+
+// see issue https://github.com/web-infra-dev/rspack/issues/5430
+it("should compile", () => {
+	if (typeof require !== "undefined") {
+		console.log("hello");
```
h-a-n-a, line 6: Would you please avoid using `console.log` here since it flushes terminal alot?

bvanjoi (approved):
I think it's acceptable to land this PR, albeit it may be removed in the near future. This test case could provide useful insights.

PR #5435, opened 2024-01-23
refactor: reimplement queue handler

JSerFeng:
## Require Documentation?

- [ ] No


crates/rspack_core/src/utils/queue.rs
```diff
@@ -1,14 +1,34 @@
...
-#[derive(Default)]
-pub struct WorkerQueue<T> {
+use rustc_hash::FxHashMap;
+use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
+
+use crate::{Compilation, ModuleIdentifier};
+
+pub struct WorkerQueue<T, K> {
   inner: VecDeque<T>,
+  queue_tx: QueueHandler<T, K>,
```
h-a-n-a, line 10: Is it possible for us to remove old `VecDeque` and just use `mpsc` here?

crates/rspack_core/src/loader/loader_runner.rs
```diff
@@ -17,7 +18,11 @@ pub struct CompilerContext {
   pub module_context: Option<Box<Context>>, // current module context
   pub module_source_map_kind: SourceMapKind,
 
-  pub queue_handler: Option<QueueHandler>,
+  pub factorize_queue: Option<FactorizeQueueHandler>,
```
h-a-n-a, line 21: Maybe init the `XXXQueue` here instead of the handler  when a new compilation is created? This could help us remove `Option`.

crates/rspack_core/src/utils/queue.rs
JSerFeng: If we use mpsc exclusively, I'm afraid the performance will decrease.

h-a-n-a: I have the same concern here. So I'm just wondering is it possible to only support adding handler for a certain queue or we have to implement them all.

JSerFeng:
!bench

crates/rspack_core/src/utils/queue.rs
JSerFeng: For now, only used 3 of 5, factorize_queue, process_dependencies_queue and build_time_execution_queue.
The reason I use separate queue_handler before, is that we can check only one channel at every iteration, instead of checking 5 channels in this PR, let me see benchmark result and then try optimizing a little bit

crates/rspack_core/src/loader/loader_runner.rs
JSerFeng: Because handler is `tx`, if we init them during compilation creation, we have to put `rx` into compilation as well, and some API like `handle_module_creation` needs &Compilation and &mut FactorizeQueue, If I change the API from taking &Compilation and &mut FactorizeQueue to only taking &mut Compilation, some other upper stream API needs to change as well because them don't have &mut Compilation :(.
I encountered this a lot before, so I choose the easier way here

h-a-n-a (approved):

crates/rspack_core/src/utils/queue.rs
h-a-n-a: The other concern that I have other than performance is that exposing every queue to the outside may break the contract which is provided by `handle_module_creation`. As this method is the entry point of building a module. We'd better use a semantic function instead of modifying the queue directly (If you can abstract the behavior as much as possible).

JSerFeng: Understand, I think it's low level part of Webpack, users should be aware of what they are doing if they want to use this

h-a-n-a (approved):

PR #5468, opened 2024-01-25
chore: bump TypeScript to 5.3.3

Boshen:


tsconfig.base.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "module": "CommonJS",
+    "module": "NodeNext",
```
Boshen, line 3: :wave:

Boshen: error TS5110: Option 'module' must be set to 'NodeNext' when option 'moduleResolution' is set to 'NodeNext'.

h-a-n-a (approved):

Issue #4450, opened 2023-10-26
support __resourceQuery for context module


ahabhgk:
```js
let a = "a"
import(`./${a}?queryString`)
```

- webpack-test/cases/parsing/issue-7778

LingyuCoder:
Fixed by #6018

PR #5476, opened 2024-01-25
feat: support resource query in context module

faga295:
#4450

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/import_scanner.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/import_scanner.rs
@@ -1,0 +1,331 @@
+use std::sync::Arc;
+
+use rspack_core::{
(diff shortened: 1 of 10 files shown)


LingyuCoder (approved):

LingyuCoder:
Thank you. It seems that the use case of issue-7778 needs to wait until oxc_resolver is upgraded to 1.5.0 (#5644). This PR can be merged first.

I will follow up on this use case in the future. In addition, it seems that the generated artifacts of rspack are different from those of webpack, and I will try to align them.

JSerFeng:
The import_scanner.rs seems not working, I don't find any place use this module

crates/rspack_plugin_javascript/src/visitors/dependency/import_scanner.rs
JSerFeng, line 1: I mean this module, any where include this module ?

PR #5499, opened 2024-01-28
fix: test jobs should not run in forked repositories

SyMind:
## Summary

Whenever I do `sync fork`, the workflow will be error as shown below.

image

To address this, we need to prevent the execution of `Run benchmark` and `Test xxx` jobs in forked repositories.

image

## Require Documentation?

- [x] No


SyMind:
!bench

jerrykingxyz:
maybe we should skip the upload which caused the test failed instead of not run test.

SyMind:
@jerrykingxyz I changed, please review again.

jerrykingxyz (approved):

Issue #5481, opened 2024-01-26
rspack_core: ustr may not be the ideal choice for module identifier


Boshen:
### Details

While upgrading to ustr 1.0 https://docs.rs/ustr/latest/ustr/, ahash upgrade yielded inconsistent hash results. 

From ahash's documentation https://github.com/tkaitchuck/ahash:

> AHash does not have a fixed standard for its output. This allows it to improve over time. 

> Because it does not have a fixed standard, different computers or computers on different versions of the code will observe different hash values. As such, aHash is not recommended for use other than in-memory maps.

Unfortunately we are currently using it as part of our hash:

https://github.com/web-infra-dev/rspack/blob/88e5557dee8cc2acce3be53cbf054065ab81610d/crates/rspack_core/src/compiler/compilation.rs#L1996

IWANABETHATGUY:
Using the `str`  may fix your issue, `module.identifier().as_str().hash(xxx)`

Boshen:
> Using the internal `str` may fix your issue, `module.identifier().as_str().hash(xxx)`

This is a footgun, how would anyone remember to use this form?

h-a-n-a:
This, for sure, is not the ultimate solution for us if portable caching is the ultimate goal of us. But it should be ok for now as we haven't got the plan in the near future to support this. Nevertheless, it still cause unstable hash in snapshots we might have to fix this.

> Using the internal `str` may fix your issue, `module.identifier().as_str().hash(xxx)`

This might be a temporary solution as it's not performant ;-( (precomputed-hash is gone)

Boshen:
@IWANABETHATGUY @h-a-n-a How do you feel about https://github.com/web-infra-dev/rspack/pull/5504, it's only a 2 line change in the ustr crate https://github.com/anderslanglands/ustr/compare/master...Boshen:ustr-fxhash:master

IWANABETHATGUY:
> @IWANABETHATGUY @h-a-n-a How do you feel about #5504, it's only a 2 line change in the ustr crate [anderslanglands/ustr@master...Boshen:ustr-fxhash:master](https://github.com/anderslanglands/ustr/compare/master...Boshen:ustr-fxhash:master)

the consistent hash function is the best way to solve this issue.

IWANABETHATGUY:
> > consistent hash function
> 
> Can you elaborate? Is my change acceptable or not?

[...]

PR #5504, opened 2024-01-29
feat(rspack_core): use consistent fxhash for ustr instead of inconsistent ahash

Boshen:
Forking ustr and let's see how it goes ...

relates #5481

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -370,9 +370,9 @@ dependencies = [
 
 [[package]]
 name = "byteorder"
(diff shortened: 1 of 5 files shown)


Boshen:
!bench

Boshen:
!eco-ci

hardfist:
I remember github url will break rust cache？

Boshen:
!bench

h-a-n-a (approved):

PR #5507, opened 2024-01-29
perf: enable "fat" LTO for release

Boshen:


Boshen:
!bench

JSerFeng:
How about enabling this only for publish, daily development needs release build sometimes

Boshen:
This took 7mins instead of 3m47s (thin lto), closing because the slowness is unbearable.

PR #5598, opened 2024-02-04
chore: 🤖 add comments when the export name is unused

IWANABETHATGUY:
## Require Documentation?

- [ ] No


crates/rspack_plugin_javascript/src/dependency/esm/harmony_import_specifier_dependency.rs
```diff
@@ -174,9 +174,6 @@ impl DependencyTemplate for HarmonyImportSpecifierDependency {
     }
 
     let ids = self.get_ids(&compilation.module_graph);
-    compilation
```
IWANABETHATGUY, line 177: this stmt doesn't generate any effects.

JSerFeng (approved):

PR #5663, opened 2024-02-18
refactor: replace asyncDependenciesBlockId with Identifier

JSerFeng:
## Summary

Replace asyncDependenciesBlockId with Identifier, this can be good for incremental build.

When Modifying a module we will check if its dependencies change, this include static imports(module identifiers) and dyn imports(async dependencies Id), using identifier for asyncDependenciesId can make sure it's the same for every build if not changed.

## Require Documentation?

- No


JSerFeng:
!canary

ahabhgk (approved):

JSerFeng:
!canary

Issue #5627, opened 2024-02-06
[Feature]: improve diagnostic when resolve failed with missing `resolve.extensions`


h-a-n-a:
### What problem does this feature solve?

For example, if a module resolution is failed as I do `import("./foo")`. 
Then we will check the files in the **current directory** with name "foo" but with different extensions or the directory named "foo" with its "index" file(according to `mainFiles`) with different extension than preconfigured resolve.extensions. This could avoid us from setting arbitrary extensions and keep rspack as more Web-spec as possible. 
Then rspack can list these suggestions in the help section of each diagnostic.

### What does the proposed API of configuration look like?

NO API change.

h-a-n-a:
To add this support, you may refer to this for details:
1. Add a new error hint helper in this file https://github.com/web-infra-dev/rspack/blob/a4b8196a493076b790313ad1deec227d3c04813c/crates/rspack_core/src/resolver/mod.rs#L70.
2. Add a test in `packages/rspack/test/diagnostics` to test out the diagnostic

luhc228:
I want to have a try.

hardfist:
@luhc228  thanks in advance

PR #5680, opened 2024-02-19
feat: add hint for missing resolve.extensions when resolving module

luhc228:
## Summary

ref: https://github.com/web-infra-dev/rspack/issues/5627

I want to handle the description file(for example <PROJECT_ROOT>/package.json) hint better, but I could not get more info from the oxc_resolver.

## Require Documentation?

- [x] No


h-a-n-a:
Hi. Thank you for the contribution! However, there might be some misunderstandings in my wordings, so I would like to make some clarifications on the original issue. 

Users not a pro in bundler might be guessing what's happening under the hood. This leads us to improve the overall diagnostic message instead of porting error messages provided by webpack as they are not as easy as possible to both READ or APPLY. This should be done brick by brick. 

When it comes to this issue, we would like to provide better diagnostic for whom is using `*.ts`,`*.tsx`, etc. Given the example in the issue, if `import("./foo")` is failed, we should see if there's any potential `"foo"` files for which extension might be missing from `resolve.extensions` and in this example, if "./foo.ts" is existing, then the bundler will give user a hint roughly like "'./foo.ts' is existing, but it's not in the `resolve.extensions` list, either replacing './foo' to './foo.ts' or adding ".ts" to `resolve.extensions`".

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,42 @@ which tries to resolve these kind of requests in the current directory too.",
...
+      .missing_dependencies
+      .iter()
+      .filter(|p| !p.ends_with("package.json"))
+      .map(|p| {
+        let path = p.to_string_lossy().to_string();
+        format!("'{}' doesn't exist", path)
+      })
+      .collect::<Vec<_>>()
+      .join("\n")
+      .red();
```
h-a-n-a, line 189: Let's use plain color here for now.

packages/rspack/tests/Compiler.test.ts
```diff
@@ -335,6 +335,11 @@ describe("Compiler", () => {
 		} catch (err) {
 			expect(err.toString()).toMatchInlineSnapshot(`
 			"Error:   × Resolve error: Can't resolve './missing-file' in '<PROJECT_ROOT>/tests'
+			  help: try to resolve './missing-file' in '<PROJECT_ROOT>/tests'
```
h-a-n-a, line 338: This is too verbose, the default error message should be easy to read and the fix should be easy to apply.

luhc228:
I will fix it later when I have time. Thanks  @h-a-n-a

h-a-n-a:
> I will fix it later when I have time. Thanks @h-a-n-a

@luhc228 At your own convenience ;-).

luhc228:
> When it comes to this issue, we would like to provide better diagnostic for whom is using `*.ts`,`*.tsx`, etc. Given the example in the issue, if `import("./foo")` is failed, we should see if there's any potential `"foo"` files for which extension might be missing from `resolve.extensions` and in this example, if "./foo.ts" is existing, then the bundler will give user a hint roughly like "'./foo.ts' exists, but it's not in the `resolve.extensions` list, either replacing './foo' with './foo.ts' or adding ".ts" to `resolve.extensions`".
> 

I am a little confused. I want to confirm shall I use the [glob](https://crates.io/crates/glob) crate to find all the the `foo` files? For example:
```md
src
   ├── foo.ts
   └── index.js   # this is the entry file. in this file import 'foo'; 
```

When I use glob to search `foo`, I will get the result: `foo.txt`. Then show the hint: `either replacing './foo' with './foo.ts' or adding ".ts" to resolve.extensions".`

packages/rspack/tests/Compiler.test.ts
luhc228: I have gave the useful messages. But I am not sure whether it is good enough. Help me to review. Thanks.

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,104 @@ which tries to resolve these kind of requests in the current directory too.",
...
+            .unwrap_or_default();
+          let suggestion_path = suggestion.to_string_lossy();
+          let specifier = args.specifier;
+
+          hint.push(format!(
+          "Found the module '{suggestion_path}' exists, but it was not resolved because its extension doesn't in the `resolve.extensions` list. Here are some possible solutions:
+
+1. add the extension '.{suggestion_ext}' to `resolve.extensions` in your rspack configuration
+2. use '{suggestion_path}' instead of '{specifier}'
+3. add the file path '{suggestion_path}' to the `entry` in your rspack configuration
```
h-a-n-a, line 261: This seems not applicable to most cases, 1,2 looks good enough.

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,104 @@ which tries to resolve these kind of requests in the current directory too.",
...
+        for suggestion in suggestions {
+          let suggestion_ext = suggestion
+            .extension()
+            .map(|e| e.to_string_lossy())
+            .unwrap_or_default();
+          let suggestion_path = suggestion.to_string_lossy();
+          let specifier = args.specifier;
+
+          hint.push(format!(
+          "Found the module '{suggestion_path}' exists, but it was not resolved because its extension doesn't in the `resolve.extensions` list. Here are some possible solutions:
```
h-a-n-a, line 257:
```suggestion
          "Found the module '{suggestion_path}', but its extension is not listed in the `resolve.extensions`. Here are some possible solutions:
```

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,104 @@ which tries to resolve these kind of requests in the current directory too.",
...
+
+1. add the extension '.{suggestion_ext}' to `resolve.extensions` in your rspack configuration
```
h-a-n-a, line 259:
```suggestion
1. add the extension `".{suggestion_ext}"` to `resolve.extensions` in your rspack configuration
```

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,104 @@ which tries to resolve these kind of requests in the current directory too.",
...
+        }
+      }
+      Err(_) => normalized_path.parent(),
+    };
+
+    if parent_path.is_none() || file_name.is_none() {
+      return None;
+    }
+
+    let file_name = file_name.unwrap();
```
h-a-n-a, line 199: Use `expect` instead of `unwrap`

crates/rspack_core/src/resolver/mod.rs
```diff
@@ -169,6 +170,104 @@ which tries to resolve these kind of requests in the current directory too.",
...
+        } else {
+          is_resolving_dir = true;
+          Some(normalized_path.borrow())
+        }
+      }
+      Err(_) => normalized_path.parent(),
+    };
+
+    if parent_path.is_none() || file_name.is_none() {
+      return None;
```
h-a-n-a, line 196: Only return if there's a hint.

crates/rspack_core/src/resolver/mod.rs
luhc228: done

crates/rspack_core/src/resolver/mod.rs
luhc228: fixed

crates/rspack_core/src/resolver/mod.rs
luhc228: ok I removed it.

h-a-n-a (approved):
Thank you!

PR #5694, opened 2024-02-19
fix: hit cache when imports connection not changed

JSerFeng:
## Current issue

If module's all dependencies and blocks stay the same, Rspack skips building chunk graph. However dependencies can change easily, for example:

```js
import {foo} from 'bar';
```

there is 1 dependency which is `ImportSideEffectDependency`. However if I add one line like:

```js
import {foo} from 'bar';
console.log(foo);
```

We don't need to rebuild chunk graph, as there are no connections change. But we do have dependencies change, we have 
`ImportSideEffectDependency` and `ImportSpecifierDependency` now, so cache miss.

## Solution
I choose to sort all outgoings by there span location first, this step ensures dependency orders change can make cache invalidate.
Then I dedupe them by there target module, so I add extra import to the same module or I make access to import specifier should hit cache as well

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -504,7 +504,7 @@ impl Compilation {
       let (deps, blocks) = compalition
         .module_graph
         .get_module_dependencies_modules_and_blocks(module_identifier);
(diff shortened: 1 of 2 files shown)


LingyuCoder (approved):

LingyuCoder:
!bench

PR #5713, opened 2024-02-20
fix: hmr update modules should consider runtime

JSerFeng:
## Summary

HMR updated modules should consider every runtime

## Require Documentation?

- [ ] No


JSerFeng:
!bench

PR #5808, opened 2024-02-28
fix(rspack_loader_swc): enable SWC plugins

### LINKED ISSUE #5763: [Feature]: SWC plugins support

Additional abilities

In my case, to migrate to rspack/rsbuild I need the plugin for SWC - https://github.com/intpp/swc-plugin-jsx-control-statements

https://github.com/web-infra-dev/rspack/blob/30570575c2eb1af99144fb3a83fa6f30469bb954/crates/rspack_loader_swc/src/lib.rs#L74

It could be used as is and set through Plugin API `modifyRspackConfig`.

dm33tri:
## Summary

Fixes  #5763 

I've locked `rkyv` to version which is used in swc and wrapped tuple in a struct as it wouldn't deserialize with an older `rkyv` version

## Require Documentation?

- [X] No

### DIFF
--- a/packages/rspack/tests/ConfigTestCases.template.js
+++ b/packages/rspack/tests/ConfigTestCases.template.js
@@ -282,7 +282,7 @@ const describeCases = config => {
 												? children.reduce(
 														(all, { modules }) => all.concat(modules),
 														modules || []
-												  )
+													)
 												: modules;
 											if (
 												allModules.some(
@@ -582,7 +582,7 @@ const describeCases = config => {
 																				referencingModule.identifier
 																					? referencingModule.identifier.slice(
 																							esmIdentifier.length + 1
-																					  )
+																						)
 																					: fileURLToPath(referencingModule.url)
 																			),
 																			options,
@@ -668,9 +668,11 @@ const describeCases = config => {
 												) {
 													return testConfig.modules[module];
 												} else {
-													return require(module.startsWith("node:")
-														? module.slice(5)
-														: module);
+													return require(
+														module.startsWith("node:")
+															? module.slice(5)
+															: module
+													);
 												}
 											};
 
--- a/.gitignore
+++ b/.gitignore
@@ -226,3 +226,5 @@ justfile
 smoke-example
 diff_output
 
+# .swc plugins cache
+.swc/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,7 +18,7 @@ version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
 dependencies = [
- "gimli",
+ "gimli 0.27.3",
 ]
 
 [[package]]
@@ -95,6 +95,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
 
 [[package]]
+name = "any_ascii"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "70033777eb8b5124a81a1889416543dddef2de240019b674c81285a2635a7e1e"
+
+[[package]]
 name = "anyhow"
 version = "1.0.80"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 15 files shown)


dm33tri:
You may use this as a base or maybe I will add some test cases later

Also I had some runtime problems running pre-commit linters and didn't bother but will look into it

h-a-n-a:
Would you please add a case? ;-)

dm33tri:
> Would you please add a case? ;-)

I have added a test case

dm33tri:
I didn't notice stats snapshot failing locally because my $HOME has `.` in it and it was replacing it with `X` and all the tests failed. I've fixed this regexp:

```
// CHANGE: The time unit display in Rspack is second
.replace(/[.0-9]+(\s?s)/g, "X$1");
```

Removed old test, and updated snapshots

On my machine `Diagnostics` test set is still failing because it shows `.js` in stack traces but mine shows `.ts`

As for timeout of my test I don't know, will rerun fix it?

h-a-n-a:
> I guess I'll have to spin up a Linux VM or something, because on M1 it passes without any timeouts 😅 To check that it doesn't actually hang
> 
> And for CSS - is it just flaky, or did I actually break something?

Yes, css test seems flaky and not being fixed. No worries. We will help you out!

Would you please take a look at your earliest convenience?  @JSerFeng

dm33tri:
Issue with the test looks similar to this https://github.com/swc-project/plugins/issues/32#issuecomment-1121491533

JSerFeng:
the error ` _require(vm) called after all tests from css urls have completed`

I guess the fix could be changing the `tests/configCases/css/urls/index.js`
to this:
```
const fs = require("fs");
const path = require("path");

it("css urls should works", async () => {
	await import("./urls.css");
	const css = await fs.promises.readFile(
		path.resolve(__dirname, "bundle.css"),
		"utf-8"
	);
	expect(css).toMatchSnapshot();
});

```

JSerFeng:
Also curious how large the binary size increases to

dm33tri:
> Also curious how large the @rspack/core package size increases to

`rspack.darwin-arm64.node` in release mode is up to 45mb from 33mb

> I guess the fix could be changing the tests/configCases/css/urls/index.js

I'm able to reproduce this error on different CSS test cases (with `require(..)` or `import(..)`) after cleaning up everything (so no caches). Will try few more times to see if it's my changes

JSerFeng:
> I'm able to reproduce this error on different CSS test cases (with require(..) or import(..)) after cleaning up everything (so no caches). Will try few more times to see if it's my changes or not

If you have any trouble fixing that, feel free to tell us

dm33tri:
I see that `_require(vm) called after all tests` is not from my code and it can flak in any test.

But I can't reproduce build hanging on mac/linux. Can I try to increase timeout for `ConfigTestCases.template.js` to 30s? Perhaps wasm runtime is just too slow in action runnner

```
● ConfigTestCases › builtin-swc-loader › swc-plugin › swc-plugin should compile
  thrown: "Exceeded timeout of 10000 ms for a test while waiting for `done()` to be called.
```

dm33tri:
Seems like it helped 

Would you approve please?

xc2:
!canary

packages/rspack/tests/ConfigTestCases.template.js
```diff
@@ -49,7 +49,7 @@ const createLogger = appendTarget => {
 	};
 };
 
-const TIMEOUT = 10_000;
+const TIMEOUT = 30_000;
```
h-a-n-a, line 52: This file is not supposed to be touched. Would you please tell me more about this one?

dm33tri: For some reason builds with wasm plugins are taking too long in runners. Previously with 10s it failed both times, but with 30s it seems to pass. Maybe there is another workaround to make build pass in time?

dm33tri: This is the error:
```
● ConfigTestCases › builtin-swc-loader › swc-plugin › swc-plugin should compile
  thrown: "Exceeded timeout of 10000 ms for a test while waiting for `done()` to be called.
```

dm33tri: I'll also try to find a plugin with target `wasm32-unknown-unknown` and test if it loads faster

dm33tri: I've replaced `@swc/plugin-transform-imports` with the simple `@swc/plugin-remove-console` and undid the timeout change. Please rerun the tests to see if it will work

h-a-n-a: Looks good, let's wait for the CI result

dm33tri: Should I also repush to undo linter changes on? 😅 They were added automatically with commit hooks

h-a-n-a (approved):
Thank you!

PR #5840, opened 2024-03-04
fix: should not generate minify plugin based on devtool

JSerFeng:
## Require Documentation?

- [ ] No


packages/rspack/tests/configCases/plugins/minify-with-devtool-false/index.js
```diff
@@ -0,0 +1,10 @@
+const lib = require("./lib");
+const fs = require('fs')
+const path = require('path')
+
+it("minify-plugin", () => {
+	// console.log("lib:", lib);
```
SyMind, line 6: This comment appears to be unused and could potentially be removed.

SyMind (approved):

Issue #5867, opened 2024-03-06
[Bug]: the type of webpack-sources is any


Amour1688:
### System Info

Every System

### Details

![image](https://github.com/web-infra-dev/rspack/assets/31695475/a0a8716e-bf18-4497-b2de-e062c4bbef3d)

![image](https://github.com/web-infra-dev/rspack/assets/31695475/1fdc0058-eb3f-4bc0-93d3-26bcf2f4005d)

### Reproduce Steps

Every repo

SyMind:
First, we should fix the `@types/webpack-sources`. I have already addressed the PR here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/69710.

PR #5868, opened 2024-03-06
fix: move `@types/webpack-sources` to dependencies (close: #5867)

Amour1688:
## Require Documentation?

- [ ] No


packages/rspack/src/exports.ts
```diff
@@ -46,8 +46,7 @@ export const WebpackError = Error;
 
 export type { Watching } from "./Watching";
 
-const sources = require("webpack-sources"); // use require to avoid wrong types, @types/webpack-sources is outdate
-export { sources };
+export * as sources from "webpack-sources";
```
ahabhgk, line 49: @types/webpack-sources seems outdated

alfaproject:
Why not fix/contribute the types in https://github.com/DefinitelyTyped/DefinitelyTyped

chenjiahan:
Agreed, we should contribute to `@types/webpack-sources`.

PR #5896, opened 2024-03-07
fix: css loading runtime order

ahabhgk:
## Summary

the order of `loadCssChunkCallback` and `webpackJsonpCallback` matters in rspack, cause rspack experiments css use `loadCssChunkCallback` to load css chunk instead of `loadCssChunkData + css head{meta data}` like webpack, if `loadCssChunkCallback` is before `webpackJsonpCallback`, umd library will have the correct exports since the exports is returned in `webpackJsonpCallback`, but in `loadCssChunkCallback` it will return `undefined`, so if `loadCssChunkCallback` is after `webpackJsonpCallback`, the umd library exports will be `undefined`

## Require Documentation?

- [x] No


ahabhgk:
BTW, `pnpm test:ng` failed, caused by `globalContext` is reuse between `runner.run()`, in webpack `globalContext` and `moduleScope` should be one instance per `runner.run()` instead of per `runner`. cc @LingyuCoder

LingyuCoder (approved):

ahabhgk:
!canary

PR #5904, opened 2024-03-08
fix: avoid invoking `toJSON` method of `additionalData`

### LINKED ISSUE #5600: [Bug]: TypeError: Cannot read properties of undefined (reading 'input')
### System Info

@rspack/core: 0.5.3

### Details

```json
// package.json
{
    "dependencies": {
        "postcss-loader": "8.1.0",
        "tailwindcss": "3.4.1",
        "postcss": "8.4.33",
    }
}
```

```js
// rspack.config.js
module.exports = {
  entry: path.resolve(__dirname, "./index.js"),
  module: {
    rules: [{
      test: /\.css$/,
      use: [{
        loader: "postcss-loader",
        options: {
          postcssOptions: {
            plugins: [{
                postcssPlugin: 'tailwind-auto-import',
                prepare() {
                  return {
                    Once: (root, {
                      AtRule
                    }) => {
                      const tailwindBase = new AtRule({
                        name: 'tailwind',
                        params: 'base',
                      });
                      root.prepend(tailwindBase);
                    }
                  }
                }
              },
              require('tailwindcss')
            ]
          }
        }
      }]
    }]
  },
}

// index.js
import './index.css'

// index.css
body {
  background: wheat;
}
```

### Reproduce link

_No response_

### Reproduce Steps

`npx rspack build` and it will throw:

```txt
let inputId = inputs.get(value.input)
                              ^
TypeError: Cannot read properties of undefined (reading 'input')
```

NotEvenANeko:
## Summary

Pass `additionalData` as a pointer to Rust side, and call `napi_create_reference` with it, then pass it back to JS side on next JS loader call, without invoking the `toJSON` method. 
Added `NormalModuleAdditionalDataHook` to convert `napi_value` into rust data structure if it's necessary. In this example, it's `CssExtractJsonDataList`.

closes #5600

## Require Documentation?

- [x] No


crates/rspack_binding_options/src/options/raw_module/js_loader.rs
```diff
@@ -416,6 +411,80 @@ pub async fn run_builtin_loader(
   JsLoaderContext::try_from(&mut cx).map_err(|e| Error::from_reason(e.to_string()))
 }
 
+pub struct AdditionalDataRef {
```
h-a-n-a, line 414: First of all, this implementation is great. However it's not error-prone. I would recommend you to use `napi::Ref<()>` as the type to store the `AdditionalData`. You may wish to take [this](https://github.com/web-infra-dev/rspack/blob/ee1a6893f9a248a9378c2f1ac08b572798fc5334/crates/rspack_fs_node/src/node.rs#L5) as an example.

NotEvenANeko: I use the `napi::Ref<()>` to hold the JS value, but it doesn't implement neither `Copy` nor `Clone` trait, so I wrap it with `Arc<Mutex<_>>`. And I need to implement `Send` and `Sync` for it due to `napi::Env` is not thread safe.

The `AdditionalDataRef` is renamed to `JsValueRef` and moved to `src/options/raw_module/js_value_ref.rs`.

crates/rspack_binding_options/src/options/raw_module/js_value_ref.rs
```diff
@@ -0,0 +1,74 @@
+use std::{
```
h-a-n-a, line 1: This file would be quite general. Would you please move this to `rspack_napi_shared/src/js_values` ?

crates/rspack_binding_options/src/options/raw_module/js_value_ref.rs
```diff
@@ -0,0 +1,74 @@
...
+use napi::bindgen_prelude::*;
+use napi::{Env, NapiValue, Ref};
+
+pub struct JsValueRef<T: NapiValue> {
+  env: Env,
+  ref_: Arc<Mutex<Ref<()>>>,
+  _phantom: PhantomData<T>,
+}
+
+unsafe impl<T: NapiValue> Send for JsValueRef<T> {}
```
h-a-n-a, line 15: This data structure is not `Send` and `Sync`. As you can not access `env` and the data behind this on the other thread.

crates/rspack_binding_options/src/options/raw_module/js_value_ref.rs
```diff
@@ -0,0 +1,74 @@
...
+    Self {
+      env: self.env,
+      ref_: self.ref_.clone(),
+      _phantom: PhantomData,
+    }
+  }
+}
+
+impl<T: NapiValue> Drop for JsValueRef<T> {
+  fn drop(&mut self) {
```
h-a-n-a, line 64: Reference counter should be decreased here with NAPI API.

crates/rspack_binding_options/src/options/raw_module/js_value_ref.rs
```diff
@@ -0,0 +1,74 @@
...
+  marker::PhantomData,
+  sync::{Arc, Mutex},
+};
+
+use napi::bindgen_prelude::*;
+use napi::{Env, NapiValue, Ref};
+
+pub struct JsValueRef<T: NapiValue> {
+  env: Env,
+  ref_: Arc<Mutex<Ref<()>>>,
```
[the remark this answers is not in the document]
NotEvenANeko: The `Ref<T>` type is not implement `Clone` trait, so I can't store it in the `AdditionalData` type of `LoaderContext` which requires the type to be `AnyClone + Send + Sync`. Wrap it with a `Arc<Mutex<T>>` will make it clonable.

And it seems that I have to store the `napi::Env` in the struct if I want to unreference the ref in `Drop` trait, but `napi::Env` is not thread safe meanwhile `JsValueRef` should be thread safe to pass around threads. `AdditionalData` requires its generic implements `Send` and `Sync` too.

That will be great if I can store the value to a Rust-owned value by using `serde_json::Value` for example, but it'll crash on some object with circular reference and can only store JSON types.

Or I can make the `additionalData` a JsFunction which returns the actual additional data, and wrap it into a `ThreadsafeFunction`, execute it to get the `additionalData` when pass it back to next loader?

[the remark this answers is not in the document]
h-a-n-a: I would suggest you to create a `ThreadsafeJsValueRef`, which is only a wrapper around a the original `JsValueRef` and make this one `Send`, `Sync` and `Clone`. Roughly like `ThreadsafeJsValueRef(Arc<JsValueRef>>)`. Be sure not to derive `Clone` for `JsValueRef`. This keeps the generality for the `JsValueRef`. Then you can pass this wrapper between threads. This struct's only method is to unwrap it's content into `JsValueRef`. For the `drop` method of which, it's basically checking reference counter of `Arc` and calls the `unref` when this is the only reference. `FromNapiValue` and `ToNapiValue` should be implemented for this type too.

h-a-n-a:
@NotEvenANeko Are you still working on this one?

NotEvenANeko:
I'm sorry I'm quite busy recently, and I'll continue on this one few days later when I'm free.

crates/rspack_napi/src/threadsafe_js_value_ref.rs
```diff
@@ -0,0 +1,149 @@
...
+  fn clone(&self) -> Self {
+    Self {
+      inner: self.inner.clone(),
+    }
+  }
+}
+
+impl<T: NapiValue> Drop for ThreadsafeJsValueRef<T> {
+  fn drop(&mut self) {
+    if Arc::strong_count(&self.inner) == 1 {
```
h-a-n-a, line 29: It's not sound here. `Arc::strong_count` uses `Relax` ordering. You can leverage `Arc`'s drop when dropping the inner content. For example: `Arc<NewType<Mutex<...>>>` and implement `fn drop` for `NewType`.

h-a-n-a (approved):
Thanks for the great work!
Requesting @JSerFeng for changes related to `CssExtractRspackPlugin`.

h-a-n-a:
!canary

JSerFeng (approved):

PR #5909, opened 2024-03-10
fix(typing): fix partial ts-expect-error in  packages/rspack 

gaoachao:
part of [#4640](https://github.com/web-infra-dev/rspack/issues/4640)

## Summary
This pr fixes all ts-expect-error in packages/rspack/src/Compiler.ts and packages/rspack/src/Compilation.ts. Errors are most related to
* Align two type `WatchOptions` present in rspack
* Solve the type problems caused by optional `args` in type `LogFunction`
* Solve some type problems concerning optional or required attributes
  

## Test Plan
```
cd packages/rspack
pnpm build
```

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack/src/Compilation.ts
+++ b/packages/rspack/src/Compilation.ts
@@ -67,7 +67,7 @@ export interface Asset {
 }
 export interface LogEntry {
 	type: string;
-	args: any[];
+	args?: any[];
 	time?: number;
 	trace?: string[];
 }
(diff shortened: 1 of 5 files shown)


Boshen:
I reverted the change because there is an API breaking change:

```
watch(
		watchOptions: WatchOptions,
		handler: (error?: Error, stats?: Stats) => void
	): Watching | void // <----- should not return void
```

gaoachao:
Got it, thank you very much. I'll see if there is any other way to only return Watching and fix @ts-expect-error at the same time. Then PR again.

Boshen:
> Got it, thank you very much. I'll see if there is any other way to only return Watching and fix @ts-expect-error at the same time. Then PR again.

That would be great!

PR #5915, opened 2024-03-11
feat: lazy compilation

JSerFeng:
## Summary

What is lazy compilation ?

The easiest mental model would be:

Imagine some code that you may not use at the first run, for example, when you click a button, load another page, if you don't click that button at all when you develop other stuff, that page is useless, so we don't need to build it. We build it once we do execute the click.

We try to make entries and dynamic imports as split point, for example, consider:

```
A -dyn import()-> B -dyn import()-> C
```

When we compile A, we make B just an empty file, pretend user never write any code in it.

```
A -dyn import() -> B'(empty content)
```

Once we access B, which means we do need to build B, we fill B with its original content, pretend that user write content immediately

## Require Documentation?

- [ ] No

### DIFF
--- a/crates/rspack_binding_options/src/options/raw_builtins/mod.rs
+++ b/crates/rspack_binding_options/src/options/raw_builtins/mod.rs
@@ -4,6 +4,7 @@ mod raw_copy;
 mod raw_css_extract;
 mod raw_html;
 mod raw_ignore;
+mod raw_lazy_compilation;
 mod raw_limit_chunk_count;
 mod raw_mf;
 mod raw_progress;
@@ -14,7 +15,7 @@ mod raw_to_be_deprecated;
 
 use napi::{bindgen_prelude::FromNapiValue, Env, JsUnknown};
 use napi_derive::napi;
-use rspack_core::{BoxPlugin, Define, DefinePlugin, PluginExt, Provide, ProvidePlugin};
+use rspack_core::{BoxPlugin, Define, DefinePlugin, Plugin, PluginExt, Provide, ProvidePlugin};
 use rspack_error::Result;
 use rspack_ids::{
   DeterministicChunkIdsPlugin, DeterministicModuleIdsPlugin, NamedChunkIdsPlugin,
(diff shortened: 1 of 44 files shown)


JSerFeng:
!canary

JSerFeng:
!canary

JSerFeng:
!canary

JSerFeng:
!canary

JSerFeng:
!canary

crates/rspack_binding_options/src/options/raw_builtins/mod.rs
```diff
@@ -142,6 +145,7 @@ pub enum BuiltinPluginName {
   // rspack js adapter plugins
   // naming format follow XxxRspackPlugin
   JsLoaderRspackPlugin,
+  LazyCompilation,
```
ahabhgk, line 148:
```suggestion
  LazyCompilationPlugin,
```

hardfist:
we can try it on rspress and see the performance gain, cc @chenjiahan

chenjiahan:
@Timeless0911 can you try to enable lazy compilation in Rspress and give some feedbacks? ❤️

Timeless0911:
Sync some context for public, now two issues have been discovered yet

1. HMR does not take effect when first screen is loaded, causing the page to be actively refreshed to see it
2. The optimize effect of huge doc website projects is not obvious

@JSerFeng is tracking on it.

ahabhgk (approved):

JSerFeng:
> HMR does not take effect when first screen is loaded, causing the page to be actively refreshed to see it

rsbuild has its own entry for hmr runtime, and that entry has been lazy compiled because of lazyCompilation, so the hmr didn't work as expected because its runtime has not been injected.

> The optimize effect of huge doc website projects is not obvious

After fixing some other bugs that causing the performance issue, the build time decreased from 4 seconds to less than 0.5 seconds, is's obvious

Timeless0911:
> > the hmr didn't work as expected because its runtime has not been injected.

Can we do something in Rsbuild or Rspack to make it work? Refresh manuall really hurts DX.

JSerFeng:
rsbuild could encapsulate this part of the logic and put rsbuild's own entry into `test`.
eg.

```
module.exports = {
  experiments: {
    lazyCompilation: {
      test(module) {
        const name = module.nameForCondition()!;
        const isHmrEntry =
                name.includes('/client/hmr') ||
                name.includes('/client/reactRefreshEntry.js');
        return !isHmrEntry;
      }
    }
  }
}
```

PR #5957, opened 2024-03-15
feat: support filename function

branchseer:
## Summary

Allow `filename` to be a function in these places:
- `output.filename`
- `output.chunkFilename`
- `output.cssFilename`
- `output.cssChunkFilename`
- `compilation.getPath(filename, data)`
- `compilation.getPathWithInfo(filename, data)`
- `compilation.getAssetPath(filename, data)`
- `compilation.getAssetPathWithInfo(filename, data)`

Most changed files in the PR are refactoring (in)direct callers of  `Filename::render` due to its return type changed from  `String` to `rspack_error::Result<String>` (functions could throw).

Other notable changes:
- Add type `ByRef` that implements `Hash` and `Eq` by js reference.
  This is essential as filenames need equality check in some places.
- Add type `ThreadSafeFunctionWithRef` that can be called in the node thread without causing deadlocks.
  This is to avoid `Filename::render` becoming `async`. The reason I choose to make `Filename::render`  blocking instead of `async` is `async` would infect all the (in)direct callers, and require much more refactors than changing the return type. I doubt that's worth the effort since js filename functions are usually very lightweight.
- Allow `entry` to be a function.
   In webpack tests, most tests for filename functions also defines `entry` as function, not object. We have to support this to unlock them.

## Require Documentation?

- [ ] No


crates/rspack_napi/src/threadsafe_function/mod.rs
```diff
@@ -67,7 +80,7 @@ impl<T: 'static, R> ThreadsafeFunction<T, R> {
   }
 
   fn call_with_return<D: 'static + FromNapiValue>(&self, value: T) -> Receiver<Result<D>> {
-    let (tx, rx) = tokio::sync::oneshot::channel::<Result<D>>();
+    let (tx, rx) = oneshot::channel::<Result<D>>();
```
branchseer, line 83: [`tokio::sync::oneshot::Receiver::blocking_recv`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.blocking_recv) panics when called within an async context.

Replacing it with [`onehost::Receiver::recv`](https://docs.rs/oneshot/latest/oneshot/struct.Receiver.html#method.recv).

crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs
```diff
@@ -117,64 +117,57 @@ impl RuntimeModule for GetChunkFilenameRuntimeModule {
         }
       });
 
-    let mut dynamic_filename: Option<&Filename> = None;
+    let mut dynamic_filename: Option<&str> = None;
     let mut max_chunk_set_size = 0;
-    let mut chunk_filenames = IndexMap::new();
+    let mut chunk_filenames = IndexMap::<&Filename, IndexSet<&ChunkUkey>>::new();
```
branchseer, line 122: This is where `Filename` is needed to impl `Hash` and `Eq`.

crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs
```diff
@@ -117,64 +117,57 @@ impl RuntimeModule for GetChunkFilenameRuntimeModule {
...
-            let should_update = match dynamic_filename {
-              Some(dynamic_filename) => match chunk_set.len().cmp(&max_chunk_set_size) {
-                Ordering::Less => false,
-                Ordering::Greater => true,
-                Ordering::Equal => match filename_template
-                  .template()
-                  .len()
-                  .cmp(&dynamic_filename.template().len())
-                {
+            if let Some(filename_template) = filename.template() {
```
branchseer, line 138: Related implementation in webpack: https://github.com/webpack/webpack/blob/e4fee1527b5444da931be6694227143f1cba53c4/lib/runtime/GetChunkFilenameRuntimeModule.js#L65

packages/rspack/src/config/normalization.ts
```diff
@@ -104,7 +104,9 @@ export const getNormalizedRspackOptions = (
 		entry:
 			config.entry === undefined
 				? { main: {} }
-				: getNormalizedEntryStatic(config.entry),
+				: getNormalizedEntryStatic(
+						typeof config.entry === "function" ? config.entry() : config.entry
+					),
```
branchseer, line 109: Allowing `config.entry` to be a function returning `EntryStatic`. 

Good enough to unlock the filename function tests, but not complete:
in webpack `config.entry` returning `Promise<EntryStatic>` is also supported.

crates/rspack_plugin_runtime/src/runtime_plugin.rs
```diff
@@ -235,16 +233,14 @@ impl Plugin for RuntimePlugin {
     }
 
     if runtime_requirements.contains(RuntimeGlobals::GET_CHUNK_SCRIPT_FILENAME) {
-      let chunk_filename = compilation.options.output.chunk_filename.template();
-      if FULL_HASH_PLACEHOLDER.is_match(chunk_filename) || HASH_PLACEHOLDER.is_match(chunk_filename)
+      if matches!(compilation.options.output.chunk_filename.template(), Some(template) if has_hash_placeholder(template))
```
branchseer, line 236: Related webpack implementation:

https://github.com/webpack/webpack/blob/e4fee1527b5444da931be6694227143f1cba53c4/lib/RuntimePlugin.js#L264

h-a-n-a:
Thanks for the contribution! I will get you back when I'm free ;-)

crates/rspack_binding_values/src/path_data.rs
```diff
@@ -10,6 +12,47 @@ pub struct PathData {
...
+  }
+}
+
+#[napi(object)]
+pub struct ChunkPathData {
+  pub id: Option<String>,
+  pub name: Option<String>,
+  pub hash: Option<String>,
+  pub content_hash: Option<HashMap<String, String>>,
+}
```
branchseer, line 39: Name consistent with the one in webpack's dts `https://github.com/webpack/webpack/blob/e4fee1527b5444da931be6694227143f1cba53c4/types.d.ts#L1423`.

More fields to be added. Existing ones are enough for most use.

We can't reuse https://github.com/web-infra-dev/rspack/blob/9c25b4d51e36d8f8a734874318d460201ae38695/crates/rspack_binding_values/src/chunk.rs#L9 because it has non-optional `inner*` fields, which would reject user-supplied `PathData` at https://github.com/web-infra-dev/rspack/blob/9c25b4d51e36d8f8a734874318d460201ae38695/crates/rspack_binding_values/src/compilation.rs#L378

crates/rspack_binding_values/src/filename.rs
```diff
@@ -0,0 +1,109 @@
...
+use serde::Deserialize;
+
+use crate::{JsAssetInfo, PathData as JsPathData};
+
+/// A js filename value. Either a string or a function
+///
+/// The function type is generic. By default the function type is thread-safe and implements `Hash` and `Eq` by reference.
+#[derive(Debug)]
+pub struct JsFilename<
+  F = ByRef<ThreadSafeFunctionWithRef<(JsPathData, Option<JsAssetInfo>), String>>,
```
h-a-n-a, line 20: I might not recommend you to create an additional isomorphic tsfn. By default, `JsFunction` and `ThreadsafeFunction` are two different kind of stuff. Trying to merge these two together would cause its state and types being implicit and cause potential deadlocks.

crates/rspack_napi/src/by_ref.rs
```diff
@@ -0,0 +1,107 @@
...
+use derivative::Derivative;
+use napi::{
+  bindgen_prelude::{FromNapiValue, ToNapiValue, TypeName, ValidateNapiValue},
+  Env, JsFunction, JsNumber, Ref,
+};
+
+/// Wrapper for napi types that provides by-reference equality and hash
+#[derive(Derivative, Debug, Clone)]
+#[derivative(Eq, PartialEq, Hash)]
+pub struct ByRef<T> {
```
h-a-n-a, line 12: I didn't see why this is necessary. Would you please make some clarifications?

crates/rspack_napi/src/threadsafe_function/mod.rs
```diff
@@ -25,6 +28,16 @@ pub struct ThreadsafeFunction<T: 'static, R> {
   _data: PhantomData<R>,
 }
 
+impl<T, R> Debug for ThreadsafeFunction<T, R> {
+  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    f.debug_struct("ThreadsafeFunction")
```
h-a-n-a, line 33: This is not intentional, right? Seems not necessary.

crates/rspack_napi/src/threadsafe_function/with_ref.rs
```diff
@@ -0,0 +1,109 @@
...
+use super::ThreadsafeFunction;
+use crate::errors::NapiResultExt;
+
+/// Thread safe function type that
+/// - implements `ToNapiValue`, and
+/// - can be synchronously called from any thread, including the node thread
+#[derive(Derivative)]
+#[derivative(Debug)]
+#[derivative(Debug(bound = ""))]
+pub struct ThreadSafeFunctionWithRef<T: 'static, R> {
```
h-a-n-a, line 17: I would prefer `JsFunctionRef` and `ThreadsafeFunction`. It's not necessary to create a isomorphic struct and changing `JsFilename` to an `enum` seems enough.
You can cut and paste my old implementation to crate `rspack_napi` and share between crates.

crates/rspack_binding_values/src/path_data.rs
h-a-n-a:
```suggestion
pub struct JsChunkPathData {
  pub id: Option<String>,
  pub name: Option<String>,
  pub hash: Option<String>,
  pub content_hash: Option<HashMap<String, String>>,
}
```

Prefixed with `Js` to make things more clear.

branchseer: Resolved in 0baf67455a1ed6dc2ab2f302024d093a18be7af5

crates/rspack_napi/src/by_ref.rs
branchseer: This makes [the tsfn in `JsFilename`]( https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_binding_values/src/filename.rs#L20 ) implements `Hash` and `Eq`, allowing it to be [converted to `Arc<dyn FilenameFn>`](https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_binding_values/src/filename.rs#L41).  `FilenameFn` requires `Hash` and `Eq` because that `Filename` needs to be hashmap keys [here](https://github.com/web-infra-dev/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs#L122) (this is consistent with the [webpack implementation](https://github.com/webpack/webpack/blob/0bc85d127c14347d646f1354a6cbf2d1142905d8/lib/runtime/GetChunkFilenameRuntimeModule.js#L48))

crates/rspack_napi/src/threadsafe_function/mod.rs
branchseer: This is to make it easier for structs/enums that contain  `ThreadsafeFunction`  to implement `Debug`, from `ThreadSafeFunctionWithRef` all the way up to `Filename`.

crates/rspack_napi/src/threadsafe_function/with_ref.rs
branchseer: Sorry I don't quite understand the comments. Could you elaborate more about avoiding isomorphic tsfn and changing `JsFilename` to an `enum`? Thanks.

Meanwhile here's a more detailed explanation on the current design. Hope it clears things up a little bit.
## Why ThreadSafeFunctionWithRef

The ultimate purpose of `ThreadSafeFunctionWithRef` is for `JsFilenameFn` to be both  `Send + Sync` and non-async.
- Field `tsfn: ThreadsafeFunction<T, R>`  is for blocking calls from non-node threads;
- Field `fn_ref: Ref<()>`  is for blocking calls in the node thread.

so I think **both `ThreadsafeFunction` and  `Ref<()>` are needed, unless we make `JsFilenameFn` async**.

`fn_ref` also make it possible to implement `ToNapiValue`, which is required by `#[napi(object)]`.

I don't think we can replace `Ref<()>` with  `JsFunctionRef` because `JsFunctionRef` can't be [safely dropped from non-node threads](https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_fs_node/src/node.rs#L25), whereas `Ref<()>` [can](https://docs.rs/napi/latest/src/napi/js_values/value_ref.rs.html#56-61) as it holds zero refcount in `ThreadSafeFunctionWithRef` (we have tsfn to strongly hold the function).
## Why make function type generic in `JsFilename` and `Filename`

We want to avoid creating tsfn when possible, and want the `render` function to be generic on the error type.
- In [`JsCompilation::get_path` and its family](https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_binding_values/src/compilation.rs#L360-L405), filename functions are immediately invoked. Creating tsfn is not needed. `JsFilename<JSFunction>` will suffice. We also need `render`'s error type to be `napi::Error` here,  so that these methods can rethrow if the filename function throws.
- As seen in this PR, in many places template-only filenames are passed to `compilation.get_path` or its family. In these cases, the result should always succeed. Generic function type allows us to use `NoFilenameFn` as the function type, whose error type `Infallible`, which allows us to unwrap without handling errors ([`ResultInfallibleExt::always_ok`](https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_util/src/infallible.rs#L10)).

crates/rspack_napi/src/by_ref.rs
h-a-n-a: It still seems not necessary to implement this for filename, a filename enum can accept both string and function variant looks enough. Webpack didn't seem necessary doing anything with its being the `HashMap` key, you can implement this with other struct roughly like a `Vec<(Filename, Chunks)>`.

crates/rspack_napi/src/threadsafe_function/with_ref.rs
h-a-n-a: Here's my point. The `ThreadsafeFunction` and `JsFunction` are two different kinds of stuff. For `ThreadsafeFunction`, it's ok to be called with thread safety and can be dropped and destroyed on the other thread. However, `JsFunction` is not. Mixing these implementations would cause the program hard to debug in the future. I would recommend you to use either one of them, and not try to make an abstraction around these two things.

crates/rspack_napi/src/threadsafe_function/mod.rs
h-a-n-a: Pointers should not be exposed. Printing `ThreadsafeFunction(...)` seems enough if you would like to implement `Debug`.

crates/rspack_napi/src/by_ref.rs
branchseer: Webpack is grouping chunks by filenames [here](https://github.com/webpack/webpack/blob/0bc85d127c14347d646f1354a6cbf2d1142905d8/lib/runtime/GetChunkFilenameRuntimeModule.js#L60-L64). And currently rspack [does the same](https://github.com/web-infra-dev/rspack/blob/4ccff45dde2ff9b26649ce99ef95ec41525428ad/crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs#L135-L139), and later in [here](https://github.com/web-infra-dev/rspack/blob/4ccff45dde2ff9b26649ce99ef95ec41525428ad/crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs#L242-L249) the grouped chunks are iterated to generated `static_urls`, which is also grouped by `filename`.

I don't see how grouping is possible if `Filename` doesn't impl `Eq`, unless we only group by filename strings and consider all filename functions distinct to each other?

crates/rspack_napi/src/threadsafe_function/mod.rs
branchseer: Resolved in d9ff5eea6c77da93af96d2fdd22af158725b788b

crates/rspack_napi/src/threadsafe_function/with_ref.rs
branchseer: Ok I now understand your concern. I have removed `ThreadsafeFunctionWithRef` and replaced its usages with vanilla `ThreadsafeFunction` in b15d622957a84c0ebf1066b5e58b43f5a44fe11f.

Now the thread-safe version of `Filename` can't be called on the node thread. Fortunately there is no such case according to the tests, and even if there is, tsfn would [detect and report the deadlock](https://github.com/nodejs/node/commit/45676be3dc27683e0a4cd76f981beecf60ed22fb), so that would be easy to debug.

crates/rspack_binding_values/src/filename.rs
```diff
@@ -0,0 +1,107 @@
...
+impl<F: ToNapiValue> ToNapiValue for JsFilename<F> {
+  unsafe fn to_napi_value(
+    env: napi::sys::napi_env,
+    val: Self,
+  ) -> napi::Result<napi::sys::napi_value> {
+    Either::to_napi_value(env, val.0)
+  }
+}
+
+impl<'de> Deserialize<'de> for JsFilename {
```
h-a-n-a, line 65: Would you please remove this? Deserialization seems not necessary anymore. You can remove every deserialize derives above.

crates/rspack_napi/src/threadsafe_function.rs
```diff
@@ -67,7 +73,7 @@ impl<T: 'static, R> ThreadsafeFunction<T, R> {
   }
 
   fn call_with_return<D: 'static + FromNapiValue>(&self, value: T) -> Receiver<Result<D>> {
-    let (tx, rx) = tokio::sync::oneshot::channel::<Result<D>>();
+    let (tx, rx) = oneshot::channel::<Result<D>>();
```
h-a-n-a, line 76: Would you please revert this? You can create the Tokio handle from callsite.

packages/rspack-test-tools/src/processor/normal.ts
```diff
@@ -23,7 +23,10 @@ export class RspackNormalProcessor extends BasicTaskProcessor<ECompilerType.Rspa
 	constructor(protected _normalOptions: IRspackNormalProcessorOptions) {
 		super({
 			compilerType: ECompilerType.Rspack,
-			findBundle: (context, options) => options.output?.filename,
+			findBundle: (context, options) => {
+				const filename = options.output?.filename;
+				return typeof filename === "string" ? filename : undefined;
```
h-a-n-a, line 28: Is this used?

h-a-n-a: cc @LingyuCoder

LingyuCoder: Yes, this looks good to me, findBundle() is used to get the asset file to run, if filename is not a string, an error will be caused

branchseer: Yes. The type of `options.output?.filename` is now changed `string | function`, but `findBundle` is defined to return `string[] | string | void` in [here](https://github.com/web-infra-dev/rspack/blob/c6b58310cfe9db5fc6ef2013177f1f8303d1ce47/packages/rspack-test-tools/src/processor/basic.ts#L26).

crates/rspack_napi/src/threadsafe_function.rs
branchseer: Do you mean [`Handle::block_on`](https://docs.rs/tokio/1.36.0/tokio/runtime/struct.Handle.html#method.block_on)? That still panics if called within an async context.

Panic message in webpack test:
```
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
                              ⋮ 17 frames hidden ⋮                              
18: <rspack_binding_values::filename::ThreadSafeFilenameFn as rspack_core::options::filename::LocalFilenameFn>::call::h6eaff8e5a90d822a
    at /Users/branchseer/code/rspack-contrib/crates/rspack_binding_values/src/filename.rs:85
19: <alloc::sync::Arc<dyn rspack_core::options::filename::FilenameFn+Error = miette::eyreish::Report> as rspack_core::options::filename::LocalFilenameFn>::call::h1f487ba22170ed27
    at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/options/filename.rs:116
20: rspack_core::options::filename::Filename<F>::render::h68b6f950e537a0ea
    at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/options/filename.rs:164
21: rspack_core::compiler::compilation::Compilation::get_path_with_info::h49aa4a7fc1b89094
    at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/compiler/compilation.rs:1346
22: rspack_plugin_javascript::plugin::impl_plugin_for_js_plugin::<impl rspack_core::plugin::api::Plugin for rspack_plugin_javascript::plugin::JsPlugin>::render_manifest::{{closure}}::hdbeee98f9fb3a496
    at /Users/branchseer/code/rspack-contrib/crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs:263
                               ⋮ 1 frame hidden ⋮                               
Message:  Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
Location: crates/rspack_binding_values/src/filename.rs:85

```

crates/rspack_binding_values/src/filename.rs
branchseer: Deserialization is used here: https://github.com/web-infra-dev/rspack/blob/c6b58310cfe9db5fc6ef2013177f1f8303d1ce47/crates/rspack_testing/src/run_fixture.rs#L20

h-a-n-a: I see... this is supposed to be removed in the future. I will initiate another PR to fix this.

crates/rspack_napi/src/threadsafe_function.rs
h-a-n-a: > Do you mean [`Handle::block_on`](https://docs.rs/tokio/1.36.0/tokio/runtime/struct.Handle.html#method.block_on)? That still panics if called within an async context.
> 
> Panic message in webpack test:
> 
> ```
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>                               ⋮ 17 frames hidden ⋮                              
> 18: <rspack_binding_values::filename::ThreadSafeFilenameFn as rspack_core::options::filename::LocalFilenameFn>::call::h6eaff8e5a90d822a
>     at /Users/branchseer/code/rspack-contrib/crates/rspack_binding_values/src/filename.rs:85
> 19: <alloc::sync::Arc<dyn rspack_core::options::filename::FilenameFn+Error = miette::eyreish::Report> as rspack_core::options::filename::LocalFilenameFn>::call::h1f487ba22170ed27
>     at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/options/filename.rs:116
> 20: rspack_core::options::filename::Filename<F>::render::h68b6f950e537a0ea
>     at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/options/filename.rs:164
> 21: rspack_core::compiler::compilation::Compilation::get_path_with_info::h49aa4a7fc1b89094
>     at /Users/branchseer/code/rspack-contrib/crates/rspack_core/src/compiler/compilation.rs:1346
> 22: rspack_plugin_javascript::plugin::impl_plugin_for_js_plugin::<impl rspack_core::plugin::api::Plugin for rspack_plugin_javascript::plugin::JsPlugin>::render_manifest::{{closure}}::hdbeee98f9fb3a496
>     at /Users/branchseer/code/rspack-contrib/crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs:263
>                                ⋮ 1 frame hidden ⋮                               
> Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
> Message:  Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
> Location: crates/rspack_binding_values/src/filename.rs:85
> ```

Yes, the solution maybe getting the handle from the main thread and use it later in functions like [this](https://github.com/web-infra-dev/rspack/blob/2a195b76aded8e133d0e53965c1cce47e6684e3f/crates/rspack_binding_options/src/options/raw_devtool.rs#L97)

h-a-n-a:
@ahabhgk Would you please help double check the API part at your convenience?

h-a-n-a (dismissed):
Thank you so much for the wonderful work you've done! We really appreciate your contribution ;-)

crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs
```diff
@@ -1,14 +1,15 @@
-use std::{cmp::Ordering, fmt};
+use std::{cmp::Ordering, collections::HashMap, fmt};
```
ahabhgk, line 1: use `rustc_hash::FxHashMap` for better perf

ahabhgk (commented):
Generally looks good to me, one small nitpick

hardfist:
@branchseer you can add document change in this pr, since we have moved the website repo into rspack repo

branchseer:
> @branchseer you can add document change in this pr, since we have moved the website repo into rspack repo

@hardfist done in b38608afa97c46b9bc33bb2361da9f3ead114037.

crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs
branchseer: Done in 4873c8254da72d798ac0da2599a529abadeb4ae2.

ahabhgk (approved):
Thanks!

PR #5958, opened 2024-03-18
fix: should call compile hook

ahabhgk:
## Summary

- beforeCompile and afterCompile is not needed for rust side
- should call compile hook

## Require Documentation?

- [ ] No


h-a-n-a (approved):

packages/rspack/tests/Stats.test.js
```diff
@@ -176,7 +176,6 @@ describe("Stats", () => {
 		<t> finish make hook: X ms
 		<t> finish compilation: X ms
 		<t> seal compilation: X ms
-		<t> afterCompile hook: X ms
```
h-a-n-a, line 179: Forgot this line. Is this intentional?

ahabhgk: it's intentional, we should log this in js side after remove it from rust side, but afterCompile is not affected for performance, so it should be ok to just remove it

PR #5960, opened 2024-03-18
refactor: this compilation hook

ahabhgk:
## Require Documentation?

- [ ] No

### DIFF
--- a/crates/node_binding/src/plugins/interceptor.rs
+++ b/crates/node_binding/src/plugins/interceptor.rs
@@ -8,8 +8,8 @@ use napi::{
 use rspack_binding_values::{JsBeforeResolveArgs, JsBeforeResolveOutput, JsCompilation};
 use rspack_core::{
   BeforeResolveArgs, Compilation, CompilationParams, CompilationProcessAssetsHook,
-  CompilerCompilationHook, CompilerMakeHook, CompilerShouldEmitHook, MakeParam,
-  NormalModuleFactoryBeforeResolveHook,
+  CompilerCompilationHook, CompilerMakeHook, CompilerShouldEmitHook, CompilerThisCompilationHook,
+  MakeParam, NormalModuleFactoryBeforeResolveHook,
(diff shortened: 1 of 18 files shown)


crates/rspack_core/src/compiler/mod.rs
```diff
@@ -141,7 +144,9 @@ where
     // Fake this compilation as *currently* rebuilding does not create a new compilation
```
h-a-n-a, line 144: This comment is stale?

ahabhgk: Not sure, not the author, I can't understand what `Fake this compilation as *currently* rebuilding` means, but `does not create a new compilation` is not stale 😂

h-a-n-a: I dug into our codebase and found: https://github.com/branchseer/rspack/blob/bf1ac34f02b2676ac8806fcda449aa1543b45b04/crates/rspack_core/src/compiler/hmr.rs#L71. Just want to make sure the new compilation get updated to the `JsCompilation`(this would be better to add a comment somewhere, maybe in other PRs), otherwise it's a UB.

crates/node_binding/src/plugins/interceptor.rs
```diff
@@ -219,18 +229,40 @@ define_register!(
...
 );
 
+#[async_trait]
+impl AsyncSeries2<Compilation, CompilationParams> for CompilerThisCompilationTap {
+  async fn run(
+    &self,
+    compilation: &mut Compilation,
+    _: &mut CompilationParams,
+  ) -> rspack_error::Result<()> {
+    // SAFETY: `Compiler` will not be moved, as it's stored on the heap.
```
h-a-n-a, line 253: This safety comment I wrote is a little bit long and almost look annoying something. Should be shorten in the future 🥲.

h-a-n-a (approved):

PR #5993, opened 2024-03-21
docs: update documentation contribution

hardfist:
## Summary
update document contribution & skip document ci

## Require Documentation?

- [x] No


.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ on:
     types: [opened, synchronize]
     paths-ignore:
       - "**/*.md"
+      - "website/**/*"
```
chenjiahan, line 17: If we use `paths-ignore` here, the required CI will never be passed:

image

chenjiahan: that's why Rsbuild uses an extra script to support this

https://github.com/web-infra-dev/rsbuild/blob/main/scripts/skipCI.js

chenjiahan (approved):

PR #6021, opened 2024-03-25
fix: stablize chunk id

### LINKED ISSUE #6020: [Bug]: Chunk Id unstable causing HMR reload
### System Info

No need

### Details

ChunkId is unstable, when changing the code, there is a possibility that can causing ChunkId inconsistent, even the change is only some `console.log`.

If chunk id changes, means many runtime output can change, for example __webpack_require__.e('foo_js-bar_js') can become __webpack_require__.e('bar_js'), causing many module changes, some module may not have HMR boundary and cannot accept hot update, so the page can full reload

### Reproduce link

_No response_

### Reproduce Steps

internal repo.

This issue is just a record, and will be fixed by me

JSerFeng:
## Summary

fix #6020

## Require Documentation?

- [ ] No


JSerFeng:
!bench

JSerFeng:
!canary

JSerFeng:
!canary

ahabhgk (approved):

hardfist:
@JSerFeng can you add some description about the root cause of this bug

hardfist:
@JSerFeng can you add some description about the root cause of this bug

JSerFeng:
> @JSerFeng can you add some description about the root cause of this bug

There are 2 reasons:
1: The algorithm of find graph roots was unstable, if you tried to find the root modules in the same chunk twice, you may get different results, this is because in Webpack, all dependencies or HashSet iter order is consistent, but in Rspack, HashSet iter order is not stable, hit some edge cases, get different results
2: Even the chunk roots modules are the same, there is a possibility that 2 chunks are totally equal, thus when we sort the chunks, the 2 chunks' order are unstable too, so their name can be unstable

stevenpetryk:
I think there are still some issues here—I'm able to trigger a full reload when I change JSX attributes in any way (such as removing/adding a classname), but still get proper fast refreshes when changing something like:

```diff
-<p>foo</p>
+<p>foobar</p>
```

Issue #3522, opened 2023-06-12
NormalModuleReplacementPlugin is not a constructor


zackarychapple:
### System Info

all

### Details

As part of the environment file replacement in the Angular cli there is the call below: 

`new NormalModuleReplacementPlugin(new RegExp('^' + key.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&') + '$'), value).apply(compiler);`

### Reproduce Steps

Enable file replacement in the Angular CLI based builder for Rspack

yharaskrik:
Note on this one: It is not limited to just Angular, in an Nx workspace when we are using the webpack executor currently with say Nest, we can use fileReplacements as well (which we do). I THINK this is the only issue that is preventing us from using rspack in production for our nest servers.

hardfist:
> Note on this one: It is not limited to just Angular, in an Nx workspace when we are using the webpack executor currently with say Nest, we can use fileReplacements as well (which we do). I THINK this is the only issue that is preventing us from using rspack in production for our nest servers.

it seems alias works too https://github.com/nrwl/nx-labs/pull/231/files

yharaskrik:
> > Note on this one: It is not limited to just Angular, in an Nx workspace when we are using the webpack executor currently with say Nest, we can use fileReplacements as well (which we do). I THINK this is the only issue that is preventing us from using rspack in production for our nest servers.
> 
> @yharaskrik it seems alias works too https://github.com/nrwl/nx-labs/pull/231/files

I am actually just looking into that, that PRs change isn't live yet it seems and I am trying to follow to associated Rspack issue saying that a relative alias key works but I cannot get it to work.

zackarychapple:
bump

zackarychapple:
bump

zackarychapple:
bump

zackarychapple:
Very nice!  Thank you

PR #6023, opened 2024-03-25
refactor: normal module factory after resolve hook

ahabhgk:
## Require Documentation?

- [ ] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #6028, opened 2024-03-26
feat(rspack): NormalModuleReplacementPlugin

xc2:
## Summary

fixes:  #3522

thanks #6023

## Require Documentation?

- [x] No

### DIFF
--- a/packages/rspack/src/exports.ts
+++ b/packages/rspack/src/exports.ts
@@ -99,6 +99,8 @@ export { LoaderTargetPlugin } from "./lib/LoaderTargetPlugin";
 
 export { EnvironmentPlugin } from "./lib/EnvironmentPlugin";
 
+export { NormalModuleReplacementPlugin } from "./lib/NormalModuleReplacementPlugin";
+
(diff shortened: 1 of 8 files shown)


ahabhgk:
What about implement it in rust side and expose it to js side by builtin plugin, which has better performance

ahabhgk:
I have an idea about this plugin, when the second argument is a string, builtin plugin will have better performance, when it is a function, js plugin (queried hook) will reduce the amount of rust <-> js call, so js plugin will have better performance. Since this is compatible with the PR's implementation, so let's merge this and we can implement the optimization in the future

ahabhgk (approved):
Thanks

xc2:
> What about implement it in rust side and expose it to js side by builtin plugin, which has better performance

sry i just seen this. 

>  when it is a function, js plugin (queried hook) will reduce the amount of rust <-> js call,

yes I was wondering if it's worth creating a Rust implementation of this plugin, considering that it seems to be just a wrapper for nmf's xResolve hooks.

a replacement function is more common, and the webpack document of this plugin is also guiding one to modify `resource.request` and `createData.resource` instead of returning a new value.

PR #6067, opened 2024-03-29
fix: emit asset file with buffer can use in asset emitted hook correctly

LingyuCoder:
## Checklist


LingyuCoder:
!canary

ahabhgk (approved):

PR #6085, opened 2024-04-01
perf(core): Use unicase for compare_ids

quininer:
## Summary

The current implement of `compare_ids` always alloc memory for `a`, `b` unnecessarily. we can use `unicase` to implement 0-allocation compare. this is expected to be good for performance.
The `unicase` crate is already in our dependency tree, so it's not a new dependency.

This PR also modifies `compare_numbers` to Rusty implementation.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3913,6 +3913,7 @@ dependencies = [
  "rustc-hash",
  "sugar_path",
  "swc_core",
+ "unicase",
 ]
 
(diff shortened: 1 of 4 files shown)


LingyuCoder:
!bench

LingyuCoder:
Thank you, please sign the CLA

LingyuCoder (approved):

PR #5490, opened 2024-01-26
fix: add sourcemap file to chunk.auxiliaryFiles (#3253)

flyingalex:
## Summary
fix: add sourcemap file to chunk.auxiliaryFiles (#3253)

## Test Plan
I created a demo to test it
https://github.com/flyingalex/rspack-for-react-test-verification
## Require Documentation?

- [x] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

Issue #3253, opened 2023-05-22
add sourcemap file to chunk.auxiliaryFiles


9aoy:
### What problem does this feature solve?

When I use rspack-manifest-plugin. I find that the sourcemap file is missing in manifest, then I found that it gets sourcemap file from chunk.auxiliaryFiles. But In rspack, it doesn't output sourcemap file to chunk.auxiliaryFiles.

<img width="716" alt="image" src="https://github.com/web-infra-dev/rspack/assets/22373761/3b0ffa15-6589-475b-baf5-87b155ec7c90">

The behavior in webpack:

<img width="476" alt="image" src="https://github.com/web-infra-dev/rspack/assets/22373761/c2f94777-6adf-41f2-82c9-603cf7fbe7f5">

### What does the proposed API of configuration look like?

the same with webpack.

hardfist:
bump

Hamzakh777:
Assign it to me @hardfist

Hamzakh777:
After some investigation, it seems that the cause of this issue:
- `rspack-manifest-plugin` gets applied twice https://github.com/hardfist/rspack-manifest-plugin/blob/master/src/index.ts#L73 which results in `emitCountMap` https://github.com/hardfist/rspack-manifest-plugin/blob/master/src/index.ts#L11 having count of 2 instead of 1.
- It's been called twice because `html-webpack-plugin` uses a child compiler, and when that child compiler runs, the `beforeRun` hook on the `rspack-manifest-plugin` gets called again.

flyingalex:
Hi, I am following this issue and want to fix it, 

I roughly feel it's the process_assets_stage_dev_tooling method issue in `rspack_plugin_devtool` crate, how I can run related tests after I change the logic? checked `rspack_plugin_devtool` seems no tests there

Any suggestion would be very appreciated, thx

flyingalex:
@9aoy seems already fixed by the core team, please double check

9aoy:
> @9aoy seems already fixed by the core team, please double check

@hulin32 looks not working in rspack v0.5.8 😢

hulin32:
@9aoy hey,  if you are talking about `rspack-manifest-plugin`, then it works, you can see  sourcemapfile in manifest, but its not using chunk.auxiliaryFiles

PR #6125, opened 2024-04-06
fix: add sourcemap file to chunk.auxiliaryFiles

hulin32:
## Summary
fix: add sourcemap file to chunk.auxiliaryFiles (#3253), follow up for #5490

## Test Plan
I created a demo to test it
https://github.com/hulin32/rspack-for-react-test-verification
## Require Documentation?

- [x] No


crates/rspack_plugin_devtool/src/lib.rs
```diff
@@ -570,6 +570,11 @@ async fn process_assets(&self, compilation: &mut Compilation) -> Result<()> {
     compilation.emit_asset(source_filename.to_owned(), source_asset.clone());
     if let Some((source_map_filename, source_map_asset)) = source_map {
       compilation.emit_asset(source_map_filename.to_owned(), source_map_asset.clone());
+      compilation.chunk_by_ukey.values_mut().for_each(|v| {
+        if v.files.contains(&source_filename) {
+          v.auxiliary_files.insert(source_map_filename.to_owned());
+        }
+      });
```
SyMind, line 577: Inserting `auxiliary_files` here is not a good choice. It is better to add them before the following code snippet:

```
mapped_asstes.push(MappedAsset {
    asset: (filename, asset.clone()),
    source_map: Some((source_map_filename, source_map_asset)),
});
```

Additionally, you should remove the following TODO comments:

```rust
// TODO
// chunk.auxiliary_files.add(filename);
```

SyMind:
Please create a unit test within the directory `packages/rspack/tests/cases/chunks`.

crates/rspack_plugin_devtool/src/lib.rs
hulin32: thx a lot for your suggestion, will look into it

hulin32: Hey @SyMind, Could you give me some guidance?

I checked the codes, and I found the borrowing rule in that method is not allowing me to change auxiliary_files, do you have any ideas? the only way I can see right now is to change `auxiliary_files` definition to `RefCell<HashSet<String>>`, but  it also will change a lot of places, not sure if it's the good direction

SyMind: I haven't considered it thoroughly. let's stick with your revisions.

SyMind:
!bench

SyMind:
!bench

SyMind:
I ran the performance test and it looks like there are no issues. Then solve the other TODOs and we can be merged.

hulin32:
> Please create a unit test within the directory `packages/rspack/tests/cases/chunks`.

@SyMind dumb question, I wrote the test, but how can I run the test in this folder to test it on my local ?
I am trying with this from [doc](https://web-infra-dev.github.io/rspack-dev-guide/testing/intro.html):
```shell
# In root path
./x build -a
./x test js
```
but not work, thanks

SyMind:
@hulin32 The document is outdated. Please execute the `pnpm run test` command on `packages/rspack` now to perform the test. We will update the document later.

hulin32:
@SyMind Hey, sorry for bother you,  just for local test right now, I am putting below test codes in `Errors.test.js` file, and I want to use imported `webpack`  to compile the file, then check the `stats` to see if `auxiliaryFiles` are good, but seems it isn't using the rspack from source codes, I put a  print log `println!` in `process_assets` method which from `crates/rspack_plugin_devtool/src/lib.rs`, and ran `cargo build && pnpm run build:cli:debug` in root to compile it, then go to `packages/rspack` to run `pnpm run test tests/Errors.test.js`, but the `println!` log is not shown, so for me seems its not using the local one, any suggestions? thx a lot

```js
it("should add source map file to auxiliaryFiles", async () => {
const stats = await new Promise((resolve, reject) => {
  const compiler = webpack({
    entry: {
		main: path.resolve(__dirname, 'cases/chunks/auxiliary-files/one.js')
	},
    devtool: 'source-map',
  });
  try {
    compiler.run((bailedError, stats) => {
      if (bailedError) {
        return reject(bailedError);
      }
      compiler.close(closeError => {
        if (closeError) {
          return reject(closeError);
        }
        resolve(stats);
      });
    });
  } catch (err) {
    // capture sync thrown errors
    reject(err);
  }
});
console.log("statssss", stats.toJson({ errorDetails: false }));
});
```

SyMind:
@hulin32 It's possible that `stats.errors` contains some error messages that could help diagnose the problem.

SyMind:
This is our current issue. Contributing a test case is quite difficult, so I will add the relevant unit tests.

SyMind (approved):

hulin32:
@SyMind awesome 👍, thanks for your help, also learn something here

PR #6216, opened 2024-04-12
feat: support function filename and chunkFilename

### LINKED ISSUE #6215: [Feature]: Support function type for CssExtractRspackPlugin.filename and CssExtractRspackPlugin.chunkFilename

Allow the `filename` and `chunkFilename` to use functions to align with the APIs from `mini-css-extract-plugin`. Currently, these can only be configured with `string`. Functions allow for more flexibility to handle special cases for generated files.

The rationale is the same as `output.filename` and `output.chunkFilename`, mentioned here:
https://github.com/web-infra-dev/rspack/issues/3483 

The new APIs would align with those used in [mini-css-extract-plugin](https://www.npmjs.com/package/mini-css-extract-plugin#filename)

JSerFeng:
## Summary

close #6215

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3318,7 +3318,6 @@ name = "rspack_plugin_extract_css"
 version = "0.1.0"
 dependencies = [
  "async-trait",
- "dashmap",
  "once_cell",
  "regex",
  "rspack_core",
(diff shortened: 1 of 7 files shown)


SyMind (approved):

JSerFeng:
!canary

PR #6247, opened 2024-04-16
fix(core): update module federation runtime

ScriptedAlchemy:
## Summary

Upgrading module federation runtime tools

There are some bugs in this version of federation which is quite old, one blocking issue is how this version works with node targets, which was patched a few months ago 

## Checklist


packages/rspack/src/container/default.runtime.js
```diff
@@ -91,10 +91,11 @@ module.exports = function () {
 				for (let stage of stages) {
 					if (typeof stage === "object" && stage !== null) {
 						const { name, version, factory, eager } = stage;
+						const options = { version, scope, get: factory };
```
ahabhgk, line 94: need a little patch here for 0.1.6

ahabhgk (dismissed):

ahabhgk (dismissed):

ahabhgk (approved):

PR #6250, opened 2024-04-16
fix: napi value convert error in css-loader

ahabhgk:
## Summary

before `pathdata.chunk.contentHash` is only `HashMap<String, String>`, but in practice (css-loader) it can be a `String`

## Checklist


ahabhgk:
!canary

h-a-n-a (dismissed):

h-a-n-a (approved):

PR #6265, opened 2024-04-17
fix: JsPathData type

ahabhgk:
## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #6274, opened 2024-04-17
fix(binding): fix type error of binding.d.ts; add test for binding.d.ts

xc2:
In addtion to https://github.com/web-infra-dev/rspack/pull/6265

Fix https://github.com/jupyterlab/jupyterlab/pull/16005#issuecomment-2061169949

Also included:

- Test for `binding.d.ts` (just a simple but useful tsc call)
- custom dts header of binding.d.ts for type re-use

### Related error

```log
binding.d.ts:751:13 - error TS2304: Cannot find name 'JsFilename'.

751   filename: JsFilename
                ~~~~~~~~~~

binding.d.ts:752:18 - error TS2304: Cannot find name 'JsFilename'.

752   chunkFilename: JsFilename
                     ~~~~~~~~~~

binding.d.ts:1078:34 - error TS2552: Cannot find name 'PathData'. Did you mean 'JsPathData'?

1078   filename: string | ((pathData: PathData, assetInfo?: JsAssetInfo) => string)
                                      ~~~~~~~~

binding.d.ts:1079:39 - error TS2552: Cannot find name 'PathData'. Did you mean 'JsPathData'?

1079   chunkFilename: string | ((pathData: PathData, assetInfo?: JsAssetInfo) => string)
                                           ~~~~~~~~

binding.d.ts:1081:37 - error TS2552: Cannot find name 'PathData'. Did you mean 'JsPathData'?

1081   cssFilename: string | ((pathData: PathData, assetInfo?: JsAssetInfo) => string)
                                         ~~~~~~~~

binding.d.ts:1082:42 - error TS2552: Cannot find name 'PathData'. Did you mean 'JsPathData'?

1082   cssChunkFilename: string | ((pathData: PathData, assetInfo?: JsAssetInfo) => string)
                                              ~~~~~~~~


Found 6 errors in the same file, starting at: binding.d.ts:751
```

## Checklist

- [x] Documentation updated (or not required). no need


xc2:
!canary

xc2:
!canary

xc2:
!canary

crates/node_binding/binding.d.ts
```diff
@@ -1,6 +1,15 @@
 /* auto-generated by NAPI-RS */
 /* eslint-disable */
 
+/* -- banner.d.ts -- */
+export type JsFilename =
```
h-a-n-a, line 5: IMO, You can create a fake object to generate this type on rust, instead of manually add types here.

xc2: This idea also came to me yesterday as it would centralize the type declarations more - I really agree the idea.

However I found that it requires a dummy struct for all those types as napi only supports type overrides for function args, function returns and struct fields.

The dummy struct will be like:

```rs
#[napi(object, object_to_js = false)]
pub struct DummyType {
  #[napi(ts_type = "string | ((pathData: JsPathData, assetInfo?: JsAssetInfo) => string)")]
  pub JsFilename: JsFilename
}
```

Which generates:

```ts
export interface DummyType {
  JsFilename: string | ((pathData: JsPathData, assetInfo?: JsAssetInfo) => string)
}
```

When referencing this type, type overriding config will also be added for reference:

```rs
#[napi(object, object_to_js = false)]
pub struct RawCssExtractPluginOption {
  #[napi(ts_type = "DummyType[\"JsFilename\"]")]
  pub filename: JsFilename,
}
```

And this generates:

```ts
export interface RawCssExtractPluginOption {
  filename: DummyType["JsFilename"]
}
```

The cons of this idea are:

1. `#[napi(ts_type = "DummyType[\"JsFilename\"]")]` is still required, and looks ugly
2. more important, I think `JsFilename` and other types like `JsPathData` is of equivalent status, they should be same form in type declaration file but not some `DummyType["XXX"]` while some `YYY`.

Thus i think if we could generate a reusable custom type decleration without a dummy struct, it would be a good. idea.

h-a-n-a: This is indeed a little bit cumbersome to have this manually implemented like this, also it's kind of error prone. I would like to propose some type generation to napi  to generate typings correctly. 

The simple solution might be exposing a macro attr like how it was implemented on the `wasm_bindgen`: https://rustwasm.github.io/docs/wasm-bindgen/reference/attributes/on-rust-exports/typescript_custom_section.html, for example:
```rs
#[napi(typescript_custom_section)]
const TS_APPEND_CONTENT: &'static str = r#"
export type JsCoords = { "latitude": number, "longitude": number, }; 
"#;
```

To make it better to use in most common scenarios, I also would like to add type generation to type alias, for example:
```rs
#[napi(js_name = "JsCoords", ts_type = "{ ... }")]
type Coords = SomeType;
```

Nevertheless, let's merge this first and try to solve this problem in future PRs when the final solution lands.

h-a-n-a:
Thank you

h-a-n-a (approved):

PR #6286, opened 2024-04-18
chore: add jsdom web runner

LingyuCoder:
## Checklist


LingyuCoder:
!test-ng

LingyuCoder:
!test-ng

packages/rspack/tests/hotCases/plugins/banner/test.filter.js
```diff
@@ -0,0 +1 @@
+module.exports = () => false;
```
LingyuCoder, line 1: This case wrote by myself use global directly and will always need to use vm.runInThisContext, disable it and find a better way in other pull request

hardfist (approved):

PR #6375, opened 2024-04-26
feat(copyRspackPlugin): support transform configuration

9aoy:
## Summary

implement the transform option in the CopyRspackPlugin. 

related issue: https://github.com/web-infra-dev/modern.js/issues/5610

## Checklist


h-a-n-a:
It's also a bit of weird that `includeForgottenExports` doesn't bundling the types exported from other packages into the api report file, in that the type `RawCopyRspackPluginOptions` is not being included. This could be related to the limit to the api-extractor as stated in https://github.com/microsoft/rushstack/issues/3430.

crates/rspack_plugin_copy/src/lib.rs
```diff
@@ -244,6 +256,38 @@ impl CopyRspackPlugin {
...
 
+    if let Some(transform) = &pattern.transform {
+      if let Some(absolute_filename) = absolute_filename.to_str() {
+        match transform {
+          Transformer::Fn(transformer) => {
+            let transformed = transformer(
+              match &source {
+                RawSource::Source(code) => code.to_owned(),
+                RawSource::Buffer(buffer) => String::from_utf8(buffer.to_owned())
+                  .unwrap_or_else(|err| panic!("Transform buffer source error: {err}")),
```
h-a-n-a, line 267: What if the `source` is a binary here? IIRC, `copy-webpack-plugin` transforms both `string` and `Buffer`.

h-a-n-a (approved):

Issue #6242, opened 2024-04-15
Improve diagnostics if `builtins.define` is failed to parse


h-a-n-a:
Improve diagnostic for the `builtins.define`. Parse should never end up in a program panic. Check out the original issue down below.

> I've closed this issue for not a bug. Feel free to reopen it if rspack should report a friendly message here to improve dx.

_Originally posted by @xc2 in https://github.com/web-infra-dev/rspack/issues/6227#issuecomment-2055115882_

PR #6382, opened 2024-04-27
feat: improve diagnostics if expressions passed into `DefinePlugin` are failed to parse

shulaoda:
## Summary

#6242 

## Checklist


crates/rspack_swc_visitors/src/define.rs
```diff
@@ -37,10 +36,14 @@ pub fn define(opts: &Define) -> impl Fold {
...
           &mut vec![],
         )
-        .unwrap_or_else(|_| panic!("builtins.define: Failed to parse {:?}", target))
       };
 
-      (NodeIgnoringSpan::owned(*target), *replacement)
+      if let (Ok(target), Ok(replacement)) = (target, replacement) {
+        Some((NodeIgnoringSpan::owned(*target), *replacement))
+      } else {
+        eprintln!("builtins.define: Failed to parse {:?}", raw_target);
```
h-a-n-a, line 44: This would better be a diagnostic, and more specifically it's a warning since I didn't see any error emitted in `webpack.DefinePlugin`(however we use swc parse in rspack, which does not block the build and you just keep the `target` or `replacement` as-is). Leveraging `eprintln` ends up errors not being showed in the `Stats`.

h-a-n-a: `builtins.define` is also a deprecated option, please use `DefinePlugin` instead.

shulaoda: if you set plugin `new rspack.DefinePlugin({ "__DEV__": "😄" })`, then it will cause errors and print logs. 
So according to your idea, should I diagnose it before it compiles and runs?
image

shulaoda: Or conduct a diagnosis before it executes `define`?

h-a-n-a: I suggest to pass [`diagnostic`](https://github.com/web-infra-dev/rspack/blob/0e438151e9742e182bb75c9393c4689d890d975e/crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs#L100) to `define`.

h-a-n-a:
Would you add some snapshots to `rspack-test-tools/tests/diagnostics/module-parse-failed`?

shulaoda:
> Would you add some snapshots to `rspack-test-tools/tests/diagnostics/module-parse-failed`?

done it

crates/rspack_swc_visitors/src/define.rs
```diff
@@ -12,35 +14,52 @@ use swc_core::ecma::visit::Fold;
...
+          // Push the error to diagnostics
+          diagnostics.push(Box::new(
+            TraceableError::from_source_file(
+              &fm,
+              span.lo.0.saturating_sub(1) as usize,
+              span.hi.0.saturating_sub(1) as usize,
+              "DefinePlugin warning".into(),
+              format!("failed to parse {:?}", text),
+            )
+            .with_help(Some("Consider wrapping it with JSON.stringify."))
```
h-a-n-a, line 47:
```suggestion
            .with_help(Some("Consider wrapping it with `JSON.stringify(...)` if a string is expected."))
```

h-a-n-a (approved):

h-a-n-a:
Thank you!

PR #6398, opened 2024-04-29
refactor: diff context module lazy once mode

ahabhgk:
## Checklist


crates/rspack_core/src/options/module.rs
```diff
@@ -74,7 +74,7 @@ impl From<&str> for DynamicImportMode {
       "weak" => DynamicImportMode::Weak,
       "eager" => DynamicImportMode::Eager,
       "lazy" => DynamicImportMode::Lazy,
-      "lazyOnce" => DynamicImportMode::LazyOnce,
+      "lazy-once" => DynamicImportMode::LazyOnce,
```
LingyuCoder, line 77: Is this a breaking change?

ahabhgk: Should be a bugfix

LingyuCoder (approved):

PR #6387, opened 2024-04-28
fix(config): snapshot.module accept `{ timestamp: true}`

SoonIter:
## Summary

close https://github.com/web-infra-dev/rspack/issues/6381

`{ timestamp: true}` should has the same meaning of `{ timestamp: true, hash: false }`

We need to process this default value when we pass it to rust side.

image

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

Issue #6381, opened 2024-04-26
[Bug]: setting snapshot options cause a build error of type AssertionError [ERR_ASSERTION]


pushkargupta9807:
### System Info

  System:
    OS: macOS 14.4.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 26.47 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    Yarn: 1.23.19 - /opt/homebrew/bin/yarn
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
  npmPackages:
    @rspack/cli: 0.6.2 => 0.6.2 
    @rspack/core: 0.6.2 => 0.6.2 

### Details

setting the snapshot config to following :

    snapshot: {
      module: {
        timestamp: true,
      },
      resolve: {
        timestamp: true,
      },
    },

cause the below assertion error. 

`AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  (0, assert_1.default)(!(0, util_1.isNil)(resolveTimestamp) &&
      !(0, util_1.isNil)(resolveHash) &&
      !(0, util_1.isNil)(moduleTimestamp) &&
      !(0, util_1.isNil)(moduleHash))

    at getRawSnapshotOptions (/Users/pushkargupta/source/modular-packages/node_modules/@rspack/core/dist/config/adapter.js:581:26)
    at getRawOptions (/Users/pushkargupta/source/modular-packages/node_modules/@rspack/core/dist/config/adapter.js:34:19)
    at Compiler._Compiler_getInstance (/Users/pushkargupta/source/modular-packages/node_modules/@rspack/core/dist/Compiler.js:484:51) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}

Node.js v20.11.1
✨  Done in 11.63s. Peak memory usage 30.25MB.`

### Reproduce Steps

Please refer to details section.

PR #6399, opened 2024-04-29
fix(config): support snapshot config {timestamp: true} and {hash: true}

SoonIter:
## Summary

#6381  #6387 

`{ hash: true }` `{ timestamp: true }` and `{ timestamp: true, hash: true }`  should be three different behaviors.
I think this is a very misleading configuration.

https://webpack.js.org/configuration/other-options/#builddependencies

image

cases

`{}` and `{ timestamp: false, hash: false }` ->  `{ timestamp: false, hash: false }`
`{ hash: true }` -> `{ timestamp: false, hash: true }`
`{ timestamp: true }` -> `{ timestamp: true, hash: false }`
`{ timestamp: true, hash: true }`

## Checklist

### DIFF
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -191,14 +191,26 @@ const applySnapshotDefaults = (
 	snapshot: SnapshotOptions,
 	{ production }: { production: boolean }
 ) => {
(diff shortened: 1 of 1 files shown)


ahabhgk (approved):
Thanks, bro

PR #6409, opened 2024-04-29
fix: remove outdated TODOs in runtime code

chenjiahan:
## Summary

Remove the TODOs in the runtime code, because:

- The TODO was originally added in 2022, and the current runtime code seems to be the same as webpack, so the TODO should be unnecessary now.
- The runtime code will be bundled with the user code, when users see the TODOs, it does not feel good.

Links:

- Initial PR: https://github.com/web-infra-dev/rspack/commit/f2bb4a8970d91955f3549454a63a396f44ba7ccd#diff-27fe28b51ff18339baefeed38d4aca5712e10ce0b51bf8db459677f2f0a7f5dfR14-R18
- webpack runtime: https://github.com/webpack/webpack/blob/c586c7b1e027e1d252d68b4372f08a9bce40d96c/lib/hmr/HotModuleReplacement.runtime.js#L30-L32

## Checklist


LingyuCoder (approved):

chenjiahan:
I will fix the snapshot tomorrow..

LingyuCoder (approved):

PR #6411, opened 2024-04-29
fix: allow all characters for webpackChunkName

### LINKED ISSUE #6358: [Bug]: webpackChunkName not used if name contains period
### System Info

`rspack` v0.6.3

### Details

The magic comment in this example will be respected by Webpack, but not by `rspack`
```js
import(/* webpackChunkName: "component.Mod" */ './mod');
```

See linked repro for full example

### Reproduce link

https://github.com/JohnDaly/rspack-magic-comment-repro

### Reproduce Steps

1. `npm run dev`
2. `npm run build`
3. Notice that the assets generated by Webpack use the `webpackChunkName`. The assets generated by `rspack`  do not

xc2:
## Summary

Webpack allows any string for webpackChunkName: https://github.com/webpack/webpack/blob/c586c7b1e027e1d252d68b4372f08a9bce40d96c/lib/javascript/JavascriptParser.js#L4584-L4587

fixes #6358

Just a quick fix.

### Known issues (Edge cases)

- will not work with texts containing the character that is same as the quote character, e.g, `"a\\"b"` (rust's regex does not support backreference)
- still does not support all js expressions like `Boolean(1)` which seems to be allowed in webpack

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/webpack_comment.rs
+++ b/crates/rspack_plugin_javascript/src/webpack_comment.rs
@@ -88,7 +88,7 @@ fn add_magic_comment_warning(
 // _5 for true/false
 // TODO: regexp/array
 static WEBPACK_MAGIC_COMMENT_REGEXP: Lazy<regex::Regex> = Lazy::new(|| {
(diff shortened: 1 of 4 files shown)


ahabhgk (approved):
I am contemplating the use of JSON for data parsing (JSON based + Regex support for webpackInclude/Exclude, or just a fast JS parser). While it has its drawbacks, it should be adequate for the majority of use cases.

PR #6431, opened 2024-05-01
perf: improve determine_export_assignments

quininer:
## Summary

This is an optimization for `TypeScript` repo build hotspot. I did some performance tests before the 5-1, but now the results have been destroyed along with harddisk.

* Use .chain instead of .clone + .push
* eliminate Atom drop by avoiding redundant Atom clone
* eliminate Atom clone
* rewrite collect hidden_exports

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
+++ b/crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
@@ -307,11 +307,13 @@ impl HarmonyExportImportedSpecifierDependency {
     let hidden_exports = self
       .discover_active_exports_from_other_star_exports(module_graph)
       .map(|other_star_exports| {
-        other_star_exports.names.into_iter()
-            .take(other_star_exports.names_slice)
-            .filter(|name| !ignored_exports.contains(name))
-            .cloned()
-            .collect::<HashSet<_>>()
+        other_star_exports
+          .names
+          .into_iter()
+          .take(other_star_exports.names_slice)
+          .filter(|name| !ignored_exports.contains(name))
+          .cloned()
+          .collect::<HashSet<_>>()
       });
     if !no_extra_exports && !no_extra_imports {
       return StarReexportsInfo {
@@ -1355,7 +1357,7 @@ fn determine_export_assignments<'a>(
   let mut names = Vec::new();
   let mut hints = HashSet::default();
   let mut dependency_indices =
-      Vec::with_capacity(dependencies.len() + additional_dependency.is_some() as usize);
+    Vec::with_capacity(dependencies.len() + additional_dependency.is_some() as usize);
 
   for dependency in dependencies.iter().chain(additional_dependency.iter()) {
     if let Some(module_identifier) = module_graph.module_identifier_by_dependency_id(dependency) {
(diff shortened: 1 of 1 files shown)


crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
```diff
@@ -1350,39 +1345,36 @@ pub struct StarReexportsInfo {
...
-  for dependency in dependencies.iter() {
-    dependency_indices.push(names.len());
+  for dependency in dependencies.iter().chain(additional_dependency.iter()) {
     if let Some(module_identifier) = module_graph.module_identifier_by_dependency_id(dependency) {
       let exports_info = module_graph.get_exports_info(module_identifier);
       for export_info_id in exports_info.exports.values() {
         let export_info = module_graph.get_export_info_by_id(export_info_id);
         // SAFETY: This is safe because a real export can't export empty string
-        let export_info_name = export_info.name.clone().unwrap_or_default();
+        let export_info_name = export_info.name.as_ref().expect("export name is empty");
```
quininer, line 1366: I'm not sure we need to be consistent with original behavior.

crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
```diff
@@ -1350,39 +1345,36 @@ pub struct StarReexportsInfo {
...
-  if let Some(additional_dependency) = additional_dependency {
-    dependencies.push(additional_dependency);
-  }
-
+) -> (Vec<&'a Atom>, Vec<usize>) {
   // https://github.com/webpack/webpack/blob/ac7e531436b0d47cd88451f497cdfd0dad41535d/lib/dependencies/HarmonyExportImportedSpecifierDependency.js#L109
-  // js `Set` keep the insertion order, use `IndexSet` to align there behavior
-  let mut names: IndexSet<Atom, BuildHasherDefault<FxHasher>> = IndexSet::default();
-  let mut dependency_indices = vec![];
+  // js `Set` keep the insertion order
```
quininer, line 1354: currently consumer do not seem to rely on insert order.

quininer:
!bench

quininer:
!bench

crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
```diff
@@ -1350,39 +1349,34 @@ pub struct StarReexportsInfo {
...
+          && !names.contains(export_info_name)
         {
-          names.insert(export_info_name.clone());
-          let cur_len = dependency_indices.len();
-          dependency_indices[cur_len - 1] = names.len();
+          names.insert(export_info_name);
         }
       }
     }
+    dependency_indices.push(names.len());
```
JSerFeng, line 1378: I think this is not the same behavior as before right ?

crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_imported_specifier_dependency.rs
quininer: We can't use Cow, but I can use static empty Atom instead. but it would be better if it could be guaranteed not to be None.

JSerFeng: Expect other_export_info, I don't see any where that export_info.name could be None, so it looks ok to me

JSerFeng (approved):

PR #6434, opened 2024-05-02
fix(rspack_plugin_mf): add chunk js matcher runtime to federation runtime

ScriptedAlchemy:
## Summary

Exposes internal chunk matcher that is used for ensureChunkHandlers. This allows one to patch js chunk handlers at runtime using the same logic that rspack uses internally. (chunk_matcher.render('chunkid')) 

This allows me to reliably implement node federation support without adding a new chunk loading target in rust. 

Original implementation: https://github.com/module-federation/core/blob/main/packages/enhanced/src/lib/container/runtime/FederationRuntimeModule.ts#L53

## Checklist


crates/rspack_plugin_mf/src/container/module_federation_runtime_plugin.rs
```diff
@@ -41,20 +42,46 @@ impl RuntimeModule for FederationRuntimeModule {
...
+fn federation_runtime_template(chunk: &Chunk, compilation: &Compilation) -> String {
   let federation_global = format!("{}.federation", RuntimeGlobals::REQUIRE);
+
+  let condition_map =
+    compilation
+      .chunk_graph
+      .get_chunk_condition_map(&chunk.ukey, compilation, chunk_has_js);
+  let has_js_matcher = compile_boolean_matcher(&condition_map);
+
+  let chunk_matcher = if matches!(has_js_matcher, BooleanMatcher::Condition(true)) {
```
ScriptedAlchemy, line 62: Allows me to implement custom chunk handlers in JS runtime. This was the final part - currently i cannot tell if a chunk is actual js or an internal/virtual chunk. So my chunk handler has to pass a empty chunk object callback if it doesnt know what it is, this is error prone since it may be a real error that i swallow. 

this passes the chunk match logic to a function i can access off webpack require to validate chunk ids are js chunks.

ahabhgk:
Could you add a test under `packages/rspack-test-tools/tests/container-1-5`, just in case of regression

ScriptedAlchemy:
> Could you add a test under `packages/rspack-test-tools/tests/container-1-5`, just in case of regression

can do!

ahabhgk (approved):

PR #6443, opened 2024-05-06
perf: on-demand sourcemap generation

h-a-n-a:
## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/lib.rs
+++ b/crates/rspack_loader_swc/src/lib.rs
@@ -80,9 +80,9 @@ impl Loader<LoaderRunnerContext> for SwcLoader {
       swc_options
     };
 
-    let source_map_kind: &SourceMapKind = match swc_options.config.source_maps {
-      Some(SourceMapsConfig::Bool(false)) => &SourceMapKind::None,
-      _ => &loader_context.context.module_source_map_kind,
+    let source_map_kind: SourceMapKind = match swc_options.config.source_maps {
+      Some(SourceMapsConfig::Bool(false)) => SourceMapKind::empty(),
+      _ => loader_context.context.module_source_map_kind,
     };
 
     let source = content.try_into_string()?;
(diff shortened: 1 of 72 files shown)


h-a-n-a:
!bench

h-a-n-a:
!bench

h-a-n-a:
!bench

crates/rspack_loader_swc/src/lib.rs
```diff
@@ -122,9 +122,9 @@ impl Loader<LoaderRunnerContext> for SwcLoader {
         .as_ref()
         .map(|v| matches!(v, OutputCharset::Ascii)),
       source_map_config: SourceMapConfig {
-        enable: !matches!(source_map_kind, SourceMapKind::None),
-        inline_sources_content: true,
-        emit_columns: matches!(source_map_kind, SourceMapKind::SourceMap),
+        enable: source_map_kind.source_map(),
```
h-a-n-a, line 125: Only generate sourcemap if it's `module`.

crates/rspack_loader_swc/src/lib.rs
```diff
@@ -122,9 +122,9 @@ impl Loader<LoaderRunnerContext> for SwcLoader {
...
+        inline_sources_content: source_map_kind.source_map(),
+        emit_columns: !source_map_kind.cheap(),
```
h-a-n-a, line 127: Do not emit columns if it's `cheap`.

h-a-n-a:
!bench

SyMind (approved):

pushkargupta9807:
can we please have more info on this perf improvement PR  ? For instance how can we use it , examples etc

chenjiahan:
@h-a-n-a please check this ⬆️

h-a-n-a:
> can we please have more info on this perf improvement PR ? For instance how can we use it , examples etc

This is an internal optimisation where source maps should always be generated on-demand by `builtin:swc-loader` with the respect to the compiler's source map options. I didn't make a real world benchmark, but this will improve the case where `options.devtool` set to `false`.

PR #6444, opened 2024-05-06
fix: consistent md4 hash in localNameIdent

### LINKED ISSUE #6435: [Bug]: [hash] for css modules (experiments.css) doesn't match Webpack
### System Info

System:
    OS: macOS 14.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 105.77 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.2 - ~/.asdf/installs/nodejs/20.12.2/bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.5.0 - ~/.asdf/plugins/nodejs/shims/npm
    Watchman: 2024.01.22.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 124.0.6367.94
    Safari: 17.3
  npmPackages:
    @rspack/cli: latest => 0.6.1
    @rspack/core: latest => 0.6.1

### Details

Using the new experiments.css approach (generator) for css modules generates different hashes to Webpack as can be seen on the below screenshot (left Rspack, right Webpack), this was reproduced by using the https://github.com/rspack-contrib/rspack-examples/tree/main/rspack/css-parser-generator-options example and adding an identical `weback.config.js` (only changing Rspack imports to Webpack equivalents).

Screenshot 2024-05-02 at 10 11 29

This isn't particularly impacting me, but discovered when trying to generate a mappings json file of the classes using `postcss`/`postcss-modules` (our existing solution) and getting different hashes to the ones produced by Rspack. They are also different to the Webpack generated ones too, and am struggling to see how both Rspack and Webpack are generating these hashes (as don't seem to just be a simple md4 hash like you'd be able to regenerate using https://github.com/webpack/loader-utils for example). 

Any help understanding how these are generated and/or how I could generate matching mapping files would be appreciated. 

### Reproduce link

_No response_

### Reproduce Steps

1. Clone https://github.com/rspack-contrib/rspack-examples/tree/main/rspack/css-parser-generator-options
2. Update the css modules `generator.localIdentName` to `[hash]-[local]` (anything containing `[hash]` will do)
3. Duplicate `rspack.config.js` as `webpack.config.js` and update the imports (Rspack & HtmlRspackPlugin) to Webpack equivalents 
4. Run each and take note of the generated hash

LingyuCoder:
## Summary

Fix #6435
- HashSalt=None should not be hashed
- Use module.context when hash the localNameIdent, align with webpack
- Add `./` to relative path, align with webpack
- Can not use `xxx.hash(hasher)` because it add length to hash, use `hasher.write(bytes)` directly.

## Checklist


crates/rspack_hash/src/lib.rs
```diff
@@ -78,7 +78,9 @@ impl RspackHash {
 
   pub fn with_salt(function: &HashFunction, salt: &HashSalt) -> Self {
     let mut this = Self::new(function);
-    salt.hash(&mut this);
+    if !matches!(salt, HashSalt::None) {
```
LingyuCoder, line 81: This line makes lots of snapshot changes, but `salt=None` should not be hashed

ahabhgk (approved):

Issue #5097, opened 2023-12-22
remove Rspack deps to reduce Rspack install size


hardfist:
people care about install size and Rspack | Rsbuild install size is much bigger than Vite or other bundlers, even we can't make our install size the same as Vite, we should try our best to reduce install size
here are some deps we can remove

- [ ] browserslist(2.22Mb): no need when we remove builtins.presetEnv
- [x] compare-versions(59kb): no necessary, we only needs limited functionality, handwritten function is enough
- [ ] enhanced-resolve(260kb): no need when we support using native-loader
- [ ] graceful-fs(31kb): no need 
- [ ] noe-async(291kb): no need, handwritten async workflow
- [ ] webpack-sources(90kb): seems we only need type?
- [x] terminal-link(180kb):  no need, handwritten color 
- [ ] zod(613kb): very large, but we need this functionality, better alternative?
- [ ] zod-validation-error(643kb): may include zod size again? very large
- [x] events(82kb): can be handwritten

xc2:
just tried `config/zod.ts` -> json schema -> ajv standalone code, resulting in ~500k mangled js code, whose size is very near zod.

```js
import { zodToJsonSchema } from "zod-to-json-schema";
import Ajv from "ajv";
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);
const standaloneCode = require("ajv/dist/standalone");
const { rspackOptions } = await import("@rspack/core/zod"); // exports hacked

const jsonSchema = zodToJsonSchema(rspackOptions, "rspackOptions");

const ajv = new Ajv({ code: { source: true, esm: true, es5: false } });
const validate = ajv.compile(jsonSchema);
const moduleCode = standaloneCode(ajv, validate);

console.log(moduleCode.length);
```

tatchi:
I've heard about https://github.com/fabian-hiller/valibot which apparently could be a smaller alternative to zod

PR #6462, opened 2024-05-07
perf(core): use prebundle to reduce dependencies

chenjiahan:
## Summary

Use prebundle to reduce dependencies of `@rspack/core` (12 deps -> 6 deps)

Screenshot 2024-05-07 at 14 16 11

How it works:

- Use [prebundle](https://github.com/rspack-contrib/prebundle) to bundle dependencies and output to the `compiled` folder.
- Use [tsc-alias](https://www.npmjs.com/package/tsc-alias) to alias the import path to the `compiled` folder.

resolve https://github.com/web-infra-dev/rspack/issues/5097

## Checklist


LingyuCoder (approved):

chenjiahan:
!canary

alfaproject:
How is this better? Now I have duplicate code in my node modules, ehhh

chenjiahan:
Bundle dependency is a very practical optimization method, and we can see that projects like Next.js and Vite have adopted it.

- Next.js bundles dependencies and output to the `compiled` folder, just like Rspack does. see https://github.com/vercel/next.js/tree/canary/packages/next/src/compiled

- Vite bundles most of dependencies and outputs them as chunks, see: https://github.com/vitejs/vite/blob/main/packages/vite/rollup.config.ts#L65-L73

There are many benefits of bundle dependencies, including faster dependency installation and prevention of supply chain attacks caused by upgrades of indirect dependencies.

Regarding your concerns, while there may be duplicate code in node_modules, in most cases the code you need to download is less than before, because the bundling process removes unnecessary files.

chenjiahan:
According to my calculations, this PR has reduced the install size by 800kB, and we can further optimize it.

PR #5075, opened 2023-12-20
chore: force build JS packages on CI

h-a-n-a:
## Summary

Try to fix https://github.com/web-infra-dev/rspack/actions/runs/7243540726/job/19730646297

## Require Documentation?

- [ ] No
- [ ] Yes, the corresponding rspack-website PR is \_\_

PR #6465, opened 2024-05-07
fix: remove --force flag to fix prebundle

chenjiahan:
## Summary

Remove --force flag to fix prebundle.

The `--force` was introduced in https://github.com/web-infra-dev/rspack/pull/5075, and it doesn't seem necessary.

When using `--force` option, the `@rspack/core` package will be built multiple times:

- the first time:  run build script (`tsc` + `tsc-alias`), the outputs are correct.
- the second time: only run `tsc` (triggered by `tsc -b --force` of other packages), the outputs are incorrect because the `tsc-alias` is skipped.

So we should remove the `--force` flag to fix this.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "x": "zx x.mjs",
     "dev": "pnpm --filter @rspack/cli run dev",
     "clean": "pnpm --filter @rspack/cli run clean",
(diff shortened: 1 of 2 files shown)


chenjiahan:
!eco-ci

LingyuCoder (approved):

PR #6508, opened 2024-05-10
fix: make code-splitting deterministic

JSerFeng:
## Summary

Rspack offers 2 approaches for handling hot module replacement (HMR):
1. Fast path: Only modifies content without changing dependencies, allowing for the reuse of the chunk graph.
2. Slow path: Module dependencies or blocks have changed, requiring regeneration of the chunk graph.

Currently, we encounter issues with both scenarios:

1. Users may trigger the slow path even when not altering dependencies (e.g., adding a div in a React app). This occurs because we detect changes in dependencies; for instance, adding a new ImportSpecifierDependency triggers the slow path.

2. When encountering the slow path, the generated chunk graph differs from the original one.

In certain instances, the chunk graph may vary between builds even if there are no modifications to the module graph.
## What does this PR fix

make code-splitting deterministic, use IndexMap to preserve the iter order

fix `has_module_graph_change`, currently it cannot detect the change when dependencies' order changes

Align exports_info.set_unknown_target with webpack

Use IndexMap to make order of iterating exports deterministic

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/hmr.rs
+++ b/crates/rspack_core/src/compiler/hmr.rs
@@ -92,8 +92,6 @@ where
         self
           .compilation
           .swap_make_artifact_with_compilation(&mut new_compilation);
-        new_compilation.entries = std::mem::take(&mut self.compilation.entries);
-        new_compilation.global_entry = std::mem::take(&mut self.compilation.global_entry);
         new_compilation.lazy_visit_modules =
           std::mem::take(&mut self.compilation.lazy_visit_modules);
         new_compilation.file_dependencies = std::mem::take(&mut self.compilation.file_dependencies);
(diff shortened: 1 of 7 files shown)


JSerFeng:
!bench

hardfist:
seems need a deterministic test?

JSerFeng:
> seems need a deterministic test?

This is a partial fix, split_chunks still generates inconsistent output between 2 builds in this pr sometimes, but I'm not able to make a minimal repro.

There are more fix soon

JSerFeng:
!bench

JSerFeng:
!bench

crates/rspack_core/src/compiler/hmr.rs
```diff
@@ -92,8 +92,6 @@ where
         self
           .compilation
           .swap_make_artifact_with_compilation(&mut new_compilation);
-        new_compilation.entries = std::mem::take(&mut self.compilation.entries);
```
LingyuCoder, line 95: why entries need to be removed

ahabhgk (approved):

crates/rspack_core/src/compiler/hmr.rs
JSerFeng: Because EntryPlugin will insert entries in hmr, causing 2 duplicated entries

Issue #6513, opened 2024-05-11
[Bug]: `import "binaryen"` throws error "'import.meta' cannot be used outside of module code" when minimize is on


xc2:
### System Info

```
  System:
    OS: macOS 14.5
    CPU: (12) arm64 Apple M2 Max
    Memory: 97.03 MB / 32.00 GB
    Shell: 3.7.1 - /opt/homebrew/bin/fish
  Binaries:
    Node: 20.12.2 - ~/.managed/n/bin/node
    Yarn: 1.22.21 - ~/.managed/n/bin/yarn
    npm: 10.5.0 - ~/.managed/n/bin/npm
    pnpm: 8.15.4 - ~/.managed/n/bin/pnpm
    bun: 1.0.34 - /opt/homebrew/bin/bun
    Watchman: 2024.03.18.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 124.0.6367.158
    Safari: 17.5
  npmPackages:
    @rspack/cli: ^0.6.3 => 0.6.5 
    @rspack/core: ^0.6.3 => 0.6.5 
```

### Code

```js
import a from 'binaryen';

console.log(a);
```

### Configuration

```js
module.exports = {
  mode: "none",
  optimization: {
    // when minimize is on, importing binaryen will produce error "'import.meta' cannot be used outside of module code"
    minimize: true,
  },
}
```

### Got

```

> rspack-repro@1.0.0 build:rspack /Users/kfll/Developer/rspack-repro-binaryen
> rspack -c rspack.config.js

ERROR in × JavaScript parsing error: 'import.meta' cannot be used outside of module code.
   ╭─[4:1]
 4 │  * SPDX-License-Identifier: Apache-2.0
 5 │  */
 6 │ var JC=(()=>{var vA=import.meta.url;return async fun...

```

### Expected

There's `type:"module"` in `binaryen/package.json` so it should be considered as module code

### More

This only happens when minimize is on

### Reproduce link

https://github.com/xc2/rspack-repro-binaryen

### Reproduce Steps

```shell
pnpm i
pnpm build:rspack | less
```

h-a-n-a:
Try to enable `module` in SWC js minifier plugin.

PR #6521, opened 2024-05-13
fix(minimizer): resolve error in minimizing esm asset file

xc2:
## Summary

let swc minimizer determine the module type if it is unknown - align with webpack and terser-webpack-plugin.

Fixes: #6513 

## Checklist

- [x] Documentation updated (or not required) - No need.


h-a-n-a (approved):
Thank you!

PR #6561, opened 2024-05-16
test: move normal case with rspack.config.js

LingyuCoder:
## Summary

- Move normal cases with rspack.config.js to config cases. Normal cases should not use config file.

## Checklist


packages/rspack-test-tools/tests/configCases/resolve/axios-0.27.2/rspack.config.js
```diff
@@ -1,4 +1,6 @@
 /** @type {import("@rspack/core").Configuration} */
 module.exports = {
-	target: "web"
+	resolve: {
+		aliasFields: ["browser"]
```
LingyuCoder, line 4: Normal cases has this aliasFields and no target by default, this file is not used before

h-a-n-a (approved):

Issue #6306, opened 2024-04-20
[Bug]: ts-loader runs into an error with rspack


pushkargupta9807:
### System Info

System:
OS: Linux 6.2 Ubuntu 20.04.6 LTS (Focal Fossa)
CPU: (32) x64 AMD EPYC 7763 64-Core Processor
Memory: 113.67 GB / 125.78 GB
Container: Yes
Shell: 5.0.17 - /bin/bash
Binaries:
Node: 20.11.0 - /usr/local/share/nvm/versions/node/v20.11.0/bin/node
Yarn: 1.23.19 - /usr/bin/yarn
npm: 10.2.4 - /usr/local/share/nvm/versions/node/v20.11.0/bin/npm
Watchman: 20240317.093234.0 - /usr/local/bin/watchman
Browsers:
Chrome: 124.0.6367.60
npmPackages:
@rspack/cli: 0.6.2 => 0.6.2
@rspack/core: 0.6.2 => 0.6.2

### Details

So running into the below error when using ts-loader for some files. (for other files I am using builtin:swc-loader) . 

```
ERROR in ../../someFile.graphql.ts
  × Module build failed:
  ╰─▶   × TypeError: Cannot read properties of undefined (reading 'buildMeta')
        │     at setModuleMeta (/workspaces/teams-modular-packages/node_modules/ts-loader/dist/index.js:62:31)
        │     at makeSourceMapAndFinish (/workspaces/teams-modular-packages/node_modules/ts-loader/dist/index.js:56:5)
        │     at successLoader (/workspaces/teams-modular-packages/node_modules/ts-loader/dist/index.js:39:5)
```
And this error is coming from the following piece of code in ts-loader:

```
function setModuleMeta(loaderContext, instance, fileVersion) {
    // _module.meta is not available inside happypack
    if (!instance.loaderOptions.happyPackMode &&
        loaderContext._module.buildMeta !== undefined) {
        // Make sure webpack is aware that even though the emitted JavaScript may be the same as
        // a previously cached version the TypeScript may be different and therefore should be
        // treated as new
        loaderContext._module.buildMeta.tsLoaderFileVersion = fileVersion;
    }
}
```
I do not run into this error with webpack. 
Currently I am getting around this issue with setting happyPackMode as true for rspack. 

### Reproduce Steps

try using ts-loader for some .tsx. file -> should repro this error.

h-a-n-a:
`LoaderContext` does not support `_module` yet, but PR is welcome

[...]

Issue #5766, opened 2024-02-25
[Bug]: support svg-chunk-webpack-plugin


singh-jay:
### System Info

System:
    OS: macOS 14.3.1
    CPU: (8) arm64 Apple M1
    Memory: 43.17 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.5.0 - /opt/homebrew/bin/node
    npm: 10.2.4 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 121.1.62.165
    Chrome: 119.0.6045.199
    Safari: 17.3.1

### Details

I'm trying to migrate our projects to rspack, everything is working fine except the svg sprites. Tried different plugins but got a similar error. I tried to find a solution to use svg sprites but could not, any guide on how to use svg-sprite with rspack will be helpful.

Getting error -
/node_modules/svg-chunk-webpack-plugin/lib/loader.js:20
this._module.factoryMeta = this._module.factoryMeta || {};
^

TypeError: Cannot read properties of undefined (reading 'factoryMeta')
at Object.SvgChunkWebpackLoader

### Reproduce link

https://github.com/singh-jay/svg-chunk-plugin-test

### Reproduce Steps

1. clone the repo
2. run bun install
3. run bun dev

LingyuCoder:
Rspack does not support _module and factoryMeta yet, if you are willing, welcome to contribute

PR #6631, opened 2024-05-27
feat: support `_module` of LoaderContext

CPunisher:
## Summary

Related: https://github.com/web-infra-dev/rspack/issues/5766, https://github.com/web-infra-dev/rspack/issues/6306

Clone some exposable properties from the `Module` into the `CompilerModuleContext`, then convert it into a `JsModule` and pass it into the `LoaderContext`. For fields dynamically added by plugins or loaders, they are only recorded in the JS side of the `Compilation` and associated with the `ModuleIdentifier`.

However, this is still not enough to completely resolve the aforementioned two issues. Based on my tests, there are additional compatibility problems. For example:

1. ts-loader modifies `compilation.errors`, which is not supported yet.
2. The svg plugin calls `compilation.chunkGraph.getOrderedChunkModulesIterable`, which is not supported yet.
3. ~~The source of imported svg files includes semicolons in the result, causing the svg plugin to fail to parse~~ (~~this occurs during the build's parse phase and seems to be a bug~~ A svg module with no type set will be recognized as “javascript/auto” and will trigger rspack inner formatter.).
4. Overwriting the `_module` fields does not write back to the rust, including modifications to buildInfo, buildMeta, and factoryMeta. This might affect https://github.com/webpack/webpack/issues/12202#issuecomment-745537821.

## Checklist


LingyuCoder:
Thanks for your contribution. Perhaps need to add some test cases and flush the snapshot. Please refer to [this](https://www.rspack.dev/contribute/development/testing-rspack) for details

packages/rspack/src/Module.ts
```diff
@@ -48,26 +49,33 @@ export class Module {
...
 
-	get resource(): string | undefined {
-		return this.#inner.resource;
+		const customModule = compilation?.getCustomModule(module.moduleIdentifier);
+		this.buildInfo = customModule?.buildInfo || {};
+		this.buildMeta = customModule?.buildMeta || {};
 	}
 
-	get originalSource(): Source | null {
+	originalSource(): Source | null {
```
LingyuCoder, line 78: seems to be a breaking change to plugin developer, cc @h-a-n-a

CPunisher: I think rspack makes the mistake. See https://github.com/webpack/webpack/blob/ac8e83a4a2984ed977fd6c23a609da5cb43da492/lib/NormalModule.js#L1388

packages/rspack/src/Compilation.ts
```diff
@@ -354,6 +363,17 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 		);
 	}
 
+	getCustomModule(moduleIdentifier: string) {
```
h-a-n-a, line 366: This should be private

packages/rspack/src/Module.ts
```diff
@@ -48,26 +49,33 @@ export class Module {
...
 
+	context?: string;
+	resource?: string;
+	request?: string;
+	userRequest?: string;
 	rawRequest?: string;
 
-	static __from_binding(module: JsModule) {
-		return new Module(module);
+	buildInfo: Record<string, any>;
```
h-a-n-a, line 58: Please add some explanation to the mutability of the field. It looks like this does not affect the rust side right?

packages/rspack/src/Module.ts
```diff
@@ -48,26 +49,33 @@ export class Module {
...
+	buildMeta: Record<string, any>;
```
h-a-n-a, line 59: Add some comment as well

packages/rspack/src/Module.ts
CPunisher: Yes. I have observed that when plugins or loaders dynamically add fields to `buildInfo` or `buildMeta`, these fields are primarily used as markers and are utilized later on the JavaScript side. Therefore, there is no need to pass them back to Rust. 
For the fields that `buildInfo` inherently has and that are related to the build, perhaps they can be exposed and written back in the future version. And by using a Proxy to combine the two, this could be transparent to plugin developers.

packages/rspack/src/Compilation.ts
CPunisher: This is called in Module's constructor. I change it to `__getCustomModule` to mark it as internal method.

crates/node_binding/binding.d.ts
```diff
@@ -377,7 +377,7 @@ export interface JsLoaderContext {
    * @internal
    */
   diagnosticsExternal: ExternalObject<'Diagnostic[]'>
-  _moduleIdentifier: string
```
h-a-n-a, line 380: Can we keep this field to the next minor/major version to avoid possible breaking changes?

packages/rspack/src/Compilation.ts
```diff
@@ -362,6 +375,17 @@ BREAKING CHANGE: Asset processing hooks in Compilation has been merged into a si
 		);
 	}
 
+	__getCustomModule(moduleIdentifier: string) {
```
h-a-n-a, line 378: prefix it with `__internal__`

h-a-n-a: Add comment: 
```
/**
* Note: This is not a webpack public API, maybe removed in future.
*
* @internal
*/
```

h-a-n-a:
Thank you!

PR #6674, opened 2024-05-30
feat: lazy compilation ignore global entry

JSerFeng:
## Summary

lazy compilation ignore global entry, so that react-refresh entry and hmr entry can be applied correctly

## Checklist


SyMind (approved):

packages/playground/playwright.config.ts
```diff
@@ -5,7 +5,7 @@ const TIMEOUT = 2 * 60 * 1000;
 
 export default defineConfig<RspackOptions>({
 	// Look for test files in the "fixtures" directory, relative to this configuration file.
-	testDir: "./cases",
```
ahabhgk, line 8: 😅

JSerFeng: 😖

Issue #4823, opened 2023-11-30
optimization.chunkIds supports 'natural'


hardfist:
### What problem does this feature solve?

align with webpack's default value   https://github.com/webpack/webpack/blob/main/lib/config/defaults.js#L1319

### What does the proposed API of configuration look like?

same as webpack

hardfist:
this is low priority and seldom used

LingyuCoder:
been support by #6685

PR #6685, opened 2024-05-30
feat: add `optimization.moduleIds: 'natural'`

colinaaa:
## Summary

Add [`optimization.moduleIds: 'natural'`](https://webpack.js.org/configuration/optimization/#optimizationmoduleids).

Also, set the default value to `'natural'` when `mode === 'none'`.

> Maybe this is a breaking change?

## Checklist


packages/rspack/src/config/defaults.ts
```diff
@@ -893,9 +893,10 @@ const applyOptimizationDefaults = (
 	D(optimization, "removeAvailableModules", true);
 	D(optimization, "removeEmptyChunks", true);
 	D(optimization, "mergeDuplicateChunks", true);
-	F(optimization, "moduleIds", (): "named" | "deterministic" => {
+	F(optimization, "moduleIds", (): "named" | "natural" | "deterministic" => {
 		if (production) return "deterministic";
-		return "named";
+		if (development) return "named";
+		return "natural";
```
LingyuCoder, line 899: It is a breaking change, should implement "moduleIds=natural" without changing the default value. We will change it in next minor version.

chenjiahan:
Related issue: https://github.com/web-infra-dev/rspack/issues/4823

packages/rspack/src/config/defaults.ts
colinaaa: The changes to defaults have been reverted. Also added a config test. PTAL~

LingyuCoder (approved):

PR #6686, opened 2024-05-30
docs: update description for module  config

leimonio:
## Summary

Improve documentation for `module.generator["css/auto"].exportsOnly` rule

## Checklist


website/docs/en/config/module.mdx
```diff
@@ -532,7 +532,8 @@ Customize how CSS export names are exported to javascript modules, such as keepi
 - **Type:** `boolean`
 - **Default:** `true` for node environments, `false` for web environments.
 
-Avoid generating and loading a stylesheet and only embed exports from CSS into output javascript files.
+The default behaviour is to generate a CSS file for the stylesheets, and load this file as an external resource in the html `<head>` of the template.
+You can override this behaviour, by embedding exported CSS into the bundled javascript file.
```
chenjiahan, line 536: I think `exportsOnly` is the same as [css-loader's exportOnlyLocals](https://github.com/webpack-contrib/css-loader?tab=readme-ov-file#exportonlylocals).

If `exportsOnly` is enabled,  it doesn't embed CSS but only exports the identifier mappings. We can document this.

leimonio: You're right, I'll update this in the docs.

leimonio: One more thing I noticed is that, when this option is `true` it will generate the `css` file and then inject in the html template. Is this also a responsibility of the css loader (it's not clear from the docs there), or is there any internal rspack logic to also add this functionality?

leimonio:
@chenjiahan can you please have another look now?

chenjiahan (approved):

PR #6697, opened 2024-05-31
feat: root module is less prone to be wrapped in IIFE

fi3ework:
## Summary

Sync up porting https://github.com/webpack/webpack/pull/18348 and https://github.com/webpack/webpack/pull/18349.

Main changes:
- Extract `find_new_name` from method to function
- Remove conflicts between through variables in non-inlined modules and module scope inlined modules which will be accessible from non-inlined modules after module renaming. The renaming  logic is heavily based on the existing concatenated methods. 

## Checklist


fi3ework:
!canary

crates/rspack_plugin_javascript/src/plugin/mod.rs
```diff
@@ -711,6 +750,189 @@ impl JsPlugin {
...
+      else {
+        continue;
+      };
+
+      let code = rendered_module;
+      let cm: Arc<swc_core::common::SourceMap> = Default::default();
+      let fm = cm.new_source_file(
+        FileName::Custom(m.identifier().to_string()),
+        code.source().to_string(),
+      );
```
fi3ework, line 780: not sure if it's the right to declare a fm

ahabhgk:
!bench

ahabhgk:
Generally looks good to me, the benchmark shows in an actual case (arco-pro) this is acceptable, but performance has regressed significantly when there are lots of javascript module (threejs 10x), can this be further improved?

ahabhgk:
There isn't a way to get the AST from the module, we don't store the AST on the module, we analyze dependencies from AST (for javascript) then drop it, especially you need to parse the generated code of the module (source has been modified from the original source, so does AST)

ahabhgk:
For caching, something like https://github.com/webpack/webpack/blob/34e2561addb0f65a7a6fb0ce7ae1aea4cd1d599f/lib/javascript/JavascriptModulesPlugin.js#L592 but works for caching the inline module result looks good to me

fi3ework:
!bench

fi3ework:
@ahabhgk 
The benchmark got better, especially in HRM case. I tested it locally, the upper is with cache and below without. Seen that following HMR speed got faster as cache hits.

image

crates/rspack_plugin_javascript/src/plugin/mod.rs
```diff
@@ -39,9 +52,60 @@ use crate::runtime::{
...
 
+#[derive(Debug, Clone)]
+struct WithCode<T> {
+  code: String,
+  value: T,
+}
+
+#[derive(Debug, Default)]
+struct RenameModuleCache {
+  inlined_modules_to_info: FxDashMap<String, WithCode<InlinedModuleInfo>>,
```
ahabhgk, line 63: Let's use `module.build_info.hash` as the key here instead of module identifier, so the cache can invalidate when the original source of the module changes

ahabhgk: `(module.identifier, module.build_info.hash)`

fi3ework: If we using `module.build_info.hash` here, how to clear the outdated cache of a certain module which could be edited by user for many times.

plugin-test/css-extract/cases/new-url/expected/09a1a1112c577c279435.png
ahabhgk, line 1: Why this png -58.6KB, is this expected?

crates/rspack_plugin_javascript/src/plugin/mod.rs
ahabhgk: I see.. there is `code.source() == ident_info_with_code.code` to do the invalidate, LGTM

fi3ework: Could we compare `build_info.hash` instead of source code? Seems like the overhead is cost less, the source code could be giant.

plugin-test/css-extract/cases/new-url/expected/09a1a1112c577c279435.png
fi3ework: Weird, I did't expect to change it and it's unrelated. Checked the history of this file https://github.com/web-infra-dev/rspack/commits/2adc1b87ddaf70a73e46db30e451013e5ce10b5c/plugin-test/css-extract/cases/new-url/expected/09a1a1112c577c279435.png, I decide to revert it to the original version as current image on `main` is broken.

fi3ework: it's already the original version now.

crates/rspack_plugin_javascript/src/plugin/mod.rs
fi3ework: Done, use hash instead of source code to compare.

ahabhgk (dismissed):

ahabhgk:
Something wrong in our CI, @SyMind is looking at it

fi3ework:
Patched a commit lost in force-pushing that fix the wrong usage of ReplaceSource, test cases're updated alongside.

fi3ework:
@SyMind Still failing with a retry just now, is this CI issue still being resolved? 🥲

SyMind:
@fi3ework I am trying to update the branch to get the latest code. Please wait for the CI to run again.

fi3ework:
@ahabhgk The review was dismissed, PTAL again.

ahabhgk (approved):

PR #6698, opened 2024-05-31
chore: ignore json named exports warnings for non-strict ESM

ahabhgk:
## Summary

Strict ESM will still reported as error

## Checklist


chenjiahan (commented):
awesome 👍🏻

chenjiahan (approved):

PR #6703, opened 2024-06-02
fix: should insert semi after variable decl

h-a-n-a:
## Summary

- Fixed a case where semicolon is not inserted after a variable declaration
- Fixed a case of define plugin using with object

## Checklist


h-a-n-a:
!eco-ci

h-a-n-a:
!eco-ci

LingyuCoder (approved):

PR #6706, opened 2024-06-03
fix: allow return outside functions in scripts

h-a-n-a:
## Summary

- Allowed return outside of function for script ("javascript/dynamic" and "javascript/auto")
- Removed `fn_bind` in ESModule parsing, which is a mistake

## Checklist


packages/rspack-test-tools/tests/statsAPICases/to-string.js
```diff
@@ -11,27 +11,13 @@ module.exports = {
...
-		       4 │
-		       5 │ // Test CJS top-level return
-		       6 │ return;
-		         · ───────
-		         ╰────
-		      
-		  help: 
-		        You may need an appropriate loader to handle this file type.
-
-		Rspack compiled with 1 error (9f555e6a76ee5e3780ec)"
```
h-a-n-a, line 34: This diagnostic is not correct. It's `"javascript/auto"`, which can contain top-level return if it is regarded as a `"script"`

LingyuCoder (approved):

PR #6721, opened 2024-06-04
feat: align `AssetInfo` type with webpack

wxiaoyun:
## Summary
- compilation.updateAsset does not add additional fields to AssetInfo in Rspack. This PR aims to resolve this/

Type definition of AssetInfo in webpack:
```typescript
type AssetInfo = KnownAssetInfo & Record<string, any>;
declare interface KnownAssetInfo {
  /**
   * true, if the asset can be long term cached forever (contains a hash)
   */
  immutable?: boolean;
  /**
   * whether the asset is minimized
   */
  minimized?: boolean;
  /**
   * the value(s) of the full hash used for this asset
   */
  fullhash?: string | string[];
  /**
   * the value(s) of the chunk hash used for this asset
   */
  chunkhash?: string | string[];
  /**
   * the value(s) of the module hash used for this asset
   */
  modulehash?: string | string[];
  /**
   * the value(s) of the content hash used for this asset
   */
  contenthash?: string | string[];
  /**
   * when asset was created from a source file (potentially transformed), the original filename relative to compilation context
   */
  sourceFilename?: string;
  /**
   * size in bytes, only set after asset has been emitted
   */
  size?: number;
  /**
   * true, when asset is only used for development and doesn't count towards user-facing assets
   */
  development?: boolean;
  /**
   * true, when asset ships data for updating an existing application (HMR)
   */
  hotModuleReplacement?: boolean;
  /**
   * true, when asset is javascript and an ESM
   */
  javascriptModule?: boolean;
  /**
   * object of pointers to other assets, keyed by type of relation (only points from parent to child)
   */
  related?: Record<string, string | string[]>;
}
```

Type definition of AssetInfo in Rspack:
```rust
#[derive(Debug, Default, Clone)]
pub struct AssetInfo {
  /// if the asset can be long term cached forever (contains a hash)
  pub immutable: bool,
  /// whether the asset is minimized
  pub minimized: bool,
  /// the value(s) of the full hash used for this asset
  // pub full_hash:
  /// the value(s) of the chunk hash used for this asset
  pub chunk_hash: HashSet<String>,
  /// the value(s) of the module hash used for this asset
  // pub module_hash:
  /// the value(s) of the content hash used for this asset
  pub content_hash: HashSet<String>,
  /// when asset was created from a source file (potentially transformed), the original filename relative to compilation context
  // pub source_filename:
  /// size in bytes, only set after asset has been emitted
  // pub size: f64,
  /// when asset is only used for development and doesn't count towards user-facing assets
  pub development: bool,
  /// when asset ships data for updating an existing application (HMR)
  pub hot_module_replacement: bool,
  /// when asset is javascript and an ESM
  pub javascript_module: Option<bool>,
  /// related object to other assets, keyed by type of relation (only points from parent to child)
  pub related: AssetInfoRelated,
  /// the asset version, emit can be skipped when both filename and version are the same
  /// An empty string means no version, it will always emit
  pub version: String,
  pub source_filename: Option<String>,
}
```

So, the type definition of AssetInto in Rspack is strictly `KnownAssetInfo`, but not `KnownAssetInfo & Record<string, any>`;

This PR aligns the behavior to webpack

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2884,6 +2884,7 @@ dependencies = [
  "rspack_regex",
  "rustc-hash",
  "serde",
+ "serde_json",
 ]
 
 [[package]]
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -237,6 +237,13 @@ export interface JsAssetInfo {
   javascriptModule?: boolean
   /** related object to other assets, keyed by type of relation (only points from parent to child) */
   related: JsAssetInfoRelated
+  /**
+   * Webpack: AssetInfo = KnownAssetInfo & Record<string, any>
+   * But Napi.rs does not support Intersectiont types. This is a hack to store the additional fields
+   * in the rust struct and have the Js side to reshape and align with webpack
+   * Related: packages/rspack/src/Compilation.ts
+   */
+  extras: Record<string, any>
 }
 
 export interface JsAssetInfoRelated {
--- a/crates/rspack_binding_values/Cargo.toml
+++ b/crates/rspack_binding_values/Cargo.toml
@@ -17,3 +17,4 @@ rspack_napi  = { path = "../rspack_napi" }
 rspack_regex = { path = "../rspack_regex" }
 rustc-hash   = { workspace = true }
 serde        = { workspace = true }
+serde_json   = { workspace = true }
--- a/crates/rspack_binding_values/src/asset.rs
+++ b/crates/rspack_binding_values/src/asset.rs
@@ -12,6 +12,7 @@ impl From<JsAssetInfoRelated> for rspack_core::AssetInfoRelated {
     }
   }
 }
+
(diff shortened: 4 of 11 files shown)


packages/playground/cases/lazy-compilation/ignore-entry/index.test.ts
```diff
@@ -1,7 +1,7 @@
 import { expect, test } from "@/fixtures";
 
 test("should load success", async ({ page }) => {
-	await page.waitForTimeout(1000);
+	await page.waitForTimeout(2000);
```
wxiaoyun, line 4: It seems 1000 is too short and will cause test to fail

wxiaoyun:
Not sure how else to fix other than increasing the timeout :skull:

h-a-n-a (approved):

h-a-n-a:
Thank you!

PR #6733, opened 2024-06-05
fix: code splitting won't invalidate in some cases

JSerFeng:
## Summary

`changed` and `updated` can be unexpectedly re-assign

## Checklist


ahabhgk (approved):
看不懂思密达

hardfist:
can we add some tests about this edge case

JSerFeng:
> can we add some tests about this edge case
here https://github.com/web-infra-dev/rspack/pull/6744

PR #6741, opened 2024-06-05
chore(infra): migrate to biome for import order sorting

SoonIter:
## Summary

#6539

1. translate eslint configuration to biome
2. fix prettier to format tsx files

## Checklist


biome.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema": "https://biomejs.dev/schemas/1.8.0/schema.json",
+  "formatter": {
+    "enabled": false
+  },
+  "organizeImports": {
+    "enabled": true,
+    "ignore": [
```
chenjiahan, line 8: Can we reuse the .gitignore file, so we do not need to configure so many ignore rules

https://github.com/web-infra-dev/rsbuild/blob/main/biome.json#L14-L19

packages/rspack-dev-server/etc/api.md
```diff
@@ -18,44 +18,44 @@ export { Configuration }
 
 // @public (undocumented)
 interface ResolvedDevServer extends Configuration {
-    // (undocumented)
-    allowedHosts: "auto" | string[] | "all";
```
h-a-n-a, line 22: IMO, api-extractor generated files are supposed to be ignored as well

chenjiahan (approved):

chenjiahan:
Awesome 👍

PR #6749, opened 2024-06-06
perf: code-splitting should not regenerate chunk

JSerFeng:
## Summary

Every time when iteratoring a block, there will be a new chunk created, though the chunk is not used, but it will stay in memory

## Checklist


SyMind (approved):

SyMind:
!bench

Issue #5729, opened 2024-02-21
Use parser::get_member_expression_info to parse member expression


LingyuCoder:
Merge this pr first to solve the issue, and we need to change to parser::get_member_expression_info in the future after aligning with webpack

_Originally posted by @LingyuCoder in https://github.com/web-infra-dev/rspack/pull/5712#discussion_r1496805721_

PR #6752, opened 2024-06-07
refactor: harmony import dependency parser plugin

ahabhgk:
## Summary

- add diff test for runtime condition (`innerGraph: false`, will enable it once innerGraphVisitor is refactored to parser plugin)
- refactor harmony import dependency parser plugin, fix an issue which cause the above added test failed
  - other parser plugin also need to refactor, currently there are so many additional state exist just for workaround (and I hate workaround...

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3401,7 +3401,6 @@ dependencies = [
  "rspack_regex",
  "rspack_util",
  "rustc-hash",
- "serde",
  "serde_json",
  "sugar_path",
  "swc_core",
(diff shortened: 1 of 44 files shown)


LingyuCoder (approved):

LingyuCoder:
Has parser::get_member_expression_info been full aligned with webpack? not sure if https://github.com/web-infra-dev/rspack/issues/5729 can be solved?

ahabhgk:
Fully aligned, will check it in next week

Issue #6648, opened 2024-05-22
[Bug]: TypeError: t.executeQuery is not a function


elliotwestlake:
### Version

```sh
System:
    OS: macOS 14.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 127.39 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 125.0.6422.61
    Safari: 17.1
  npmPackages:
    @rsbuild/core: ^0.6.15 => 0.6.15 
    @rsbuild/plugin-react: ^0.6.15 => 0.6.15
```

### Details

When trying to use urql with rsbuild we are running into the following error

![image](https://github.com/web-infra-dev/rsbuild/assets/17042376/25f78d46-5626-4702-acc3-b67700879071)

### Reproduce link

https://github.com/elliotwestlake/rsbuild-urql-reproduction

### Reproduce Steps

1. run `pnpm dev`
2. look in console and see the error
```[tasklist]
### Tasks
```

PR #6760, opened 2024-06-08
fix: Add `in_function_scope` for walking method function

CPunisher:
## Summary

closes https://github.com/web-infra-dev/rspack/issues/6648.

Maybe some test cases need to be added but I can't find where to add them.

```ts
import { f1 as a } from "./utils";

function authExchange() {
  return {
    mutate() {
      var a = 1;
      return a + 1;
    },
    appendHeaders() {
      return a();
    },
  };
}
```

In this case, if there is no `in_function_scope` for method functions, the variable declaration `var a = 1` will be defined in the scope of the `authExchange` function. When walking `a` identifier in `appendHeaders`, the `a` defined in mutate will be found first. Then, during code generation, the `a` in `appendHeaders` will not be recognized as an import specifier and will not be replaced with harmony import.

This change is aligned with webpack https://github.com/webpack/webpack/blob/941d3a6017cec12545118da2c2048a5449394d27/lib/javascript/JavascriptParser.js#L3059, where the ast node type of method declaration is function expression.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
@@ -594,8 +594,13 @@ impl<'parser> JavascriptParser<'parser> {
         self.walk_prop_name(&method.key);
         let was_top_level = self.top_level_scope;
         self.top_level_scope = TopLevelScope::False;
-        // FIXME: maybe we need in_function_scope here
-        self.walk_function(&method.function);
+        self.in_function_scope(
+          true,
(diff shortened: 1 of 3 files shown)


h-a-n-a:
You may add one here under `packages/rspack-test-tools/tests/normalCases/parsing`.

ahabhgk (approved):
LGTM, what do you think @h-a-n-a?

h-a-n-a (approved):

h-a-n-a:
Looks good! Thank you!

PR #6783, opened 2024-06-12
fix: increase stack size limit to 4mb

h-a-n-a:
## Summary

Increase stack size limit to 4mb for deep recursions.
Limited to **production builds**, otherwise it's too error-prone.

## Checklist

### DIFF
--- a/crates/node_binding/src/lib.rs
+++ b/crates/node_binding/src/lib.rs
@@ -16,9 +16,11 @@ use rspack_error::Diagnostic;
 use rspack_fs_node::{AsyncNodeWritableFileSystem, ThreadsafeNodeFS};
 
 mod compiler;
-mod env;
 mod panic;
 mod plugins;
+
+#[cfg(not(debug_assertions))]
+mod env;
 
 use plugins::*;
 use rspack_binding_options::*;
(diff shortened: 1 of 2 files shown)


h-a-n-a:
!eco-ci

crates/node_binding/src/lib.rs
```diff
@@ -181,6 +184,13 @@ enum TraceState {
 
 #[ctor]
 fn init() {
+  #[cfg(not(debug_assertions))]
```
9aoy, line 187: > Limited to `production builds`, otherwise it's too error-prone.

Is this restriction only available during `rspack build`, not `rspack dev`? Or rspack build debug / release binding?

9aoy (approved):

crates/node_binding/src/lib.rs
markmssd: I was wondering the same thing. For example in https://github.com/web-infra-dev/rspack/issues/6785, it's happening with dev as well. If that env applies to production only, does it imply we won't be able to start a dev if we use highlight.js?

h-a-n-a: It’s not related to rspack mode

chenjiahan: Just confirmed that the `production builds` means the production build of Rspack bindings. So this fix should resolve #6785 in both dev/prod modes.

Issue #6737, opened 2024-06-05
[Bug]: Bundle Analyzer issue


elsassph:
### System Info

```
System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 259.19 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.16.1 - ~/.nvm/versions/node/v18.16.1/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v18.16.1/bin/yarn
    npm: 9.5.1 - ~/.nvm/versions/node/v18.16.1/bin/npm
    pnpm: 8.15.6 - ~/.nvm/versions/node/v18.16.1/bin/pnpm
    bun: 1.0.15 - ~/.bun/bin/bun
  Browsers:
    Chrome: 125.0.6422.142
    Safari: 17.5
  npmPackages:
    @rspack/cli: ^0.7.1 => 0.7.1 
    @rspack/core: ^0.7.1 => 0.7.1 
```

### Details

When updating from 0.6.x to 0.7.x the analyzer report regressed to show a partial view of the modules (size is mostly identical).

Right: 0.6.3, left: 0.7.1
![image](https://github.com/web-infra-dev/rspack/assets/297963/e7d2bb07-4e78-4b61-adce-68ce45edae1c)

Some specifics of the configuration:
- `concatenateModules: true`
- `externalHelpers: true`

### Reproduce link

https://github.com/elsassph/test-rspack-compat

### Reproduce Steps

```
npx rspack --analyze
```

LingyuCoder:
Unable to run your repro. There is no rule for jsx files in rspack.config.js. However, there are jsx files inside `@lightningjs/solid/dist/source/components`.

elsassph:
@LingyuCoder thanks for checking - I indeed had a patched `@lightningjs/solid` library so I pushed a change to ignore that for now (BTW the nasty "panic" error still happens with a JSX handler). The issue reported now appears with a clean checkout.

LingyuCoder:
> @LingyuCoder thanks for checking - I indeed had a patched `@lightningjs/solid` library so I pushed a change to ignore that for now (BTW the nasty "panic" error still happens with a JSX rule). The issue reported now appears with a clean checkout.

Panic should has been fixed by https://github.com/web-infra-dev/rspack/pull/6778. 

[...]

PR #6784, opened 2024-06-12
feat: concatenated modules stats

LingyuCoder:
## Summary

should analyze concated modules by webpack-bundler-analyzer

- [x] support `module.modules[]`, it is used for drill-down analysis of concated modules
- [x] support `module.depth`, it is used while parsing IIFE to get parsed size and gzip size

fix: https://github.com/web-infra-dev/rspack/issues/6737

image

## Checklist


JSerFeng (approved):

crates/rspack_core/src/stats.rs
```diff
@@ -500,7 +503,32 @@ impl Stats<'_> {
     });
 
     // TODO: a placeholder for concatenation modules
```
ahabhgk, line 505: We can remove the TODO now 😄

LingyuCoder: I will remove it in https://github.com/web-infra-dev/rspack/pull/6799/files#diff-16aee29ffef703f493d3e45add4619f21709212d700fadbf55e8d479b2eedb37L505

PR #6795, opened 2024-06-13
chore: mark defferent between webpack defaults

SyMind:
## Checklist


packages/rspack/src/config/defaults.ts
```diff
@@ -185,11 +186,13 @@ const applyExperimentsDefaults = (
 	experiments: ExperimentsNormalized,
 	{ cache }: { cache: boolean }
 ) => {
-	D(experiments, "lazyCompilation", false);
+	D(experiments, "lazyCompilation", undefined);
 	D(experiments, "asyncWebAssembly", false);
-	D(experiments, "css", true); // we not align with webpack about the default value for better DX
+	// IGNORE(experiments.css): Rspack enables this feature by default for better DX
```
hardfist, line 191: rspack will switch to false when reach 1.0 and css-extract is stable enough

packages/rspack/src/config/defaults.ts
```diff
@@ -349,6 +354,7 @@ const applyModuleDefaults = (
 		D(module.generator["css/module"], "esModule", true);
 	}
 
+	// IGNORE(module.defaultRules): Rspack does not support `rule.assert`
```
hardfist, line 357: what is rule.assert, can you add webpack link about it

LingyuCoder (approved):

PR #6801, opened 2024-06-13
feat: lightningcss minimizer

ahabhgk:
## Summary

```js
module.exports = {
  optimization: {
    minimize: [new rspack.LightningCssMinimizerRspackPlugin()],
  }
}
```

Two noticeable new features:

- Eliminate unused css declaration for CSS Modules: we already have tree shaking for CSS Modules, but that only remove the code in JS file, with [`unusedSymbols`](https://lightningcss.dev/minification.html#unused-symbols) now we can also eliminate the unused declaration in CSS file, including class name, id, and keyframes.
- Lowering css syntax in rust side: now we can use [`browserslist`](https://lightningcss.dev/transpilation.html#browser-targets) to lowering css syntax while minimize css files, this avoid use an additional loader (postcss-loader) to do the lowering

## Checklist


SyMind (approved):

website/docs/en/plugins/rspack/lightning-css-minimizer-rspack-plugin.mdx
```diff
@@ -0,0 +1,70 @@
...
+```css
+.a{color: red}
+```
+
+### browserslist
+
+- **Type:** `string[]`
+- **Default:** `["defaults"]`
+
+Browserslist targets to compile the CSS for.
```
SyMind, line 70: Specifies the minimum version of each browser you want to support.

Issue #6820, opened 2024-06-16
[Bug]: Segmentation fault and bus error on production builds


JohnDaly:
### System Info

```
npmPackages:
  @rspack/cli: 0.7.3 => 0.7.3
  @rspack/core: 0.7.3 => 0.7.3
  @rspack/plugin-react-refresh: 0.7.3 => 0.7.3
```

### Details

When using `rspack` 0.7.3, I've experienced `Segmentation fault` and `Bus error` crashes when doing production builds.

I built a debug build of `rspack` 0.7.3 locally and got this output when running `rust-lldb -- node /path/to/rspack build`:
```
Process 48663 stopped
 * thread #83, name = 'tokio-runtime-worker', stop reason = EXC_BAD_ACCESS (code=2, address=0x30d67fa58)
     frame #0: 0x000000014de6e69c rspack.darwin-arm64.node`rspack_plugin_javascript::utils::eval::eval_binary_expr::eval_binary_expression::h07bcadfd518a6b43(scanner=<unavailable>, expr=<unavailable>) at [eval_binary_expr.rs:380](http://eval_binary_expr.rs:380/)
    377     }
    378     
    379     #[inline]
 -> 380     pub fn eval_binary_expression(
    381       scanner: &mut JavascriptParser,
    382       expr: &BinExpr,
    383     ) -> Option<BasicEvaluatedExpression> {
 Target 0: (node) stopped.
```

### Reproduce Steps

I don't have a simple repro for this yet.

JohnDaly:
Likely related to:
- https://github.com/web-infra-dev/rspack/issues/6775
- https://github.com/web-infra-dev/rspack/issues/6785

h-a-n-a:
Would you please attach the dependencies to the issue?

PR #6851, opened 2024-06-18
refactor: use iteration to evaluate binary expressions

h-a-n-a:
## Summary

Use iteration instead of recursion for binary expression evaluations. 
As nested binary expressions are most likely to be an unbalanced binary tree,
we can iterate from the leaf node all the way up to the top-level to get the evaluation result.

Stack size env is removed as well.

closes https://github.com/web-infra-dev/rspack/issues/6820

## Checklist


h-a-n-a:
!eco-ci

LingyuCoder (approved):

PR #6860, opened 2024-06-19
feat: expose global

LingyuCoder:
## Summary

- Add `ExposeGlobalPlugin` which controled by `experiments.rspackFuture.exposeGlobal`
- Add `ExposeGlobalRuntimeModule` which define the exposed global object, depends on `experiments.rspackFuture.exposeGlobal.global` as name of global variable (`__RSPACK__` by default) and `output.uniqueName` as name of scope variable (the full hash by default)
- Modify `RspackVersionRuntimeModule` and save the version into the exposed global object, we can add more info just like this in the future.

```diff
var version = "$VERSION$";
+ if (typeof __webpack_require__.rg === "object") {
+	__webpack_require__.rg.version = version;
+ }
__webpack_require__.rv = function () {
	return version;
};
```

## Checklist


crates/rspack_plugin_runtime/src/runtime_module/runtime/expose_global.js
```diff
@@ -0,0 +1,13 @@
+var globalName = "$GLOBAL$";
+var uniqueName = $UNIQUE_NAME$;
+var global = __webpack_require__.g;
+if (global[globalName] && typeof global[globalName] !== "object") {
+  console.warn("[RSPACK] Expose global failed: Global variable `" + globalName + "` already exists and its type is not `Object`.");
```
chenjiahan, line 5: Maybe we should avoid injecting `console` to the output, some JS runtime may not provide `console` object.

chenjiahan: I prefer to inject minimal runtime code to include Rspack information to avoid affecting the user's bundle size and code readability.

PR #6861, opened 2024-06-19
fix: export default when environment supports const

### LINKED ISSUE #4031: [Tracking]: Tracking the difference runtime result between Rspack and Webpack
### What subject does this issue tracking?

# [Runtime Diff]
image

- [x] no 'use strict' in esm module codegen
- [x] #6402
image

- [x] missing binding annotation for export esm variable
- [x] module codegen shouldn't be  transformed to es5 by default
- [x] instance of is translated to function by swc in rspack
- [x] 'use strict' optimization if all modules are strict module
- [ ] default export misalign
image
- [x] #4047
- [x] missing constPlugin cause import extra react module
image
- [x] #4118

ahabhgk:
## Summary

This have small different for TDZ, const is more correct than var

resolve #4031

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -866,6 +866,7 @@ export interface RawEntryPluginOptions {
 }
 
 export interface RawEnvironment {
+  const?: boolean
   arrowFunction?: boolean
 }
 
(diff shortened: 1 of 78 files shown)


LingyuCoder (dismissed):

LingyuCoder (approved):

SyMind (approved):

SoonIter:
document needs to be updated
https://www.rspack.dev/config/output#outputenvironment

PR #6864, opened 2024-06-19
perf(rspack_plugin_javascript): use Rayon to parse modules parallelly

fi3ework:
## Summary

Improve perf issue introduced in #6697. Use Rayon to parallelly parse modules.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/mod.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/mod.rs
@@ -1,6 +1,8 @@
 use std::collections::{HashMap, HashSet};
 use std::ops::Deref;
 use std::sync::Arc;
+
+use rayon::prelude::*;
 pub mod api_plugin;
 mod drive;
 mod flag_dependency_exports_plugin;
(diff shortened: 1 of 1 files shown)


fi3ework:
!bench

fi3ework:
The performance regression relieved than before.

fi3ework:
!bench

crates/rspack_plugin_javascript/src/plugin/mod.rs
```diff
@@ -786,150 +795,203 @@ impl JsPlugin {
   ) -> Result<HashMap<String, Arc<dyn Source>>> {
     let mut inlined_modules_to_info: HashMap<String, InlinedModuleInfo> = HashMap::new();
```
h-a-n-a, line 796: I was wondering if the key can be a direct `Identifier`. Its performance is supposed to be better than `string`.

fi3ework: Done. Good suggestion.

fi3ework:
!bench

fi3ework:
!bench

ahabhgk (approved):

Issue #6436, opened 2024-05-02
[Bug]: Angular bundle size 2x bigger than with Webpack


edbzn:
### System Info

 System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 25.70 GB / 30.98 GB
    Container: Yes
    Shell: 5.8 - /usr/bin/zsh
  Binaries:
    Node: 20.11.0 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    pnpm: 8.15.1 - /usr/local/bin/pnpm
    bun: 1.0.22 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 121.1.62.165
    Chrome: 121.0.6167.184
  npmPackages:
    @rspack/cli: ^0.6.3 => 0.6.3 
    @rspack/core: ^0.6.3 => 0.6.3 

### Details

Hi,

I'm trying Rspack with Angular, and all the critical features seem to be working. However, the bundle size generated by Rspack is 2x bigger than base builders. Here's the [configuration](https://github.com/edbzn/ng-rspack/blob/main/rspack.config.js) I'm using. The generated main.js file is around 421K in my example against around 250K for Webpack or Esbuild. You can see the build logs [here](https://github.com/edbzn/ng-rspack/blob/main/build.log#L766). Do you have an idea why, and how I can get approximatively the same bundle size?

Thanks!

### Reproduce link

https://github.com/edbzn/ng-rspack/blob/main

### Reproduce Steps

Rspack bundle size: 

```
pnpm nx build && ls -lah dist/ng-rspack/rspack
```

Webpack bundle size: 

```
pnpm nx build-wp && ls -lah dist/ng-rspack/webpack
```

hardfist:
bump

hardfist:
@edbzn after https://github.com/web-infra-dev/rspack/pull/6888 merged, rspack output seems smaller than angular now, can you help check the result using  1.0.0-alpha.0

edbzn:
[...]

PR #6888, opened 2024-06-23
feat: add `factoryMeta` to `JsModule` and optimize `InnerGraphPlugin` for variable decl with iife

CPunisher:
## Summary

releated: https://github.com/web-infra-dev/rspack/issues/6436

After my debugging, I find some causes: 

1. In the reproduction of this issue, a loader that marks the module as side effect free and a corresponding optimization loader were configured, but factoryMeta is currently not being passed back to the js side.
2. Tree shaking fails for the following code:
```js
// index.js
import { B } from "./b";
let obj = /*#__PURE__*/ (() => {
	return B;
})();

// b.js
const B = 1;
export { B };
```
webpack output: 
```js
/******/ (() => { // webpackBootstrap
/******/ 	"use strict";
var __webpack_exports__ = {};


let obj = /*#__PURE__*/ (/* unused pure expression or super */ null && ((() => {
	return B;
})()));

/******/ })()
;
```
while rspack output:
```js
(() => { // webpackBootstrap
"use strict";
var __webpack_modules__ = ({
"996": (function (__unused_webpack_module, __webpack_exports__, __webpack_require__) {
  B: function() { return B; }
});
const B = 1;
}),

});
// Some runtime...

var __webpack_exports__ = {};
/* harmony import */var _b__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(996);
let obj = /*#__PURE__*/ (/* unused pure expression or super */ null && ((() => {
	return _b__WEBPACK_IMPORTED_MODULE_0__.B;
})()));
})()
;
//# sourceMappingURL=main.js.map
```

The cause of the problem is that InnerGraphPlugin does not set the top symbol to the variable's name before  `init_.visit_children_with` in `visit_var_declarator`, which leads to the lack of dependency relationship between the usage of the identifiers in decl init function and the usage of the declared variable. Instead, it directly considers the identifiers in iife as used in top level, thus preventing optimization.

3. Differences between swc minifier and terser:

|                | origin | swc(TerserPlguin.swcMinify)                               | terser(TerserPlugin)             |
| -------------- | ------ | --------------------------------- | ------------------- |
| webpack        | 2.1MB  | 231KB     | 217KB |
| rspack(before) | 2.1MB  | 319KB | 310KB   |
| rspack(after)  | 2.1MB  | 281KB | 194KB   |

I will also debug the swc minifier if I have time.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/inner_graph_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/inner_graph_plugin.rs
@@ -380,7 +380,6 @@ impl<'a> Visit for InnerGraphPlugin<'a> {
           self.clear_symbol_if_is_top_level();
         }
         _ => {
-          init_.visit_children_with(self);
           if is_pure_expression(init, self.unresolved_ctxt, self.comments.as_ref()) {
             self.set_symbol_if_is_top_level(symbol);
             let start = init.span().real_lo();
@@ -396,7 +395,10 @@ impl<'a> Visit for InnerGraphPlugin<'a> {
                 }
               },
             ));
+            init_.visit_children_with(self);
             self.clear_symbol_if_is_top_level();
+          } else {
+            init_.visit_children_with(self);
           }
         }
       }
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -360,6 +360,10 @@ export interface JsExecuteModuleResult {
   id: number
 }
 
+export interface JsFactoryMeta {
+  sideEffectFree?: boolean
+}
+
 export interface JsLoaderContext {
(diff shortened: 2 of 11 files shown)


ahabhgk (approved):
Thanks

chenjiahan:
Awesome 👍

crates/rspack_plugin_javascript/src/plugin/inner_graph_plugin.rs
```diff
@@ -396,7 +395,10 @@ impl<'a> Visit for InnerGraphPlugin<'a> {
                 }
               },
             ));
+            init_.visit_children_with(self);
```
ahabhgk, line 398: This make the output size much smaller in some cases, good catch! Thanks again!

PR #6917, opened 2024-06-25
docs: review devtool and entry config docs

Timeless0911:
## Summary

- review docs

  - Entry
  - Mode
  - Cache
  - Devtool
  - Performance

- modify the style of PropertyType component

## Checklist


Timeless0911:
It seems we have a `PropertyType` component to write default value and types for some properties, but it is only be used in several docs. We can sort out all common components the entire document needs and then make a unified replacement after all documents are ready. @SyMind @chenjiahan

website/docs/en/config/devtool.mdx
```diff
@@ -35,18 +36,18 @@ export type Devtool =
...
 
-`inline` indicates whether the generated Source Map is inlined into the end of the file via the data url.
+- `inline` indicates whether the generated Source Map is inlined into the end of the file via the data url.
 
-`nosources` is used to control whether the generated Source Map contains source code content to reduce the size of the generated Source Map.
+- `nosources` is used to control whether the generated Source Map contains source code content to reduce the size of the generated Source Map.
 
-`hidden` is used to control whether the end of the generated file contains the `# sourceMappingURL=... ` annotation, browser developer tools and VS Code etc. will look for the Source Map by the path or data url of this annotation in order to map the product's row number back to its location in the source code during debugging.
+- `hidden` is used to control whether the end of the generated file contains the `# sourceMappingURL=...` annotation. The browser developer tools and VSCode etc. will look for the Source Map by the path or data url of this annotation in order to map the product's row number back to its location in the source code during debugging.
```
chenjiahan, line 53: VS Code is correct.

```suggestion
- `hidden` is used to control whether the end of the generated file contains the `# sourceMappingURL=...` annotation. The browser developer tools and VS Code etc. will look for the Source Map by the path or data url of this annotation in order to map the product's row number back to its location in the source code during debugging.
```

website/docs/en/config/entry.mdx
```diff
@@ -7,27 +7,7 @@ import { ApiMeta, Stability } from '../../../components/ApiMeta';
...
-
-  type EntryObject = {
-    [k: string]: EntryItem | EntryDescription;
-  };
-
-  type Config = {
-    entry: EntryItem | EntryObject;
-  };
-  ```
+- **Type:** `string | string[] | Record<string, string | string[] | EntryDescription> | Function`
```
chenjiahan, line 10: Function?

website/docs/en/config/entry.mdx
```diff
@@ -106,23 +86,15 @@ When you set `entry` to an object, you can set the value of the entry to a descr
 
 #### EntryDescription.import
 
-- **Type:**
+- **Type:** `string | string[]`
```
chenjiahan, line 89: I prefer using the `ts` code block, as it can tell the type name and maybe helpful for TypeScript users.

website/docs/en/config/entry.mdx
```diff
@@ -145,19 +117,21 @@ module.exports = {
...
 How this entry load other chunks.
 
+#### EntryDescription.wasmLoading
+
+<ApiMeta addedVersion={'0.2.6'} />
+
+- **Type:** `false | 'fetch' | 'async-node' | string`
+- **Default:** `undefined`
+
+How this entry load WebAssembly modules.
```
chenjiahan, line 133: Is `EntryDescription.wasmLoading` supported?

image

website/docs/zh/config/entry.mdx
```diff
@@ -106,23 +86,15 @@ module.exports = {
...
 
-- **类型：**
-
-  ```ts
-  type EntryItem = string[] | string;
-  ```
+- **类型：** `string | string[]`
 
-- **默认值：** `'./src/index.js'`
+- **默认值：** `./src/index.js`
```
chenjiahan, line 91: The original value is more accurate.

```suggestion
- **默认值：** `'./src/index.js'`
```

website/docs/en/config/entry.mdx
Timeless0911: it is in types of the config, i'm bot sure

chenjiahan: Maybe the type should be removed util Rspack support it

website/docs/en/config/entry.mdx
Timeless0911: It supports (entrychunkname: string) => Record<string, string | string[] | EntryDescription>, I see in other docs use Function to specify such scerios.

website/docs/en/config/entry.mdx
Timeless0911: Most of the documents I see are written directly inline code, so for simple types, I use Inline code.

website/docs/en/config/entry.mdx
chenjiahan: Get, we can add the function usage to current document.

website/docs/en/config/entry.mdx
chenjiahan: Yeah, simple types can just use inline code. But we don't need to change the existed code block back to inline code, which reduces the potentially useful information.

website/docs/en/config/entry.mdx
Timeless0911: image
It already exists and this part seems to be copied from webpack.

Timeless0911:
Updated with the review suggestions, you can rereview it tomorrow. @chenjiahan

website/docs/en/config/entry.mdx
chenjiahan: get

chenjiahan (approved):

chenjiahan:
👍

PR #6942, opened 2024-06-26
docs: polish up external config

fi3ework:
## Summary

Changes to note:

- Didn't document out `byLayer`, this isn't supported yet.

## Checklist

### DIFF
--- a/website/docs/en/config/externals.mdx
+++ b/website/docs/en/config/externals.mdx
@@ -25,7 +25,7 @@ For example, to include [jQuery](https://jquery.com/) from a CDN instead of bund
 ></script>
 ```
 
-```javascript title=rspack.config.js
+```js title="rspack.config.js"
 module.exports = {
   //...
   externals: {
@@ -52,7 +52,7 @@ Depending on the [externalsType](#externalstype), this could be the name of the 
 
 You can also use the shortcut syntax if you're defining only 1 external:
 
-```javascript
+```js title="rspack.config.js"
 module.exports = {
   //...
   externals: 'jquery',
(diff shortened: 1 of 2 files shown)


website/docs/en/config/externals.mdx
```diff
@@ -105,13 +133,166 @@ module.exports = {
...
+  },
+};
+```
+
+This syntax is used to describe all the possible ways that an external library can be made available. `lodash` here is available as `lodash` under AMD and CommonJS module systems but available as `_` in a global variable form. `subtract` here is available via the property `subtract` under the global `math` object (e.g. `window['math']['subtract']`).
+
+### function
+
+- **Type:**
+  - `function ({ context, request, contextInfo, getResolve }, callback)`
```
chenjiahan, line 152: Can we use a TypeScript code block to show the function type?

fi3ework: We could, but using inline code for complex types is widely used in all the Rspack documentation.
Code block will add a new line, which is unexpected IMO. If we're looking for better syntax highlighting, I think it's better to seek the support from Rspress natively. I'd prefer to keep the behavior here 🤔.

chenjiahan: I think most inline complex types are copied from the webpack documentation, which should be updated to TypeScript code blocks in the future.

Rspress has no plan to support inline code syntax highlighting, there is no markdown syntax to specify language for inline code. And inline code are usually some Illegal code snippets.

fi3ework: Makes seem, it seems we're looking for an inline styled code block? I tested on Shiki playground (https://shiki.style/guide/#playground). It has good tolerance on the incomplete syntax.

chenjiahan: Nice 👍 Maybe Rspress can support inline code highlighting one day, but not in the short term

chenjiahan (approved):

PR #6944, opened 2024-06-27
fix: sources types

### LINKED ISSUE #5867: [Bug]: the type of webpack-sources is any

SyMind:
## Summary

Fix #5867

## Checklist

### DIFF
--- a/packages/rspack-test-tools/src/processor/hook.ts
+++ b/packages/rspack-test-tools/src/processor/hook.ts
@@ -1,8 +1,7 @@
 import path from "path";
-import { Compilation, Compiler } from "@rspack/core";
+import { Compilation, Compiler, sources } from "@rspack/core";
 import { getSerializers } from "jest-snapshot";
 import { PrettyFormatOptions, format as prettyFormat } from "pretty-format";
-import { Source } from "webpack-sources";
 
 import { TTestContextOptions, TestContext } from "../test/context";
 import {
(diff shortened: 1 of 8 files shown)


packages/rspack/src/exports.ts
```diff
@@ -45,7 +45,7 @@ export const WebpackError = Error;
 
 export type { Watching } from "./Watching";
 
-const sources = require("webpack-sources"); // use require to avoid wrong types, @types/webpack-sources is outdate
+import sources = require("webpack-sources");
```
chenjiahan, line 48: Can we use `import ... from` here?

packages/rspack-test-tools/src/processor/hook.ts
chenjiahan, line 1: `webpack-sources` can be moved to `devDependencies` as it is prebundled.

packages/rspack/src/exports.ts
SyMind: noop. it will enforce some problems.

The `webpack-sources.d.ts` file uses ESM to declare the module, while `webpack-sources` is implemented with CommonJS.

Using ESM import will cause misalignment between the source and the types defined.

If we declare a namespace and export default in `webpack-sources.d.ts` to fit CommonJS source, `api-extractor` will not support it.

Therefore, use `import sources = require("webpack-sources");` here.

chenjiahan (approved):

PR #6960, opened 2024-06-27
feat!: enable lightning css minimizer as default css minimizer

ahabhgk:
## Checklist


JSerFeng (approved):

chenjiahan:
Should we remove the `SwcCssMinimizerRspackPlugin` plugin?

hardfist:
> Should we remove the `SwcCssMinimizerRspackPlugin` plugin?

I think we should keep it, consider lightningcss maybe strict than swc minifier in some cases which may cause problem for user migration. and even we want to remove it it would be better deprecate it first

chenjiahan:
We can enable the errorRecovery option of LightningCSS by default to align with SWC CSS and other CSS minimizers. Turbopack also enabled it. @ahabhgk what's your opinion

Issue #6874, opened 2024-06-20
[Bug]: Evaluating context request with conditional expression


h-a-n-a:
### System Info

Rspack 0.7.*

### Details

Failed to evaluate context request with conditional expression.

Take this as an example:

```js
let content = "foo"
import(`${true ? "a" : "b"}-${content}`)
```

This was turned into a request:
```js
__webpack_require__("context")(`./a-0${content}`)
```

`0` is most likely introduced by const plugin

### Reproduce Steps

See details above.

PR #6963, opened 2024-06-27
refactor: move context dependency `walk_expression` into `create_context_dependency`

CPunisher:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/6874

The cause of the problem is the direct walking of import call expression, which led to a conflict between the replacement generated by this walk and the replacement handled by ConstPlugin.

Webpack handles the import call expression in three cases:
TemplateLiteral: walk the parts with odd indices, see https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/dependencies/ContextDependencyHelpers.js#L151
Wrapped expr: process the inner expressions, see https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/dependencies/ContextDependencyHelpers.js#L218
Other cases: directly walk the import call expression, see https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/dependencies/ContextDependencyHelpers.js#L241

My solution is to use a visitor to traverse the expression and find the corresponding `Expr` for `BasicEvaluatedExpressionVisitor` based on span. A better solution might be to add a reference to the corresponding `Expr` for `BasicEvaluatedExpressionVisitor` like webpack does. I tried this but couldn't resolve the lifetime issues.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/context_dependency_helper.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/context_dependency_helper.rs
@@ -4,8 +4,13 @@ use itertools::Itertools;
 use once_cell::sync::Lazy;
 use regex::Regex;
 use rspack_core::parse_resource;
+use rspack_core::SpanExt;
 use rspack_error::Severity;
 use rspack_util::json_stringify;
+use swc_core::common::Spanned;
+use swc_core::ecma::ast::Expr;
+use swc_core::ecma::visit::Visit;
+use swc_core::ecma::visit::VisitWith;
 
 use super::create_traceable_error;
 use crate::utils::eval::{BasicEvaluatedExpression, TemplateStringKind};
@@ -15,6 +20,7 @@ const DEFAULT_WRAPPED_CONTEXT_REGEXP: &str = ".*";
 
 pub fn create_context_dependency(
   param: &BasicEvaluatedExpression,
+  expr: &Expr,
   parser: &mut crate::visitors::JavascriptParser,
 ) -> ContextModuleScanResult {
   if param.is_template_string() {
@@ -79,6 +85,12 @@ pub fn create_context_dependency(
         }
       }
     }
+
+    let mut walker = ExprSpanFinder {
+      targets: vec![&parts[1]],
+      on_visit: |n| parser.walk_expression(n),
+    };
+    expr.visit_with(&mut walker);
 
     if parser.javascript_options.wrapped_context_critical {
       let range = param.range();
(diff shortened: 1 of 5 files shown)


crates/rspack_plugin_javascript/src/visitors/dependency/context_dependency_helper.rs
```diff
@@ -160,14 +182,21 @@ pub fn create_context_dependency(
       ));
     }
 
+    if let Some(wrapped_inner_expressions) = param.wrapped_inner_expressions() {
+      let mut walker = BasicEvaluatedExpressionVisitor {
+        targets: wrapped_inner_expressions.iter().collect_vec(),
```
h-a-n-a, line 187: Allocation doesn't seem necessary.

h-a-n-a (approved):
Thank you!

h-a-n-a:
Do you need a double check? @ahabhgk

crates/rspack_plugin_javascript/src/visitors/dependency/context_dependency_helper.rs
```diff
@@ -190,3 +220,23 @@ pub fn create_context_dependency(
...
   }
 }
+
+struct BasicEvaluatedExpressionVisitor<'a, F: FnMut(&Expr)> {
+  targets: Vec<&'a BasicEvaluatedExpression>,
+  on_visit: F,
+}
+
+impl<'a, F: FnMut(&Expr)> Visit for BasicEvaluatedExpressionVisitor<'a, F> {
+  fn visit_expr(&mut self, n: &Expr) {
```
ahabhgk, line 230: Walk is different with swc visitor, parser plugins' goal is replace all the swc visitors, so this needs to rewrite to walk_expression

ahabhgk (changes requested):

crates/rspack_plugin_javascript/src/visitors/dependency/context_dependency_helper.rs
CPunisher: 🤔 I use this visitor to find the `Expr` ast node corresponding to the BasicEvaluatedExpression and then call `parser.walk_expression`. how would this visitor be rewritten to `walk_expression`?

ahabhgk: I think there could have some walk around to implement this and avoid the life time issue, I will take a look next week

ahabhgk: I decide to not implement `setExpression` for BasicEvaluatedExpression, which will introduce lots of lifetime for the parser, basically only context dependency is using `setExpression`, which compare to the current solution it has no different in most cases, and in some cases, it will have slight better performance, but I think it's acceptable

ahabhgk (approved):

PR #6980, opened 2024-06-28
docs: module variables

LingyuCoder:
## Summary

Update documents of module variables

## Checklist

### DIFF
--- a/website/docs/en/api/modules/module-variables.mdx
+++ b/website/docs/en/api/modules/module-variables.mdx
@@ -1,15 +1,129 @@
 import WebpackLicense from '@components/webpack-license';
+import Columns from '@components/Columns';
 import { ApiMeta } from '../../../../components/ApiMeta';
 
 <WebpackLicense from="https://webpack.docschina.org/api/module-variables/" />
 
 # Module Variables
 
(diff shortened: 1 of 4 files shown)


website/docs/en/api/modules/module-variables.mdx
```diff
@@ -126,52 +240,371 @@ componentsContext.keys().forEach(fileName => {
 
 `import.meta.webpackContext()` streamlines the process of module importation especially when you have a lot of files to manage. When using it, please avoid matching unnecessary files, as this might lead to significantly increased build time and output size.
 
-### import.meta.webpackHot (webpack-specific)
+### import.meta.webpackHot (Rspack/Webpack-specific)
```
chenjiahan, line 243: How about removing `(Rspack/Webpack-specific)` from the title? It makes the title too long

LingyuCoder: This is used to distinguish from the standard. Perhaps we can try to use some label components

LingyuCoder: I will handle this in an other pr

chenjiahan: How about move to comment:

```md
### import.meta.webpackHot

> This is an Rspack/Webpack specific API.
```

chenjiahan (approved):

PR #6991, opened 2024-07-01
chore: disable renovate update for swc

hardfist:
## Summary
swc may introduce breaking during patch update which is not safe to do auto bump for rspack, 
we need to do manually update for swc related crated

## Checklist

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -67,7 +67,8 @@
       ],
       "matchPackageNames": [
         "styled_components"
-      ]
+      ],
+      "enable": false // manually update swc and related crate since it may contain breaking change in minor | patch version
     },
     {
       "groupName": "napi",


.github/renovate.json
```diff
@@ -67,7 +67,8 @@
       ],
       "matchPackageNames": [
         "styled_components"
-      ]
+      ],
+      "enable": false
```
chenjiahan, line 71: how about use json5 ext and add some comments

chenjiahan: see https://github.com/web-infra-dev/rsbuild/blob/main/.github/renovate.json5

hardfist: json5 is better, I will change to json5

hardfist: changed to json5 now

chenjiahan (approved):

PR #6992, opened 2024-07-01
chore: add deprecated notice for rspack-plugin-minify

hardfist:
## Summary
since we're compatible with terser-webpack-plugin now, rspack-minify-plugin could be deprecated

## Checklist


chenjiahan (approved):

chenjiahan:
The related documentations should be updated too, such as https://www.rspack.dev/config/optimization#optimizationminimizer

Issue #6878, opened 2024-06-20
[Bug]:  Debugging in browser with rspack serve shows multiple copies of a source file.


pushkargupta9807:
### System Info

Browser : edge

### Details

[Bug]:  Debugging in browser with rspack serve shows multiple copies of a source file. 
Say a source file is included in 2 different bundles. Then the file tree under the sources tab in the browser while debugging would show 2 copies of the same file each belonging to the different bundle. 
![image](https://github.com/web-infra-dev/rspack/assets/93791633/b9956b47-debf-4d12-b360-28c9940e9c46)

This confusing to devs. With webpack it shows only 1 copy, as shown below:
![image](https://github.com/web-infra-dev/rspack/assets/93791633/82b32a81-e21b-4dda-9ad3-0c4a62a39cde)

 

Please find the repro repo and steps in next section

### Reproduce link

https://github.com/pushkargupta9807/rspack-repro-latest

### Reproduce Steps

1. yarn install
2. for rspack  run : yarn dev:rspack 
3. Go to browser Devtools-> Sources tab and observe the multiple copies of the same files as shown in the image above in the details section
4. for webpack run : yarn dev:webpack
5. observe  only 1 copy of the files are displayed in the file tree under the sources tab.

PR #6993, opened 2024-07-01
fix: align source name conflict handling 

CPunisher:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/6878

The processing flow of webpack is as follows:
1. Webpack treats each element in `sourcemap.sources` as a module. ([Code](https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/SourceMapDevToolPlugin.js#L95))
2. It calculates the module -> sourceName mapping and stores it in `moduleToSourceNameMapping`. ([Code](https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/SourceMapDevToolPlugin.js#L287))
3. It checks and resolves conflicts in the values of `moduleToSourceNameMapping`. ([Code](https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/SourceMapDevToolPlugin.js#L344))
4. `sourceMap.sources = modules.map(m => moduleToSourceNameMapping.get(m));` ([Code](https://github.com/webpack/webpack/blob/8241da7f1e75c5581ba535d127fa66aeb9eb2ac8/lib/SourceMapDevToolPlugin.js#L401))

The issue with rspack is that it directly handles conflicts between `sourcemap.sources`. Even if the same source points to the same module, it will execute the fallback logic and append a query `?xxxx` to the original result.

## Checklist


SyMind:
!bench

crates/rspack_plugin_devtool/src/source_map_dev_tool_plugin.rs
```diff
@@ -191,167 +193,170 @@ impl SourceMapDevToolPlugin {
...
+            None => ModuleOrSource::Source(source),
+          }
+        } else {
+          ModuleOrSource::Source(source.to_string())
+        };
+        (source.to_string(), module_or_source)
+      })
+      .collect::<Vec<_>>()
+      .into_iter()
+      .collect::<HashMap<_, _>>();
```
SyMind, line 218: Directly use `collect::<HashMap<_, _>>()` and remove `collect::<Vec<_>>()`.

crates/rspack_plugin_devtool/src/source_map_dev_tool_plugin.rs
```diff
@@ -191,167 +193,170 @@ impl SourceMapDevToolPlugin {
...
+            s,
+            output_options,
+            self.namespace.as_str(),
+          );
+          (module_or_source, source_name)
         })
-        .collect::<Vec<_>>(),
+        .collect::<Vec<_>>()
+        .into_iter()
+        .collect::<HashMap<_, _>>(),
```
SyMind, line 236: Directly use `collect::<HashMap<_, _>>()` and remove `collect::<Vec<_>>()`.

SyMind (approved):

SyMind:
@CPunisher Thank you for your excellent work!

PR #7003, opened 2024-07-01
perf: make picking concatenable modules parallel

JSerFeng:
## Summary

Detect which modules can be concatenated in parallel

## Checklist


JSerFeng:
!bench

ahabhgk (approved):

PR #7009, opened 2024-07-02
docs: improve externals plugin doc

### LINKED ISSUE #7006: docs: add api description of externalsPlugin
After some exploration, I found that it can be implemented through the ExternalsPlugin plug-in, but there is no detailed documentation explaining the role of the parameters. I'm not sure if I should use it this way

https://www.rspack.dev/config/plugins.html#externalsplugin  

```js
new ExternalsPlugin('commonjs', (dep, callback) => {
        if (dep.dependencyType === 'url' && dep.request.startsWith('/')) {
          return callback(null, dep.request);
        }
        callback(null);
      }),
```
For example, what does the `callback(null)` above represent, what does the second parameter mean, and what is the third parameter?

_Originally posted by @noyobo in https://github.com/web-infra-dev/rspack/issues/4518#issuecomment-1791833533_

SyMind:
## Summary

Fix #7006

## Checklist

### DIFF
--- a/website/docs/en/plugins/webpack/externals-plugin.mdx
+++ b/website/docs/en/plugins/webpack/externals-plugin.mdx
@@ -4,11 +4,95 @@ import { ApiMeta } from '@components/ApiMeta.tsx';
 
 <ApiMeta addedVersion={'0.3.3'} />
 
-This plugin can be used to add external dependencies.
+This plugin allows you to specify external dependencies that should not be bundled into the output files. This is particularly useful for libraries that are already available globally or managed by other scripts.
+The [`externalsType`](/config/externals#externalstype) and [`externals`](/config/externals#externals-1) configurations leverage the plugin internally. Therefore, you can utilize the respective functionality directly through these configuration options without needing to use the plugin separately.
 
 ```js
 new rspack.ExternalsPlugin(type, externals);
 ```
 
-- [type](/config/externals#externalstype)
-- [externals](/config/externals#externals-1)
+## Options
+
+### type
+
+**Type:**
+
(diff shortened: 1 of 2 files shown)


website/docs/en/plugins/webpack/externals-plugin.mdx
```diff
@@ -4,11 +4,94 @@ import { ApiMeta } from '@components/ApiMeta.tsx';
...
+
+type ModuleFactoryCreateDataContextInfo = {
+  issuer: string;
+  compiler: string;
+};
+```
+
+**Prevent bundling** of certain `import`ed packages and instead retrieve these _external dependencies_ at runtime.
+
+For more details, refer to [externals](/config/externals#externals-1).
```
chenjiahan, line 97: When should users use this plugin instead of the `externals` config, can we explain this in the document.

SyMind: The `externals` configuration uses the `ExternalsPlugin` internally. Therefore, users can always use the `externals` option directly.

```ts
if (options.externals) {
	new ExternalsPlugin(options.externalsType, options.externals).apply(
		compiler
	);
}
```

chenjiahan: Get, can we add it to document so users can understand this

SyMind: @chenjiahan i complement more info.

chenjiahan (approved):

chenjiahan:
Nice 👍

PR #7026, opened 2024-07-03
refactor: remove dead code of still valid module hook

LingyuCoder:
## Summary

- Still valid module hook has been removed, remain code is all dead
- Rspack should have different cache design against webpack, so remove this hook, should add if needed in the future

## Checklist


LingyuCoder:
seems can not remove directly because [eslint-webpack-plugin](https://github.com/webpack-contrib/eslint-webpack-plugin/blob/master/src/index.js#L121) depend on this hook

PR #7039, opened 2024-07-04
feat(rspack_core): nmf resolve in schema hook

ScriptedAlchemy:
## Summary

Resolve in schema hook for normal module factory

allows me to support https imports 
## Checklist


crates/rspack_loader_runner/src/scheme.rs
```diff
@@ -114,7 +118,12 @@ pub fn get_scheme(specifier: &str) -> Scheme {
     }
   }
 
-  Scheme::from(specifier[..i].to_ascii_lowercase().as_str())
+  let scheme_str = &specifier[..i].to_ascii_lowercase();
+  if scheme_str == "http" || scheme_str == "https" {
```
h-a-n-a, line 122: I think this would be better to move to `From<&str>` implementation for `Scheme`.

ScriptedAlchemy: I'll take a look. Still learning rust haha

ahabhgk: I refactored a little bit, this is included

ahabhgk (approved):

Issue #6821, opened 2024-06-17
[Bug]: DefinePlugin in webpack will throw a warning for the same key, and rspack will choose to override it.


nanianlisao:
### System Info

 System:
    OS: macOS 13.2
    CPU: (8) arm64 Apple M1
    Memory: 92.13 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 18.17.0 
    npm: 9.6.7
    pnpm: 8.9.0 
    bun: 1.0.0
  Browsers:
    Chrome: 126.0.6478.61
    Safari: 16.3

### Details

https://github.com/webpack/webpack/blob/34e2561addb0f65a7a6fb0ce7ae1aea4cd1d599f/lib/DefinePlugin.js#L659-L674

### Reproduce Steps

empty

h-a-n-a:
This is a known issue, as we haven't fully aligned with webpack. Will be fixed, just stay tuned ;-)

nanianlisao:
@h-a-n-a 
Thanks for the reply. I hope this priority can be raised slightly, because this is a relatively hidden but serious bug, especially in rspack and webpack, `process.env.NODE_ENV` will be injected after `optimization.nodeEnv` is turned on.

In addition, if it is convenient, can you provide us with which issue is it duplicated? I will close this issue and subscribe to related content

PR #7045, opened 2024-07-04
feat!: detect conflicting values in `DefinePlugin`

h-a-n-a:
## Summary

Conflict values in `DefinePlugin` are being detected.
Previously, if keys were the same, rspack will use the new value to override the previous one.

closes https://github.com/web-infra-dev/rspack/issues/6821

## Checklist


crates/rspack_binding_options/src/options/raw_builtins/mod.rs
```diff
@@ -512,6 +516,35 @@ impl BuiltinPlugin {
...
+  pub fn append_to(
+    self,
+    plugins: &mut Vec<BoxJavascriptParserPlugin>,
+    builtin_plugins: &mut Vec<Box<dyn Plugin>>,
+  ) -> rspack_error::Result<()> {
+    match self.name {
+      JsParserPluginName::DefinePlugin => {
+        let plugin = DefinePlugin::new(downcast_into(self.options)?);
+        plugins.push(Arc::new(plugin.clone()));
+        builtin_plugins.push(Box::new(plugin));
```
h-a-n-a, line 541: `trait JavaScriptParserPlugin` is not derived from `trait Plugin`

crates/node_binding/src/lib.rs
```diff
@@ -38,16 +38,22 @@ impl Rspack {
     env: Env,
     options: RawOptions,
     builtin_plugins: Vec<BuiltinPlugin>,
+    parser_plugins: Vec<JsParserPlugin>,
```
ahabhgk, line 41: I think there could have a way to just use BuiltinPlugin here instead of introducing a new JsParserPlugin

h-a-n-a: Any advice to implement this?

ahabhgk (approved):

PR #7052, opened 2024-07-05
fix: truncateArgs is not a function

SyMind:
## Summary

Fix

```
Error:   × TypeError: truncateArgs is not a function
```

## Checklist

### DIFF
--- a/packages/rspack/src/node/nodeConsole.ts
+++ b/packages/rspack/src/node/nodeConsole.ts
@@ -8,8 +8,9 @@
  * https://github.com/webpack/webpack/blob/main/LICENSE
  */
 
-const util = require("util");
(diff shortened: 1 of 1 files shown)


packages/rspack/src/logging/truncateArgs.ts
```diff
@@ -87,4 +87,4 @@ const truncateArgs = (args, maxLength) => {
 	});
 };
 
-export { truncateArgs };
+export = truncateArgs;
```
hardfist, line 90: this is a little bit tricky, why not use` import {  truncateArgs } from './logging/truncateArgs'`

SyMind: Fixed

JSerFeng (approved):

PR #7058, opened 2024-07-05
chore: Depend on `@swc/counter`

kdy1:
## Summary

`@swc/counter` is a package to count NPM download count of the SWC project. **Note: This is not enforcement**

See: https://github.com/swc-project/swc/discussions/9155

- Next.js PR: https://github.com/vercel/next.js/pull/67504
- fe-farm PR: https://github.com/farm-fe/farm/pull/1541

## Checklist

### DIFF
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -95,6 +95,7 @@
     "@module-federation/runtime-tools": "0.2.3",
     "@rspack/binding": "workspace:*",
     "@rspack/lite-tapable": "workspace:*",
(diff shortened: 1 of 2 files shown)


chenjiahan:
We prefer to minimize npm dependencies to improve installation speed 😂 Depending on an empty npm package doesn't seem to generate value for the users.

If it's to count the downloads number of SWC, I think you can add up the number of @rspack/core directly to SWC

kdy1:
I see, thanks!

PR #7059, opened 2024-07-06
perf: cache `compilation.entrypoints`

chenjiahan:
## Summary

Cache `compilation.entrypoints` to improve performance, it is similar to `compilation.#cachedAssets`.

The performance cost of `compilation.entrypoints` is noticeable when using [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin/blob/41f4a7b952e80eea4878d96638b4021f4a762750/index.js#L1296).

I have tested it in a MPA project (https://github.com/liangyuetian/rsbuild-hmr-slow-demo) with 200+ entries:

- without cache, it took 500ms (total) to get `compilation.entrypoints`:

截屏2024-07-06 19 31 24

- with cache, it took 2.5ms (total) to get `compilation.entrypoints`:

截屏2024-07-06 20 02 30

## Checklist


chenjiahan:
!bench

ahabhgk (approved):

PR #7065, opened 2024-07-07
chore: use catelog to unify npm version

hardfist:
## Summary
catelog is great feature which is similar with cargo's dependencies.workspace feature, we can use it to unify npm version in monorepo

## Checklist


zoolsher:
nice feature

Issue #7078, opened 2024-07-08
[Bug]: Panic occurred at runtime (rspack alpha.2)


LinusCenterstrom:
### System Info

```
System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13900K
    Memory: 16.66 GB / 63.58 GB
Binaries:
    Node: 18.18.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
Browsers:
    Edge: Chromium (126.0.2592.87)
    Internet Explorer: 11.0.22621.3527
npmPackages:
    @rspack/cli: 1.0.0-alpha.2 => 1.0.0-alpha.2
    @rspack/core: 1.0.0-alpha.2 => 1.0.0-alpha.2
```

### Details

The following error occurs when building with rspack. (@rspack/core 1.0.0-alpha.2)
This does NOT occur in 1.0.0-alpha.0 but does occur in both alpha.1 and alpha.2

```
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  assertion failed: self.max_target_is_set
Location: crates\rspack_core\src\exports_info.rs:1722

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: napi_register_module_v1
    at <unknown source file>
 2: napi_register_module_v1
    at <unknown source file>
 3: napi_register_module_v1
    at <unknown source file>
 4: napi_register_module_v1
    at <unknown source file>
 5: napi_register_module_v1
    at <unknown source file>
 6: wasmer_vm_imported_memory32_atomic_notify
    at <unknown source file>
 7: wasmer_vm_imported_memory32_atomic_notify
    at <unknown source file>
 8: napi_register_module_v1
    at <unknown source file>
 9: napi_register_module_v1
    at <unknown source file>
10: napi_register_module_v1
    at <unknown source file>
11: napi_register_module_v1
    at <unknown source file>
[...]

Issue #7077, opened 2024-07-08
 [Bug]: panic during build phase (rsbuild-alpha.5)


Verthon:
### System Info

System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 668.61 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.19.0 - ~/.nvm/versions/node/v18.19.0/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v18.19.0/bin/yarn
    npm: 10.2.3 - ~/.nvm/versions/node/v18.19.0/bin/npm
    pnpm: 8.15.1 - ~/.nvm/versions/node/v18.19.0/bin/pnpm
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.5

### Details

Sadly no reproduction repo is available as it is config used in the workplace.

RSBuild versions:
* "@rsbuild/core": "1.0.0-alpha.5",
* "@rsbuild/plugin-react": "1.0.0-alpha.5",

Here is what was printed out to the terminal:

```bash
● web ━━━━━━━━━━━━━━━━━━━━━━━━━ (70%) sealing chunk ids                                                                                                     Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  assertion failed: self.max_target_is_set
Location: crates/rspack_core/src/exports_info.rs:1722

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: _napi_register_module_v1
    at <unknown source file>
 2: _napi_register_module_v1
    at <unknown source file>
 3: _napi_register_module_v1
    at <unknown source file>
 4: _napi_register_module_v1
    at <unknown source file>
 5: _napi_register_module_v1
    at <unknown source file>
 6: _wasmer_vm_imported_memory32_atomic_notify
    at <unknown source file>
 7: _wasmer_vm_imported_memory32_atomic_notify
    at <unknown source file>
 8: _napi_register_module_v1
    at <unknown source file>
 9: _napi_register_module_v1
    at <unknown source file>
[...]

Issue #7062, opened 2024-07-07
[Bug]: Build error after upgrade to 1.0


hustcer:
### System Info

System:
    OS: macOS 14.5
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
    Memory: 543.95 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.13.1 - ~/.local/state/fnm_multishells/68643_1720315491078/bin/node
    npm: 10.5.2 - ~/.local/state/fnm_multishells/68643_1720315491078/bin/npm
    pnpm: 9.1.4 - ~/.local/state/fnm_multishells/68643_1720315491078/bin/pnpm
    bun: 1.0.33 - /usr/local/bin/bun
    Watchman: 2024.07.01.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.5

```json
    "@rsbuild/core": "1.0.0-alpha.5",
    "@rsbuild/plugin-less": "1.0.0-alpha.5",
    "@rsbuild/plugin-react": "1.0.0-alpha.5",
    "@rsbuild/plugin-sass": "1.0.0-alpha.5",
    "@rsbuild/plugin-umd": "^1.0.0",
```

### Details

It works fine with rsbuild 0.7.10, however, after upgrading to the latest version will got a build error as follows:
```console
  Rsbuild v1.0.0-alpha.5

● web ━━━━━━━━━━━━━━━━━━━━━━━━━ (70%) sealing chunk ids                                                                Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  assertion failed: self.max_target_is_set
Location: crates/rspack_core/src/exports_info.rs:1722

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: _napi_register_module_v1
    at <unknown source file>
 2: _napi_register_module_v1
    at <unknown source file>
 3: _napi_register_module_v1
    at <unknown source file>
 4: _napi_register_module_v1
    at <unknown source file>
 5: _napi_register_module_v1
[...]

Issue #7057, opened 2024-07-05
[Bug]: panic occured at runtime - sealing chunk ids


Dmitrijs999:
### System Info

  System:
    OS: macOS 13.0
    CPU: (8) arm64 Apple M2
    Memory: 136.95 MB / 8.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 21.7.2 - ~/.nvm/versions/node/v21.7.2/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v21.7.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v21.7.2/bin/npm

### Details

Trying to migrate CRA app to rsbuild, this is my config:
```typescript
import { defineConfig, loadEnv } from '@rsbuild/core';
import { pluginReact } from '@rsbuild/plugin-react';
import { pluginSvgr } from '@rsbuild/plugin-svgr';
const { publicVars } = loadEnv({ prefixes: ['REACT_APP_'] });

export default defineConfig({
  plugins: [pluginReact(), pluginSvgr({ mixedImport: true })],
  html: {
    template: './public/index.html',
  },
  output: {
    distPath: {
      root: 'build',
    },
  },
  source: {
    define: publicVars,
    alias: {
      crypto: require.resolve('crypto-browserify'),
      stream: require.resolve('stream-browserify'),
      assert: require.resolve('assert'),
      http: require.resolve('stream-http'),
      https: require.resolve('https-browserify'),
      os: require.resolve('os-browserify'),
      url: require.resolve('url'),
    },
  },
});

```
Running rsbuild build
```bash
yarn run v1.22.22
$ rsbuild build
  Rsbuild v1.0.0-alpha.4

● web ━━━━━━━━━━━━━━━━━━━━━━━━━ (70%) sealing chunk ids                                                                                                                                                           Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  assertion failed: self.max_target_is_set
Location: crates/rspack_core/src/exports_info.rs:1722

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
[...]

PR #7082, opened 2024-07-09
fix: panic when get max target in find target

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/7057
fix https://github.com/web-infra-dev/rspack/issues/7062
fix https://github.com/web-infra-dev/rspack/issues/7077
fix https://github.com/web-infra-dev/rspack/issues/7078

In webpack, module graph is always mutable and will insert the max target of export info while getting it, even during the code gen.

But in rspack, module graph should be immutable during the code gen. So can only get max target when it has been set. If it is not exists, export info should generate it but not set it into export info. 

## Checklist

### DIFF
--- a/crates/rspack_core/src/exports_info.rs
+++ b/crates/rspack_core/src/exports_info.rs
@@ -1293,11 +1293,17 @@ impl ExportInfoId {
     if !export_info.target_is_set || export_info.target.is_empty() {
       return FindTargetRetEnum::Undefined;
     }
-    let raw_target = export_info
-      .get_max_target_readonly()
-      .values()
-      .next()
-      .cloned();
+
+    let raw_target = if export_info.max_target_is_set {
+      export_info
+        .get_max_target_readonly()
+        .values()
+        .next()
+        .cloned()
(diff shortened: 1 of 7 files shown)


crates/rspack_core/src/exports_info.rs
chenjiahan, line 1: Can we add a test case for this

crates/rspack_core/src/exports_info.rs
```diff
@@ -1293,11 +1293,17 @@ impl ExportInfoId {
...
     }
-    let raw_target = export_info
-      .get_max_target_readonly()
-      .values()
-      .next()
-      .cloned();
+
+    let raw_target = if export_info.max_target_is_set {
+      export_info
+        .get_max_target_readonly()
```
h-a-n-a, line 1299: I was wondering if we can add a panic notice to this method.

h-a-n-a (approved):

JSerFeng (approved):

PR #7087, opened 2024-07-09
Chore/update swc 0 7 5

JSerFeng:
## Checklist


hardfist:
friendly notice, this should be feat other than chore, and if it contains breaking change in swc, remind add breaking change notice

PR #7088, opened 2024-07-09
perf: enable "fat" LTO for production release

chenjiahan:
### Pros

- Better performance (according to benchmark): **1% ~ 3% faster**
- Smaller binary size (rspack.darwin-arm64.node): **55.4 MB -> 51.7 MB**

### Cons

- Longer build time, only affect the production release: **4-5min -> 8-10min**

see: 

- https://github.com/web-infra-dev/rspack/pull/5507
- https://doc.rust-lang.org/cargo/reference/profiles.html#lto

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,6 +64,7 @@ jobs:
     with:
       target: ${{ matrix.array.target }}
       runner: ${{ matrix.array.runner }}
+      profile: "release-prod"
 
   release:
     name: Release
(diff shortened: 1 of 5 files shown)


chenjiahan:
!bench

JSerFeng (approved):

JSerFeng:
Cool, I thought we were already use fat in production release long ago

h-a-n-a (approved):

chenjiahan:
> I thought we were already use fat in production release long ago

Haha, I guess the reason for using "thin" before was mainly due to CI time. And it is worth to enable "fat" if it only affects the prod release

Issue #4326, opened 2023-10-16
support hideStack in WebpackError


ahabhgk:
https://github.com/webpack/webpack/blob/87660921808566ef3b8796f8df61bd79fc026108/lib/WebpackError.js#L32

- webpack-test/cases/compile/error-hide-stack

h-a-n-a:
I would like to add **part** of this to the new error refactor.

h-a-n-a:
Related: https://github.com/web-infra-dev/rspack/issues/4339

Issue #4339, opened 2023-10-17
Improve the relevant APIs of StatsError.


ityuany:
### What problem does this feature solve?

see [friendly-errors-webpack-plugin](https://github.com/geowarin/friendly-errors-webpack-plugin) , We need to implement similar functionality. This functionality requires extracting information such as StatsError stack, moduleId, etc. Currently, the api for rspack StatsError is not perfect.

### What does the proposed API of configuration look like?

like webpack StatsError

h-a-n-a:
Related https://github.com/web-infra-dev/rspack/issues/4326

hardfist:
bump

LingyuCoder:
Most of StatsError fields had been aligned with webpack:

```ts
export interface JsStatsError {
  message: string
  chunkName?: string
  chunkEntry?: boolean
  chunkInitial?: boolean
  file?: string
  moduleIdentifier?: string
  moduleName?: string
  moduleId?: string
  chunkId?: string
  details?: string
  stack?: string
  moduleTrace: Array<JsStatsModuleTrace>
}
```

Cause the error handling of rspack on the rust side cannot be completely consistent with nodejs of webpack. Can these fields meet your needs?

LingyuCoder:
I will close this issue as completed. If you meet some problem again, please just reopen this.

PR #7091, opened 2024-07-09
feat: align part of StatsError

LingyuCoder:
## Summary

related https://github.com/web-infra-dev/rspack/issues/4339

resolve https://github.com/web-infra-dev/rspack/issues/4326

### Align `stack`, `details`, `file` on `StatsError`

Allow the following javascript error to hide the stack by `hideStack`, and when `hideStack = true`, the stack info will be placed in `details`, no matter what `hideStack` is, the stack info will always be placed in `stack`.
- Throw error in js loader
- Call `this.emitError()` or `this.emitWarning()` in js loader
- Push error to `compilation.errors` or `compilation.warnings`

> Rust error has no stack, so most errors throw by rust are different with webpack

### Align chunk related fields on `StatsError`
Some errors are thrown during creating chunk asset, so they have `chunkId`, `chunkName`, `chunkEntry`, `chunkInitial` on `StatsError`

> Currently, only code splitting warnings(dependOn related) and css warnings(confilicting) has chunk info. 

## Checklist


packages/rspack/src/loader-runner/index.ts
```diff
@@ -635,7 +640,7 @@ export async function runLoaders(
 		error.message = `${error.message} (from: ${stringifyLoaderObject(
 			loaderContext.loaders[loaderContext.loaderIndex]
 		)})`;
-		hasStack && Error.captureStackTrace(error);
+		!hasStack && Error.captureStackTrace(error);
```
LingyuCoder, line 643: if error thrown by loader does not have stack, then generate it

crates/rspack_error/src/error.rs
```diff
@@ -351,3 +364,9 @@ fn _assert() {
   _assert_send_sync::<InternalError>();
   _assert_send_sync::<DiagnosticError>();
 }
+
+/// (message, stack, backtrace, hide_stack)
+#[derive(Debug, Error, Diagnostic)]
+#[diagnostic()]
+#[error("{0}\n{2}")]
+pub struct NodeError(pub String, pub Option<String>, pub String, pub Option<bool>);
```
h-a-n-a, line 372: Can we convert this to a struct? Too much elements for a tuple.

h-a-n-a (approved):

PR #7092, opened 2024-07-09
feat(create-rspack): add custom template pkg name

liangchaofei:
Automatically set the folder name entered by the user to the name of pkg.

Scenario: The package in the monorepo is usually the same as the entered folder name, so there is no need to modify it manually

将用户输入的文件夹名称自动渲染到pkg的name。

场景：monorepo中的包一般和输入的文件夹名称一样，就不需要手动修改了

### DIFF
--- a/packages/create-rspack/index.js
+++ b/packages/create-rspack/index.js
@@ -38,13 +38,15 @@ yargs(hideBin(process.argv))
 			);
 
 		await promptProjectDir();
-		let root = path.resolve(process.cwd(), getProjectDir(targetDir));
+		let projectDir = getProjectDir(targetDir);
+		let root = path.resolve(process.cwd(), projectDir);
 		while (fs.existsSync(root)) {
 			console.log(
 				`${targetDir} is not empty, please choose another project name`
 			);
 			await promptProjectDir();
-			root = path.resolve(process.cwd(), getProjectDir(targetDir));
+			projectDir = getProjectDir(targetDir);
+			root = path.resolve(process.cwd(), projectDir);
 		}
 
 		// choose template
@@ -73,7 +75,7 @@ yargs(hideBin(process.argv))
 		copyFolder(srcFolder, root, targetDir);
 		const pkgManager = getPkgManager();
 		console.log("\nDone. Now run:\n");
-		console.log(`cd ${getProjectDir(targetDir)}\n`);
+		console.log(`cd ${projectDir}\n`);
 		console.log(`${pkgManager} install\n`);
 		console.log(`${pkgManager} run dev\n`);
 	})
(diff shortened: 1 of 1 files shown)


packages/create-rspack/index.js
```diff
@@ -114,6 +114,9 @@ function copyFolder(src, dst) {
 						}
 					}
 				}
+				if (pkg.name) {
+					pkg.name = dst;
```
chenjiahan, line 118: `dst` may contain sub folder, like this:

Can you try `path.dirname(dst)`?

image

liangchaofei: > path.dirname(dst)

I think it should work like this: The user creates an empty folder, runs the CLI command in this empty folder, and then enters the desired package name. The selected template code is then copied into the current empty folder. This is also the reason behind the PR I submitted last time: why it is necessary to check if the current directory is empty. By doing it this way, we don't have to worry about whether the user's input contains subfolders; the input is just the package name, which can either include a scope or be standalone. Please evaluate whether this approach is reasonable. If it is, we can make this change.

liangchaofei: In projects with a monorepo architecture, the names of many packages are often under a scope.

chenjiahan: My concern is that this will require users to create an empty directory before running the CLI, which most users will not do.

chenjiahan: Users have already developed habits, so what we can do is try to adapt to more usage scenarios and align with other tools in the community

liangchaofei: > Users have already developed habits, so what we can do is try to adapt to more usage scenarios and align with other tools in the community

After using path.dirname(dst), for example, if the user inputs test/rspack-project, the name in the created template's package.json will be test instead of test/rspack-project. In my scenario, I want to use rspack to directly create a project's package.json name in a monorepo project to be exactly what I input. Since my input does not include subfolder names, I think it would be better to use dst directly.

chenjiahan: I think we can support both of them:

- If input is `subdir/rspack-project`, the package.json#name should be `rspack-project`, the output path is `/cwd/subdir/rspack-project'.
- If input is `@scope/rspack-project,`, the package.json#name should be `@scope/rspack-project`, the output path is `/cwd/rspack-project'.

liangchaofei: > I think we can support both of them:我认为我们可以支持他们两个：
> 
> * If input is `subdir/rspack-project`, the package.json#name should be `rspack-project`, the output path is `/cwd/subdir/rspack-project'.如果输入是 `subdir/rspack-project` ，则 package.json#name 应为 `rspack-project` ，输出路径为 `/cwd/subdir/rspack-project'。
> * If input is `@scope/rspack-project,`, the package.json#name should be `@scope/rspack-project`, the output path is `/cwd/rspack-project'.如果输入为 `@scope/rspack-project,` ，则 package.json#name 应为 `@scope/rspack-project` ，输出路径为 `/cwd/rspack-project'。

done

packages/create-rspack/index.js
```diff
@@ -69,17 +70,17 @@ yargs(hideBin(process.argv))
 
 		fs.mkdirSync(root, { recursive: true });
 		const srcFolder = path.resolve(__dirname, `template-${template}`);
-		copyFolder(srcFolder, targetDir);
+		copyFolder(srcFolder, root, targetDir);
 		const pkgManager = getPkgManager();
 		console.log("\nDone. Now run:\n");
-		console.log(`cd ${targetDir}\n`);
+		console.log(`cd ${getProjectDir(targetDir)}\n`);
```
chenjiahan, line 76: `getProjectDir(targetDir)` is called multiple times, can we use a variable to store it

packages/create-rspack/index.js
```diff
@@ -122,6 +126,29 @@ function copyFolder(src, dst) {
 	}
 }
 
+function getPkgName(targetDir) {
+  const scopeMatch = matchScopedPackageName(targetDir);
+  if (scopeMatch) {
+    return targetDir; // Scoped package name
+  } else {
```
chenjiahan, line 133: unless `else`

packages/create-rspack/index.js
```diff
@@ -122,6 +126,29 @@ function copyFolder(src, dst) {
...
+    return path.basename(targetDir); // Use the base name of the target directory
+  }
+}
+
+function getProjectDir(targetDir) {
+  const scopeMatch = matchScopedPackageName(targetDir);
+  if (scopeMatch) {
+    return scopeMatch[1]; // Subdirectory project name for scoped packages
+  } else {
```
chenjiahan, line 142: unless `else`

liangchaofei: > unless `else` 除非 `else`

done

packages/create-rspack/index.js
liangchaofei: done

chenjiahan (approved):
LGTM, thanks! ❤️

PR #7093, opened 2024-07-09
feat: support type field for jsModule

JSerFeng:
## Summary

support type field for JsModule, now we can use experiments.css and css-extract-loader together

## Checklist


ahabhgk (approved):

tests/plugin-test/css-extract/cases/together-with-experiment-css/webpack.config.js
```diff
@@ -0,0 +1,21 @@
+const { CssExtractRspackPlugin } = require("@rspack/core");
+
+/** @type {import("@rspack/core").Configuration} */
+module.exports = {
```
hardfist, line 4: I don't understand what's the behavior when they use toghether, what's the difference between use them together and use experiments.css alone?

packages/rspack/src/builtin-plugin/css-extract/loader.ts
```diff
@@ -79,10 +84,15 @@ export const pitch: LoaderDefinition["pitch"] = function (request, _, data) {
...
+		this._compiler.options.experiments.css &&
+		this._module &&
+		(this._module.type === "css" ||
+			this._module.type === "css/auto" ||
+			this._module.type === "css/global" ||
+			this._module.type === "css/module")
 	) {
 		let e = new Error(
-			"You can't use `experiments.css` and `css-extract-rspack-plugin` together, please set `experiments.css` to `false`"
+			'You can\'t use `experiments.css` (`experiments.futureDefaults` enable built-in CSS support by default) and `mini-css-extract-plugin` together, please set `experiments.css` to `false` or set `{ type: "javascript/auto" }` for rules with `mini-css-extract-plugin` in your webpack config (now `mini-css-extract-plugin` does nothing).'
```
hardfist, line 95: this is weird? why this warning can't use experiments.css with mini-css-extract-plugin together and then tests case use them together?

tests/plugin-test/css-extract/cases/together-with-experiment-css/webpack.config.js
chenjiahan: I found a use case:

Rsbuild uses `CssExtractPlugin` by default. Sometimes user want to extend some custom CSS rules, then they can use `experiments.css` to handle specific CSS modules, it is easier to config than `CssExtractPlugin`.

JSerFeng:
They can use together when you set moduleType to `javascript` to modules which are handled by css-loader, as they become normal javascript modules.

Modules whose moduleType is `css` can still work by native css support.

Just like asset module, in Webpack5 you can use type:'asset', or you can still use file-loader, url-loader, etc

For example

```
module.exports = {
  modules: {
    rules: [
      { test: /foo\.css$/, use: [ CssExtractPlugin.loader, 'css-loader' ], type: 'javascript' },
      { test: /bar\.css$/, type: 'css' },
    ]
  }
}
```

packages/rspack/src/builtin-plugin/css-extract/loader.ts
JSerFeng: Because one module with type css should not handled by this plugin, only type 'javascript' can

chenjiahan:
Set `type: 'javascript'` for `foo.css` is wired in this example, a bit confusing

JSerFeng:
> Set type: 'javascript' for foo.css is wired in this example, a bit confusing

When not enable native css, the module type for those modules are 'javascript' by default, and modules processed by css-loader are javascript

hardfist:
I think we should explain well how `experiments.css` `type:css` and `cssExtractPlugin` works together before landing this feature, otherwise users maybe confusing(I'm still confusing now)

packages/rspack/src/builtin-plugin/css-extract/loader.ts
```diff
@@ -79,10 +84,15 @@ export const pitch: LoaderDefinition["pitch"] = function (request, _, data) {
...
+		this._compiler.options.experiments.css &&
+		this._module &&
+		(this._module.type === "css" ||
+			this._module.type === "css/auto" ||
+			this._module.type === "css/global" ||
+			this._module.type === "css/module")
 	) {
 		let e = new Error(
-			"You can't use `experiments.css` and `css-extract-rspack-plugin` together, please set `experiments.css` to `false`"
+			'You can\'t use `experiments.css` (`experiments.futureDefaults` enable built-in CSS support by default) and `CssExtractRspackPlugin` together, please set `experiments.css` to `false` or set `{ type: "javascript/auto" }` for rules with `CssExtractRspackPlugin` in your webpack config (now `CssExtractRspackPlugin` does nothing).'
```
hardfist, line 95: > 'You can\'t use `experiments.css` (`experiments.futureDefaults` enable built-in CSS support by default) and `CssExtractRspackPlugin` together,

I thinks this is inaccurate, since we did use experiments.css and cssExtractRspackPlugin together in the test case, what you mean maybe "You can't use type:css with cssExtractRspackPlugin together" and experiments.css set type:css for css file by default so remember set type:"javascript/auto" for css file if you want to use cssExtractRspackPlugin to handle css files

JSerFeng: Done

JSerFeng:
docs: #7114

ahabhgk (approved):

PR #7115, opened 2024-07-11
test: enable stats test

LingyuCoder:
## Summary

Enable most of webpack stats test cases to track rspack stats changes
- Default state is `partial_pass` because it is not full aligned with webpack
- If case is full aligned, state can be modfied to `pass`

## Checklist


h-a-n-a (approved):
🚀

PR #7131, opened 2024-07-12
feat: add trace in hook macro

JSerFeng:
## Summary

Add trace for every hook, you can disable by:

```rust
#[plugin_hook(Params for Plugin, tracing=false)]
fn foo()
```

Now we can see costs for every plugin

image

## Checklist


ahabhgk (approved):

hardfist:
can we see js plugin cost

PR #7192, opened 2024-07-16
perf: reduce allocation for `TraceableError`

h-a-n-a:
## Summary

Replaced `src: String` with `src: Arc<String>`. `Arc<String>` is an internal representation of `SourceFile` source.
The type should not be `Arc<str>` as it creates additional heap allocations.

Removed duplicated `String` allocations in `TraceableError` of `CssPlugin`.

Removed normalization code for `start` and `end` position.

## Checklist

### DIFF
--- a/crates/rspack_error/src/error.rs
+++ b/crates/rspack_error/src/error.rs
@@ -7,6 +7,7 @@ use thiserror::Error;
 
 use crate::RspackSeverity;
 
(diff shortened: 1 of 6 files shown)


JSerFeng (approved):

hardfist:
maybe we should run bench on perf pr?

PR #7211, opened 2024-07-18
chore(deps): use typescript latest version and `"module": "NodeNext"`

SoonIter:
## Summary

1. have a try to use latest typescript version and `"module": "NodeNext"`
2. typescript is also a phantom dependence, we are using the surprising version of `tsc` which caused local build failed

https://github.com/web-infra-dev/rspack/pull/4650
https://github.com/web-infra-dev/rspack/issues/7156

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

Issue #7156, opened 2024-07-15
[Tracking]: Dependencies related governance [JS/pnpm] [Rust/Cargo]


SoonIter:
## Summary

Since Rspack contains a large number of phantom dependency problems, which is not conducive to developers, we need to govern the dependencies in both js and Rust sides.

### benefits

1. more stable CI test and e2e results 
2. more readable dependencies relationship and easier to update the deps version
3. this is a pre-task for https://github.com/web-infra-dev/rspack/pull/6323, better and simple dependencies graph will enhance the install speed

```[tasklist]
### Tasks
- [ ] https://github.com/web-infra-dev/rspack/pull/7094
- [ ] https://github.com/web-infra-dev/rspack/pull/7178
- [x] ci: `.pnpm/node_modules/<MODULE>/node_modules/dep_name` Snapshot serializer https://github.com/web-infra-dev/rspack/pull/8137
- [ ] https://github.com/web-infra-dev/rspack/pull/7130
- [ ] https://github.com/web-infra-dev/rspack/pull/6323
- [ ] #7208
- [ ] https://github.com/web-infra-dev/rspack/issues/7341
- [ ] #7403
- [ ] https://github.com/web-infra-dev/rspack/pull/7211
- [ ] https://github.com/web-infra-dev/rspack/pull/7235
- [x] Update the strategy of renovate in JS side https://github.com/web-infra-dev/rspack/pull/6429 https://github.com/web-infra-dev/rspack/pull/8294
- [x] Update the strategy of renovate in Rust side https://github.com/web-infra-dev/rspack/pull/6596 https://github.com/web-infra-dev/rspack/pull/8294
- [ ] Upgrade to pnpm9.5 to use "catelog:" protocol https://pnpm.io/catalogs (pending by renovate support https://github.com/renovatebot/renovate/issues/30079 ) https://github.com/web-infra-dev/rspack/pull/7065
```

SoonIter:
It should be emphasized that we have a killer tool for phantom dependencies. https://github.com/chengcyber/scan-deps

https://github.com/web-infra-dev/rspack/pull/7178
`npx scan-deps --directory alert,debug,meta,release,test --extension ts,jsx,tsx,cjs,mjs,js --debug`

SoonIter:
https://github.com/web-infra-dev/rspack/pull/7638#issuecomment-2301666065

SoonIter:
only one remaining task

[...]

PR #7235, opened 2024-07-19
chore(deps): typescript phantom dependence

SoonIter:
## Summary

typescript is also a phantom dependence, we are using the surprising version of tsc which caused local build failed

#7211 
https://github.com/web-infra-dev/rspack/pull/7211#issuecomment-2238700491
https://github.com/web-infra-dev/rspack/issues/7156

## Checklist

### DIFF
--- a/packages/rspack/src/lib/CacheFacade.js
+++ b/packages/rspack/src/lib/CacheFacade.js
@@ -59,6 +59,7 @@ class MultiItemCache {
 	getPromise() {
 		// @ts-expect-error
 		const next = i => {
+			// @ts-ignore if your typescript version >= 5.5, this line will throw an error
 			return this._items[i].getPromise().then(result => {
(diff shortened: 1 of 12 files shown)


packages/rspack/src/lib/CacheFacade.js
```diff
@@ -59,6 +59,7 @@ class MultiItemCache {
 	getPromise() {
 		// @ts-expect-error
 		const next = i => {
+			// @ts-ignore if your typescript version >= 5.5, this line will throw an error
```
chenjiahan, line 62: Can we fix this error in >= 5.5

SoonIter: this file needs to be translated to typescript, but not in this pull request

chenjiahan (approved):

PR #7243, opened 2024-07-20
ci: build linux debug binary directly for CI purpose to speedup PR check

xc2:
## Summary

Build linux debug binary directly instead of on another container

Applied for:

1. github shared runner only
1. the `CI` Workflow only
1. `Test Linux` job only that is typically run test for PR

Will reduce the execute time of build step of `Test Linux / Build` from >7m to ~5m - typically from image pull and toolchain download.

## Checklist

- [x] Tests updated (or **not required**).
- [x] Documentation updated (or **not required**).


.github/workflows/ci.yml
```diff
@@ -62,6 +62,7 @@ jobs:
     with:
       target: x86_64-unknown-linux-gnu
       profile: "debug"
+      native: ${{ needs.get-runner-labels.outputs.LINUX_RUNNER_LABELS == '"ubuntu-latest"' }}
```
jerrykingxyz, line 65: We can use `${{ startsWith(runner.name, 'GitHub Actions') }}` to check whether it runs in github runner.

chenjiahan: @jerrykingxyz can you check again?

chenjiahan: @jerrykingxyz cc

PR #7257, opened 2024-07-22
perf: a bunch of small improvement for ConcatenatedModule

CPunisher:
## Summary

😆 This is my practice using the profile tool for the first time, and here is what I do:
1. Remove unnecessary clones.
2. Remove `ModuleInfoOrReference`, as I find `ModuleInfo` is sufficient and `ModuleInfoOrReference` can be queried using the module id.
3. Replaced `LinkedHashSet` with `IndexSet`. Based on my benchmarking, `IndexSet::insert` and `IndexSet::pop` perform better.
4. Replaced the check `self.modules.iter().any(|item| Some(&item.id) == module_id_of_dep)` in each loop with a pre-constructed `HashSet` to reduce time complexity.

Since my local performance test results are very unstable, if there is a performance regression, I will further modify or close this PR.

## Checklist

### DIFF
--- a/crates/rspack_core/src/concatenated_module.rs
+++ b/crates/rspack_core/src/concatenated_module.rs
@@ -407,7 +407,7 @@ impl ConcatenatedModule {
   ) -> String {
     let mut identifiers = vec![];
     for m in modules {
-      identifiers.push(m.shorten_id.clone());
+      identifiers.push(m.shorten_id.as_str());
     }
     identifiers.sort();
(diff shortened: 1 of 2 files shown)


chenjiahan:
!bench

JSerFeng (dismissed):

JSerFeng:
Thanks, could you resolve conflicts ?

JSerFeng (approved):

JSerFeng:
Thank you !

Issue #7225, opened 2024-07-19
[Tracking]: JS -> TS refactor the remaining .js file to .ts


SoonIter:
### What subject does this issue tracking?

# JS -> TS

1. refactor the remaining .js file to .ts
2. remove all ts-expect-error
3. use typescript rather than JSDocs

https://github.com/web-infra-dev/rspack/issues/4640

```[tasklist]
### Tasks
- [ ] https://github.com/web-infra-dev/rspack/pull/7132
- [ ] https://github.com/web-infra-dev/rspack/pull/7136
- [ ] packages/rspack/src/ModuleTypeConstants.js & packages/rspack/src/Template.js
- [ ] packages/rspack/src/lib
- [ ] packages/rspack/src/config
- [ ] packages/rspack/src/util
- [ ] packages/rspack/src/loader-runner
- [ ] packages/rspack/src/builtin-plugin/css-extract/hmr
- [ ] packages/rspack-dev-server/src/ansiHTML.js
```

SoonIter:
A very huge task, pr welcome~ :D

SoonIter:
Use https://github.com/alexcanessa/typescript-coverage-report to count the typescript coverage 

## Before

<img width="400" alt="image" src="https://github.com/user-attachments/assets/318d524a-b46a-48b8-8139-4977d8982fe7">

## After #7289 

Still remaining some ts files

<img width="400" alt="image" src="https://github.com/user-attachments/assets/7f1f6e7a-e991-4aee-916a-46affa00a8da">

our goal is reach >= 97%, so pr also welcome

wxiaoyun:
I will work on src/stats/StatsFactory.ts @shulaoda

shulaoda:
> I will work on src/stats/StatsFactory.ts @shulaoda

I am working on `src/logging/truncateArgs.ts` 👀

wxiaoyun:
Thanks for leaving utils/webpack 
Please let me work on it ❤️

wxiaoyun:
I will work on `src/node/nodeConsole.ts` as well :heart:

SoonIter:
after #7437

<img width="500" alt="image" src="https://github.com/user-attachments/assets/84920338-f302-4098-be09-9e9e190a1581">

our type coverage reach 95%, any pull request to increase type coverage is acceptable.

### type coverage contribution guide

1. run `pnpm install`
2. run `pnpm run type-coverage`
3. view files with low type coverage to contribute

[...]

PR #7258, opened 2024-07-22
refactor(typescript): packages/rspack/src/template from js to ts and esm

shulaoda:
## Summary
#7225 

## Checklist

### DIFF
--- a/packages/rspack/src/Template.ts
+++ b/packages/rspack/src/Template.ts
@@ -2,7 +2,6 @@
 	MIT License http://www.opensource.org/licenses/mit-license.php
 	Author Tobias Koppers @sokra
 */
-"use strict";
 
 const START_LOWERCASE_ALPHABET_CODE = "a".charCodeAt(0);
 const START_UPPERCASE_ALPHABET_CODE = "A".charCodeAt(0);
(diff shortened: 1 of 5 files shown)


shulaoda:
The file `Template.js` seems to have been copied directly from `webpack`, and some `modules` in `Rspack` have not been adapted, so I have removed some unnecessary code.

packages/rspack/src/Template.ts
```diff
@@ -0,0 +1,219 @@
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+"use strict";
```
SoonIter, line 5:
```suggestion
```

SoonIter:
> The file `Template.js` seems to have been copied directly from `webpack`, and some `modules` in `Rspack` have not been adapted, so I have removed some unnecessary code.

it is acceptable i think, could you add `The following code is modified based on` comments at the top ?

https://github.com/web-infra-dev/rspack/blob/de876babb550365b80cf904917471a58af72e541/packages/rspack/src/ErrorHelpers.ts#L1-L9

SoonIter (approved):
thanks 🫶🏻

PR #7286, opened 2024-07-23
perf: reduce large pre-allocations for `JavascriptParser::new`

h-a-n-a:
## Summary

`AsyncDependenciesBlock` is of the size of 208 bytes. It was pre-allocated for the capacity of 256:

```
size_of::<AsyncDependenciesBlock>() = 208
```

Taking a project with 10000 modules. It will at least allocate (and destroy) the memory of `256 * 208 * 10000` bytes (507Mbs), causing a memory increment and decrement back and forth.

Since the layout `AsyncDependenciesBlock` is quite compat, two approaches remain:
- Change `Vec<AsyncDependenciesBlock>` to `Vec<Box<AsyncDependenciesBlock>>`
- Change initial capacity of the `Vec` to a smaller value

The first solution is to reduce the memory usage from `256 * 208` to `256 * 24`, giving us a significant memory reduce, however with some potential overhead for sending the value to the heap.
With the second capacity change from `256` to `64`. 

I also changed the initial capacity of other vecs, although it does not(and should not) have much memory cost difference.

### Performance for big projects
The performance difference between these two changes of our internal project of 38134 modules remain **unnoticeable**. It's around 30 seconds for each build on my computer (Apple M2 Max, 64 GB).

### Memory cost

The memory cost of this is a const value for each module. With the difference number of modules, it results in different amount of allocations created and destroyed. Taking [10000 modules case](https://github.com/web-infra-dev/rspack-ecosystem-benchmark/tree/main/cases/10000) as an example, we can notice the difference of memory being created and destroyed during each build:

Before:
img_v3_02d2_979ee573-67e2-4211-8105-58e1a8a4d0eg

After:
img_v3_02d2_a89f6588-d5ab-49b0-8877-f80ac9e4e88g

It's 12.5% improvement of the total memory being allocated.

## Clippy complaints of `Vec<Box<T>> where T: Sized`

Clippy is complaining that `T` of `Vec<T>` was already allocated on the heap. Since it's a huge struct with that much of pre-allocation, we can workaround this as it was said in the detail page: https://github.com/rust-lang/rust-clippy/issues/3530

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/make/repair/build.rs
+++ b/crates/rspack_core/src/compiler/make/repair/build.rs
@@ -138,9 +138,9 @@ impl Task<MakeTaskContext> for BuildResultTask {
     let mut queue = VecDeque::new();
     let mut all_dependencies = vec![];
     let mut handle_block = |dependencies: Vec<BoxDependency>,
-                            blocks: Vec<AsyncDependenciesBlock>,
-                            current_block: Option<AsyncDependenciesBlock>|
-     -> Vec<AsyncDependenciesBlock> {
+                            blocks: Vec<Box<AsyncDependenciesBlock>>,
+                            current_block: Option<Box<AsyncDependenciesBlock>>|
+     -> Vec<Box<AsyncDependenciesBlock>> {
       for dependency in dependencies {
         let dependency_id = *dependency.id();
         if current_block.is_none() {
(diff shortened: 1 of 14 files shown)


h-a-n-a:
!bench

h-a-n-a:
!bench

ahabhgk (approved):

PR #7287, opened 2024-07-23
perf: faster hasher for `Ukey`s

ahabhgk:
## Summary

- perfer `IdentifierMap<V>` over `FxHashMap<Identifier, V>`
- introduce `UkeyHasher` for faster hash calculate for `Ukey`
- now `Ukey`s need there own COUNTER instead of using the same one (avoid u32 overflow)
- merge `rspack_identifier` and `rspack_database` into `rspack_collections`

Further more, we can also apply this to `HashMap<DependencyId, V>`, etc.

## Checklist


ahabhgk:
!bench

h-a-n-a (approved):

PR #7301, opened 2024-07-24
perf: reduce allocation for adding dependencies

h-a-n-a:
## Summary

(Also the heavy computation)

Dependency paths returned from oxc-resolver had already been normalized.
See: https://github.com/oxc-project/oxc-resolver/blob/f4c3ef5b4765ec8fcfb0f0d74a7562a6baa1456c/src/lib.rs#L380

Reduced memory cost for about 28.7% (at peak) for the case of [10000 modules](https://github.com/web-infra-dev/rspack-ecosystem-benchmark/tree/main/cases/10000) (Rspack was compiled with debug profile).

Before:
img_v3_02d3_ac33769d-7e05-409d-b613-ef293732b7eg

After:
img_v3_02d3_02b6b4ec-5889-4c24-839d-6e57f984ea9g

## Checklist

- [X] Tests updated (or **not required**).
- [X] Documentation updated (or **not required**).

### DIFF
--- a/crates/rspack_core/src/module_factory.rs
+++ b/crates/rspack_core/src/module_factory.rs
@@ -2,7 +2,6 @@ use std::{fmt::Debug, path::PathBuf, sync::Arc};
 
 use rspack_error::{Diagnostic, Result};
 use rustc_hash::FxHashSet as HashSet;
-use sugar_path::SugarPath;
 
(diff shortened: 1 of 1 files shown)


h-a-n-a:
!bench

PR #7304, opened 2024-07-24
chore: run `prebundle` when prepare

colinaaa:
## Summary

We don't need to run prebundle for every build, but only when dependencies change.

So moving to `prepare` would work and also speed up the build.

## Checklist


chenjiahan (approved):
make sense👍

PR #7305, opened 2024-07-24
feat(dev-server): export package version

leimonio:
## Summary

I was looking in a convenient way of getting the version of the `@rsapck/dev-server` similar to what is available in `@rspack/core`

Export version from `package.json` (similar to the `@rspack/core`)

## Checklist

### DIFF
--- a/packages/rspack/src/exports.ts
+++ b/packages/rspack/src/exports.ts
@@ -1,5 +1,13 @@
-const { version: rspackVersion, webpackVersion } = require("../package.json");
-export { rspackVersion, webpackVersion as version };
+import {
+	version as _version,
+	webpackVersion as _webpackVersion
+	// @ts-ignore 'package.json' is not under 'rootDir'
+} from "../package.json";
+
+const rspackVersion = _version as string;
+const version = _webpackVersion as string;
+
+export { rspackVersion, version };
 
 export type {
 	Asset,
--- a/packages/rspack/src/exports.ts
+++ b/packages/rspack/src/exports.ts
@@ -4,6 +4,7 @@ import {
 	// @ts-ignore 'package.json' is not under 'rootDir'
 } from "../package.json";
 
+// this is a hack to be compatible with plugin which detect webpack's version
 const rspackVersion = _version as string;
 const version = _webpackVersion as string;
 
--- a/packages/rspack/tsconfig.json
+++ b/packages/rspack/tsconfig.json
@@ -18,16 +18,11 @@
       "json-parse-even-better-errors": [
         "../compiled/json-parse-even-better-errors"
       ],
-      "webpack-sources": [
-        "../compiled/webpack-sources"
-      ]
+      "webpack-sources": ["../compiled/webpack-sources"]
     }
   },
   "include": ["src", "src/**/*.json"],
-  "exclude": [
-    "src/config/schema.check.js",
-    "src/container/default.runtime.js"
-  ],
+  "exclude": ["src/config/schema.check.js", "src/container/default.runtime.js"],
   "ts-node": {
     "transpileOnly": true
   },
--- a/packages/rspack-dev-server/etc/api.md
+++ b/packages/rspack-dev-server/etc/api.md
@@ -72,6 +72,8 @@ export class RspackDevServer extends WebpackDevServer {
     // (undocumented)
     staticWatchers: FSWatcher[];
     // (undocumented)
+    static version: string;
+    // (undocumented)
     webSocketServer: WebpackDevServer.WebSocketServerImplementation | undefined;
 }
 
--- a/packages/rspack-dev-server/src/server.ts
+++ b/packages/rspack-dev-server/src/server.ts
@@ -12,6 +12,8 @@ import path from "node:path";
 import type { Server } from "node:http";
 import type { Socket } from "node:net";
 import { type Compiler, MultiCompiler } from "@rspack/core";
+// @ts-ignore 'package.json' is not under 'rootDir'
+import { version } from "../package.json";
 import type { FSWatcher } from "chokidar";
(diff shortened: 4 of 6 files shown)


SoonIter:
At present, users need to use the same version of `@rspack/core` and `@rspack/dev-server`

```json
"devDependencies": {
     "@rspack/core": "0.7.5",
     "@rspack/dev-server": "0.7.5"
}
```

Have you encountered any issues? :D

leimonio:
So getting the version from @rspack/core should do the job?

I believe even if this is the case, since the 2 modules are separate they'd still need to expose their own version.

SoonIter:
@chenjiahan cc

chenjiahan:
How about getting the version from `package.json`?

```js
import pkgJson from '@rspack/dev-server/package.json';

console.log(pkgJson.version);
```

chenjiahan:
> I believe even if this is the case, since the 2 modules are separate they'd still need to expose their own version.

Agree with this.

leimonio:
In this use case, there are different options of how ***@***.***/dev-server`
is loaded.
1. Direct import at the top of the file
2. Injected as a reference (for user to be able to customise it), which is
dynamic.

For the second case, I'd need to have programmatic access of the version
through the API.

What I'm doing at the moment is something like this, which rather hacky

version(): string {
// Resolve version from the rspack-dev-server package.json
try {
const root = findRoot(this.rspackDevServerModulePath);
const packageJsonPath = join(root, 'package.json');
const packageJson = require(packageJsonPath);
return packageJson.version;
} catch (err) {
return 'unknown';
}
}

packages/rspack-dev-server/src/index.ts
```diff
@@ -1,2 +1,7 @@
+import pkgJson from "@rspack/dev-server/package.json";
+
+const { version: rspackDevServerVersion } = pkgJson;
+export { rspackDevServerVersion };
```
chenjiahan, line 4: We can export `version` instead of `rspackDevServerVersion`

```suggestion
export { version };
```

packages/rspack/src/exports.ts
```diff
@@ -1,4 +1,6 @@
-const { version: rspackVersion, webpackVersion } = require("../package.json");
+import pkgJson from "@rspack/core/package.json";
```
chenjiahan, line 1: Relative path should be faster, no need to resolve

```suggestion
import pkgJson from "../package.json";
```

packages/rspack-dev-server/src/index.ts
```diff
@@ -1,2 +1,7 @@
+import pkgJson from "@rspack/dev-server/package.json";
```
chenjiahan, line 1:
```suggestion
import pkgJson from "../package.json";
```

packages/rspack-dev-server/src/index.ts
chenjiahan: I think there is no difference between `import { version } from '@rspack/dev-server'` and `import pkgJson from '@rspack/dev-server/package.json'`.

If we need to support `this.RspackDS.version`, maybe we need add the version to somewhere else?

packages/rspack/src/exports.ts
SoonIter: image

I think "@rspack/core/package.json" is better for maintainability and "nodejs" resolve spec

packages/rspack/tsconfig.json
```diff
@@ -18,16 +18,11 @@
...
-      ]
+      "webpack-sources": ["../compiled/webpack-sources"]
     }
   },
-  "include": ["src", "src/**/*.json"],
-  "exclude": [
-    "src/config/schema.check.js",
-    "src/container/default.runtime.js"
-  ],
+  "include": ["src", "src/**/*.json", "package.json"],
```
SoonIter, line 24:
```suggestion
  "include": ["src", "src/**/*.json"],
```

leimonio:
Export from @rspack/core exports webpack version as version, and @rspack/dev-server exports rspack dev server as version. This will be confusing, as I'd expect both modules to reference the respective package version.

packages/rspack/src/exports.ts
chenjiahan: What do you mean by `"nodejs" resolve spec`, any document for this?

chenjiahan:
`@rspack/core` exports the webpack version to be compatible with the webpack ecosystem, and `@rspack/dev-server` does not need this. Therefore the meaning of their exported `version` are different.

chenjiahan:
@leimonio I want to confirm how `this.RspackDS` is imported?

1. `import * as RspackDS from '@rspack/dev-server';`
2. `import RspackDS from '@rspack/dev-server';`

leimonio:
In my case it's a bit more abstracted in this way:

```
export class RspackDevServer implements DevServer {
  private readonly Rspack: typeof RspackCompiler;
  private readonly RspackDS: typeof RspackDS;

  id = 'rspack-dev-server';

  displayName = 'Rspack dev server';

  constructor(
    private config: Configuration,
    private rspackModulePath: string = require.resolve('@rspack/core'),
    private rspackDevServerModulePath: string = require.resolve(
      '@rspack/dev-server'
    )
  ) {
    this.Rspack = require(this.rspackModulePath);
    this.RspackDS = require(this.rspackDevServerModulePath).RspackDevServer;
  }
  ...
  } // end of class
  ```

**Comment by chenjiahan** [comment]
I see, so we need to add the version to the `RspackDevServer` class:

```js
export class RspackDevServer extends WebpackDevServer {
  static version = 'a.b.c';
}
```

And users can access the version via:

```js
this.RspackDS = require('@rspack/dev-server').RspackDevServer;
console.log(this.RspackDS.version);
```

chenjiahan (approved):
LGTM 👍🏻

PR #7307, opened 2024-07-25
chore: remove @rspack/plugin-minify

hardfist:
## Summary
@rspack/plugin-minify is not needed anymore, since we are compatible with terser-webpack-plugin

## Checklist


SoonIter (approved):

chenjiahan:
Added `breaking change` label, as this should be considered as a breaking change.

Issue #7146, opened 2024-07-14
[Bug]: Does not work with AWS Lambda with TopLevelAwait enabled (ESM)


cjnoname:
### System Info

 System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 1.97 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    npm: 10.8.1 - ~/.nvm/versions/node/v20.14.0/bin/npm
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 17.5
  npmPackages:
    @rspack/cli: 1.0.0-alpha.3 => 1.0.0-alpha.3 
    @rspack/core: 1.0.0-alpha.3 => 1.0.0-alpha.3 

### Details

When using AWS Lambda, there must be a function exported called "handler".

Normally, we can export the handler function without any issues.

However, with "TopLevelAwait" enabled, the bundled file does not work with AWS Lambda. This is because the "handler" function is exported as y = n ? Symbol.for("@reflect-metadata:registry") : void 0, which is not a regular function and therefore won't be recognized by AWS Lambda.

To be completely fair, ESModules natively support top-level await, which means we don't need to do extra work to build a separate implementation for it. Nonetheless, please help find a way to solve this issue.

Thanks,
CJ

### Reproduce link

https://github.com/cjnoname/example-esm-lambda

### Reproduce Steps

In the "lambda.ts" file:

When the line console.log(await bb()); is commented out, the handler function can be exported correctly in the build file, and it works properly on AWS Lambda.

However, when the line console.log(await bb()); is enabled, the handler function is built as y = n ? Symbol.for("@reflect-metadata:registry") : void 0, which is not a regular function and therefore won't be recognized by AWS Lambda.

Please run the npm build script to regenerate the file.

PR #7308, opened 2024-07-25
fix: add `await` to async module export

CPunisher:
## Summary

🤔 It's confusing why this code is missing.

webpack: https://github.com/webpack/webpack/blob/aa7174671e33cb61b52271d0806a0f7c9e0d08df/lib/library/ModuleLibraryPlugin.js#L83

closes https://github.com/web-infra-dev/rspack/issues/7146

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/configCases/output/library-async-module/lambda.js
+++ b/packages/rspack-test-tools/tests/configCases/output/library-async-module/lambda.js
@@ -2,11 +2,8 @@ const bb = async () => {
   return "asdsadasda";
 };
 
-console.log(await bb());
-
-console.log(123);
+await bb();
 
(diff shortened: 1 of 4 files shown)


packages/rspack-test-tools/tests/configCases/output/library-async-module/lambda.js
```diff
@@ -0,0 +1,12 @@
+const bb = async () => {
+  return "asdsadasda";
+};
+
+console.log(await bb());
+
+console.log(123);
```
ahabhgk, line 7: Let's remove these console

ahabhgk (approved):

chenjiahan:
thanks!

PR #7323, opened 2024-07-26
docs: include types from common mdx chunks

LingyuCoder:
## Summary

Add types of api data in documents. Each type should be a collapse panel and can expand by clicking, just like below:

image

## Checklist

### DIFF
--- a/website/components/Collapse.scss
+++ b/website/components/Collapse.scss
@@ -1,0 +1,14 @@
+.ant-collapse-header {
+  padding: 6px 10px !important;
+}
+
+.collapse-code-panel {
+  .ant-collapse-content-box {
+    padding: 0 !important;
+  }
+
+  .ant-collapse-content-box > div {
+    margin: 0 !important;
+    border-radius: 0 !important;
+  }
+}
--- a/website/docs/en/api/javascript-api.mdx
+++ b/website/docs/en/api/javascript-api.mdx
@@ -1,6 +1,10 @@
 import WebpackLicense from '@components/webpack-license';
-import { ApiMeta, Stability } from '@components/ApiMeta';
+import { Collapse, CollapsePanel } from '@components/Collapse';
 import { PackageManagerTabs } from '@theme';
+import { ApiMeta, Stability } from '@components/ApiMeta';
+import CompilerType from '../types/compiler.mdx';
+import WatchingType from '../types/watching.mdx';
+import StatsType from '../types/stats.mdx';
 
 <WebpackLicense from="https://webpack.js.org/api/node/" />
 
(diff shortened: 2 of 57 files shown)


website/components/Collapse.scss
```diff
@@ -0,0 +1,14 @@
+.ant-collapse-header {
+  padding: 6px 10px !important;
```
chenjiahan, line 2: the original padding of antd looks better..

website/docs/en/types/asset-info.mdx
```diff
@@ -0,0 +1,17 @@
+<>
```
chenjiahan: why do we need this fragment, it looks wired

website/docs/en/types/cache.mdx
```diff
@@ -0,0 +1,50 @@
...
+        computer: (fn: (err: Error, result: T) => void) => void,
+        callback: (err: Error, result: T) => void,
+      ): void;
+      providePromise<T>(
+        computer: (fn: (err: Error, result: T) => void) => void,
+      ): Promise<T>;
+};
+
+type CacheFacade = {
+getChildCache(name: string): CacheFacade;
```
chenjiahan: the ident is broken

website/docs/en/api/javascript-api.mdx
```diff
@@ -66,6 +70,16 @@ The `run` method is then used to kickstart all compilation work. Upon completion
 The API only supports a single concurrent compilation at a time. When using `run` or `watch`, call `close` and wait for it to finish before calling `run` or `watch` again. Concurrent compilations will corrupt the output files.
 :::
 
+<Collapse>
+  <CollapsePanel
+    className="collapse-code-panel"
```
chenjiahan, line 75: `collapse-code-panel` repeated many times, can we simplify it?

For example, replace it with `type="code"` or something else

website/components/Collapse.scss
LingyuCoder: Too big and will use too much space

website/docs/en/types/asset-info.mdx
LingyuCoder: bug of rspress

website/docs/en/types/cache.mdx
LingyuCoder: fixed, david`s solution not works

website/docs/en/api/javascript-api.mdx
LingyuCoder: this is original collapse antd component, class name is just to override its style of padding and margin. we do not need to wrap it as a new component

PR #7328, opened 2024-07-27
fix: rspack errors don't support the correct location

### LINKED ISSUE #7327: [Bug]: Rspack errors seem to not support the correct location
### Details
Related to #6051

image

### Reproduce link

_No response_

### Reproduce Steps

1. Config rspack.config.js
2. Import non-existent module

shulaoda:
## Summary

Closed #7327 

## Checklist


crates/rspack_error/src/graphical.rs
```diff
@@ -431,19 +431,23 @@ impl GraphicalReportHandler {
       self.theme.characters.hbar,
     )?;
 
+    let (source_line, source_col) = labels
+      .iter()
+      .find_map(|hl| {
+        lines
+          .iter()
+          .find(|line| line.span_applies(hl) && line.span_line_only(hl))
+          .map(|line| (line.line_number, self.visual_offset(line, hl.offset()) + 1))
```
h-a-n-a, line 440: In order to match the result of `Stats`, I think it would be better to use 0-based value for columns.

shulaoda: > In order to match the result of `Stats`, I think it would be better to use 0-based value for columns.

Is my understanding correct?

https://github.com/web-infra-dev/rspack/blob/768e43b9ed373511feacabb9dd3e34939bf8badd/crates/rspack_error/src/graphical.rs#L440-L442

h-a-n-a: This is great! Thank you.

h-a-n-a (approved):

Issue #7070, opened 2024-07-08
[Optimize]: generate stats.toJson maybe need optimization


easy1090:
### System Info

  System:
    OS: macOS 13.2.1
    CPU: (12) x64 Apple M2 Max
    Memory: 26.10 MB / 64.00 GB
    Shell: 5.8 - /usr/local/bin/zsh
  Binaries:
    Node: 16.20.0 - ~/.nvm/versions/node/v16.20.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.20.0/bin/yarn
    npm: 8.19.4 - ~/.nvm/versions/node/v16.20.0/bin/npm
    pnpm: 6.23.6 - ~/.nvm/versions/node/v18.19.0/bin/pnpm
  Browsers:
    Chrome: 126.0.6478.127
    Safari: 16.3
    
  Rspack: v0.7.4

### Details

`stats.toJson` takes 20s, statsOptions is :
```js
{
  all: false,
  chunks: true,
  modules: true,
  assets: true,
  chunkModules: true,
}

```

toJson profiling is:
<img width="1377" alt="image" src="https://github.com/web-infra-dev/rspack/assets/18437716/d7bcf2e5-9c9c-42dd-9fb1-0dd38fd98c87">

profile file:

[CPU.20240704.111213.75007.0.001.cpuprofile.zip](https://github.com/user-attachments/files/16123670/CPU.20240704.111213.75007.0.001.cpuprofile.zip)

### Reproduce Steps

- `rush update`
- `rushx build` with --json.

PR #7344, opened 2024-07-29
perf: stats to js speed

SyMind:
## Summary

Fix: #7070

Optimize the Speed of Generating Js Stats on the Rust Side:

1. Previously, I used getters to read on-demand to reduce the number of `JsObject` creations. However, tests have shown that creating objects with `getter`s performs worse than directly creating `JsObject`. Therefore, in this update, the stats options are passed to the Rust side, allowing Rust to determine whether specific data is needed.
2. Stats contain duplicate data, such as a large amount of common data between `chunkModules` and `modules`. Hence, we can use `Ref` to cache the common parts of the data, reducing the number of `JsObject` creations.

Benchmark Results Based on an Internal Project:

By invoking `stats.toJson()`, we measured the execution time in a project containing two compilations.

Before Optimization:
* First compilation: 15.705s
* Second compilation: 18.784s

After Optimization:
* First compilation: 8.475s
* Second compilation: 8.284s

## Checklist


crates/rspack_binding_values/src/value_ref.rs
```diff
@@ -0,0 +1,71 @@
+use std::ptr;
+
+use napi::sys::{napi_env, napi_value};
+use napi::{bindgen_prelude::ToNapiValue, check_status, sys, Result};
+
+pub struct Ref {
```
ahabhgk, line 6: @h-a-n-a need you checkout this

h-a-n-a: Please add some comments to this file, including the place it originally copied from and the changes applied in this file.

ahabhgk (approved):

PR #7363, opened 2024-07-30
fix: builtin:lightningcss-loader shuold keep loader query

JSerFeng:
## Summary

lightningcss-loader should keep loader query, eg: `builtin:lightningcss-loader??rules[3][1]` instead of `builtin:lightningcss-loader`, so that when used as inline-loader, options can be kept

## Checklist


packages/rspack/src/config/adapterRuleUse.ts
```diff
@@ -225,14 +225,15 @@ const getLightningcssLoaderOptions: GetLoaderOptions = (o, _) => {
 			o.targets = browserslistToTargets(browserslist(o.targets));
 		}
 
-		if (o.include) {
-			o.include = toFeatures(o.include as unknown as FeatureOptions);
+		if (o.include && typeof o.include === "object") {
+			o.include = toFeatures(o.include as unknown as FeatureOptions) as any;
```
chenjiahan, line 229: can we remove `as any`

ahabhgk (approved):

PR #7389, opened 2024-07-31
feat: add memory cache of javascript plugins

LingyuCoder:
## Summary

- Add memory cache for javascript plugins
- Add document of plugin cache

## Checklist


packages/rspack/src/lib/cache/MemoryCachePlugin.ts
```diff
@@ -0,0 +1,55 @@
...
+import { Cache } from "../Cache";
+
+export default class MemoryCachePlugin {
+	static PLUGIN_NAME = "MemoryCachePlugin";
+	apply(compiler: Compiler) {
+		const cache: Map<
+			string,
+			{
+				etag: string | null;
+				data: any;
```
chenjiahan, line 11: can we use `unknown` here? it is better than `any`

website/docs/en/api/objects/cache.mdx
```diff
@@ -0,0 +1,187 @@
...
+import Cache from '../../types/cache.mdx';
+import Columns from '@components/Columns';
+
+# Cache Object
+
+When writing Rspack plugins, you can use `compiler.getCache(name: string)` or `compilation.getCache(name: string)` to get the cache object which can share data in the build process. The cache data is stored on the `Compiler`, so it can be used in multiple `Compilation`s in the watch mode.
+
+:::warning Notice
+
+- Only available when `cache: true` which it is enabled in `mode="development"` by default.
```
chenjiahan, line 11:
```suggestion
- Only available when [cache: true](/config/cache) which it is enabled in `mode="development"` by default.
```

chenjiahan, line 11:
```suggestion
- Only available when [cache: true](/config/cache) which it is enabled in `mode="development"` by default.
```

website/docs/zh/api/objects/cache.mdx
```diff
@@ -0,0 +1,187 @@
...
+import Cache from '../../types/cache.mdx';
+import Columns from '@components/Columns';
+
+# Cache 对象
+
+在编写 Rspack 插件时，可通过 `compiler.getCache(name: string)` 或 `compilation.getCache(name: string)` 获取缓存对象，以在构建流程中共享数据。缓存数据存储在 Compiler 上，因此可在 Watch 模式下的多次 Compilation 中使用。
+
+:::warning 注意
+
+- 仅在 `cache: true` 时可用，在 `mode="development"` 时默认开启。
```
chenjiahan, line 11:
```suggestion
- 仅在 [cache: true](/config/cache) 时可用，在 `mode="development"` 时默认开启。
```

packages/rspack/src/lib/cache/MemoryCachePlugin.ts
LingyuCoder: fixed, copied from webpack type doc

chenjiahan (approved):

PR #7391, opened 2024-07-31
feat(diagnostic): improve diagnostics for swc wasm plugins

h-a-n-a:
## Checklist


crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -610,3 +642,25 @@ impl IntoSwcComments for SingleThreadedComments {
     }
   }
 }
+
+fn get_swc_core_version() -> String {
+  let workspace_toml = cargo_toml::Manifest::from_str(CARGO_TOML)
```
SoonIter, line 647: I have a small question to ask: Would the `cargo_toml` parsing here be optimized into a constant during compilation?

h-a-n-a: I think this can be achieved by using compile time environment variables: https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env

h-a-n-a: This is a good question. Lemme fix this ;-)

Cargo.toml
```diff
@@ -71,17 +71,15 @@ swc_node_comments   = { version = "=0.23.0" }
 
 [profile.dev]
 codegen-units = 16
-debug         = 2       # debug build will cause runtime panic if codegen-unints is default
+debug         = 2    # debug build will cause runtime panic if codegen-unints is default
 incremental   = true
-panic         = "abort"
```
chenjiahan, line 76: 55.9 MB -> 63.8 MB is a big degradation 😂, can you explain the benefits in more detail?

h-a-n-a: `panic = abort` aborts the program whenever a panic occurs. Only with `panic = unwind` enabled can we catch the panic and add help message to the original panic thrown by the swc. See: https://doc.rust-lang.org/cargo/reference/profiles.html#panic

chenjiahan (approved):

Cargo.toml
chenjiahan: Get, it should be helpful for error diagnosis, we can reduce the binary size in other ways

Issue #7269, opened 2024-07-22
[Tracking]: Rsnext support


hardfist:
### What subject does this issue tracking?
Run Next.js on Rspack
https://github.com/ScriptedAlchemy/rsnext/pull/1/files# the progress is tracked here
```[tasklist]
### RsNext support
- [x] compilation.modules[i].blocks support
- [ ] https://github.com/web-infra-dev/rspack/issues/8406
- [x] support module.size as function
- [ ] issue and excluded used together have some problem https://github.com/ScriptedAlchemy/rsnext/pull/1/files#diff-84d995a20cdd661a5748456a06eefe6d6a7ea79501d500462e265d7ddba6742eL576
- [ ] support missing hook https://github.com/ScriptedAlchemy/rsnext/pull/1/files#diff-84d995a20cdd661a5748456a06eefe6d6a7ea79501d500462e265d7ddba6742eL576
- [x] support compilation.entries
- [ ] support moduleGraph.getModule
- [x] support seal hook
- [ ] moduleGraph api(getresolvedModule,getIssue,getExportsInfo,getModule,getIncomingConnections,getParentModule,getOutgoingConnection)
- [ ] succeedEntry and failedEntry Hook
```

omar391:
Why is this issue closed? Its seems there's some existing task.

webbertakken:
https://github.com/ScriptedAlchemy/rsnext is archived.

There's an update here as well: https://github.com/module-federation/core/issues/3153#issuecomment-2982745135.

PR #7396, opened 2024-08-01
feat: support compilation.entries

SyMind:
## Summary

Relative #7269

This PR adds comprehensive support for the compilation.entries methods, including `get`, `set`, `delete`, `values`, `keys`.

This PR introduces two structures using NAPI: `XxxDTO` and `JsXxx`. 

- `XxxDTO` is used when a Rust instance is associated. It is primarily utilized by users for reading and modifying specific data, making it ideal to be defined as a class. Getters and setters are employed to access the original values.

- `JsXxx` is used when a Rust instance is not associated, making it naturally defined as a `JsObject`.

## Checklist


ahabhgk (approved):
Two remaining things:
1. This lead two code style for passing options to rust side: `RawXxx` and `JsXxx`, we should unify it.
2. Three node binding crate merge to one

PR #7400, opened 2024-08-01
fix: dollar sign in filename

LingyuCoder:
## Summary

the dollar sign `$` is special while doing string replacement in rust. need to escape it to `$$`.

> see [regexp doc](https://docs.rs/regex/latest/regex/struct.Regex.html#replacement-string-syntax) for more details

## Checklist

### DIFF
--- a/crates/rspack_core/src/options/filename.rs
+++ b/crates/rspack_core/src/options/filename.rs
@@ -5,8 +5,9 @@ use std::str::FromStr;
 use std::sync::Arc;
 use std::{borrow::Cow, convert::Infallible, ptr};
 
+use itertools::Itertools;
 use once_cell::sync::Lazy;
-use regex::{Captures, Regex};
+use regex::{Captures, NoExpand, Regex};
 use rspack_error::error;
 use rspack_macros::MergeFrom;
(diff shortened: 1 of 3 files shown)


crates/rspack_core/src/options/filename.rs
```diff
@@ -215,6 +216,10 @@ impl<F: LocalFilenameFn> Filename<F> {
   }
 }
 
+fn replace_all<'a>(regex: &Lazy<Regex>, haystack: &'a str, rep: &str) -> Cow<'a, str> {
+  regex.replace_all(haystack, rep.split("$").join("$$"))
```
h-a-n-a, line 220: Can this be optimized based on its call site? I believe this can be omitted with replacers without `$` in it and calling `Itertools::join` creates extra overhead of `String`s.

LingyuCoder: Seems can use NoExpand()

h-a-n-a (dismissed):
Looks great!

LingyuCoder:
duplicated with https://github.com/web-infra-dev/rspack/pull/7402

Issue #7397, opened 2024-07-31
[Bug]: @arcgis/core is not supported. Error message is displayed during compilation: Uncaught SyntaxError: Unexpected token ';' #7397


BrightX:
### Version

```sh
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
    Memory: 20.38 GB / 31.94 GB
  Browsers:
    Edge: Spartan (44.19041.3570.0), Chromium (125.0.2535.67)
    Internet Explorer: 11.0.19041.3570
  npmPackages:
    @arcgis/core: ^4.30.9 => 4.30.9
    @rsbuild/core: 1.0.1-beta.6 => 1.0.1-beta.6
```

## `yarn dev` 浏览器控制台报错：Uncaught SyntaxError: Unexpected token ';'

![image](https://github.com/user-attachments/assets/5cd7c9ae-6bd7-4c87-9308-e45a5389d814)

## `yarn build` 报错：

![image](https://github.com/user-attachments/assets/2e9a8b79-15a6-4fff-a552-0cbed4162e0e)

### Reproduce link

https://developers.arcgis.com/javascript/latest/get-started-npm/#api

### Reproduce Steps

`index.html`
```html
<div id="app">
  <div id="viewDiv"></div>
</div>
```

`index.js`
```js
import Map from "@arcgis/core/Map";
import MapView from "@arcgis/core/views/MapView";
import "@arcgis/core/assets/esri/themes/dark/main.css";
import "@esri/calcite-components/dist/calcite/calcite.css";

const map = new Map({
  basemap: "topo-vector"
});
const view = new MapView({
  container: "viewDiv", // reference the div id
  map: map,
  zoom: 4,
  center: [15, 65]
});

```

`css`
```css
html, body {
  margin: 0;
  padding: 0;
  border: none;
}

#app {
  height: 100vh;
  margin: 0;
  padding: 0;
  border: 0;
}

#viewDiv {
  height: 100vh;
}
```

SoonIter:
## investigation

I conducted a preliminary investigation. The npm package `@arcgis/core` releases the minified js product, which causes Rspack/webpack to easily make wrong analysis during scope analysis and variable replace. 
After I formatted this npm package via `npx prettier --write .`, it works

LingyuCoder:
[...]

PR #7410, opened 2024-08-01
fix: change some `walk_block_statement` to `walk_statement`

CPunisher:
## Summary

webpack only directly call `walkBlockStatement` at `walkStatement` and `walkClass`(StaticBlock).

The difference between calling directly and calling via `walk_statement` is that `walk_statement` updates the `statement_path` and `prev_statement` in `enter_statement`.

Omitting to update `prev_statement` results in `asi_safe` error, which causes an unexpected semicolon to be inserted after the comma in the `SequentialStatement`.

closes https://github.com/web-infra-dev/rspack/issues/7397

## Checklist


h-a-n-a (approved):
Thanks

Issue #7182, opened 2024-07-16
[Tracking]: enable biome lint and format rules


SoonIter:
### What subject does this issue tracking?

Recommended rules https://biomejs.dev/linter/rules/#recommended-rules

```[tasklist]
### Tasks
- [ ] https://github.com/web-infra-dev/rspack/pull/7061
- [ ] https://github.com/web-infra-dev/rspack/pull/7099
- [ ] https://github.com/web-infra-dev/rspack/pull/7103
- [ ] https://github.com/web-infra-dev/rspack/pull/7157
- [ ] https://github.com/web-infra-dev/rspack/pull/7160
- [ ] https://github.com/web-infra-dev/rspack/pull/7176
- [ ] https://github.com/web-infra-dev/rspack/pull/7181
- [ ] https://github.com/web-infra-dev/rspack/pull/7224
- [x] lint/complexity/useLiteralKeys
- [x] lint/style/useEnumInitializers
- [x] lint/a11y/noSvgWithoutTitle
- [x] lint/style/noUselessElse
- [x] lint/suspicious/noDoubleEquals
- [x] lint/complexity/noForEach
- [x] lint/style/useSelfClosingElements
- [x] lint/a11y/useButtonType
- [ ] remove "noMultipleSpacesInRegularExpressionLiterals": "off",
- [ ] remove "noStaticOnlyClass": "off",
- [ ] remove "noThisInStatic": "off",
- [ ] remove "noUselessConstructor": "off",
- [ ] remove "noUselessTernary": "off",
- [ ] remove "noBannedTypes": "off",
- [ ] remove "useOptionalChain": "off",
- [ ] remove "noUselessEmptyExport": "off",
- [ ] remove "noExtraBooleanCast": "off" -
- [ ] remove "noEmptyPattern": "off",
- [ ] remove "noConstructorReturn": "off",
- [ ] remove "noSwitchDeclarations": "off",
- [ ] remove "noInvalidUseBeforeDeclaration": "off",
- [ ] remove "noInnerDeclarations": "off",
- [ ] remove "noVoidTypeReturn": "off" -
- [ ] remove "noCommaOperator": "off",
- [ ] remove "useExponentiationOperator": "off",
- [ ] remove "useShorthandFunctionType": "off",
- [ ] remove "noArguments": "off",
- [ ] remove "noInferrableTypes": "off",
- [ ] remove "noVar": "off",
- [ ] remove "noParameterAssign": "off",
[...]

PR #7420, opened 2024-08-02
chore(infra/biome): enable noStaticOnlyClass

shulaoda:
## Summary

Related to #7182 

## Checklist

### DIFF
--- a/packages/rspack/src/NormalModule.ts
+++ b/packages/rspack/src/NormalModule.ts
@@ -63,8 +63,8 @@ const deprecateAllProperties = <O extends object>(
 	return newObj;
 };
 
-export class NormalModule {
-	static getCompilationHooks(compilation: Compilation) {
+export const NormalModule = {
+	getCompilationHooks(compilation: Compilation) {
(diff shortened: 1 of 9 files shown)


packages/rspack/src/NormalModule.ts
```diff
@@ -63,8 +63,8 @@ const deprecateAllProperties = <O extends object>(
 	return newObj;
 };
 
-export class NormalModule {
```
SoonIter, line 66: 🤨 In the fact, I don't like this rule, `NormalModule` seems useful as a type before but it is forced to use like `typeof NormalModule` now and I think it will be a little different from webpack

so it's an api change without benefits, what do you think

SoonIter (approved):

packages/rspack/src/NormalModule.ts
shulaoda: I think we don't have to follow the recommended rules for now when there is no benefit. 🤔

shulaoda: If you agree, I will close this PR. 😌

PR #7428, opened 2024-08-02
feat: support seal hook in js side

SyMind:
## Summary

feat: support seal hook in js side

relate: #7269

## Checklist


crates/node_binding/src/plugins/interceptor.rs
```diff
@@ -681,6 +684,14 @@ define_register!(
   kind = RegisterJsTapKind::CompilationAfterProcessAssets,
   skip = true,
 );
+define_register!(
+  RegisterCompilationSealTaps,
+  tap = CompilationSealTap<(), ()> @ CompilationSealHook,
+  cache = false,
+  sync = true,
```
ahabhgk, line 691: `sync` means at rust side you can't call it in async way, basically only `executeModule` need `sync: true`

ahabhgk (approved):

PR #7430, opened 2024-08-03
chore(infra/biome): enable noParameterAssign

shulaoda:
## Summary
Related to #7182 

## Checklist

### DIFF
--- a/packages/rspack-cli/src/commands/build.ts
+++ b/packages/rspack-cli/src/commands/build.ts
@@ -94,7 +94,7 @@ export class BuildCommand implements RspackCommand {
 								});
 						}
 					} else {
-						const printedStats = stats.toString(statsOptions as any);
+						const printedStats = stats.toString(statsOptions);
 						// Avoid extra empty line when `stats: 'none'`
 						if (printedStats) {
 							logger.raw(printedStats);
--- a/packages/rspack-cli/src/commands/build.ts
+++ b/packages/rspack-cli/src/commands/build.ts
@@ -94,7 +94,7 @@ export class BuildCommand implements RspackCommand {
 								});
 						}
 					} else {
-						const printedStats = stats.toString(statsOptions);
+						const printedStats = stats.toString(statsOptions as any);
 						// Avoid extra empty line when `stats: 'none'`
 						if (printedStats) {
 							logger.raw(printedStats);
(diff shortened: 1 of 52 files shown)


packages/rspack-cli/src/commands/build.ts
```diff
@@ -94,7 +94,7 @@ export class BuildCommand implements RspackCommand {
 								});
 						}
 					} else {
-						const printedStats = stats.toString(statsOptions);
+						const printedStats = stats.toString(statsOptions as any);
```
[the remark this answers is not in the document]
shulaoda: At that time, I had to do this because of the TS error, but now it seems to be better?

packages/rspack-lite-tapable/src/index.ts
```diff
@@ -122,15 +122,9 @@ export class HookBase<T, R, AdditionalOptions = UnsetAdditionalOptions>
...
-			if (interceptor.register) {
-				const newOptions = interceptor.register(options);
-				if (newOptions !== undefined) {
-					options = newOptions;
-				}
-			}
-		}
-		return options;
+		return this.interceptors.reduce((options, interceptor) => {
+			return interceptor.register?.(options) || options;
```
SoonIter, line 126:
```suggestion
			return interceptor.register?.(options) ?? options;
```

I think this modification is consistent with the previous behavior

shulaoda:
Please let me know when your review is completed, and I will update the branch all at once. @SoonIter

packages/rspack/src/MultiStats.ts
```diff
@@ -65,32 +62,36 @@ export default class MultiStats {
 			warningsCount: children.every(o => o.warningsCount),
 			errors: children.every(o => o.errors),
 			warnings: children.every(o => o.warnings),
-			children
+			children,
+			context: "",
+			version: ""
```
SoonIter, line 67: Is this necessary?

shulaoda: In order to prevent errors in the TS type.

shulaoda: These attributes were used, but we did not give them initial values.

shulaoda: You can notice `childOptions` attributes `context` and `version` below: 
https://github.com/web-infra-dev/rspack/blob/d686b72a22e1f44bc7e00d4837d50251176ea177/packages/rspack/src/MultiStats.ts#L71-L90

SoonIter: not a good code now due to webpack laziness, 

I think we need to replace `options: any` in the future

SoonIter:
> Please let me know when your review is completed, and I will update the branch all at once. @SoonIter

done, I like `.reduce` function refactor which is one of the highlights in this pull request

SoonIter (approved):

PR #7437, opened 2024-08-05
refactor(type): increase type-coverage

SoonIter:
## Summary

ci(infra/type): add type-coverage

The type coverage of each file needs to be >= 80%

https://github.com/web-infra-dev/rspack/issues/7225


chenjiahan:
I think it is better to use Biome to ban `any` because it executes faster and is more accurate

chenjiahan (approved):

PR #7485, opened 2024-08-07
feat: splitChunks support usedExports

JSerFeng:
### Docs: WIP

We have runtime optimization for used exports, for example when we have two entries named `foo` and `bar`, and a shared module `shared` which exports 2 values `foo` and `bar`.

`foo` imports `foo` from `shared`

`bar` imports `bar` from `shared`

with default behavior, there will be 2 `shared` modules in the final output, one contains only `foo` export for `foo` entry, and one contains only `bar` export for `bar` entry.

This is great. But when it comes to splitChunks, there will be problem.

if we have a cacheGroup that can select `shared`, then the `shared` module maybe extracted into a separate chunk, thus the runtime optimization is gone.

Now if you enable `splitChunks.usedExports`, when `shared` module is selected by a cacheGroup, the cacheGroup will calculate all possible combinations of how exports are used in different runtime.

`shared` with `foo` exported belongs to `foo` chunk, and with `bar` exported belongs to `bar` chunk, and if you didn't set `cacheGroups.{cacheGroup}.name`, then `shared` modules will still exist for both runtime.

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -982,6 +982,7 @@ export interface RawCacheGroupOptions {
   name?: string | false | Function
   reuseExistingChunk?: boolean
   enforce?: boolean
+  usedExports?: boolean
 }
 
 export interface RawCacheGroupTestCtx {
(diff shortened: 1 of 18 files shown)


JSerFeng:
!bench

ahabhgk (approved):

PR #7489, opened 2024-08-07
fix: reuseExistingChunk should be false by default

JSerFeng:
## Summary

The default value of reuseExistingChunk should be false by default (the same with webpack)

## Checklist


SyMind (approved):

crates/rspack_binding_options/src/options/raw_split_chunks/mod.rs
```diff
@@ -211,7 +211,7 @@ impl From<RawSplitChunksOptions> for rspack_plugin_split_chunks::PluginOptions {
               .automatic_name_delimiter
               .unwrap_or(overall_automatic_name_delimiter.clone()),
             filename: v.filename.map(Filename::from),
-            reuse_existing_chunk: v.reuse_existing_chunk.unwrap_or(true),
+            reuse_existing_chunk: v.reuse_existing_chunk.unwrap_or(false),
```
chenjiahan, line 214: This should be a breaking change.

 And webpack seems to use `true` by default?

https://webpack.js.org/plugins/split-chunks-plugin/#splitchunkscachegroupscachegroupreuseexistingchunk

image

JSerFeng: I think this is a ambiguous description, `reuseExistingChunk` is `true` only for `default` and `defaultVendor` cacheGroups, and `undefined` for other cache groups, you can see here https://github.com/webpack/webpack/blob/main/lib/optimize/SplitChunksPlugin.js#L779.

packages/rspack-test-tools/tests/configCases/split-chunks/custom-filename/a.js
```diff
@@ -17,8 +17,8 @@ it("should have files", () => {
...
 	expect(files).toContain("b.js");
 	expect(files).toContain("common-common1_js.js");
-	expect(files).toContain("otherName1.js");
-	expect(files).toContain("otherName2.js");
+	expect(files).toContain("common-common2_js.js");
+	expect(files).toContain("common-common3_js.js");
 	expect(files).toContain("shared-shared-shared1_js.js");
-	expect(files).toContain("shared-theName.js");
+	expect(files).toContain("shared-shared-shared2_js.js");
 });
```
JSerFeng, line 24: because reuseExistingChunk is false now, module in `otherName` chunk is moved to new splited chunk, so there is no `otherName` chunk anymore, `shared-theName` is the same

crates/rspack_binding_options/src/options/raw_split_chunks/mod.rs
chenjiahan: Get, I also wonder why not enable reuseExistingChunk for all cacheGroups by default. Is there any reason?

JSerFeng: Not sure, but I guess if all cache groups are reusing existing chunk, some chunks may not meet user's expectation

chenjiahan: This doesn't make sense. Rspack has always enabled `reuseExistingChunk` by default, and no one seems to have reported any issues with it. Perhaps we can test the impact of this in a real project.

JSerFeng: This option has no actual optimization, the most differences are just output names, you can see an example here

https://github.com/webpack/webpack.js.org/issues/2122#issuecomment-388609306

Also the reuseExistingChunk will have extra check, so the perf is a little bit slow,(just a little bit

chenjiahan: Get, so it has no effect on the bundle size

chenjiahan (approved):

crates/rspack_binding_options/src/options/raw_split_chunks/mod.rs
chenjiahan, line 1: Can we add documentation for `reuseExistingChunk` to make it more clear?

JSerFeng: Yeah, I'll modify them

website/docs/zh/plugins/webpack/split-chunks-plugin.mdx
```diff
@@ -283,3 +283,24 @@ module.exports = {
   },
 };
 ```
+
+#### splitChunks.cacheGroups.\{cacheGroup\}.reuseExistingChunk
+
+- **类型：** `boolean`
+- **默认值：** `false`
+
+是否重用现有的 chunk。举例
```
chenjiahan, line 292:
```suggestion
是否重用现有的 chunk。如果当前 chunk 包含已从主 bundle 中拆分出的模块，则它将被重用，而不是生成新的模块，这可能会影响 chunk 最终的文件名。举例：
```

website/docs/zh/plugins/webpack/split-chunks-plugin.mdx
```diff
@@ -283,3 +283,24 @@ module.exports = {
...
+chunk Foo: [ module A, module B ]
+chunk Bar: [ module B ]
+
+cacheGroup: {
+  test: /B/,
+  chunks: 'all'
+}
+```
+
+在 chunk Foo 和 chunk Bar 中的 module B 由于 cacheGroup 的配置，会被拆分到一个新的 chunk 中，该 chunk 只包含 module B，这个新 chunk 和 chunk Bar 包含的 module 完全一样，因此可以直接复用 chunk Bar。
```
chenjiahan, line 304:
```suggestion
由于 cacheGroup 的配置，在 chunk Foo 和 chunk Bar 中的 module B 会被拆分到一个新的 chunk 中，该 chunk 只包含 module B，这个新 chunk 和 chunk Bar 包含的 module 完全一样，因此可以直接复用 chunk Bar。
```

website/docs/zh/plugins/webpack/split-chunks-plugin.mdx
```diff
@@ -283,3 +283,24 @@ module.exports = {
...
+
+如果设置 reuseExistingChunk 为 `false` 那么 chunk Bar 和 chunk Foo 中的 module B 会被移到新 chunk 中，chunk Bar 由于不包含任何 module，会作为空 chunk 被删除。
```
chenjiahan, line 306:
```suggestion
如果设置 reuseExistingChunk 为 `false`，那么 chunk Bar 和 chunk Foo 中的 module B 会被移到新 chunk 中，chunk Bar 由于不包含任何 module，会作为空 chunk 被删除。
```

website/docs/en/plugins/webpack/split-chunks-plugin.mdx
```diff
@@ -293,3 +293,24 @@ module.exports = {
   },
 };
 ```
+
+#### splitChunks.cacheGroups.\{cacheGroup\}.reuseExistingChunk
+
+- **Type:** `boolean`
+- **Default** `false`
+
+Whether to reuse existing chunks when possible. For example:
```
chenjiahan, line 302:
```suggestion
Whether to reuse existing chunks when possible. If the current chunk contains modules already split out from the main bundle, it will be reused instead of a new one being generated. This can affect the resulting file name of the chunk. For example:
```

website/docs/zh/plugins/webpack/split-chunks-plugin.mdx
JSerFeng: Do we need the word `bundle` here, we don't have any explanation about what is bundle actually

chenjiahan: I copied this line from the webpack documentation, and webpack has a glossary for `bundle`. We can add it as well.

https://webpack.js.org/glossary/#b

JSerFeng: How about

`如果新 chunk 包含的模块和原 chunk 中拆分出的模块，则原 chunk 会被复用，不会生成新的模块，这可能会影响 chunk 最终的文件名。举例：`

chenjiahan: LGTM 😄

PR #7491, opened 2024-08-08
fix: make sure it would not read from property from an `undefined` value in DefinePlugin

xfsnowind:
## Summary

In the different environment, the keyword `this` would be assigned with different values.  One case I met is when I used the lib `cypress-rspack-dev-server` to run the cypress test, it reported the `undefined` error:

image

Add the question mark to all the properties to avoid this error.

## Checklist


h-a-n-a:
`plugin.apply` is supposed to receive a `Compiler`. Would you please share a MRE?

xfsnowind:
> `plugin.apply` is supposed to receive a `Compiler`. Would you please share a MRE?

OK, thanks for reply, will pass a compiler to `raw`

PR #7496, opened 2024-08-08
chore: remove useless swc_core features

JSerFeng:
## Summary

Remove some useless features of swc_core

## Checklist


Cargo.lock
chenjiahan, line 1: maybe this change can be marked as `perf` instead of `chore`? It should reduce the binary size.

h-a-n-a:
Looks good

Cargo.lock
JSerFeng: I tried diff the size using cargo-float but see no differences, these features may already be ignored by the compiler I suppose

chenjiahan: get

chenjiahan (approved):

PR #7513, opened 2024-08-09
fix(modern-module): correct use bailout reason

fi3ework:
## Summary

Should access `optimization_bailout` from mgm to get the final bailout reason vec.

## Checklist

### DIFF
--- a/crates/rspack_plugin_library/src/modern_module_library_plugin.rs
+++ b/crates/rspack_plugin_library/src/modern_module_library_plugin.rs
@@ -75,12 +75,17 @@ impl ModernModuleLibraryPlugin {
       .iter()
       .filter(|id| !concatenated_module_ids.contains(id))
       .filter(|id| {
-        let module = module_graph
-          .module_by_identifier(id)
+        let mgm = module_graph
+          .module_graph_module_by_identifier(id)
           .expect("should have module");
-        module
-          .get_concatenation_bailout_reason(&module_graph, &compilation.chunk_graph)
-          .is_none()
+        let reasons = &mgm.optimization_bailout;
+        reasons
+          .iter()
+          // We did want force concatenate entry point here.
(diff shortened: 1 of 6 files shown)


crates/rspack_plugin_library/src/modern_module_library_plugin.rs
```diff
@@ -75,12 +75,16 @@ impl ModernModuleLibraryPlugin {
...
+          .module_graph_module_by_identifier(id)
           .expect("should have module");
-        module
-          .get_concatenation_bailout_reason(&module_graph, &compilation.chunk_graph)
-          .is_none()
+        let reasons = &mgm.optimization_bailout;
+        reasons
+          .iter()
+          // We did want force concatenate entry point here.
+          .filter(|r| !r.contains("Module is an entry point"))
```
chenjiahan, line 85: Should we make this string a constant?

```js
pub const MODULE_BAILOUT_REASON: &str = "Module is an entry point";
```

chenjiahan: Also used in https://github.com/web-infra-dev/rspack/blob/6f19caa348d0ce2d9133b185281a18a08703b0d8/crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs#L857

fi3ework: Yes, this is copied from concatenation plugin. Since modern-module is still in actively development and there's a test case to ensure the message string consistency. I'd like to add an TODO for now, and turn it into a variable if the current implementaion is stable in the future. Is that OK?

```rs
// TODO: use constant variable to identify the reason.
```

chenjiahan: LGTM

JSerFeng (approved):

Issue #7490, opened 2024-08-07
[Bug]: watcher blocks CPU after each rebuild


vkurchatkin-miro:
### System Info

ystem:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 167.59 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.0 - ~/.volta/tools/image/node/20.11.0/bin/node
    Yarn: 4.1.0 - ~/.volta/tools/image/yarn/1.22.10/bin/yarn
    npm: 10.2.4 - ~/.volta/tools/image/node/20.11.0/bin/npm
  Browsers:
    Chrome: 127.0.6533.99
    Safari: 17.5
  npmPackages:
    @rspack/cli: 1.0.0-beta.1 => 1.0.0-beta.1
    @rspack/core: 1.0.0-beta.1 => 1.0.0-beta.1
    @rspack/plugin-react-refresh: 1.0.0-beta.1 => 1.0.0-beta.1

### Details

In our setup after each interactive rebuild the watcher is restarted and blocks the CPU for ~2 seconds. Subsequently HMR or live reloading is delayed for this amount of time.

It looks like watching is restarted from scratch after every rebuild here: https://github.com/web-infra-dev/rspack/blob/2bd905d47e731772365637188c929f5371e60267/packages/rspack/src/node/NodeWatchFileSystem.ts#L66

while it is not necessary and the watcher instance can be reused (`watchpack` supports `watch` to be called again so that it can only create watchers that don't exists). 

### Reproduce Steps

N/A

vkurchatkin-miro:
This fixed the issue for me:

```diff
index c78273a3f..906abb084 100644
--- a/packages/rspack/src/node/NodeWatchFileSystem.ts
+++ b/packages/rspack/src/node/NodeWatchFileSystem.ts
@@ -62,8 +62,9 @@ export default class NodeWatchFileSystem implements WatchFileSystem {
                if (typeof callbackUndelayed !== "function" && callbackUndelayed) {
                        throw new Error("Invalid arguments: 'callbackUndelayed'");
                }
-               const oldWatcher = this.watcher;
-               this.watcher = new Watchpack(options);
+               if (!this.watcher) {
+                       this.watcher =  new Watchpack(options);
+               }

                if (callbackUndelayed) {
                        this.watcher.once("change", callbackUndelayed);
[...]

PR #7522, opened 2024-08-09
feat: expose added/removed compilation.*_dependencies to js side

jerrykingxyz:
## Summary

Currently we can only get all of *_dependencies in both js and rust compilation which will make HMR renew watcher very slow. https://github.com/web-infra-dev/rspack/issues/7490

This PR will 
1. Add a new struct `incremental_info` to collect files added and removed during rebuild.
2. Update compilation.*_dependencies  in rust side to expose all file, added file, removed file.
``` rust
struct Compilation {
  pub fn file_dependencies(
      &self,
    ) -> (
      // all files
      impl Iterator<Item = &PathBuf>,
      // added file
      impl Iterator<Item = &PathBuf>,
      // removed file
      impl Iterator<Item = &PathBuf>,
    ) { ... }
}
```
3. Merge node binding `get_*_dependencies` method to dependenciesDTO. And you can access them by
``` js
compilation.#inner.dependencies().addedFileDependencies
compilation.#inner.dependencies().removedFileDependencies
```

## Checklist


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -329,72 +329,100 @@ impl Compilation {
...
+  ) -> (
+    impl Iterator<Item = &PathBuf>,
+    impl Iterator<Item = &PathBuf>,
+    impl Iterator<Item = &PathBuf>,
+  ) {
+    let all_files = self
       .make_artifact
       .file_dependencies
       .files()
-      .chain(
```
jerrykingxyz, line 337: file_dependencies in module_executor has been added to make_artifact by generate module when import_module executed.

crates/rspack_binding_values/src/dependency.rs
jerrykingxyz, line 1: Dependencies is a top-level structure in rspack_core , so move it to the top level of rspack_binding_values ​​to avoid conflicts with compilation.*_dependencies

LingyuCoder (approved):

Issue #7477, opened 2024-08-06
[Bug]: Rspack produces non-working bundles for simple code pattern


jerrys-msft:
### System Info

**System:**
  OS: Windows 11 10.0.22631
  CPU: (12) x64 Intel(R) Core(TM) i7-8086K CPU @ 4.00GHz
  Memory: 14.21 GB / 31.93 GB
**Binaries:**
  Node: 18.12.0 - C:\Program Files\nodejs\node.EXE
  npm: 8.19.2 - C:\Program Files\nodejs\npm.CMD
**Browsers:**
  Edge: Chromium (127.0.2651.74)
  Internet Explorer: 11.0.22621.3527

### Details

Rspack produces non-working bundles for the following code pattern:

``` JavaScript
// file.mjs
export const foo = 'methodName';
export class Base {
  [foo]() {
    console.log('Hello from Base!');
  }
}

// index.mjs
import { foo, Base } from './file.mjs'
class Derived extends Base {
  [foo]() {
    super[foo](); // <-- ERROR HERE
    console.log('Hello from Derived!');
  }
}
const instance = new Derived();
instance[foo]();
```

At runtime, the bundle fails at the indicated line. Reproduced on Windows with Rspack 1.0.0-beta.3.

### Reproduce link

https://github.com/jerrys-msft/rspack-repro

### Reproduce Steps

The following commands reproduce the bug on Windows:

```
> npm install
> npx rspack
> node dist\bundle.js
```

To produce the correct output, run the sample without bundling:

```
> node .
```

EDIT: Simplified sample code.

PR #7530, opened 2024-08-11
fix: check SuperProp in walk_call_expression

CPunisher:
## Summary

Webpack reguard `super[foo]()` as MemberExpression, but swc distinguishes them. So there should be an extra check.

closes https://github.com/web-infra-dev/rspack/issues/7477

## Checklist


JSerFeng:
Thanks 🥺

JSerFeng (approved):

PR #7533, opened 2024-08-12
fix: remove useless "module" external runtime

fi3ework:
## Summary

The code deleted in this pull request is located at [here](https://github.com/webpack/webpack/blob/001cab14692eb9a833c6b56709edbab547e291a1/lib/ExternalModule.js#L362) in webpack. However, the code will never be called in Rspack because it is not fully implemented. I am deleting it now to reduce the runtime code for Rslib, and we can reimplement it when we are ready.

Additionally, I attempted to delete the corresponding code in webpack, but no CI tests were broken.

## Checklist

### DIFF
--- a/crates/rspack_core/src/external_module.rs
+++ b/crates/rspack_core/src/external_module.rs
@@ -281,14 +281,10 @@ impl ExternalModule {
                 )
                 .boxed(),
               );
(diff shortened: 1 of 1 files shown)


JSerFeng:
rebased main, ci error is fixed on main branch

JSerFeng (approved):

PR #7544, opened 2024-08-13
perf: use browserslist-rs for lightningcss

JSerFeng:
## Summary

use browserslist-rs instead of browserslist on js side, so that we can gain slightly more initial setup performance

## Checklist


crates/rspack_binding_options/src/options/raw_builtins/raw_lightning_css_minimizer.rs
```diff
@@ -79,17 +79,15 @@ impl TryFrom<RawLightningCssMinimizerRspackPluginOptions> for PluginOptions {
...
-          samsung: t.samsung,
-        }),
+        targets: value
+          .minimizer_options
+          .targets
+          .map(|t| {
+            rspack_loader_lightningcss::lightningcss::targets::Browsers::from_browserslist(t)
+          })
+          .transpose()
+          .map_err(|e| rspack_error::error!("Failed to parser browserslist: {}", e))?
```
h-a-n-a, line 89:
```suggestion
          .map_err(|e| rspack_error::error!("Failed to parse browserslist: {}", e))?
```

packages/rspack-test-tools/tests/__snapshots__/StatsAPI.test.js.snap
```diff
@@ -52,7 +52,7 @@ Object {
         "main.js",
       ],
       "filteredModules": undefined,
-      "hash": "6264df4ebe23175182cf",
+      "hash": "333e59d94047c9b4ec6c",
```
JSerFeng, line 55: ChunkHash changes because browserslist-rs and browserslist have different output for `fully supports es6`, but they should be the same in the future when they updates to the same db version

ahabhgk (approved):

chenjiahan:
Note: This is a breaking change (targets?: `Targets | string[] | string` -> `string[] | string`)

But I think nobody uses the `Targets` type, which depends on installing the lightningcss package and makes no sense.

PR #7556, opened 2024-08-13
feat: unified use of rspack_glob

SyMind:
### Background

In this PR, we unified the glob crate used in Rspack into a new crate named `rspack_glob`, which affects the `SideEffectsFlagPlugin` and `CopyRspackPlugin`.

The code is copied from the `glob` crate with some modifications, some modifications following the approach from biome:
https://github.com/biomejs/biome/blob/main/crates/biome_service/src/matcher/pattern.rs

1. Added support for the `{a,b}` syntax.
2. Removed the syntax validation for `**` with preceding and trailing `/`. Instead of throwing an error, it now replaces `**` with `*`.

### Result

We do not need to unify the implementation of `glob` within Rspack:

1. In `SideEffectsFlagPlugin`, the `glob` library is used to match the expressions configured in `sideEffects` with the import filenames. There is no need to interact with the file system I/O. The requirements are fixed, such as case sensitivity and the necessity of delimiters.

2. In `CopyRspackPlugin`, the `glob` library uses user-configured expressions to traverse and find files matching the specified directory. To achieve this, it minimizes the I/O process as much as possible. In the JS ecosystem, `fast-glob` determines whether a `glob` expression is dynamic or static and performs partial matching of the `glob`. The Rust `glob` crate splits the `glob` based on delimiters.

3. In `CopyRspackPlugin`, it is expected to have more configuration options for `glob` matching, and its requirements are not fixed.

In summary, to ensure the best performance for `sideEffects` matching and to satisfy the flexible needs of `CopyRspackPlugin`, unification is unnecessary.

## Checklist


SyMind:
!bench

LingyuCoder (dismissed):

LingyuCoder (approved):

SyMind:
@shulaoda I made some mistakes. I habitually thought that `fast_glob` and `glob_match` had the same performance. I only tested the performance of `glob_match` running once, which is slower than the `matches` method of `glob`.
But `fast_glob` is faster than the `matches` method of `glob`, it seems that I need to look at this PR again.

But unfortunately, `fast_glob` is missing the following configuration. Can you add it?
- case_sensitive
- require_literal_leading_dot

shulaoda:
> But unfortunately, `fast_glob` is missing the following configuration. Can you add it?
> 
> * case_sensitive
> * require_literal_leading_dot

It's my pleasure.

shulaoda:
I'm sorry, due to some reasons, I may not be able to complete this task until this weekend. @SyMind

chenjiahan:
That's totally fine, take your time 😄

SyMind:
@shulaoda In the end, we found we should not let `fast-glob` support `case_sensitive` and `require_literal_leading_dot` feature. It will reduce performance.

Another point is that the performance of `fast-glob` can also be optimized. In Rspack, the `glob_match_with_brace` method is used. This method preprocesses the glob expression and then analyzes the existing optional branches, which are the sub-items of the brace expressions.

When this method matches a glob without internal braces, its performance is even lower than `glob_match`. I think this preprocessing can be canceled and start only when `{` is encountered.

Here is my attempt: [https://github.com/SyMind/glob-match/blob/main/src/lib.rs#L236](https://github.com/SyMind/glob-match/blob/main/src/lib.rs#L236)

shulaoda:
> When this method matches a glob without internal braces, its performance is even lower than `glob_match`. I think this preprocessing can be canceled and start only when `{` is encountered.
> 
> Here is my attempt: https://github.com/SyMind/glob-match/blob/main/src/lib.rs#L236

I will try to further optimize, thank you for your attempt.

PR #7567, opened 2024-08-14
feat(core): don't inject bundlerInfo in library mode

### LINKED ISSUE #7563: don't include rspack_version runtime into library mode
inject rspack_version runtime in library seems meaningless and it's bad for shaking,  consider don't inject runtime_code for library

hardfist:
## Summary
fix #7563, don't inject bundlerInfo in library mode

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/configCases/output/library-truthy/index.js
+++ b/packages/rspack-test-tools/tests/configCases/output/library-truthy/index.js
@@ -1,0 +1,7 @@
+it('should not inject bundlerInfo when library is truthy', () => {
+    const content = require('fs').readFileSync(
+		require('path').resolve(__dirname, "bundle0.js"),
+		"utf-8"
+	);
+    expect(content).not.toContain('__webpack_require__' + '.rv')
+})
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -46,6 +46,7 @@ import type {
 	Optimization,
 	Performance,
 	ResolveOptions,
+	RspackFutureOptions,
 	RuleSetRules,
 	SnapshotOptions
 } from "./zod";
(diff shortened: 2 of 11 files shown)


packages/rspack/src/config/defaults.ts
chenjiahan, line 1: The PR title seems incorrect

packages/rspack-test-tools/tests/configCases/output/library-truthy/index.js
```diff
@@ -0,0 +1,7 @@
+it('should not inject bundlerInfo when library is truthy', () => {
+    const content = require('fs').readFileSync(
+		require('path').resolve(__dirname, "bundle0.js"),
+		"utf-8"
+	);
+    expect(content).not.toContain('__webpack_require__' + '.rv')
```
hardfist, line 6: cause the code itself is bundled in test code, so we can't use __webpack_require__.rv directly here

fi3ework: Could test it in a hook but current implementation is concise enough (https://github.com/web-infra-dev/rspack/blob/98c0c1357eb36830bf3b0e290fc068899a59a150/packages/rspack-test-tools/tests/configCases/library/modern-module-export-char/rspack.config.js)

fi3ework (dismissed):

packages/rspack-test-tools/tests/configCases/output/library-truthy/index.js
hardfist: yes, but it's kind of weird to put test in config hook, because we can't put it('xxx') there, better solution would be we support some placeholder for testing like `__TEST_BUNDLE_CONTENT__` to save fs call here

LingyuCoder: perhaps we can use `test.config.js` to inject some global variables into the runner's global context

LingyuCoder (dismissed):

LingyuCoder (approved):

PR #7569, opened 2024-08-14
refactor: module hash

ahabhgk:
## Summary

introduce module (chunk graph module) hash to optimize codegen phase when rebuild, this add a new phase `create_module_hashes` to calculate module hash

- for now, we use it as cache key in codegen phase
- other phases like `process_runtime_requirements` will also use it as cache key in future PR
- will improve it to incremental to avoid the cost of cache read and cache key compare
- will improve it to only calculate affected modules hash when rebuild (need some refactor in make phase to achieve that), for now we calculate all modules hash when rebuild
- `create_module_hashes` cause start time slower a bit, but this is necessary and will improve it from other phases

## Checklist


ahabhgk:
!bench

LingyuCoder (approved):

PR #7570, opened 2024-08-14
refactor: drop support for non-UTF8 paths

h-a-n-a:
## Summary

Enforce paths passed to rspack to UTF-8. 

According to [`OsString`](https://doc.rust-lang.org/stable/std/ffi/struct.OsString.html):
> - On Unix systems, strings are often arbitrary sequences of non-zero bytes, in many cases interpreted as UTF-8.
> - On Windows, strings are often arbitrary sequences of non-zero 16-bit values, interpreted as UTF-16 when it is valid to do so.
> - In Rust, strings are always valid UTF-8, which may contain zeros.

Two paths types are received by rspack:
- Paths passed as `String` from JavaScript runtime (UTF-8 encoded)
- Paths passed from system APIs are platform specific (could be non UTF-8 encoded). This might happen on unix system and windows ([WTF-8](https://simonsapin.github.io/wtf-8) encoding) if the path contains a [surrogate byte sequence](https://simonsapin.github.io/wtf-8/#surrogate-byte-sequence). Most of the time it's UTF-8 encoded.

<details>
<summary>Performance</summary>

#### Ad-hoc test : String reference -> PathBuf / Utf8PathBuf -> String

1. `PathBuf::from` takes inputs without UTF-8 validations, `to_string_lossy` validates if `PathBuf` is a valid UTF-8 string.
2. `Utf8PathBuf::from` takes inputs without UTF-8 validations, `to_string` creates a `formatter` and call the `as_str` to validate(not if rustc version is >= 1.74 stable, https://github.com/camino-rs/camino/pull/93). (It's even faster to call `as_str().to_string()` instead)

```rs
let mut group = c.benchmark_group("match");
    for i in [10, 100, 1000, 10000] {
        let p = "i".repeat(i);
        group.bench_with_input(BenchmarkId::new("utf8 lossy", i), &p, |b, i| {
            b.iter(|| {
                let a = PathBuf::from(i);
                a.to_string_lossy().to_string();
            })
        });

        group.bench_with_input(BenchmarkId::new("utf8 camino", i), &p, |b, i| {
            b.iter(|| {
                let a = Utf8PathBuf::from(i);
                a.to_string();
            })
        });
    }
```

```
match/utf8 lossy/10     time:   [33.520 ns 33.636 ns 33.757 ns]
match/utf8 camino/10    time:   [39.122 ns 39.266 ns 39.405 ns]

match/utf8 lossy/100    time:   [67.696 ns 68.091 ns 68.521 ns]
match/utf8 camino/100   time:   [42.497 ns 42.864 ns 43.411 ns]

match/utf8 lossy/1000   time:   [428.14 ns 428.97 ns 429.80 ns]
match/utf8 camino/1000  time:   [146.31 ns 146.57 ns 146.82 ns]

match/utf8 lossy/10000  time:   [3.1279 µs 3.1323 µs 3.1368 µs]
match/utf8 camino/10000 time:   [617.73 ns 623.08 ns 628.75 ns]
```

#### Comparison: PathBuf -> Utf8PathBuf -> String vs. PathBuf -> String

This is the performance generated with `bench` command as shown below. 
Always choose not to call `assert_utf8` if these paths are only read once. UTF-8 validation is costy. 
For example: `file_dependencies`, etc. Check out this commit for the changes I made: https://github.com/web-infra-dev/rspack/pull/7570/commits/58b8b95005fbc2f3e0226f07f2231abc3231d5e6

</details>

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,7 +391,8 @@ dependencies = [
 [[package]]
 name = "camino"
 version = "1.1.7"
-source = "git+https://github.com/h-a-n-a/camino.git?branch=perf-as-str#0b3c118bacf96c8c4b4903a1c29e7d821c501a1b"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0ec6b951b160caa93cc0c7b209e5a3bff7aae9062213451ac99493cd844c239"
 
 [[package]]
 name = "cargo_toml"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -391,8 +391,7 @@ dependencies = [
 [[package]]
 name = "camino"
 version = "1.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e0ec6b951b160caa93cc0c7b209e5a3bff7aae9062213451ac99493cd844c239"
+source = "git+https://github.com/h-a-n-a/camino.git?branch=perf-as-str#0b3c118bacf96c8c4b4903a1c29e7d821c501a1b"
 
 [[package]]
 name = "cargo_toml"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -389,6 +389,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "camino"
+version = "1.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e0ec6b951b160caa93cc0c7b209e5a3bff7aae9062213451ac99493cd844c239"
+
+[[package]]
 name = "cargo_toml"
 version = "0.20.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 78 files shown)


h-a-n-a:
!benccch

h-a-n-a:
!bench

h-a-n-a:
!bench

h-a-n-a:
!bench

h-a-n-a:
This will not be considered as a breaking change. As we does not support Rust API in the first place and strings passed from NAPI runtime are always UTF-8 encoded.

h-a-n-a:
!bench

h-a-n-a:
!bench

h-a-n-a:
Turns out the regression is caused by https://github.com/web-infra-dev/rspack/commit/a5e321288eb8fc95dc0606375ba11fe39678f2a5#commitcomment-145406470. With the new camino implementation, the performance is actually even better than the old test: https://github.com/web-infra-dev/rspack/pull/7570#issuecomment-2288534475

h-a-n-a:
!bench

h-a-n-a:
bumped camino to support: https://github.com/camino-rs/camino/pull/93

JSerFeng (approved):

PR #7571, opened 2024-08-14
test(plugin): test html plugin with basic cases of html-webpack-plugin

LingyuCoder:
## Summary

Test HtmlRspackPlugin with [basic test cases](https://github.com/jantimon/html-webpack-plugin/blob/main/spec/basic.spec.js) of [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin).

## Checklist


packages/rspack-test-tools/tests/builtinCases/plugin-html/basic/__snapshots__/output.snap.txt
```diff
@@ -4,7 +4,7 @@
 <head>
     <meta charset="utf-8">
     <title>rspack</title>
-  <script src="/runtime.js" defer></script><script src="/main.js" defer></script><link href="/main.css" rel="stylesheet" /></head>
+  <script defer="defer" src="/runtime.js"></script><script defer="defer" src="/main.js"></script><link href="/main.css" rel="stylesheet" /></head>
   <body>
```
hardfist, line 8: it seems defer shouldn't have "defer" value here

hardfist (approved):

Issue #7422, opened 2024-08-02
[Tracking]: Improve dependency structure specification


shulaoda:
### What problem does this feature solve?

Among many dependencies, they mark the scope of the dependency as `span`, `range`, or `start&end`. This behavior is very chaotic in Dependency.

https://github.com/web-infra-dev/rspack/blob/f37741c316c30e6234cfc504efbabea53a26df75/crates/rspack_plugin_javascript/src/dependency/esm/harmony_export_header_dependency.rs#L11-L16

https://github.com/web-infra-dev/rspack/blob/f37741c316c30e6234cfc504efbabea53a26df75/crates/rspack_plugin_javascript/src/dependency/esm/harmony_import_dependency.rs#L43-L54

https://github.com/web-infra-dev/rspack/blob/f37741c316c30e6234cfc504efbabea53a26df75/crates/rspack_core/src/dependency/cached_const_dependency.rs#L7-L12

This `0-based` range is based on the `1-based` SWC's `Span` conversion, and I think we should name it `range` which used in `webpack` to avoid ambiguity.

In addition, I think it may be better to name the types of `loc` and `range` as `DependencyLocation` and `DependencyRange`.

What's your opinion? @h-a-n-a 

### What does the proposed API of configuration look like?

```rust
pub struct Dependency {
  loc: DependencyLocation,
  range: DependencyRange
}
```
```[tasklist]
### Tasks
- [ ] https://github.com/web-infra-dev/rspack/issues/7336
- [ ] https://github.com/web-infra-dev/rspack/pull/7576
- [ ] https://github.com/web-infra-dev/rspack/pull/7635
- [ ] https://github.com/web-infra-dev/rspack/pull/7641
- [ ] https://github.com/web-infra-dev/rspack/pull/7671
- [ ] https://github.com/web-infra-dev/rspack/pull/7841
- [ ] https://github.com/web-infra-dev/rspack/pull/7871
- [ ] https://github.com/web-infra-dev/rspack/pull/7892
```

h-a-n-a:
Thanks for filing this issue! The dependency struts are indeed chaotic and a messed-up.

For the status-quo, we do over-used the word `Span`. To me, `Span` is more likely to be a specific word using to **describe the internal data of a compiler**(SWC in this case) and if this was to used in other cases, it will definitely mess up the definition. 

[...]

PR #7576, opened 2024-08-14
refactor: improve esm harmony dependencies structure

shulaoda:
## Summary
Related to #7422 

This job is a bit large, and I have temporarily managed dependencies that already have `loc`. I need to evaluate whether the current approach is acceptable.

## Checklist


shulaoda:
The `DependencyLocation` in `webpack` is as follows. At present, I have only considered `RealDependencyLocation` because I have not encountered the situation of `SyntheticDependencyLocation` yet, and it seems that this can be considered to be ignored.
```rust
/**
 * @typedef {object} RealDependencyLocation
 * @property {SourcePosition} start
 * @property {SourcePosition=} end
 * @property {number=} index
 */

/**
 * @typedef {object} SyntheticDependencyLocation
 * @property {string} name
 * @property {number=} index
 */

/** @typedef {SyntheticDependencyLocation | RealDependencyLocation} DependencyLocation */
```

h-a-n-a (approved):

PR #7585, opened 2024-08-15
fix: "module-import" simply falls back to "module" when bails

fi3ework:
## Summary

update: port https://github.com/webpack/webpack/pull/18696.

---

Port https://github.com/webpack/webpack/pull/18660, please take a look at the description from the webpack PR. This is my first time editing binding related code, hope didn't break any implicit things.

## Checklist


crates/rspack_binding_options/src/options/mod.rs
```diff
@@ -106,6 +108,7 @@ impl TryFrom<RawOptions> for CompilerOptions {
       dev_server: Default::default(),
       profile: value.profile,
       bail: value.bail,
+      externals_presets,
```
ahabhgk, line 111: Should not pass externals_presets to CompilerOptions

crates/rspack_core/src/external_module.rs
```diff
@@ -150,6 +151,31 @@ impl ExternalModule {
...
+          ExternalTypeEnum::Import => "import".to_string(),
+          ExternalTypeEnum::Module => "module".to_string(),
+        };
+      } else {
+        let presets = &dependency_meta.externals_presets;
+        if presets.web {
+          resolved_external_type = "module".to_string();
+        } else if presets.web_async {
+          resolved_external_type = "import".to_string();
+        } else if presets.node
```
ahabhgk, line 167: externals_presets should not effect the external_type here, externalsPresets is just a "preset" of externals, it should be resolved at config-level, instead of the core logic

fi3ework: As reference of https://github.com/webpack/webpack/pull/18620#issuecomment-2276069970.

> 1. Work with `externalsPresets`.
>    - when `externalsPresets` is a node.js-related presets, fallback to `node-commonjs`;
>    - when `externalsPresets === 'web'`, fallback to `module`;
>    - when `externalsPresets === 'web-async'`, fallback to `import`;
> 
>    If `externalsPresets` is not specified, we could set the fallback to `commonjs` (🤔 I'm not sure, this might break `module-import`'s semantics)
> 1. Adding `module-import-node-commonjs`, `module-import-import`, `module-import-module`, `module-import-commonjs`, maybe too verbose.
> 3. Adding a new config `exteralsFallback`, but this will only work with `"module-import"`, but adding an option on the top is a high tradeoff.

Plan 2 or 3 is better at decoupling the configs. Which one do u think is better (we still have chance to change the implementation before webpack's next release), or other better ideas?

ahabhgk: https://github.com/webpack/webpack/blob/main/lib/WebpackOptionsApply.js#L143-L155

Try using a function for dependency type awareness external type

crates/rspack_core/src/external_module.rs
```diff
@@ -469,21 +468,15 @@ impl Module for ExternalModule {
...
       "this" => build_result.build_info.strict = false,
       "system" => build_result.build_meta.exports_type = BuildMetaExportsType::Namespace,
       "script" | "promise" => build_result.build_meta.has_top_level_await = true,
-      "module" | "import" | "module-import" => match self.get_module_import_type(external_type) {
-        "module" => build_result.build_meta.exports_type = BuildMetaExportsType::Namespace,
-        "import" => {
-          build_result.build_meta.has_top_level_await = true;
-          build_result.build_meta.exports_type = BuildMetaExportsType::Namespace;
-        }
-        r#type => panic!(
```
ahabhgk, line 482: This should just ignore the request, instead of panic, module-import should only care about import and import(), for other dependency type, it's user's job to tell the compiler what external type to use (by config

ahabhgk (approved):

PR #7587, opened 2024-08-15
fix(node-stuff): use more conventional shim method

fi3ework:
## Summary

In previous implementation, when there's a `.` in the path, the `__webpack_fileURLToPath__(temp + '/..').slice(0, -1)
` will change like this:

```txt
file:///Users/abc/.pnpm/index.js

-- after __webpack_fileURLToPath__(temp + '/..') -->

/Users/abc/.pnpm/index.js/..

-- after slice(0, -1) -->

/Users/abc/.pnpm/index.js/.
```

Use the commonly used and battle tested way, also used by tsup https://github.com/egoist/tsup/blob/49c11c3073ce977a01c84e7848fc070d5de0a652/assets/esm_shims.js.

## Checklist


Timeless0911 (approved):

Timeless0911:
I need a canary version now.

PR #7588, opened 2024-08-15
feat(html): support js template and loader template

LingyuCoder:
## Checklist


zackarychapple:
Very nice site update! Great work

LingyuCoder:
The javascript function template has been support in https://github.com/web-infra-dev/rspack/pull/7704. Template with loaders need to be compiled by child compiler which may lead huge performance regression. We will try to find a better way to handle it, but before that you can juest use HtmlWebpackPlugin directly.

PR #7599, opened 2024-08-16
feat(rspack_plugin_swc_js_minimizer): add minify option

fi3ework:
## Summary

Add a new JS minimize option to support alter `minify` config in swc codegen, previously it was hard coded as `true`.

The usage is to enable some compress features, but without minify the whitespace in Rslib for now.

## Checklist


JSerFeng:
LGTM, @ahabhgk

crates/rspack_plugin_swc_js_minimizer/src/lib.rs
```diff
@@ -36,6 +36,7 @@ static JAVASCRIPT_ASSET_REGEXP: LazyLock<Regex> =
 
 #[derive(Debug, Hash)]
 pub struct PluginOptions {
+  pub minify: Option<bool>,
```
ahabhgk, line 39: Let's put this into minimizerOptions, since this is an option pass to swc minifier instead of consuming by the plugin

ahabhgk (dismissed):

ahabhgk (approved):

PR #7606, opened 2024-08-19
perf: warn case sensitive plugin

ahabhgk:
## Summary

For [9000](https://github.com/rspack-contrib/rspack-incremental-bench/tree/2966a89339890c414dcf265fb862a23033d4e5e2/cases/9000) case (without case sensitive path, which is the common case in development), 8ms -> 4ms

And now this plugin will increase 1ms per 2000 modules, which is acceptable for now, we can consider to move it as production only plugin by default when it becomes the bottleneck

## Checklist

### DIFF
--- a/crates/rspack_plugin_warn_sensitive_module/Cargo.toml
+++ b/crates/rspack_plugin_warn_sensitive_module/Cargo.toml
@@ -5,11 +5,14 @@ license     = "MIT"
 name        = "rspack_plugin_warn_sensitive_module"
 repository  = "https://github.com/web-infra-dev/rspack"
 version     = "0.1.0"
+
 [dependencies]
-rspack_core  = { version = "0.1.0", path = "../rspack_core" }
-rspack_error = { version = "0.1.0", path = "../rspack_error" }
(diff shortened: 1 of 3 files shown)


ahabhgk:
!bench

ahabhgk:
!bench

JSerFeng (approved):

crates/rspack_plugin_warn_sensitive_module/Cargo.toml
```diff
@@ -5,11 +5,14 @@ license     = "MIT"
 name        = "rspack_plugin_warn_sensitive_module"
 repository  = "https://github.com/web-infra-dev/rspack"
 version     = "0.1.0"
+
 [dependencies]
-rspack_core  = { version = "0.1.0", path = "../rspack_core" }
-rspack_error = { version = "0.1.0", path = "../rspack_error" }
-rspack_hook  = { version = "0.1.0", path = "../rspack_hook" }
-tracing      = { workspace = true }
+rspack_collections = { path = "../rspack_collections" }
```
hardfist, line 10: this need version, otherwise it can't be published

PR #7623, opened 2024-08-20
chore(infra/biome): enable noGlobalEval

shulaoda:
## Summary
Related to #7182 

I don't know if this is the right thing to do. Why use `eval`? Is there any special situation?

## Checklist

### DIFF
--- a/biome.jsonc
+++ b/biome.jsonc
@@ -48,9 +48,6 @@
 			"performance": {
 				"noDelete": "off"
 			},
-			"security": {
-				"noGlobalEval": "off"
-			},
 			"suspicious": {
 				"noFallthroughSwitchClause": "off",
 				"noConfusingVoidType": "off",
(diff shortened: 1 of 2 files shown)


shulaoda:
@SoonIter cc

SoonIter:
const dynamicImport = (specifier) => eval(`import(${specifier})`)
or  const dynamicImport = new Function('specifier', 'return import(specifier)');`

https://github.com/TypeStrong/ts-node/blob/ddb05ef23be92a90c3ecac5a0220435c65ebbd2a/src/test/resolver.spec.ts#L76-L81

it is a trick to avoid tsc transform the  dynamic import

SoonIter:
> it is a trick to avoid tsc transform the dynamic import

Sorry, this code is moved from the webpack and should be used to be compatible with node8.

this trick can be used in two cases: 

1. avoid `import` from being transformed by tsc to`Promise.resolve().then(() => require('foo'))`
2. node8

SoonIter (approved):

SoonIter:
image
You can check the dist. If the import in dist is not transformed by tsc, it can be merged.

shulaoda:
@SoonIter same

SoonIter:
LGTM ：D

PR #7630, opened 2024-08-20
perf: improve performance of eval-source-map

JSerFeng:
## Summary

using simd_json to convert string to json string, the improvement is small, but better than current, especially if the number of modules is super large.

encoding url always allocate new string, but in most cases there is no replacement at all, so using Cow to avoid unnessary allocation here.

here is the improvement I saw with my MacBook Pro M1, using this [repo](https://github.com/rspack-contrib/rspack-incremental-bench), CASE=9000

## Before
image

## After
image

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1170,6 +1170,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "float-cmp"
+version = "0.9.0"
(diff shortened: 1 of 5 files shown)


JSerFeng:
!bench

SyMind (approved):

PR #7631, opened 2024-08-20
refactor(core): remove unnecessary enum for resolver

hardfist:
## Summary
we only have one resolver implementation now so it's unnecessary to use enum to wrap resolver

## Checklist

### DIFF
--- a/crates/rspack_core/src/resolver/resolver_impl.rs
+++ b/crates/rspack_core/src/resolver/resolver_impl.rs
@@ -77,8 +77,8 @@ impl<'a> ResolveInnerOptions<'a> {
 ///
 /// Internal caches are shared.
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


crates/rspack_core/src/resolver/resolver_impl.rs
```diff
@@ -77,8 +77,8 @@ impl<'a> ResolveInnerOptions<'a> {
 ///
 /// Internal caches are shared.
 #[derive(Debug)]
-pub enum Resolver {
-  RspackResolver(rspack_resolver::Resolver),
+pub struct Resolver {
+  resolver: rspack_resolver::Resolver,
```
hardfist, line 81: will add pnp_manifest field in the following pr

ahabhgk (approved):

PR #7622, opened 2024-08-20
refactor: remove extra CommonJs dependency locations

shulaoda:
## Summary
Related to #7422 

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #7635, opened 2024-08-21
refactor: dependency location

shulaoda:
## Summary
Related to #7622 
> I will start governance again, and some dependencies can be directly addressed using `RealDependencyLocation`, as its location is actually determined. And in some places, we may need to use `SyntheticDependencyLocation` and `DependencyLocation`.
> 
> ```rust
> struct RealDependencyRange {
>   pub end: u32,
>   pub start: u32,
> }
> 
> struct SyntheticDependencyName {
>   pub name: String
> }
> 
> enum DependencyLocation {
>   Real(RealDependencyRange),
>   Synthetic(SyntheticDependencyName),
> }
> ```

Looks good. We might need `index` for `SyntheticDependencyName` as well in some places.

_Originally posted by @h-a-n-a in https://github.com/web-infra-dev/rspack/issues/7622#issuecomment-2300714328_
            

## Checklist


crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -0,0 +1,118 @@
...
+        start.line, start.column, end.line, end.column
+      ));
+    }
+
+    None
+  }
+}
+
+#[derive(Debug, Clone)]
+pub struct SyntheticDependencyName {
```
h-a-n-a, line 61: `SyntheticDependencyLocation`?

h-a-n-a: Let's align the name if there's nothing different between us and webpack.

crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -0,0 +1,118 @@
+use std::sync::Arc;
+
+use derivative::Derivative;
+
+#[derive(Derivative)]
+#[derivative(Debug, Clone)]
+pub struct RealDependencyRange {
```
h-a-n-a, line 7: What about `RealDependencyLocation`?

crates/rspack_core/src/dependency/dependency_location.rs
shulaoda: It seems that this structure has been implemented in the past, now let's temporarily use this name. I will resolve the conflict issue in the next PR, which won't be too long.

h-a-n-a: I see.

h-a-n-a (approved):
Thanks again!

PR #7637, opened 2024-08-21
perf: optimize small size runtime spec

ahabhgk:
## Summary

In most cases runtime spec should be small, it's no need to do heap alloc for small size runtime spec

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -332,6 +332,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "borsh"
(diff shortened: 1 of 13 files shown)


ahabhgk:
!bench

JSerFeng:
How about changing the `Arc<str>` to `smol_str` in RuntimeSpec, runtime string should be small as well

ahabhgk:
!bench

hardfist:
!bench

Issue #2236, opened 2023-03-14
[Feature Request]: Yarn PnP workspace


MarshallChen:
### System Info

  System:
    OS: macOS 12.6.4
    CPU: (8) x64 Intel(R) Core(TM) i7-4770HQ CPU @ 2.20GHz
    Memory: 2.54 GB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 14.17.4 - /private/var/folders/wv/hb8dy4gs4xq51jx7k9bbtthm0000gn/T/xfs-ce0be687/node
    Yarn: 3.4.1 - /private/var/folders/wv/hb8dy4gs4xq51jx7k9bbtthm0000gn/T/xfs-ce0be687/yarn
    npm: 6.14.14 - ~/.nvm/versions/node/v14.17.4/bin/npm

### Details

is yarn workspace pnp mode supported already or its not support currently. I got lots of errors like this.

```bash

error[internal]: Resolve error
   ┌─ src/components/SomeComponent.js:8:1
   │
 8 │ ╭ import React, {
 9 │ │   useCallback,
10 │ │   useMemo,
11 │ │   useState,
12 │ │   useRef,
13 │ │ } from 'react';
   │ ╰───────────────^ Failed to resolve react in /my/local/absolute/path/to/some/other/Component
```

### Reproduce Steps

```
NODE_ENV=development yarn rspack serve --config ./rspack.config.js
```

bvanjoi:
This feature is not yet supported

MarshallChen:
@bvanjoi Thank you for your prompt response. Could you please let me know if this feature is planned for development and if so, can you provide an estimated timeline for release?

bvanjoi:
> @bvanjoi Thank you for your prompt response. Could you please let me know if this feature is planned for development and if so, can you provide an estimated timeline for release?

We have no plans to develop this for the time being as there are very few people using PnP

MarshallChen:
Thats bad news, I don't know how big the PnP community is, but since webpack v5 supports yarn PnP by default which means if you guys plan to support this it would be awesome.

bvanjoi:
> Thats bad news, I don't know how big the PnP community is, but since webpack v5 supports yarn PnP by default which means if you guys plan to support this it would be awesome.

[...]

PR #7639, opened 2024-08-21
feat(core): add yarn pnp support

hardfist:
# Summary

This pull request implements the [Yarn PnP](https://yarnpkg.com/features/pnp) feature and closes https://github.com/web-infra-dev/rspack/issues/2236. Special thanks to @arcanis for implementing PnP resolution in [rspack_resolver](https://github.com/web-infra-dev/rspack-resolver).

### Intercepting FS on the Rust Side

A significant part of Webpack’s implementation is that it leverages PnP runtime to intercept all filesystem calls in Node.js via the .pnp.cjs file, correctly handling Yarn’s zip file paths as regular files. However, intercepting all filesystem calls on the JavaScript side incurs a significant performance cost in Rspack. Therefore, for performance reasons, Rspack intercepts standard filesystem calls in a custom native FS implementation in rust side.

### Enabling PnP Only When resolve.enablePnp is Set

For performance optimization, Yarn PnP is enabled by default only when running in a PnP environment (when `process.versions.pnp` is truthy). This avoids unnecessary overhead for users not utilizing Yarn PnP.

## Limitations

After discussions with @arcanis, supporting multiple Yarn PnP resolver instances would complicate the implementation and negatively impact cache efficiency. Since multiple PnP instances are rarely used in most Yarn PnP applications, Rspack supports only a single PnP resolver instance. Consequently, only the root `resolve.enablePnp` setting is effective, and settings within module.rule have no impact. This approach suffices for most Yarn PnP applications.

## Implementation Caveats

Yarn PnP requires filesystem interception for zip file calls, coupled the filesystem implementation with `resolve.enablePnP` configuration. Introducing a root `enablePnP` setting might decouple the filesystem and resolver but could also add confusion and complexity to the configuration. Therefore, I have adhered to the current implementation.

## Checklist


hardfist:
!bench

packages/rspack/src/ResolverFactory.ts
```diff
@@ -9,29 +9,34 @@ type ResolveOptionsWithDependencyType = Resolve & {
...
+	context: string;
 
 	static __to_binding(
 		resolver_factory: ResolverFactory
 	): binding.JsResolverFactory {
 		return resolver_factory.#binding;
 	}
 
-	constructor() {
+	constructor(context: string) {
```
hardfist, line 21: this seems break ResolverFactory api, I'm not sure whether this api is exposed to user side

hardfist: it's indeed a breaking js api of webpack, need find a better way

packages/rspack/etc/api.md
```diff
@@ -9822,10 +9822,12 @@ type ResolveRequest = BaseResolveRequest & Partial<ParsedIdentifier>;
 
 // @public (undocumented)
 class ResolverFactory {
```
hardfist, line 9824: @h-a-n-a is this real public api? how users can access this api?

crates/node_binding/src/resolver_factory.rs
```diff
@@ -26,48 +27,66 @@ impl JsResolverFactory {
...
+  #[napi(
+    ts_args_type = "type: string, options?: RawResolveOptionsWithDependencyType,pnpManifestPath: string|undefined"
+  )]
   pub fn get(
     &mut self,
     r#type: String,
     raw: Option<RawResolveOptionsWithDependencyType>,
+    context: String,
   ) -> napi::Result<JsResolver> {
+    let pnp_manifest_path = find_closest_pnp_manifest_path(context);
```
h-a-n-a, line 75: This API seems to have a recursive I/O operation. Is it possible to make it opt-in?

packages/rspack-test-tools/tests/configCases/resolve/pnp/.yarn/install-state.gz
hardfist, line 1: @arcanis what's usage of this file, will this file affect resolve behavior?

crates/node_binding/src/resolver_factory.rs
hardfist: then we may need feature flag in rspack like
```js
module.exports = {
   pnp: true | false
}
``

Comment (hardfist) [reply crates/node_binding/src/resolver_factory.rs:75]: actually this can be passed from js side by `require('module').findPnpApi` and enhanced-resolve call it for every resolve call(which is time consuming, https://github.com/webpack/enhanced-resolve/blob/247edebc909040ff99200108f7165e57f4fac2b1/lib/ResolverFactory.js#L138),but better performance for monorepo(like multi .pnp.cjs support)
I think we can support bultin resolve plugin so uers don't use pnp shouldn't pay the cost of pnpapi call
```js
module.exports = {
   resolve: {
      plugins: ['builtin:resolve-pnp-pugin']
   }
}
```

crates/node_binding/binding.d.ts
```diff
@@ -136,7 +136,7 @@ export class JsResolver {
 
 export class JsResolverFactory {
   constructor()
-  get(type: string, options?: RawResolveOptionsWithDependencyType): JsResolver
+  get(type: string, options?: RawResolveOptionsWithDependencyType,pnpManifestPath: string|undefined): JsResolver
```
arcanis, line 139: Do you plan to move the `pnpManifestPath` into the option bag (here and in the other places where the manifest path is passed as free function)? Ideally the PnP-specific references should be kept to a minimum imo.

crates/node_binding/src/resolver_factory.rs
```diff
@@ -26,48 +27,66 @@ impl JsResolverFactory {
...
+    resolve_options: Resolve,
+    pnp_manifest_path: Option<PathBuf>,
+  ) -> Arc<ResolverFactory> {
     match &self.resolver_factory {
       Some(resolver_factory) => resolver_factory.clone(),
       None => {
-        let resolver_factory = Arc::new(ResolverFactory::new(resolve_options));
+        let pnp_manifest = pnp_manifest_path
+          .and_then(|x| load_pnp_manifest(x).ok())
+          .map(Arc::new);
```
arcanis, line 40: nit: Perhaps this function should return a `Result`, or panic if the PnP file exists but can't be loaded (since presumably it'd cause all resolutions to fail with cryptic "module not found" errors)?

crates/node_binding/src/resolver_factory.rs
arcanis: Note that we could make the implementation smarter, and only check directories that haven't been checked before, assuming we keep a cache of some sort (a `HashMap<Path, Option<Path>>` would be enough); this would trim most of the recursion.

packages/rspack-test-tools/tests/configCases/resolve/pnp/.yarn/install-state.gz
arcanis: No, it can be removed - it's just a bunch of precomputed internal data structures that make Yarn faster.

packages/rspack/src/ResolverFactory.ts
hardfist: it seem this is not a public api

crates/node_binding/binding.d.ts
hardfist: no, don't want to change resolveOption api signature, I'm not sure whether it's worthy to support manifest per request or manifest per project. if we need to support manifest per project, we need to pass it to resolveOptions at least in rust side.

crates/node_binding/src/resolver_factory.rs
hardfist: thanks for tip, yes, it need some api change

packages/rspack/etc/api.md
h-a-n-a: This is not according to the generated `api.md`

crates/node_binding/src/resolver_factory.rs
hardfist: do you think the cache logic should be implemented in `find_closest_pnp_manifest_path` or outside

arcanis:
Bump, I still have it in my backlog (I've been in on-and-off paternity leave since August, so my bandwidth has been limited - but I'm still very interested to bring it to the finish line).

hardfist:
> Bump, I still have it in my backlog (I've been in on-and-off paternity leave since August, so my bandwidth has been limited - but I'm still very interested to bring it to the finish line).

thank you, we'll continue investigating yarn pnp support

ZeWaka:
bump

crates/node_binding/src/resolver_factory.rs
```diff
@@ -21,8 +21,8 @@ pub struct JsResolverFactory {
 #[napi]
 impl JsResolverFactory {
   #[napi(constructor)]
-  pub fn new() -> napi::Result<Self> {
-    let input_filesystem = Arc::new(NativeFileSystem {});
+  pub fn new(enable_pnp: bool) -> napi::Result<Self> {
```
arcanis, line 24: nit: No strong opinion, but I'd suggest to make an option bag struct rather than pass `enable_pnp` as a raw boolean; if you need to add other options (which for a JS resolver is conceivable), you won't have to significantly change the API.

crates/rspack_core/src/options/resolve/mod.rs
```diff
@@ -95,6 +95,8 @@ pub struct Resolve {
   pub description_files: Option<DescriptionFiles>,
   /// If enforce_extension is set to EnforceExtension::Enabled, resolution will not allow extension-less files. This means require('./foo.js') will resolve, while require('./foo') will not.
   pub enforce_extension: Option<EnforceExtension>,
+  /// enable yarn pnp
+  pub enable_pnp: Option<bool>,
```
arcanis, line 99:
```suggestion
  /// If set, Yarn PnP resolution will be supported.
  pub enable_pnp: Option<bool>,
```

crates/rspack_fs/src/native_fs.rs
```diff
@@ -54,14 +70,56 @@ impl WritableFileSystem for NativeFileSystem {
...
   fn read(&self, path: &Utf8Path) -> Result<Vec<u8>> {
+    if self.options.enable_pnp {
+      let path = path.as_std_path();
+      let buffer = match VPath::from(path)? {
+        VPath::Zip(info) => self.pnp_lru.read(info.physical_base_path(), info.zip_path),
+        VPath::Virtual(info) => std::fs::read(info.physical_base_path()),
+        VPath::Native(path) => std::fs::read(&path),
+      };
+      return buffer.map_err(Error::from);
+    }
```
arcanis, line 99: I wonder if I could/should write a macro in the `pnp` crate that would make this code less boilerplate-y.

arcanis (commented):
Looks great to me! Just a couple of nits.

crates/node_binding/src/resolver_factory.rs
hardfist: actually pnp support is a special case and is global setting, all other js resolverOptions is passed using https://github.com/web-infra-dev/rspack/blob/3eaae99bb796ca5487d720947dcc73bd704c8467/packages/rspack/src/ResolverFactory.ts#L23, which supports other complex resolver setting and support multi resolver Instance with different resolve options

website/docs/en/config/resolve.mdx
```diff
@@ -369,3 +369,10 @@ When enabled, symlinked resources are resolved to their real path, not their sym
 - **Type:** `Record<string, Resolve>`.
 
 Customize the Resolve configuration based on the module type.
+
+## resolve.pnp
+
+- **Type:** `boolean`
+- **Default:** `!!process.versions.pnp`
+
+When enabled, it will enable [Yarn PnP](https://yarnpkg.com/features/pnp) resolution, It's enabled by default when [`!!process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) is true which means running in Yarn PnP environments.
```
chenjiahan, line 378:
```suggestion
When enabled, it will enable [Yarn PnP](https://yarnpkg.com/features/pnp) resolution.

It's enabled by default if [`!!process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) is `true`, which means the application is running in Yarn PnP environments.

Example:

```js title="rspack.config.js"
module.exports = {
  // ...
  resolve: {
    pnp: true,
  }
};
```
```

File: website/docs/zh/config/resolve.mdx [hunk 10]
```diff
@@ -369,3 +369,10 @@ module.exports = {
 - **类型：** `Record<string, Resolve>`.
 
 依据模块类型自定义 Resolve 配置。
+
+## resolve.pnp
+
+- **类型:** `boolean`
+- **默认值:** `!!process.versions.pnp`
+
+启用时，将使用 [Yarn PnP](https://yarnpkg.com/features/pnp) 算法进行路径解析，当 [`!!process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) 为 true 时，即在 Yarn PnP 环境中运行时，它默认是启用的。
```
chenjiahan, line 378:
````suggestion
启用时，将使用 [Yarn PnP](https://yarnpkg.com/features/pnp) 算法进行路径解析，当 [`!!process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) 为 true 时，即在 Yarn PnP 环境中运行时，它默认是启用的。
```
```suggestion
启用时，将使用 [Yarn PnP](https://yarnpkg.com/features/pnp) 算法进行路径解析。

当 [`!!process.versions.pnp`](https://yarnpkg.com/advanced/pnpapi#processversionspnp) 为 `true` 时（即应用在 Yarn PnP 环境中运行时），它默认是启用的。

```js title="rspack.config.js"
module.exports = {
  // ...
  resolve: {
    pnp: true,
  }
};
```
````

NyanHelsing:
nice work!!!

PR #7641, opened 2024-08-21
refactor: resolve conflicts for dependency location

shulaoda:
## Summary
Related to #7635 and #7422 

## Checklist

### DIFF
--- a/crates/rspack_core/src/chunk_group.rs
+++ b/crates/rspack_core/src/chunk_group.rs
@@ -16,20 +16,9 @@ use crate::{ChunkLoading, ChunkUkey, Compilation};
 use crate::{LibraryOptions, ModuleIdentifier, PublicPath};
 
 #[derive(Debug, Clone)]
-pub struct SyntheticDependencyLocation {
-  pub name: String,
-}
-
-#[derive(Debug, Clone)]
-pub enum OriginLocation {
-  Real(DependencyLocation),
-  Synthetic(SyntheticDependencyLocation),
-}
-
-#[derive(Debug, Clone)]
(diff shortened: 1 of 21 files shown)


shulaoda:
@h-a-n-a cc

crates/rspack_core/src/chunk_group.rs
```diff
@@ -15,21 +15,10 @@ use crate::{
...
-pub enum OriginLocation {
-  Real(DependencyLocation),
-  Synthetic(SyntheticDependencyLocation),
-}
-
 #[derive(Debug, Clone)]
 pub struct OriginRecord {
   pub module_id: Option<ModuleIdentifier>,
-  pub loc: Option<OriginLocation>,
+  pub loc: Option<DependencyLocation>,
```
h-a-n-a, line 21: If we call `loc` here, should we just align every `range` with `loc` as well?

shulaoda: We refer to `RealDependencyLocation` as range in `Dependency` because it serves as a `span`, utilizing its `start` and `end`, and can also be converted to `loc`.  

Here is the combination of `RealDependencyLocation` and `SyntheticDependencyLocation`, which is directly used as the `loc` for the `chunk` info, so it is not named as `range`.

shulaoda: What's your opinion?

h-a-n-a: That makes sense to me. Thanks for the explanation.

h-a-n-a (approved):

PR #7653, opened 2024-08-22
feat: add minify option to lightningcss-loader

witsaint:
## Summary

When we use buildin:lightningcss，The production code can only be compressed from the rule，Code space cannot be compressed, such as cssnano's normalizewhitespace feature

for the following reason

image

 The first argument to the `stylesheet.to_css`  method is a fixed false，

So，When we use rspack or rebuild, do we use `lightningcss` compression capabilities to the user,

At the same time, rsbuild can increase the scalability when built-in that

[rsbuild issure#3247](https://github.com/web-infra-dev/rsbuild/issues/3247)

## Checklist

- [✅] Tests updated (or not required).
- [✅] Documentation updated (or not required).


crates/node_binding/binding.d.ts
```diff
@@ -1319,6 +1319,7 @@ export interface RawLightningCssBrowsers {
 }
 
 export interface RawLightningCssMinimizerOptions {
+  Minify?: boolean
```
chenjiahan, line 1322: Should be `Minify`, have you updated this file manually? it is generated automatically.

crates/rspack_plugin_lightning_css_minimizer/src/lib.rs
```diff
@@ -233,7 +234,7 @@ async fn process_assets(&self, compilation: &mut Compilation) -> Result<()> {
             .map_err(|e| error!(e.to_string()))?;
           let result = stylesheet
             .to_css(PrinterOptions {
-              minify: true,
+              minify: minimizer_options.minify.unwrap_or(false),
```
chenjiahan, line 237: Why change this option to default `false`?

chenjiahan: A minimizer should minify code by default

chenjiahan: I think we do not need to add a `minify` option to the minimizer plugin, it does not make sense

witsaint: Thanks. I have checked out

chenjiahan (approved):
LGTM👍🏻

This option is useful when using `style-loader`, as the built-in `LightningCssMinimizerRspackPlugin` can not handle CSS files injected by `style-loader`, and users can use this option to minify CSS.

PR #7666, opened 2024-08-23
fix: `Path` pitfall for loader type

h-a-n-a:
## Summary

There's an pitfall for [`Path::ends_with`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.ends_with). It only matches by components. We need to either convert to `&str` or use `Path::extension`.

Use `Path::as_str` to match extensions instead of `Path::extension` for performance. The second one analyzes `components` first.

## Checklist

### DIFF
--- a/tests/webpack-test/cases/loaders/cjs-loader-type/index.js
+++ b/tests/webpack-test/cases/loaders/cjs-loader-type/index.js
@@ -1,0 +1,15 @@
+it("should pass package.json type to loader", function () {
+	expect(require("cjs/loader.js!")).toBe("commonjs");
+	expect(require("./loader.js!")).toBe("undefined");
+});
+
+it("should pass 'commonjs' type to loader for .cjs", function () {
+	expect(require("cjs/loader.cjs!")).toBe("commonjs");
+	expect(require("./loader.cjs!")).toBe("commonjs");
+	// ORIGINAL WEBPACK COMMENT: TODO need fix in v8 https://github.com/nodejs/node/issues/35889
+	// ORIGINAL WEBPACK COMMENT: TODO otherwise this test case cause segment fault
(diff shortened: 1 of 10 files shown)


tests/webpack-test/cases/loaders/cjs-loader-type/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+it("should pass 'commonjs' type to loader for .cjs", function () {
+	expect(require("cjs/loader.cjs!")).toBe("commonjs");
+	expect(require("./loader.cjs!")).toBe("commonjs");
+	// ORIGINAL WEBPACK COMMENT: TODO need fix in v8 https://github.com/nodejs/node/issues/35889
+	// ORIGINAL WEBPACK COMMENT: TODO otherwise this test case cause segment fault
+	// Turned on this as rspack checks extensions for loader type.
+	// So this will not fall into dynamic import which causes segment fault.
+	// See: crates/rspack_binding_options/src/plugins/js_loader/resolver.rs
+	expect(require("esm/loader.cjs!")).toBe("commonjs");
```
h-a-n-a, line 14: Turned on a case that is failed in webpack. As we have a different strategy on loader type matching.

h-a-n-a: This is the test case for the CI issue: https://github.com/web-infra-dev/rspack-ecosystem-ci/actions/runs/10521239196/job/29151639120#step:6:805

h-a-n-a: In rsbuild, they relied on `type: "module"` and loader extension `.cjs` to correctly determine loader types.

LingyuCoder (approved):

PR #7693, opened 2024-08-27
fix(core): don't panic when load config failed

hardfist:
## Summary
Rspack shouldn't panic when load config failed

image
this error message can be improved 

## Checklist


chenjiahan (commented):
Can you share a screenshot of current error stacks?

hardfist:
> Can you share a screenshot of current error stacks?

done, but the message still can be improved in following pr

chenjiahan (approved):

PR #7705, opened 2024-08-27
feat: support `"loose-unrecognized-keys"` for config validation

h-a-n-a:
## Summary

Support `"loose-unrecognized-keys"` for `RSPACK_CONFIG_VALIDATE`.

- `"loose-silent"`: ignore every config validation errors
- `"loose"`: warn every config validation errors
- `"loose-unrecognized-keys"`: only warn unrecognized keys(this is for compatibility for webpack configs) and treat other errors as `"strict"` mode does.
- `"strict"`: return every config validation errors

There's no breaking changes in this PR: `"loose"` is the default value of `RSPACK_CONFIG_VALIDATE` in rspack-cli. `"strict"` is the default value for calling compiler programmatically.

## Checklist

### DIFF
--- a/packages/rspack/src/util/validate.ts
+++ b/packages/rspack/src/util/validate.ts
@@ -59,7 +59,7 @@ function toValidationError(error: z.ZodError): ValidationError {
 	const issueSeparator = "$issue$";
 	const prefixSeparator = "$prefix$";
 	const validationErr = fromZodError(error, {
-		prefix: "Configuration error",
+		prefix: "Invalid configuration object. Rspack has been initialized using a configuration object that does not match the API schema.",
 		prefixSeparator,
 		issueSeparator
 	});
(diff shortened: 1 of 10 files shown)


packages/rspack/src/util/validate.ts
```diff
@@ -13,28 +13,66 @@ export function validate<T extends z.ZodType>(opts: any, schema: T) {
...
+		friendlyErr = toValidationError(res.error);
+		throw friendlyErr;
 	}
 }
 
+function toValidationError(error: z.ZodError): ValidationError {
+	const issueSeparator = "$issue$";
+	const prefixSeparator = "$prefix$";
+	const validationErr = fromZodError(error, {
+		prefix: "Configuration error",
```
chenjiahan, line 62: Can we align with webpack here:

`Invalid configuration object. Rspack has been initialized using a configuration object that does not match the API schema.`

h-a-n-a: Fixed: 
image

chenjiahan (approved):

PR #7706, opened 2024-08-27
docs(homepage): lazy load images

chenjiahan:
## Summary

Lazy load images for performance.

- before:

Screenshot 2024-08-27 at 17 27 36

- after:

Screenshot 2024-08-27 at 17 29 59

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#loading

## Checklist


SoonIter:
it is a track worth recording :D

PR #7710, opened 2024-08-28
feat(rspack-cli): support more flags for command serve

### LINKED ISSUE #7573: [Feature]: support more command-line flags

Rspack currently missing some cli flags which cause some migration pain for users, see https://birtles.blog/2024/08/14/lessons-learned-switching-to-rspack/#command-line-flags

> Rspack CLI includes a built-in dev server but unfortunately, the command-line flags differ from webpack-dev-server in a few places, notably missing options like --port and --hot/--no-hot.
To handle such cases I ended up using various --env values and checking them in rspack.config.js’s [configuration function](https://rspack.dev/config/index#exporting-a-configuration-function). This seems like a potentially unnecessary point of friction that could be eased in a future release.

we can align more flags  to make migration easier

seems no need to support all flags(some flags are rarely used and hard to understand), but we can support some common flags like
```
--port
--hot | --no-hot
--host
```

shulaoda:
## Summary

closes #7573 

## Checklist

### DIFF
--- a/packages/rspack-cli/src/commands/serve.ts
+++ b/packages/rspack-cli/src/commands/serve.ts
@@ -14,7 +14,30 @@ export class ServeCommand implements RspackCommand {
 		cli.program.command(
 			["serve", "server", "s", "dev"],
 			"run the rspack dev server.",
-			commonOptions,
+			yargs =>
+				commonOptions(yargs).options({
+					hot: {
+						coerce: arg => {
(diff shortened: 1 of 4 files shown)


h-a-n-a:
Thank you! Would you please add a test case?

h-a-n-a (approved):
Thank you!

PR #7713, opened 2024-08-28
test(infra): improve e2e test stability

### LINKED ISSUE #7683: Infra: improve e2e test stability
some of the e2e tests are very unstable which cause pain for release workflow
https://github.com/web-infra-dev/rspack/actions/runs/10553187120/job/29234939205#step:8:122

shulaoda:
## Summary

closes #7683 

It seems that the issue is caused by the CI attempting to perform a hot update on the `index.js` module, but the update fails because it is the entry module.

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/hotCases/css/recovery/index.js
+++ b/packages/rspack-test-tools/tests/hotCases/css/recovery/index.js
@@ -1,4 +1,4 @@
-import "./index.css";
+import "./main";
 
 it("css recovery", done => {
 	NEXT(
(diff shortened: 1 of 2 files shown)


shulaoda:
The first pipeline test passed, I will rebase and try again.

fi3ework:
But why does it only have a small probability of failure, and can only be reproduced in CI?

shulaoda:
> But why does it only have a small probability of failure, and can only be reproduced in CI?

I'm not sure yet.

chenjiahan (approved):
Thanks! hope the case will be stable after merging this 😄

chenjiahan:
If this doesn't resolve the issue, we can investigate further.

PR #7721, opened 2024-08-29
perf: incremental build chunk graph

JSerFeng:
## Summary

make code-splitting incremental, rebuild can cost less time than first build, only rebuild chunks that are needed

## Checklist

### DIFF
--- a/crates/rspack_core/src/build_chunk_graph/incremental.rs
+++ b/crates/rspack_core/src/build_chunk_graph/incremental.rs
@@ -1,0 +1,600 @@
+use std::{collections::HashSet, hash::BuildHasherDefault};
+
+use num_bigint::BigUint;
+use rspack_collections::{IdentifierHasher, IdentifierIndexSet, UkeySet};
+use rspack_error::Result;
+use tracing::instrument;
+
+use super::code_splitter::{CgiUkey, CodeSplitter, DependenciesBlockIdentifier};
+use crate::{
+  incremental::IncrementalPasses, is_runtime_equal, AsyncDependenciesBlockIdentifier,
+  ChunkGroupUkey, ChunkUkey, Compilation, GroupOptions, ModuleIdentifier, RuntimeSpec,
+};
+
+#[derive(Debug, Clone)]
+pub enum ChunkReCreation {
+  Entry(String),
+  Normal(NormalChunkRecreation),
+}
+
+#[derive(Debug, Clone)]
+pub struct NormalChunkRecreation {
+  pub block: AsyncDependenciesBlockIdentifier,
+  module: ModuleIdentifier,
+  pub cgi: CgiUkey,
+  chunk: ChunkUkey,
+}
+
+impl ChunkReCreation {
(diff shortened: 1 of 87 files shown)


JSerFeng:
!bench

crates/rspack_core/src/build_chunk_graph/incremental.rs
```diff
@@ -0,0 +1,605 @@
...
+
+    let modules = if compilation
+      .incremental
+      .can_read_mutations(IncrementalPasses::MAKE)
+    {
+      let affected_lock = compilation
+        .incremental
+        .mutations_read(IncrementalPasses::MAKE)
+        .expect("should get mutation")
+        .get_affected_modules_with_module_graph(&module_graph);
```
ahabhgk, line 445:
```suggestion
    let modules = if let Some(mutations) = compilation
      .incremental
      .mutations_write()
    {
      let affected_lock = mutations.get_affected_modules_with_module_graph(&module_graph);
```

ahabhgk: And if you have the write permission, then you need to collect all the mutations in your pass, currently we don't need to collect these for chunk graph, we have a separate global effect pass to find the changed modules by compare the all modules' blocks and module_id, but we might need to collect them if we can remove the separate pass

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -115,7 +115,7 @@ pub struct CompilationHooks {
 }
 
 #[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
-pub struct CompilationId(u32);
+pub struct CompilationId(pub u32);
```
ahabhgk, line 118: I'm more prefer to keep this private, and add a method to get the `u32`

crates/rspack_core/src/dependencies_block.rs
```diff
@@ -43,14 +43,26 @@ pub fn dependencies_block_update_hash(
 }
 
 #[derive(Debug, Clone, Copy, Hash, Eq, PartialEq, Ord, PartialOrd)]
-pub struct AsyncDependenciesBlockIdentifier(Identifier);
+pub struct AsyncDependenciesBlockIdentifier(pub Identifier);
```
ahabhgk, line 46: same here

crates/rspack_core/src/old_cache/local/code_splitting_cache.rs
```diff
@@ -38,7 +39,12 @@ where
     return Ok(());
   }
 
-  if !compilation.has_module_import_export_change() {
+  let has_change = compilation.has_module_import_export_change();
```
ahabhgk, line 42: IMO we can remove this once incremental build chunk graph get stable enough, thoughts?

JSerFeng: Agree

ahabhgk (approved):

PR #7739, opened 2024-08-30
perf: source map dev tool plugin

SyMind:
## Summary

Use Rayon for the process of converting source and source map to buffer.

## Checklist


SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

crates/rspack_plugin_devtool/src/source_map_dev_tool_plugin.rs
```diff
@@ -341,189 +340,193 @@ impl SourceMapDevToolPlugin {
...
-          None => data,
+
+        let mut asset = compilation
+          .assets()
+          .get(&filename)
+          .unwrap_or_else(|| {
+            panic!(
+              "expected to find filename '{}' in compilation.assets, but it was not present",
+              &filename
+            )
```
JSerFeng, line 367: Is there a way we make this diagnostic instead of panic

SyMind: Panic is better. `filename` should always be found in `assets`. Using `unwrap_or_else` instead of `expect` provides more information when a panic occurs.

JSerFeng (approved):

Issue #7719, opened 2024-08-28
[Bug]: export of comma-separated strings from scss modules does not work


dartess:
### System Info

```
  System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 140.42 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.1 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.4.0 - ~/Library/pnpm/pnpm
    bun: 1.0.28 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 128.1.69.153
    Chrome: 128.0.6613.85
    Chrome Canary: 130.0.6683.2
    Edge: 111.0.1661.54
    Safari: 17.6
    Safari Technology Preview: 18.0
  npmPackages:
    @rspack/cli: ^1.0.0 => 1.0.0 
    @rspack/core: ^1.0.0 => 1.0.0 
    @rspack/plugin-react-refresh: ^1.0.0 => 1.0.0 
```

### Details

After updating to version 1, this code stopped working:

```scss
$inter: 'Inter', sans-serif;

:export {
  inter: $inter;
}
```

exported data from this module is incorrect (does not contain `inter`)

### Reproduce link

https://github.com/dartess/rspack-import-scss-module

### Reproduce Steps

1. run `dev:rspack`
2. open localhost

expected result
`'Inter', sans-serif`

actual result
`undefined`

webpack and rspack@0.7.5 work fine

dartess:
fixed, thanks

PR #7743, opened 2024-08-30
fix: escape css ident

inottn:
## Summary

#7719 

## Checklist

### DIFF
--- a/crates/rspack_plugin_css/src/parser_and_generator/mod.rs
+++ b/crates/rspack_plugin_css/src/parser_and_generator/mod.rs
@@ -38,6 +38,9 @@ use crate::{
 static REGEX_IS_MODULES: LazyLock<Regex> =
   LazyLock::new(|| Regex::new(r"\.module(s)?\.[^.]+$").expect("Invalid regex"));
 
+static REGEX_IS_COMMENTS: LazyLock<Regex> =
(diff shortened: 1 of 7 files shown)


ahabhgk:
Looks good to me, could you add a test case, just for avoiding regression

crates/rspack_plugin_css/src/parser_and_generator/mod.rs
```diff
@@ -38,6 +38,9 @@ use crate::{
 static REGEX_IS_MODULES: LazyLock<Regex> =
   LazyLock::new(|| Regex::new(r"\.module(s)?\.[^.]+$").expect("Invalid regex"));
 
+static REGEX_IS_COMMENTS: LazyLock<Regex> =
+  LazyLock::new(|| Regex::new(r"/\*.*?\*/").expect("Invalid regex"));
```
JSerFeng, line 42: Can this match comments with multiple lines ?

```suggestion
  LazyLock::new(|| Regex::new(r"/\*[\s\S]*?\*/").expect("Invalid regex"));
```

JSerFeng (changes requested):

crates/rspack_plugin_css/src/parser_and_generator/mod.rs
inottn: Thanks, I agree.

JSerFeng (approved):

JSerFeng:
Thanks !

Issue #7419, opened 2024-08-02
[Bug]: Panic Module should be added before using (Location: crates/rspack_core/src/chunk_graph/chunk_graph_module.rs:74)


hverlin:
### System Info

`@rsbuild/core: 1.0.1-beta.9` (but happens on earlier version as well)

### Details

When I start `rsbuild`, it needs to run the `compiling...` step twice (I am not sure exactly why)

I have noticed that rspack will crash if I try to access the page before the second `compiling...` step is done. If I wait, the page will load fine.

```
start   Compiling...
● web ━━━━━━━━━━━━━━━━━━━━━━━━━ (100%) emitting after emit                                                                                                                ready   Compiled in 13.2 s (web)
start   Compiling...
● web ━━━━━━━━━━━━━━━━━━━━━━━━━ (70%) sealing dependencies                                                                                                                Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  Module(builtin:react-refresh-loader!builtin:swc-loader??ruleSet[1].rules[3].use[0]!/<..>./some-file.js) should be added before using
Location: crates/rspack_core/src/chunk_graph/chunk_graph_module.rs:74

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
Run with RUST_BACKTRACE=full to include source snippets.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: _ZN6v8impl12_GLOBAL__N_123FunctionCallbackWrapper6InvokeERKN2v820FunctionCallbackInfoINS2_5ValueEEE
    at <unknown source file>
 2: _ZN2v88internal25FunctionCallbackArguments4CallENS0_15CallHandlerInfoE
    at <unknown source file>
 3: _ZN2v88internal12_GLOBAL__N_119HandleApiCallHelperILb0EEENS0_11MaybeHandleINS0_6ObjectEEEPNS0_7IsolateENS0_6HandleINS0_10HeapObjectEEENS8_INS0_20FunctionTemplateInfoEEENS8_IS4_EEPmi
    at <unknown source file>
 4: _ZN2v88internal21Builtin_HandleApiCallEiPmPNS0_7IsolateE
    at <unknown source file>
[...]

PR #7761, opened 2024-09-02
fix(make): isolated module clean incomplete

jerrykingxyz:
## Summary

Fix the step of clean isolated module in `cutout` will only clean one layer module.

There are a example:
1. if we have a module graph `a` -> `b` -> `c`
2. Edit module `a` so that it does not depend on module `b`

Before this PR the module_graph will only remove module `b`, because of call `revoke_modules` after module isolated check.

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/make/cutout/clean_isolated_module.rs
+++ b/crates/rspack_core/src/compiler/make/cutout/clean_isolated_module.rs
@@ -31,8 +31,6 @@ impl CleanIsolatedModule {
   }
 
   pub fn fix_artifact(self, artifact: &mut MakeArtifact) {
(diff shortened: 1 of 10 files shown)


jerrykingxyz:
> may fix #7419 can you release a canary ?

@JSerFeng  1.0.1-canary-d6c74be-20240902095124

JSerFeng:
Can we add a test, for example using a plugin hook to print all modules after rebuild ?

jerrykingxyz:
> Can we add a test, for example using a plugin hook to print all modules after rebuild ?

Good Idea, i will add it

JSerFeng (approved):

PR #7762, opened 2024-09-02
perf: use itoa for integer to string conversion

shulaoda:
## Summary

Related to https://github.com/rolldown/rolldown/pull/2107

## Checklist

### DIFF
--- a/crates/rspack_binding_values/src/stats.rs
+++ b/crates/rspack_binding_values/src/stats.rs
@@ -328,7 +328,7 @@ impl From<(String, rspack_core::LogType)> for JsStatsLogging {
         args: Some(vec![format!(
           "{}: {} ms",
           label,
-          secs * 1000 + subsec_nanos as u64 / 1000000
+          itoa::Buffer::new().format(secs * 1000 + subsec_nanos as u64 / 1000000)
         )]),
         trace: None,
       },
@@ -355,8 +355,8 @@ impl From<(String, rspack_core::LogType)> for JsStatsLogging {
           } else {
             hit as f32 / total as f32 * 100_f32
           },
-          hit,
-          total,
+          itoa::Buffer::new().format(hit),
+          itoa::Buffer::new().format(total),
         )]),
         trace: None,
       },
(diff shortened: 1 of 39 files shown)


shulaoda:
@chenjiahan cc

chenjiahan:
!bench

crates/rspack_binding_values/src/stats.rs
```diff
@@ -328,7 +328,7 @@ impl From<(String, rspack_core::LogType)> for JsStatsLogging {
         args: Some(vec![format!(
           "{}: {} ms",
           label,
-          secs * 1000 + subsec_nanos as u64 / 1000000
+          itoa::Buffer::new().format(secs * 1000 + subsec_nanos as u64 / 1000000)
```
ahabhgk, line 331: Can we have a util function for this, `itoa(xx)` instead of `itoa::Buffer::new().format(xx)`

shulaoda: Of course, just a moment.

crates/rspack_util/src/itoa.rs
```diff
@@ -0,0 +1,8 @@
+pub use itoa::Buffer;
+
+#[macro_export]
+macro_rules! itoa {
```
ahabhgk, line 4: I prefer a function here (if possible) instead of a macro

shulaoda: The function can cause `temporary value` issues, and using `to_owned` may result in some performance loss. I am not sure how to better implement it.

crates/rspack_macros/src/merge.rs
```diff
@@ -86,8 +86,14 @@ fn body(data: Data, chose_base: Option<&Attribute>) -> Result<TokenStream> {
               .enumerate()
               .map(|(i, f)| {
                 let field_name = &f.ident;
-                let a_name = Ident::new(&format!("a_{i}"), field_name.span());
-                let b_name = Ident::new(&format!("b_{i}"), field_name.span());
+                let a_name = Ident::new(
+                  &format!("a_{}", itoa::Buffer::new().format(i)),
```
ahabhgk, line 90: This is inside macro, which runs in build time, not affect to performance, let's keep it as is

crates/rspack_util/src/itoa.rs
ahabhgk: I see

ahabhgk (approved):
Thanks!

Issue #5316, opened 2024-01-12
Align module exports with webpack


JohnDaly:
### System Info

System:
OS: macOS 12.4
CPU: (10) arm64 Apple M1 Max
Memory: 22.38 GB / 64.00 GB
Shell: 5.8.1 - /bin/zsh
Binaries:
Node: 20.5.0 - ~/Library/Caches/fnm_multishells/61564_1700104358947/bin/node
Yarn: 1.22.17 - ~/.yarn/bin/yarn
npm: 9.8.0 - ~/Library/Caches/fnm_multishells/61564_1700104358947/bin/npm
pnpm: 8.10.5 - ~/Library/Caches/fnm_multishells/61564_1700104358947/bin/pnpm
bun: 0.8.1 - ~/.bun/bin/bun
Browsers:
Chrome: 119.0.6045.105
Safari: 15.5

### Details

I've noticed that `rspack` build artifacts are typically larger than what `webpack` generates.

One of the contributing factors is how exports are generated.

Example file:
```js
export const variableWithLongNameOne = 'one';
export const variableWithLongNameTwo = 'two';
export const variableWithLongNameThree = 'three';
export const variableWithLongNameFour = 'four';
export const variableWithLongNameFive = 'five';
export const variableWithLongNameSix = 'six';
```

`webpack` will generate the following code:
```js
/*!**********************!*\
  !*** ./src/index.js ***!
  \**********************/
__webpack_require__.r(__webpack_exports__);
/* harmony export */ __webpack_require__.d(__webpack_exports__, {
/* harmony export */   variableWithLongNameFive: () => (/* binding */ variableWithLongNameFive),
/* harmony export */   variableWithLongNameFour: () => (/* binding */ variableWithLongNameFour),
/* harmony export */   variableWithLongNameOne: () => (/* binding */ variableWithLongNameOne),
/* harmony export */   variableWithLongNameSix: () => (/* binding */ variableWithLongNameSix),
/* harmony export */   variableWithLongNameThree: () => (/* binding */ variableWithLongNameThree),
/* harmony export */   variableWithLongNameTwo: () => (/* binding */ variableWithLongNameTwo)
/* harmony export */ });
const variableWithLongNameOne = 'one';
const variableWithLongNameTwo = 'two';
const variableWithLongNameThree = 'three';
const variableWithLongNameFour = 'four';
const variableWithLongNameFive = 'five';
const variableWithLongNameSix = 'six';
```

`rspack` generates this code:
[...]

PR #7764, opened 2024-09-02
feat: runtime template with dojang

LingyuCoder:
## Summary

Use [dojang](https://github.com/rspack-contrib/rspack-dojang) to render runtime modules, related to https://github.com/web-infra-dev/rspack/issues/5316

- Use dojang because it supports custom function, and we can use the same code in template to generate different result by inject functions according to `output.environment`, and also html plugin is using it so the package size will not increase.
- Inject runtime globals as parameters by default.
- Dojang only support string literal so array of strings need to be joined with `,`.

relate to https://github.com/rspack-contrib/rspack-dojang/pull/3

for example, the on chunk loaded runtime module can be rendered by template blow:

```ejs
var deferred = [];
<%= ON_CHUNKS_LOADED %> = <%= basicFunction("result, chunkIds, fn, priority") %> {
	if (chunkIds) {
		priority = priority || 0;
		for (var i = deferred.length; i > 0 && deferred[i - 1][2] > priority; i--)
			deferred[i] = deferred[i - 1];
		deferred[i] = [chunkIds, fn, priority];
		return;
	}
	var notFulfilled = Infinity;
	for (var i = 0; i < deferred.length; i++) {
		<%= destructureArray("chunkIds, fn, priority", "deferred[i]") %>
		var fulfilled = true;
		for (var j = 0; j < chunkIds.length; j++) {
			if (
				(priority & (1 === 0) || notFulfilled >= priority) &&
				Object.keys(<%= ON_CHUNKS_LOADED %>).every(<%= returningFunction(ON_CHUNKS_LOADED + "[key](chunkIds[j])", "key") %>)
			) {
				chunkIds.splice(j--, 1);
			} else {
				fulfilled = false;
				if (priority < notFulfilled) notFulfilled = priority;
			}
		}
		if (fulfilled) {
			deferred.splice(i--, 1);
			var r = fn();
			if (r !== undefined) result = r;
		}
	}
	return result;
};


```

## Checklist


GiveMe-A-Name:
Continue work another PR https://github.com/web-infra-dev/rspack/pull/8851

PR #7775, opened 2024-09-03
docs: fix migrate-from-builtins.css example codes

noshower:
## Checklist


chenjiahan (approved):
Thanks!

PR #7784, opened 2024-09-04
fix(dev-server): pin webpack-dev-server version

### LINKED ISSUE #7782: [Bug]: dev server cannot start
### System Info

-

### Details

webpack-dev-server has changed its implementation in webpack/webpack-dev-server#5275 and released in 5.1.0 which removed some internal hooks.

A temporary fix would be lock webpack-dev-server to 5.0.4.

### Reproduce link

_No response_

### Reproduce Steps

1. run `rspack dev`
2. get
```
[rspack-cli] TypeError: this.setupHostHeaderCheck is not a function
```

hardfist:
## Summary
fix #7782
rspack-dev-server use some internal api of webpack-dev-server, which cause it's easily broken if webpack-dev-server do some internal refactor
so the temporary solution is to pin the webpack-dev-server version
we need redesign rspack-dev-server implemention and shouldn't rely on internal api

## Checklist

### DIFF
--- a/packages/playground/package.json
+++ b/packages/playground/package.json
@@ -33,7 +33,7 @@
     "vue": "^3.4.21",
     "vue-loader": "^17.3.1",
     "css-loader": "^6.11.0",
-    "webpack-dev-server": "^5.0.4",
+    "webpack-dev-server": "5.0.4",
     "ws": "^8.16.0"
   }
 }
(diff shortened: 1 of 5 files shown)


chenjiahan (approved):

jerrykingxyz (approved):

Austaras:
It would be interesting if rspack team could rewrite this part in rust too.

PR #7788, opened 2024-09-04
fix(dev-server): prebundle webpack-dev-server and pin webpack-dev-middleware

hardfist:
## Summary
related to #7782 and #7676

new published webpack-dev-server and webpack-dev-middleware can easily break rspack-cli cause there're no tests in webpack-dev-server & webpack-dev-middle repo and rspack-dev-server use some internal apis of these two packages.

so we need to pin webpack-dev-server  & webpack-dev-middleware version to avoid accidentally breaking, 
but since webpack-dev-middleware is indirect dependency of webpack-dev-server, there're no easy way to pin it without bundling webpack-dev-server.

This pr is  a temporarily solution, and the long term solutions is refactor rspack-dev-server implementation and maybe add tests in webpack-dev-server and webpack-dev-middleware about rspack compatibility

## Checklist

### DIFF
--- a/packages/rspack-dev-server/package.json
+++ b/packages/rspack-dev-server/package.json
@@ -19,7 +19,7 @@
   },
   "scripts": {
     "prepare": "prebundle",
-    "build": "tsc -b ./tsconfig.build.json",
+    "build": "tsc -b ./tsconfig.build.json && tsc-alias -p tsconfig.build.json",
     "dev": "tsc -b -w ./tsconfig.build.json",
     "test": "rimraf .test-temp && cross-env NO_COLOR=1 node --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --colors",
     "api-extractor": "api-extractor run --verbose",
--- a/packages/rspack-dev-server/package.json
+++ b/packages/rspack-dev-server/package.json
@@ -47,7 +47,8 @@
     "jest-serializer-path": "^0.1.15",
     "prebundle": "^1.1.0",
     "typescript": "5.0.2",
-    "webpack-dev-server": "5.0.4"
+    "webpack-dev-server": "5.0.4",
+    "tsc-alias": "^1.8.8"
   },
   "dependencies": {
     "chokidar": "^3.6.0",
--- a/packages/rspack-dev-server/package.json
+++ b/packages/rspack-dev-server/package.json
@@ -43,6 +43,8 @@
     "@types/connect-history-api-fallback": "1.5.4",
     "@types/express": "4.17.21",
     "@types/mime-types": "2.1.4",
+    "@types/express-serve-static-core": "4.19.5",
+    "@types/serve-static": "1.15.7",
     "@types/ws": "8.5.10",
     "jest-serializer-path": "^0.1.15",
     "prebundle": "^1.1.0",
--- a/packages/rspack-dev-server/package.json
+++ b/packages/rspack-dev-server/package.json
@@ -18,8 +18,9 @@
     "./package.json": "./package.json"
   },
   "scripts": {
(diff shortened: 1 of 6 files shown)


hardfist:
@alexander-akait is it acceptable that we add rspack compat tests in webpack-dev-server and webpack-dev-middleware repos or any other suggestions?

packages/rspack-dev-server/package.json
```diff
@@ -18,8 +18,9 @@
     "./package.json": "./package.json"
   },
   "scripts": {
+    "prepare": "prebundle",
     "build": "tsc -b ./tsconfig.build.json",
-    "dev": "tsc -w -b ./tsconfig.build.json",
+    "dev": "tsc -b -w ./tsconfig.build.json",
```
chenjiahan, line 23: we need to add tsconfig#paths, alias `import 'webpack-dev-server'` to `import '../compiled/webpack-dev-server'`

Example: https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/tsconfig.json#L7-L10

packages/rspack-dev-server/package.json
```diff
@@ -18,8 +18,9 @@
     "./package.json": "./package.json"
   },
   "scripts": {
+    "prepare": "prebundle",
     "build": "tsc -b ./tsconfig.build.json",
```
chenjiahan, line 22: Use `tsc-alias` to apply alias paths

```suggestion
    "build": "tsc -b ./tsconfig.build.json && tsc-alias -p tsconfig.build.json",
```

packages/rspack-dev-server/package.json
hardfist: fixed

packages/rspack-dev-server/package.json
```diff
@@ -42,19 +43,38 @@
...
+    "ipaddr.js": "^2.1.0",
+    "launch-editor": "^2.6.1",
+    "open": "^10.0.3",
+    "p-retry": "^6.2.0",
+    "schema-utils": "^4.2.0",
+    "selfsigned": "^2.4.1",
+    "serve-index": "^1.9.1",
+    "sockjs": "^0.3.24",
+    "spdy": "^4.0.2"
   },
```
markmssd, line 78: Have you considered leveraging [resolutions](https://pnpm.io/package_json#resolutions) to pin the indirect dependency `webpack-dev-middleware`? It would look look this:
```suggestion
  },
  "resolutions": {
    "webpack-dev-server/webpack-dev-middleware": "7.4.2"
  },
```

SoonIter: The users of resolution are consumers, not pkg publishers. 
We don't want all consumers to need resolution.

SoonIter (changes requested):
`require.resolve` is not modified by `tsc-alias`

https://github.com/web-infra-dev/rspack/blob/aadd2c1a7ef7f78f29cf317fb7f1bb5e82c1f700/packages/rspack-dev-server/src/server.ts#L79

chenjiahan (approved):

hardfist:
close since this is too complex and hacky, waiting for rspack-dev-server refactor

hardfist:
> @hardfist what is broken, I think we can fix them

@alexander-akait  thanks, it' broken cause rspack-dev-server use some private api of webpack-dev-server(we're refactoring the rspack-dev-server to use the public api https://github.com/web-infra-dev/rspack/blob/main/packages/rspack-dev-server/src/server.ts#L159-L174), so it's actually nothing wrong with webpack-dev-server itself, but it would be nice if we can test the new version of webpack-dev-server compat before users are hitting the problems.

alexander-akait:
@hardfist I am fine with adding cross tests :+1:

PR #7794, opened 2024-09-04
refactor: clean dev server code

LingyuCoder:
## Summary

- should not override private methods, just add some alias
- override the `module.require.resolve` in NodeModule instance of `webpack-dev-server` to redirect webpack server module request to rspack

## Checklist

### DIFF
--- a/packages/rspack-dev-server/src/server.ts
+++ b/packages/rspack-dev-server/src/server.ts
@@ -17,25 +17,10 @@ import WebpackDevServer from "webpack-dev-server";
 // @ts-ignore 'package.json' is not under 'rootDir'
 import { version } from "../package.json";
 
-import { addResolveAlias } from "./alias";
+import { addResolveAlias, removeResolveAlias } from "./alias";
 import type { DevServer, ResolvedDevServer } from "./config";
 import { applyDevServerPatch } from "./patch";
 
-// override
-addResolveAlias("webpack-dev-server", {
-	"../client/clients/SockJSClient": require.resolve(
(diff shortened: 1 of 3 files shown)


packages/rspack-dev-server/src/server.ts
```diff
@@ -13,23 +13,31 @@ import type { Server } from "node:http";
...
+import WebpackDevServer from "webpack-dev-server";
 // @ts-ignore 'package.json' is not under 'rootDir'
 import { version } from "../package.json";
 
+import { addResolveAlias } from "./alias";
 import type { DevServer, ResolvedDevServer } from "./config";
 import { applyDevServerPatch } from "./patch";
 
+// override
+addResolveAlias("webpack-dev-server", {
```
ahabhgk, line 25: Can we only override this before the call site and then remove the override after the call site, just to control the effect as small as possible and not affect other resolve

LingyuCoder: This override only affect the require.resolve of `webpack-dev-server`. Has any other place use it too?

LingyuCoder: Oh I see, when required `rspack-dev-server` but not use it, then starting a webpack-dev-server will be affected. I will find a better way to handle this

LingyuCoder: Seems that only need to override before `addAdditionalEntries `, this method is sync so override before it and resume after it will make the effect as small as possible.

ahabhgk (dismissed):
LGTM

ahabhgk (approved):

PR #7828, opened 2024-09-06
fix: use SourceMapSource in js side to support map method

SyMind:
## Summary

The `CompatSource` in Rust is unnecessary and can be simplified. Below are the steps detailing the changes for converting between Rust and JavaScript sources.

1. Rust to JavaScript
    * If the source is a buffer or doesn't have a source map, JavaScript will expose it using `RawSource`.
    * If the source has a source map, JavaScript will expose it using `SourceMapSource`.
    
2. JavaScript to Rust
    * The conversion process remains the same as above.

Fix: #7891

## Checklist


SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

SyMind:
!bench

packages/rspack/src/util/source.ts
```diff
@@ -1,79 +1,49 @@
...
-					return JSON.parse(source.map.toString("utf-8"));
-				}
-
-				return null;
-			}
-		});
+		return new SourceMapSource(
+			source.source,
+			"inmemory://from rust",
+			source.map ? JSON.parse(source.map) : null
```
h-a-n-a, line 17: Can we lazily call `JSON.parse`?

h-a-n-a:
Most part of this PR is a refactor and due to this PR including a fix, I would recommend you to use 'fix' as the title.

packages/rspack/src/util/source.ts
```diff
@@ -1,79 +1,49 @@
...
-				if (source.map) {
-					return JSON.parse(source.map.toString("utf-8"));
-				}
-
-				return null;
-			}
-		});
+		return new SourceMapSource(
+			source.source,
+			"inmemory://from rust",
```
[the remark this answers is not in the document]
SyMind: `filename` is only used when an inner source map exists in `SourceMapSource`.

packages/rspack/src/util/source.ts
SyMind: We can. I fix it.

h-a-n-a (approved):

PR #7834, opened 2024-09-09
chore(infra): add rust codecov

### LINKED ISSUE #7090: integrate with code coverage
we did lots of refactor and may have some legacy code which is not removed and we also could have some code which should be covered by tests, use code coverage service could help us spot these issues

SoonIter:
## Summary

close #7090 

We may also need the coverage of js jest test.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -335,30 +335,25 @@ jobs:
   # codecov often fails, use another workflow for retry
   rust_test_codecov_upload:
     name: Rust Test Coverage Upload
-    needs: [rust_changes, rust_test]
-    if: ${{ needs.rust_changes.outputs.changed == 'true' }}
+    if: github.repository_owner == 'web-infra-dev'
+    needs: [rust_test]
     runs-on: ubuntu-latest
-    # Check if the event is not triggered by a fork by checking whether CODECOV_TOKEN is set
-    env:
-      CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
     steps:
       - name: Checkout
-        if: env.CODECOV_TOKEN
         uses: actions/checkout@v4
 
       - name: Download coverage file
-        if: env.CODECOV_TOKEN
         uses: actions/download-artifact@v4.1.7
         with:
           name: codecov
 
       - name: Upload to codecov.io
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -320,6 +320,64 @@ jobs:
...
+          shared-key: check
+
+      - name: Install llvm-tools-preview
+        run: rustup component add llvm-tools-preview
+      - uses: cargo-bins/cargo-binstall@v1.10.2
+      - run: cargo binstall --no-confirm cargo-llvm-cov
+
+      - name: Run test coverage
+        # reason for excluding https://github.com/napi-rs/napi-rs/issues/2200
+        run: cargo llvm-cov --workspace --exclude rspack_binding_options --exclude rspack_node --lcov --output-path lcov.info
```
chenjiahan, line 347: Why do we need to run test in two different jobs (rust_test and rust_test_codecov), can we unify them?

.github/workflows/ci.yml
```diff
@@ -320,6 +320,64 @@ jobs:
...
+
+      - name: Upload Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: codecov
+          path: lcov.info
+
+  rust_test_codecov_upload:
```
chenjiahan, line 355: Why split rust_test_codecov and rust_test_codecov_upload, can we merge them?

SoonIter:
😂 7.05%

image

.github/workflows/ci.yml
SoonIter: done, merged in `rust_test` with `cargo test`

.github/workflows/ci.yml
SoonIter: `# codecov often fails, use another workflow for retry` 
https://github.com/web-infra-dev/rspack-resolver/blob/16eb56eef68d651edec703b94022ff7f3e46d585/.github/workflows/codecov.yml#L44

Migrated from `rspack-resolver` The generated file is huge, and the api of codecov does not seem to be very stable.

SoonIter:
can be added with rust/js unit test refactor
There is no need of codecov at present

We are looking for a way to calculate the coverage rate of the **intergration/e2e test**

PR #7836, opened 2024-09-09
fix(external_module): use "const" if supported when concatenation

fi3ework:
## Summary

Align with https://github.com/webpack/webpack/blob/039e8c96338ee98cf37d5035cdd30073a96b4ec5/lib/ExternalModule.js#L878. No coveraged test found in webpack.

## Checklist

### DIFF
--- a/crates/rspack_core/src/external_module.rs
+++ b/crates/rspack_core/src/external_module.rs
@@ -71,10 +71,17 @@ impl ExternalRequestValue {
   }
 }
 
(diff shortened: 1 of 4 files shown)


JSerFeng:
Note: We can safely use `const` or `let` here, because external module has definite runtime condition, so the `require` won't be wrapped in `if` statement, which is like

```
if ("main" == __webpack_require__.j) {
  var foo = require('./foo')
}
```

So LGTM

JSerFeng (approved):

PR #7856, opened 2024-09-10
docs: add NIO logo at built with rspack

littleVege:
## Summary
NIO has several projects now using rsbuild, wish to join "build with rspack"

## Checklist


chenjiahan (approved):

chenjiahan:
Thanks! We are pleased that NIO is using Rspack and Rsbuild.

PR #7866, opened 2024-09-11
fix: trusted type with devtool eval

LingyuCoder:
## Summary

- While using eval devtool plugin, `eval(module)` should be `eval(__webpack_require__.ts(module))` to make sure that the module script is a trusted script
- Add CreateScriptUrlDependency to handle worker scripts.
- Can use runtimeRequirementsInTree instead of runtimeRequirementsInModule to add runtime globals because these plugins only check the `compilation.output.trustedTypes` without any module level logic

## Checklist


SyMind (approved):

pushkargupta9807:
@LingyuCoder @SyMind I just upgraded from 1.04 to 1.0.5 and started running into the below error when initializing the worker script. 

image

it is a simple worker init code as below which was working fine in the previous version and I suspect this change could have broken it. 

image

Can you please advise? Thanks.

LingyuCoder:
> @LingyuCoder @SyMind I just upgraded from 1.04 to 1.0.5 and started running into the below error when initializing the worker script.
> 
> image
> 
> it is a simple worker init code as below which was working fine in the previous version and I suspect this change could have broken it.
> 
> image
> 
> Can you please advise? Thanks.

Can you print the `import.meta.url` here? And try to run the same `new URL` in console, not sure whether this is caused by wrong `import.meta.url` or trusted type.

LingyuCoder:
Seems there is a bug in CreateScriptUrlRuntimeModule [here](https://github.com/web-infra-dev/rspack/blob/0683b3e018f6e4967b5019f0053dc5ecbff46f0e/crates/rspack_plugin_runtime/src/runtime_module/create_script_url.rs#L40). I will try to fix it.

pushkargupta9807:
Thank you @LingyuCoder for the super fast turn around fix.  
However I wanted to let you know that I checked and nowhere in my code I use the RuntimeGlobals.createScriptUrl . 
(we do have a MF related plugin which uses it, but I have never included that plugin in our rspack builds).

I do have output.trustedTypes set . I disabled it , still ran into the same error. 

Is there a temporary hack/ workaround you can suggest to get it working ? i.e to avoid hitting that problematic code.

LingyuCoder:
> Thank you @LingyuCoder for the super fast turn around fix. However I wanted to let you know that I checked and nowhere in my code I use the RuntimeGlobals.createScriptUrl . (we do have a MF related plugin which uses it, but I have never included that plugin in our rspack builds).
> 
> I do have output.trustedTypes set . I disabled it , still ran into the same error.
> 
> Is there a temporary hack/ workaround you can suggest to get it working ? i.e to avoid hitting that problematic code.

You can try using [`compilation.hooks.runtimeModule`](https://rspack.dev/api/plugin-api/compilation-hooks#runtimemodule) to override the code of `CreateScriptUrlRuntimeModule` with `__webpack_require__.tu = function(url) { return url; }`.

Not sure whether there are other bugs. Maybe the compiled code and the parameters of `new URL` need to be provided to locate the error.

pushkargupta9807:
@LingyuCoder I just tried the 1.0.11 v with the above fix , but I am still running into that same issue. I filed the following bug to track the same.
 https://github.com/web-infra-dev/rspack/issues/8134

Issue #7336, opened 2024-07-28
[Tracking]: better diagnostic report of tapable error


shulaoda:
### What subject does this issue tracking?

Related to #6051 

This is a huge and complex task, and it took me a long time to understand how to complete it. This task involves many dependency types, and I need to break them down and gradually cover them.

```[tasklist]
### Tasks
- [ ] https://github.com/web-infra-dev/rspack/pull/7337
- [ ] https://github.com/web-infra-dev/rspack/pull/7374
- [ ] https://github.com/web-infra-dev/rspack/pull/7871
- [x] HarmonyExportExpressionDependency、HarmonyExportHeaderDependency、HarmonyExportImportedSpecifierDependency、HarmonyImportSideEffectDependency、HarmonyImportSpecifierDependency
- [x] HarmonyAcceptDependency、HarmonyExportSpecifierDependency、ModuleArgumentDependency
- [ ] CommonJsExportRequireDependency、CommonJsExportsDependency、CommonJsFullRequireDependency、CommonJsRequireDependency、CommonJsSelfReferenceDependency、CommonJsRequireContextDependency
- [ ] CssComposeDependency、CssImportDependency、CssLocalIdentDependency、CssUrlDependency
- [ ] WebpackIsIncludedDependency、PureExpressionDependency
- [ ] TODO: Others...
```

PR #7871, opened 2024-09-12
feat(rspack_plugin_javascript): improve provide dependency

shulaoda:
## Summary

Related to #7336 and #7422 

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/dependency/esm/provide_dependency.rs
+++ b/crates/rspack_plugin_javascript/src/dependency/esm/provide_dependency.rs
@@ -1,8 +1,7 @@
 use itertools::Itertools;
-use rspack_core::DependencyType;
 use rspack_core::{
-  create_exports_object_referenced, module_raw, Compilation, ExtendedReferencedExport, ModuleGraph,
-  NormalInitFragment, RuntimeSpec, UsedName,
+  create_exports_object_referenced, module_raw, Compilation, DependencyRange, DependencyType,
+  ExtendedReferencedExport, ModuleGraph, NormalInitFragment, RuntimeSpec, UsedName,
 };
 use rspack_core::{AsContextDependency, Dependency, InitFragmentKey, InitFragmentStage};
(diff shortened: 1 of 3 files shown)


shulaoda:
I introduced `DependencyRange` because some dependencies's `span` are only used as `range` and don't require `to_loc`. This approach reduces memory usage since `RealDependencyLocation` has an additional `source: Option<Arc<dyn SourceLocation>>`.

crates/rspack_plugin_javascript/src/dependency/esm/provide_dependency.rs
```diff
@@ -38,6 +35,10 @@ impl Dependency for ProvideDependency {
     &self.id
   }
 
+  fn loc(&self) -> Option<String> {
+    Some(self.request.to_string())
```
h-a-n-a, line 39: I don't get this. Why do this return a clone of `request`?

shulaoda: The loc here is `SyntheticDependencyLocation`, and since the `request` and `SyntheticName` are the same, it has not been repeatedly introduced.

shulaoda: `ProvidePlugin` essentially dynamically introduces a `provide_dependency` without a real location.

shulaoda: Perhaps we should express it as `ProvidePlugin:request_name`?

shulaoda: My negligence, I have always believed that we should use the real location of dependency.

h-a-n-a: > we should use the real location of dependency.

LGTM

h-a-n-a (approved):

PR #7878, opened 2024-09-12
refactor: css-extract using parseMeta to store css info

JSerFeng:
## Summary

css-extract uses `additionalData` to parse necessary data to Rust, but `additionalData` can be removed by other user's plugin, so it's not safe.

Now we introduce a new internal field for passing data, other loaders cannot modify it

## Checklist


h-a-n-a:
Would you please add a test just to avoid regressions?

JSerFeng:
> Would you please add a test just to avoid regressions?

Done

h-a-n-a (approved):

Issue #7888, opened 2024-09-13
[Bug]: Panic if resolved module is deleted


JSerFeng:
### System Info

not important

### Details

Delete one file, but keep importer unchanged, there will be a panic

```
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  should have module
Location: crates/rspack_core/src/compiler/make/cutout/has_module_graph_change.rs:23

Backtrace omitted.
```

You can reproduce this using `rspack serve`, make sure you have a module called 'foo.js' in the module graph, then delete it.

### Reproduce Steps

Creating 2 module index.js and foo.js

index.js imports foo.js

run `rspack serve`

deleted foo.js

PR #7890, opened 2024-09-13
fix: should skip if module already be removed

JSerFeng:
## Summary

fix: #7888

If module is already removed, should not panic

## Checklist


hardfist:
why this is not merged?

JSerFeng:
Cause I can't find the minimal repro and this optimization is about to remove when incremental code splitting is stabilized

PR #7892, opened 2024-09-13
refactor: use `realDependencyLocation` instead of `range` in dependency

shulaoda:
## Summary

Related to #7422 

## Checklist

### DIFF
--- a/crates/rspack_core/src/dependency/dependency_location.rs
+++ b/crates/rspack_core/src/dependency/dependency_location.rs
@@ -2,6 +2,30 @@ use std::{fmt, sync::Arc};
 
 use derivative::Derivative;
 
+#[derive(Debug, Clone)]
+pub struct DependencyRange {
+  pub start: u32,
+  pub end: u32,
+}
+
+impl From<(u32, u32)> for DependencyRange {
+  fn from(range: (u32, u32)) -> Self {
+    Self {
+      start: range.0,
+      end: range.1,
+    }
+  }
+}
+
+impl From<swc_core::common::Span> for DependencyRange {
+  fn from(span: swc_core::common::Span) -> Self {
+    Self {
+      start: span.lo.0.saturating_sub(1),
+      end: span.hi.0.saturating_sub(1),
+    }
+  }
+}
+
 #[derive(Derivative)]
 #[derivative(Debug, Clone, Hash)]
 pub struct RealDependencyLocation {
--- a/crates/rspack_core/src/dependency/dependency_location.rs
+++ b/crates/rspack_core/src/dependency/dependency_location.rs
@@ -2,36 +2,6 @@ use std::{fmt, sync::Arc};
 
 use derivative::Derivative;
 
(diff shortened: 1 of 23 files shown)


shulaoda:
I introduced `DependencyRange` because some dependencies's `span` are only used as `range` and don't require `to_loc`. This approach reduces memory usage since `RealDependencyLocation` has an additional `source: Option<Arc<dyn SourceLocation>>`.

_Originally posted by @shulaoda in https://github.com/web-infra-dev/rspack/issues/7871#issuecomment-2345210104_

LingyuCoder:
LGTM, cc @h-a-n-a

LingyuCoder (approved):

h-a-n-a:
Do you have other purpose of introducing this data structure other than its size?

shulaoda:
> Do you have other purpose of introducing this data structure other than its size?

Perhaps it can be used as a substitute for `ErrorSpan`?

h-a-n-a:
> > Do you have other purpose of introducing this data structure other than its size?
> 
> Perhaps it can be used as a substitute for `ErrorSpan`?

That's the question I was wondering. If we have `ErrorSpan`, why bother creating another `DependencyRange`? I'm still thinking about to unify these locations and ranges into `RealDependencyLocation`.

shulaoda:
> That's the question I was wondering. If we have `ErrorSpan`, why bother creating another `DependencyRange`? I'm still thinking about to unify these locations and ranges into `RealDependencyLocation`.

If an additional `usize` memory overhead is acceptable, we can do it this way. Alternatively, we could place the `Sourcemap` into the `Dependency` and use `range.to_loc(self.sourcemap)` for location transformation.

h-a-n-a:
> If an additional `usize` memory overhead is acceptable, we can do it this way. 
I do think it's non-trivial. Let's keep `RealDependencyLocation` as small as 8-bytes. 

> Alternatively, we could place the `Sourcemap` into the `Dependency` and use `range.to_loc(self.sourcemap)` for location transformation.
Your suggestion sounds promising. Would you please give it a try?

shulaoda:
> Your suggestion sounds promising. Would you please give it a try?

I will complete it in the next PR. Additionally, I propose replacing `RealDependencyLocation` and `SyntheticDependencyLocation` with `RealDependencyRange` and `SyntheticDependencyName` respectively. Our implementation ultimately differs somewhat from `Webpack`.

h-a-n-a:
> > Your suggestion sounds promising. Would you please give it a try?
> 
> I will complete it in the next PR. Additionally, I propose replacing `RealDependencyLocation` and `SyntheticDependencyLocation` with `RealDependencyRange` and `SyntheticDependencyName` respectively. Our implementation ultimately differs somewhat from `Webpack`.

LGTM. You could replace name `RealDependencyRange` with `DependencyRange` and remove `ErrorSpan`. Both of them looks good to me.

h-a-n-a (approved):
Thanks for the work!

PR #7907, opened 2024-09-15
perf(rspack_core): replace `HASH_PLACEHOLDER_REGEX` with simple string parsing

shulaoda:
## Summary

Related to https://github.com/rolldown/rolldown/pull/1496

I plan to convert all the regular expressions into string lookup, because these regular expressions are very simple and do not require regex.

https://github.com/web-infra-dev/rspack/blob/a06f078ee77c66f45eeb4a6ac35ee7b57e94eca8/crates/rspack_core/src/options/filename.rs#L19-L38

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -675,12 +675,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cow-utils"
-version = "0.1.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "417bef24afe1460300965a25ff4a24b8b45ad011948302ec221e8a0a81eb2c79"
-
-[[package]]
 name = "cpufeatures"
 version = "0.2.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


chenjiahan:
@shulaoda can you resolve the conflicts? ❤️

h-a-n-a:
Wasn't the issue in the description had be resolved? I think this is necessary.
> I'm not sure if the hash pattern appears only once. If there are cases where the pattern appears multiple times, such as [hash]-[hash:6].js, I need to optimize further.

```rust
let result = replace_all_hash_pattern("hello-[hash]-[hash:5].js", "[hash]", |n| {
    &"abcdefgh"[..n.unwrap_or(8)]
  });
 assert_eq!(result, Some("hello-abcdefgh-abcde.js".to_string()));
```

shulaoda:
> Wasn't the issue in the description had been resolved? I think this is necessary.

This is a PR from a long time ago, and I fixed the issues in the description later. Perhaps I should modify the description.

h-a-n-a (approved):

PR #7911, opened 2024-09-16
fix(core): use resource shouldn't contain query

hardfist:
## Summary
align resource in use with webpack
the use resource(doesn't contain query) is actually misleading here because it has different meaning compared with loaderContext.resource(which contains query)

## Checklist


chenjiahan (commented):
can we clarify this in the documentation

h-a-n-a (approved):

Issue #7516, opened 2024-08-09
[Tracking]: migrate development workflow to just


hardfist:
### What subject does this issue tracking?

# [migrate development workflow to just]
## background
we may need some rust workflow(like rust api bump、release) in the future, which cause we may have js workflow and rust workflow together. Currently we use `./x` to manage js workflow but it seems not idea to use `./x` to manage rust workflow. so we're considering migrate the whole development workflow to just + js-script + xtask.
## Design
our daily workflow actually contain two parts
* task implementation: task like publish | bump | debug | test and so on 
* task orchestra: how to orchestra task like run bump first and then run publish
just is good at task management and xtask is good at writing task using rust
so we  use just to do task orchestra and use js script and xtask to write task implementation.

```[tasklist]
### Tasks
- [ ] migrate current task orchestra in ./x to just
- [ ] update cicd to just command
- [ ] update contribute develop guide to just command
- [ ] remove `./x` and `./x.mjs`
- [ ] implement rust workflow (rust api release, doc generate) in just & xtask
```

chenjiahan:
Should we alias the commands in `npm scripts` to just commands?

hardfist:
> Should we alias the commands in `npm scripts` to just commands?

I think it's the reverse, just call `npm script` to do js task management

hardfist:
> Some contributors may be used to using `npm scripts`, so how can we guide them?

provide detailed guide like this ? https://github.com/biomejs/biome/blob/main/CONTRIBUTING.md#install-the-required-tools 

> I think we can alias some basic commands to just. For example, alias pnpm run setup to just setup.

we can add some basic alias like `setup`, but when we have lots of tasks like https://github.com/biomejs/biome/blob/main/justfile, duplicate alias for all tasks seem not ideal

shulaoda:
We can learn from this `xtask` case https://github.com/swc-project/swc/tree/main/xtask

hardfist:
close since not gonna use just

PR #7915, opened 2024-09-17
feat(rspack): support xtask

shulaoda:
## Summary

Related to #7516 

## Checklist

### DIFF
--- a/xtask/src/commands/api_extractor.rs
+++ b/xtask/src/commands/api_extractor.rs
@@ -1,0 +1,36 @@
+use anyhow::{Context, Result};
+use clap::{Parser, Subcommand};
+
+use crate::utils::command::run_command;
+
+#[derive(Debug, Parser)]
+pub struct ApiExtractorCmd {
+  #[clap(subcommand)]
+  sub_cmd: ApiExtractorSubCmd,
+}
+
+#[derive(Debug, Subcommand)]
+pub enum ApiExtractorSubCmd {
+  Update,
+  Ci,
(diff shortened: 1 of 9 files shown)


xtask/src/commands/api_extractor.rs
```diff
@@ -0,0 +1,36 @@
...
+          "pnpm",
+          &["--filter", "@rspack/*", "api-extractor", "--local"],
+        )?;
+      }
+      ApiExtractorSubCmd::Ci => {
+        run_command("pnpm", &["--filter", "@rspack/*", "api-extractor:ci"]).with_context(|| {
+          "Api-extractor testing failed. Did you forget to update the snapshots locally?\nRun the command below locally to fix this error (in the *ROOT* of rspack workspace).\n$ ./x api-extractor update"
+        })?;
+      }
+    }
```
hardfist, line 33: what's the benefit of run pnpm script in xtask then in justfile directly

shulaoda: You can use cmd args more flexibly. This is just a start, and perhaps `api-extractor` isn't the best choice. I expect to implement commands like `debug`, `publish`, and `version` in `xtask`.

hardfist: I'm not sure put js related task into xtask is a good idea, maybe we should start with rust related task first, @h-a-n-a what's your idea?

hardfist:
thanks, close since xtask is already introduced

PR #7927, opened 2024-09-19
fix: infer async modules when incremental enabled

ahabhgk:
## Summary

fix infer async modules when incremental enabled (**both incremental and newIncremental**)

~~This could cause a little bit perf regression when cold start but correctness first and the regression is acceptable (under 5% according to the bench)~~ shouldn't cause perf regression

## Checklist


ahabhgk:
!bench

JSerFeng (dismissed):

JSerFeng:
Can you explain the issue before ?

ahabhgk:
for example, `a -> b -> c`

in the first build, c has top-level-await, this will infer a and b are async modules (`a (async) -> b (async) -> c (async)`)

but in the second build, we remove the top-level-await inside c, the correct module graph should be `a -> b -> c`, but since we incrementally rebuild the module graph, only rebuild the modified module c, so the module graph is `a (async) -> b (async) -> c`, this is an incorrect module graph

in this PR we will fix this first in InferAsyncModulesPlugin, we find the `not has_top_level_await` modules from affected modules, and infer `is_async` for them, reset `is_async` back to `false` first, and then find the `has_top_level_await` modules from affected modules and set them and inferred modules to `is_async: true`

JSerFeng:
> for example, a -> b -> c
> 
> in the first build, c has top-level-await, this will infer a and b are async modules (a (async) -> b (async) -> c (async))

Why are `a` and `b` not included in `affectedModules`, as they have data referred from `c`

JSerFeng (approved):

PR #7939, opened 2024-09-20
feat(parser): add new javascript parser options

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/204 and more.

We'll try to merge this back into webpack after stabilized.

## Checklist


crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
```diff
@@ -268,6 +272,10 @@ impl CommonJsImportsParserPlugin {
       }
     }
 
+    if matches!(parser.javascript_options.require_dynamic, Some(false)) && !param.is_string() {
+      return Some(true);
```
ahabhgk, line 276: Failed to match a dependency here, so this should return `Some(false)`

crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
```diff
@@ -290,6 +298,10 @@ impl CommonJsImportsParserPlugin {
     parser: &mut JavascriptParser,
     ident: &Ident,
   ) -> Option<bool> {
+    if matches!(parser.javascript_options.require_as_expression, Some(false)) {
+      return Some(true);
```
ahabhgk, line 302: Same here

crates/rspack_plugin_javascript/src/parser_plugin/import_parser_plugin.rs
```diff
@@ -146,6 +146,10 @@ impl JavascriptParserPlugin for ImportParserPlugin {
       parser.blocks.push(Box::new(block));
       Some(true)
     } else {
+      if matches!(parser.javascript_options.import_dynamic, Some(false)) {
+        return Some(true);
```
ahabhgk, line 150: Same here

crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
fi3ework: Hmm, I think we could not use `Some(false)` here (others are OK). If we use `Some(false)` here, it will still fall resolve logic here https://github.com/web-infra-dev/rspack/blob/7e1b6f05255026d154401c224d4123473c54c5ea/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs#L466 and the `require` won't be preserved.

ahabhgk: I see, `require` won't be preserved should be caused by `require_as_expression_handler`, not `process_resolve`, if you try `requireContext: false` in webpack the require is also not preserved, it has the similar behavior: `require.context("./dir")` for example
- `requireContext: true`: `__webpack_require__("./dir sync recursive ^\\.\\/.*$")`
- `requireContext: false` in actual: `__webpack_require__("./. sync recursive").context("./dir")`
- `requireContext: false` in expect: `require.context("./dir")`

I'm really not sure this is a bug or a feature, since there isn't any test for `requireContext: false` in webpack.

So if it's a feature, then this is expected, `requireContext: false` / `requireDynamic: false` should be used together with `requireAsExpression: false`

if it's a bug, then we should probably fix that first

ahabhgk: BTW, `Some(true)` should change to `None`, I was wrong

fi3ework: When `requireContext: false`, `require.context` behaves like any random `require` method like `require.abc` and does not execute any special logic. However, it will throw a runtime error: `TypeError: __webpack_require__(...).context is not a function`. Users might try to hack `__webpack_require__` to implement certain methods. Based on current behavior, the functionality of `requireContext: false` seems reasonable. I prefer not to touch it.

I tested `requireDynamic` with `None` alone, all tests failed, so the flag of `requireAsExpresss` is requested for `requireDynamic`. In the latest changes, `requireDynamic` will also execute the logic of `requireAsExpresss`. Whereas these two switches remain independent.

ahabhgk: But users shouldn't hack __webpack_require__, and `requireContext: false` isn't meant to use with hack, it should work without hack, so IMO it's still a bug in `requireContext: false`

ahabhgk (dismissed):
I am ok with the current implementation, but we should keep it as experiment for now, or at least hide it from docs, and wait for implementing it in webpack to have a more stable implementation

crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
fi3ework: > But users shouldn't hack webpack_require, and requireContext: false isn't meant to use with hack, it should work without hack

Yes. What I mean is that changing `requireContext: false` might be a breaking change, previously it is a undefined behavior. Users could do anything (hack), but the impact of this is quite limited.

---

I believe that `requireContext` indicates whether the **`context` method** is specially handled by Webpack, **regardless of how the preceding `require` is processed** (methods like `requireEnsure` and `requireInclude` also behaves this way).

On the other hand, `requireAsExpress` marks whether the corresponding `require` method should be transformed into a webpack require.

> So if it's a feature, then this is expected, requireContext: false / requireDynamic: false should be used together with requireAsExpression: false

Yes! I didn't realize that before. I think this is better because it allows us to maintain the current state without causing a breaking change, while also giving us finer control over the behavior.
image

If you agree, I will update the code. `requireResolve` must be used with `requireAsExpress` to achieve preserving `resolve.resolve` as-is.

fi3ework: > But users shouldn't hack webpack_require, and requireContext: false isn't meant to use with hack, it should work without hack

Yes. What I mean is that changing `requireContext: false` might be a breaking change, previously it is a undefined behavior. Users could do anything (hack), but the impact of this is quite limited.

---

I believe that `requireContext` indicates whether the **`context` method** is specially handled by Webpack, **regardless of how the preceding `require` is processed** (methods like `requireEnsure` and `requireInclude` also behaves this way).

On the other hand, `requireAsExpress` marks whether the corresponding `require` method should be transformed into a webpack require.

> So if it's a feature, then this is expected, requireContext: false / requireDynamic: false should be used together with requireAsExpression: false

Yes! I didn't realize that before. I think this is better because it allows us to maintain the current state without causing a breaking change, while also giving us finer control over the behavior.
image

If you agree, I will update the code. `requireResolve` must be used with `requireAsExpress` to achieve preserving `resolve.resolve` as-is.

ahabhgk: I agree, let's update the code

fi3ework:
> I am ok with the current implementation, but we should keep it as experiment for now, or at least hide it from docs,

Sure. Any why to make it as experimental now? Or just using name like `experimenalRequireResolve` or `rspackRequireResolve`

ahabhgk:
An experimental warning in jsdoc and hide these options from rspack docs looks good to me

ahabhgk (approved):

Issue #7862, opened 2024-09-11
all the rust plugin stage should be changed to f64


hardfist:
currently rust plugin stage use i32 which cause it can't accept Infinity and float points in js side, which is not compatible with webpack, so the stage should be changed to f64 to be compatible with webpack

pawel-twardziak:
Can I jump on it? Rspack looks so promising :)

hardfist:
thanks in advance，go try it

hardfist:
@pawel-twardziak are you still working on it or met any problems?

pawel-twardziak:
@hardfist was busy, couldn't even kick it off :( Today and tomorrow I am on my sick leave so I cloud finally try it out. Any rush? I might need some help. But if you need it sooner feel free to give it to someone else.

pawel-twardziak:
Hi, @hardfist I've just raised the job - is that all that should get done?

hardfist:
bump

stormslowly:
in Rust side, we can't  change the stage type to f64.

if i32 changed to f64
in proc macro `define_hook` , it generates code like
```rust
        fn stage(&self) -> f64 {
            0.0
        }
                             f64 doesn't  meet the 'Eq' and 'Hash' trait in HashSet
                                                                         |
                                                                         |
                                                                        \/
        fn used_stages(&self) -> rspack_hook::__macro_helper::FxHashSet<f64> {
            rspack_hook::__macro_helper::FxHashSet::from_iter(
                self.taps.iter().map(|h| h.stage()),
            )
        }
```

and ( either the 'ord ' trait in sort_by_key  )

```rust
            let mut all_taps = std::vec::Vec::new();
            all_taps.extend(&self.taps);
            all_taps.extend(&additional_taps);
            all_taps.sort_by_key(|hook| hook.stage());
```

maybe we need to find other ways to support float stage number.

stormslowly:
@hardfist  let's try a newType

stormslowly:
rust plugin is no needs to change stage into float,  the internal usage i32 is enough.

PR #7942, opened 2024-09-20
fix(plugins): make rust plugin stage compatible with webpack

pawel-twardziak:
## Summary

see #7862 

## Checklist

### DIFF
--- a/crates/node_binding/src/plugins/interceptor.rs
+++ b/crates/node_binding/src/plugins/interceptor.rs
@@ -72,12 +72,12 @@ use rspack_plugin_javascript::{JavascriptModulesChunkHash, JavascriptModulesChun
 #[napi(object)]
 pub struct JsTap {
   pub function: JsFunction,
-  pub stage: i32,
+  pub stage: f64,
 }
 
 pub struct ThreadsafeJsTap<T: 'static, R> {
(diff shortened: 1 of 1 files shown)


hardfist:
can you add test about it, try pass Infinity to stage and see whether it works

pawel-twardziak:
@hardfist sure, will do and let you know

crates/node_binding/src/plugins/interceptor.rs
```diff
@@ -1419,7 +1419,7 @@ impl NormalModuleFactoryResolve for NormalModuleFactoryResolveTap {
   }
 
   fn stage(&self) -> i32 {
```
hardfist, line 1421: should change to f64 too

pawel-twardziak: yes, but there ara hard implications when I change it :) Let me show you @hardfist

pawel-twardziak: @hardfist When I change it, I see this while bouilding:
obraz
And it seems I need to align way more code...

stormslowly:
bump

stormslowly:
close ref https://github.com/web-infra-dev/rspack/issues/7862#issuecomment-2606463361

PR #7945, opened 2024-09-22
feat(core): support using tokio-console

hardfist:
## Summary
support debug scheduler problem using tokio console
we can spot which task takes long time using tokio-console
```sh
$ cargo install tokio-console
$ RSPACK_PROFILE=trace rspack dev -c rspack.config.js
$ tokio-console
```
image
image

## Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -30,7 +30,8 @@ rustflags = [
   "-Wclippy::same_name_method",
 
   "-Aclippy::default_constructed_unit_structs",
-  "-C", "symbol-mangling-version=v0"
+  "-C", "symbol-mangling-version=v0",
+  "--cfg", "tokio_unstable"
 ]
 
 # Fix napi breaking in test environment <https://github.com/napi-rs/napi-rs/issues/1005#issuecomment-1011034770>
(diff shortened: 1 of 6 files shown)


.cargo/config.toml
```diff
@@ -30,7 +30,8 @@ rustflags = [
   "-Wclippy::same_name_method",
 
   "-Aclippy::default_constructed_unit_structs",
-  "-C", "symbol-mangling-version=v0"
+  "-C", "symbol-mangling-version=v0",
+  "--cfg", "tokio_unstable"
```
h-a-n-a, line 34: What's this flag all about? Is it intentional?

hardfist: yes, it's required by tokio-console https://github.com/tokio-rs/console#:~:text=in%20order%20to%20collect%20task%20data%20from%20Tokio%2C%20the%20tokio_unstable%20cfg%20must%20be%20enabled.%20for%20example%2C%20you%20could%20build%20your%20project%20with

hardfist: deno also use this flag https://github.com/denoland/deno/blob/main/.cargo/config.toml#L37, so it seems safe to use it

h-a-n-a (approved):

chenjiahan:
This crate is quite big:

image

hardfist:
it's not necessary now, we can remove it

chenjiahan:
Nice 👍

PR #7812, opened 2024-09-05
fix: change javascript parser options in child compiler

ahabhgk:
## Summary

fix https://github.com/web-infra-dev/rsbuild/issues/3353

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #7946, opened 2024-09-22
fix: change javascript parser options to properly merge

inottn:
## Summary

After #7812, we should not return the default value in getRawJavascriptParserOptions, as it will break the merging logic.

I want to write test cases, but I found that the createParser hook of NormalModuleFactory is not yet supported.

## Checklist


ahabhgk (approved):
Good catch! Thanks!

PR #7961, opened 2024-09-24
perf: lazy create hash utils

chenjiahan:
## Summary

Lazy create `md4` and `xxhash64` hash utils to make the startup a little bit faster (0.3ms * 2)

Screenshot 2024-09-24 at 16 52 45

## Checklist


packages/rspack-test-tools/tests/hookCases/javascriptModulesPlugin#chunkHash/update-hash/hooks.snap.txt
```diff
@@ -38,38 +38,38 @@ Array [
...
-          55,
-          101,
-          48,
-          99,
-          48,
-          56,
-          57,
-          99,
-          48,
+          0,
```
chenjiahan, line 41: @ahabhgk Can you help to review the snapshot changes? I'm not sure what `hash.mem.data' means here.

packages/rspack/src/util/hash/md4.ts
```diff
@@ -10,14 +10,20 @@
...
+				"base64"
+			)
+		);
+		//#endregion
+	}
+
+	return create.bind(null, md4, [], 64, 32)();
```
ahabhgk, line 28: should also reuse `create.bind(null, md4, [], 64, 32)`, `[]` is a wasm instance pool

packages/rspack-test-tools/tests/hookCases/javascriptModulesPlugin#chunkHash/update-hash/hooks.snap.txt
ahabhgk: Related to the wasm instance pool, but the instance pool shouldn't affect the hash result, so the best way is adjust our test tool to snapshot the hash result instead of the hash inner structure layout

packages/rspack/src/util/hash/md4.ts
chenjiahan: Get👌 fixed

ahabhgk: The `create.bind(null, md4, [], 64, 32)` whole function can be reused, avoid repeatedly calling `.bind()`

packages/rspack-test-tools/tests/hookCases/javascriptModulesPlugin#chunkHash/update-hash/hooks.snap.txt
chenjiahan: Agreed, we can improve the test tool in another PR. The snapshot is now restored after fixing the instance pool.

packages/rspack/src/util/hash/md4.ts
chenjiahan: Better 👍

ahabhgk (approved):

PR #7963, opened 2024-09-24
perf: lazy require `browserslist` package

chenjiahan:
## Summary

In most cases, Rspack does not need to require the `browserslist` package, it is only be used if `target` contains "browserslist".

Lazy require `browserslist` can make Rspack startup **17ms** faster.

- before:

Screenshot 2024-09-24 at 17 20 31

- after:

Screenshot 2024-09-24 at 17 20 44

## Checklist


chenjiahan:
!bench

hardfist (approved):

PR #7965, opened 2024-09-24
perf: cache swc loader

CPunisher:
## Checklist

Assume: 
1. Creating an `SwcLoaderJsOptions` is costly
2. Most modules are created with same swc loader options

Cost:
1. `RwLock` and concurrent runtime.
Maybe there is a better cache strategy.

🙏 Help me run the benchmark.

### DIFF
--- a/crates/rspack_binding_options/src/plugins/js_loader/resolver.rs
+++ b/crates/rspack_binding_options/src/plugins/js_loader/resolver.rs
@@ -1,4 +1,4 @@
-use std::sync::Arc;
+use std::sync::{Arc, LazyLock};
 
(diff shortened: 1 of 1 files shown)


JSerFeng:
!bench

JSerFeng:
Thanks.

Modules in our bench cases are not that many, so the benefits may not be so obvious, but personally I think it's great for some large projects

chenjiahan:
@JSerFeng can we merge this

JSerFeng:
> @JSerFeng can we merge this

sure

JSerFeng (approved):

chenjiahan:
🎉

PR #7986, opened 2024-09-25
refactor(core): kick async task first

hardfist:
## Summary
currently we run job by FIFO order which may cause async job is blocked by sync job, so try to kick async job first so it can run parallel with sync job

## Checklist


hardfist:
!bench

PR #7994, opened 2024-09-26
docs: add ReleaseV1Announcement to 0.x

SoonIter:
Packages:0.7.5


website/theme/components/Announcement/index.tsx
```diff
@@ -1,49 +1,43 @@
...
           setDisable(true);
-          window.localStorage.setItem('disabled-hire', 'true');
-        }}
-        style={{
-          right: 10,
-          color: 'white',
-          fontSize: 18,
-          position: 'absolute',
-          cursor: 'pointer',
+          // window.localStorage.setItem(localStorageKey, 'true');
```
chenjiahan, line 37: The close icon can be mored if we do not use local storage

SoonIter: done

chenjiahan (approved):

PR #7996, opened 2024-09-26
fix(modern-module): use module style export for non entry module

fi3ework:
## Summary

- If the variable to be exported is not in the inlined module (commonly triggered by async modules), should use the "module" style export way to export it.
- Deleted the test cases of `library` in rspack test which are duplicated with existing webpack test.
- Remove `--` from `testu` command, I do not know when it became redundant and will fail the command.

## Checklist


LingyuCoder:
The `-- -u` in `testu` works well for me. It only works when using `npm run testu`. If you use `pnpm run testu`, the `--` will be converted to `"--"` and then the test will regard it as a path and run without any test cases. Perhaps we can use a bash script for compatibility.

fi3ework:
> The `-- -u` in `testu` works well for me. It only works when using `npm run testu`. If you use `pnpm run testu`, the `--` will be converted to `"--"` and then the test will regard it as a path and run without any test cases. Perhaps we can use a bash script for compatibility.

The newer version (6.2.1, we're using 6.2.0) of npm-run-all2 has fix the `--` dilemma https://redirect.github.com/bcomnes/npm-run-all2/pull/143. I'll launch a new PR to fix this. Revert the change in this PR.

LingyuCoder (approved):

PR #8001, opened 2024-09-26
fix(diagnostic): improve diagnostics for swc wasm plugins when mismatch

SyMind:
## Summary

improve diagnostics when wasm plugin binary format mismatch, redirect to https://rspack.dev/errors/swc-plugin-version

image

## Checklist


Cargo.toml
```diff
@@ -87,17 +87,15 @@ rspack_dojang = { version = "0.1.9" }
 rate-limit = { existing-packages = 70, new-packages = 70 }
 [profile.dev]
 codegen-units = 16
-debug         = 2       # debug build will cause runtime panic if codegen-unints is default
```
hardfist, line 90: no need change this file now

h-a-n-a:
What is the context?

chenjiahan:
I think the motivation for this PR is to add the test case back, @SyMind right?

chenjiahan:
We can change the PR title to make it more accurate.

SyMind:
@hardfist let me revert this PR. Initially, I thought `swc` converted panics into `Result`, which would allow us to modify the error handling. However, after reviewing the swc code, I found that no such change has been made.

Therefore, I believe I misunderstood the purpose of this PR. We need @hardfist to take a look at it.

hardfist:
> What is the context? cc @hardfist @chenjiahan
swc support throw error when wasm binary not match now, so it's safe to add this test case back without enabling catch_on_unwind
https://github.com/swc-project/swc/pull/9562/files#diff-4d01e730818bcc35a17d65dd364390283fd29e3fb1dc0e35050a67034eeb474dR111

crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -346,16 +350,37 @@ impl SwcCompiler {
...
+      match result {
+        Ok(v) => Ok(v),
+        Err(err) => {
+        let error_msg = match err.downcast_ref::<String>(){
+          Some(msg) => {
+            msg
+          },
+          None => "unknown error"
+        };
+        // FIXME: with_help has bugs, use with_help when diagnostic print is fixed
```
hardfist, line 372: ModuleBuildError will double print help msg if diagnostic contains help, which should be fixed

crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -346,16 +350,37 @@ impl SwcCompiler {
...
+        let help_msg =  indoc!{"
+          The version of the SWC Wasm plugin you're using might not be compatible with `builtin:swc-loader`.
+          The `swc_core` version of the current `rspack_core` is {swc_core_version}. 
+          Please check the `swc_core` version of SWC Wasm plugin to make sure these versions are within the compatible range.
+          Versions of `swc_core` are more likely to be located in the `Cargo.toml` file in the root directory of these plugin repositories.
+          Check out this guide as a reference for selecting the versions of SWC Wasm plugins: https://rspack.dev/errors/swc-plugin-version"};
```
chenjiahan, line 378:
```suggestion
          See this guide as a reference for selecting SWC Wasm plugin versions: https://rspack.dev/errors/swc-plugin-version"};
```

crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -346,16 +350,37 @@ impl SwcCompiler {
...
+            msg
+          },
+          None => "unknown error"
+        };
+        // FIXME: with_help has bugs, use with_help when diagnostic print is fixed
+        let help_msg =  indoc!{"
+          The version of the SWC Wasm plugin you're using might not be compatible with `builtin:swc-loader`.
+          The `swc_core` version of the current `rspack_core` is {swc_core_version}. 
+          Please check the `swc_core` version of SWC Wasm plugin to make sure these versions are within the compatible range.
+          Versions of `swc_core` are more likely to be located in the `Cargo.toml` file in the root directory of these plugin repositories.
```
chenjiahan, line 377: Maybe we can remove this line. It is hard for normal JS developers to understand, and we have mentioned this in the FAQ doc.

```suggestion
```

h-a-n-a (approved):

Issue #4241, opened 2023-09-25
Add JSDoc for all configuration options as webpack


chenjiahan:
### What problem does this feature solve?

Can we add JSDoc for all configuration options as webpack?

- webpack:

<img width="692" alt="Screenshot 2023-09-25 at 11 03 39" src="https://github.com/web-infra-dev/rspack/assets/7237365/cad22942-9c6d-4e6c-9bee-6cecab5a2b5d">

- Rspack:

![image](https://github.com/web-infra-dev/rspack/assets/7237365/2880cf43-54c0-45f8-8fbd-3323f9a816bb)

### What does the proposed API of configuration look like?

Same as webpack:

![image](https://github.com/web-infra-dev/rspack/assets/7237365/aa80951e-15ed-49f6-9c10-3b3157e1165b)

hyf0:
The type of rspack options is inferred by `zod`. It's not possible in the current time. See https://github.com/colinhacks/zod/issues/200#issuecomment-718979724 for details.

chenjiahan:
It seems that JSDoc can work in some cases:

![image](https://github.com/web-infra-dev/rspack/assets/7237365/f55f4da5-03cf-4605-b1c4-4b2a529cacfb)

https://github.com/colinhacks/zod/issues/200#issuecomment-1198922371

dinckelman:
This appears to be very much situational, because as shown below, VSCode cannot detect the comment, and Webstorm is... well, trying to do something else entirely
<img width="406" alt="Screenshot 2023-10-14 at 23 21 57" src="https://github.com/web-infra-dev/rspack/assets/12545914/ede19f05-d36e-4bdd-8693-9e1a4cef2a8f">
<img width="682" alt="Screenshot 2023-10-14 at 23 21 32" src="https://github.com/web-infra-dev/rspack/assets/12545914/d9483972-0de5-4ce3-8233-cfd8672b36b6">

hardfist:
it indeed hurt DX for users，we will try to find better ways to do this

xc2:
bump

chenjiahan:
bump

chenjiahan:
bump

chenjiahan:
@hardfist  handwritten json schema or handwritten config types?

ecraig12345:
[...]

PR #8003, opened 2024-09-27
refactor: add JSDoc to improve BannerPlugin options types

GiveMe-A-Name:
## Summary

Zod types system is not easy to understand by user. 

We tend use raw ts provider intellisense and jsDoc for user in ide, 

See https://github.com/web-infra-dev/rspack/issues/4241 more detail.

## Checklist

### DIFF
--- a/packages/rspack/etc/api.md
+++ b/packages/rspack/etc/api.md
@@ -395,44 +395,20 @@ export type Bail = z.infer<typeof bail>;
 const bail: z.ZodBoolean;
 
 // @public (undocumented)
+type BannerContent = string | BannerFunction;
+
+// @public (undocumented)
+type BannerFunction = (args: {
+    hash: string;
+    chunk: JsChunk;
+    filename: string;
+}) => string;
+
+// @public (undocumented)
 export const BannerPlugin: {
-    new (args: string | ((args_0: {
-        filename: string;
-        hash: string;
-        chunk: JsChunk;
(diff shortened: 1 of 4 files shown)


packages/rspack/src/builtin-plugin/BannerPlugin.ts
```diff
@@ -3,45 +3,48 @@ import {
...
+	/** Include all modules matching any of these conditions. */
+	include?: Rules;
+
+	/** If true, banner will not be wrapped in a comment. */
+	raw?: boolean;
+
+	/** If true, banner will be placed at the end of the output. */
+	footer?: boolean;
+
+	/** The stage of the compilation in which the banner should be injected. */
```
chenjiahan, line 40: Adding the default value to JSDoc would be helpful:

```suggestion
/**
 * The stage of the compilation in which the banner should be injected.
 * @default PROCESS_ASSETS_STAGE_ADDITIONS (-100)
 */
```

GiveMe-A-Name: Got it

packages/rspack/etc/api.md
chenjiahan, line 1: For the PR title, I prefer to use `refactor` or `feat` instead of `chore` because it can benefit the user.

For example, `feat: add JSDoc to improve BannerPlugin options types`

LingyuCoder (approved):

PR #8034, opened 2024-10-01
perf: move swc transform into another thread

hardfist:
## Summary
try to move swc into another thread to see whether performance improve

## Checklist


hardfist:
!bench

hardfist:
!bench

hardfist:
!bench

hardfist:
!bench

hardfist:
I don't know why rayon::spawn is much faster than spawn_blocking here, one guess is tokio has limited blocking threads, so it will block the blocking_threads when spawn_block is large than default blocking_threads number(512)

PR #8064, opened 2024-10-08
fix: replace deprecated fs.rmdir with recursive removal logic

nexckycort:
## Summary

This pull request addresses a deprecation warning in Node.js regarding the use of `fs.rmdir(path, { recursive: true })`. In future versions, this method will be removed.

To resolve this, I replaced `fs.rmdir` with `fs.rm(path, { recursive: true })` as recommended by the warning. This change maintains the functionality while adhering to future Node.js standards.
image

## Checklist

### DIFF
--- a/packages/rspack/src/util/fs.ts
+++ b/packages/rspack/src/util/fs.ts
@@ -10,8 +10,8 @@
 
 import assert from "node:assert";
 import type { Abortable } from "node:events";
-import path from "node:path";
 import type { RmOptions } from "node:fs";
+import path from "node:path";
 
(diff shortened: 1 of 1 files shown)


SoonIter:
this is a breaking change and it is not aligned with Webpack

It will cause fs which does not implenment `fs.rm(, {recursive: true})` error

image

https://github.com/webpack/webpack/blob/e237b580e2bda705c5ab39973f786f7c5a7026bc/types.d.ts#L10391-L10406

so my opinion is temporarily closing this change, and reopen when webpack has corresponding breaking changes of `OutputFileSystem` API first

PR #8069, opened 2024-10-09
fix: allow file names and directory names to contain brackets

### LINKED ISSUE #8056: [Bug]: CopyRspackPlugin.pattern.from not support to specify a file path not regexp
### System Info

Mac os

### Details

There appears to be a bug in Rspack's CopyPlugin where it does not properly handle patterns that specify a single file. This functionality is supported in webpack's CopyWebpackPlugin, but it seems to be missing or not working as expected in Rspack.

### Reproduce link

_No response_

### Reproduce Steps

1. Configure CopyPlugin in Rspack config
2. Attempt to copy a specific file using a pattern

```ts
{
  from: '/xx/xx.ts'
}
```

But `/xx/xx.ts*` is working

Expected Behavior:
The plugin should be able to copy a single, specific file when the pattern points to that file.

Actual Behavior:
The plugin fails to copy the file or throws an error (please specify the exact behavior you're seeing).

inottn:
## Summary
fix #8056 

## Checklist


JSerFeng (approved):

JSerFeng:
Thank you !

PR #8072, opened 2024-10-09
perf: bundle `@rspack/core` for faster startup

chenjiahan:
## Summary

- This PR uses `tsup` to bundle `@rspack/core`, this reduces the number of modules required by Node.js and speeds up Rspack startup time.
- Move runtime code to `src/runtime` folder for easier management.
- Remove some packages from `prebundle.config.js` because they can be bundled to `dist/index.js`.
- Once Rslib is ready, we will switch from tsup to Rslib.

> TODO: `tests/plugin-test/css-extract/stringifyLocal.test.js` and `packages/rspack/scripts/check-documentation-coverage.mjs` depend on internal dist files and should be fixed in separate PRs.

## Comparasion

`require('@rspack/core')` **become 30ms faster**:

- before (65~75ms):

```bash
hyperfine --warmup 3 'node ./node_modules/@rsbuild/core/bin/rsbuild.js build'
Benchmark 1: node ./node_modules/@rsbuild/core/bin/rsbuild.js build
  Time (mean ± σ):     345.4 ms ±  23.0 ms    [User: 385.2 ms, System: 65.3 ms]
  Range (min … max):   328.7 ms … 405.7 ms    10 runs
```

Screenshot 2024-10-09 at 17 50 37

- after (35~40ms):

```bash
hyperfine --warmup 3 'node ./node_modules/@rsbuild/core/bin/rsbuild.js build'
Benchmark 1: node ./node_modules/@rsbuild/core/bin/rsbuild.js build
  Time (mean ± σ):     308.8 ms ±   4.0 ms    [User: 350.4 ms, System: 58.6 ms]
  Range (min … max):   304.6 ms … 319.0 ms    10 runs
```

Screenshot 2024-10-09 at 18 09 26

## Checklist


chenjiahan:
!eco-ci

chenjiahan:
The devserver CI failed because we need to update the module path snapshots. I will update them after releasing a new Rspack version.

image

fi3ework:
> Remove some packages from prebundle.config.js because they can be bundled to dist/index.js.

Do we have guidelines for deciding when to use `prebundle` or direct bundle? I assume we should use `prebundle` when direct bundling isn't possible, as it can handle more edge cases.

chenjiahan:
Currently we have two cases to use `prebundle`:

1. Use `prebundle` for chunk splitting. For example, we want to load `browserslist` on demand, so we can use prebundle to output a separate `browserslist/index.js` chunk. This can be removed after Rslib supports chunk splitting.
2. Use `prebundle` to bundle the types of thrid-party packages. For example, we need the types of zod and `prebundle` can handle it.

fi3ework (approved):

Issue #7914, opened 2024-09-17
[Bug]: The source map is not working on Sentry


cjnoname:
### System Info

  System:
    OS: macOS 15.0
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.03 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.8.0 - ~/.nvm/versions/node/v22.8.0/bin/node
    npm: 10.8.3 - ~/.nvm/versions/node/v22.8.0/bin/npm
    bun: 1.1.27 - /usr/local/bin/bun
  Browsers:
    Chrome: 128.0.6613.138
    Safari: 18.0

### Details

The source maps stopped working on Sentry. After discussing with their team, we identified that the issue was caused by an invalid source map generated by Rspack.

When testing the source map with source-map-explorer or other tools, the following error appeared:

'Your source map refers to generated column 53099 on line 39, but the source only contains 53098 columns on that line. Please verify that you are using the correct source map.'

Approximately 70% of the source map files we tested are working, while 30% have failed.

### Reproduce Steps

https://github.com/cjnoname/rspack-reproduce

npm run s

hardfist:
please provide rspack version

cjnoname:
> please provide rspack version

1.0.5 the latest

cjnoname:
> please provide rspack version

I could send you the buggy soucemap files and mjs files via email or whatever tools but I don't want to share them with public.

gitim:
I think this is the same issue as https://github.com/web-infra-dev/rspack/pull/7647. we found that source map stopped generated properly since 1.0.1 version. @cjnoname, you can try the 1.0.0 version of rspack to identify when it was broken.

gitim:
I think this is the same issue as https://github.com/web-infra-dev/rsbuild/issues/3433 we found that source map stopped generated properly since 1.0.1 version. @cjnoname, you can try the 1.0.0 version of rspack to identify when it was broken.

cjnoname:
@hardfist 

This is a significant issue since Sentry is the primary monitoring tool for many companies, and I hope someone can assist in looking into it.

Verifying the issue is straightforward - a single command line should work:

npx source-map-explorer index.mjs index.mjs.map

hardfist:
> @hardfist
> 
> This is a significant issue since Sentry is the primary monitoring tool for many companies, and I hope someone can assist in looking into it.
> 
[...]

PR #8080, opened 2024-10-10
fix: source map is not works with Sentry

SyMind:
## Summary

Fix: #7914

Relative: https://github.com/web-infra-dev/rspack/issues/7914

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -318,9 +318,9 @@ dependencies = [
 
 [[package]]
 name = "better_scoped_tls"
-version = "0.1.1"
+version = "0.1.2"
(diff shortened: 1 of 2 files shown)


chenjiahan (commented):
Should we add a test case to avoid source map regression?

SyMind:
@chenjiahan There is a test case in swc. I don't think we should add related test cases in Rspack because this is a matter for swc.

But I'm releasing a canary version for Rspack to give users the opportunity to check.

SyMind:
!eco-ci

chenjiahan (approved):

PR #8089, opened 2024-10-11
refactor: remove connection id

jerrykingxyz:
## Summary

Currently we use `ModuleIdentifier` to access `Module` and `ModuleGraphModule`, but need to convert the dependency_id to the connection_id to access the `Connection`, which will cause additional performance loss.

``` rust
impl ModuleGraph {
    fn connection_by_dependency_id(&self, dep_id: &DependencyId) -> Option<Connection> {
        let con_id = self.dependency_id_to_connection_id.get(dep_id)?;
        self.connections.get(con_id)
    }
}
```

This PR will remove ConnectionId and use DependencyId to access Connection.

``` rust
impl ModuleGraph {
    fn connection_by_dependency_id(&self, dep_id: &DependencyId) -> Option<Connection> {
        self.connections.get(dep_id)
    }
}
```

NOTE: There are some inactive connections in the module graph that cannot be found from the dependency ID. This is useful when you need to print debug information, but it makes the module graph more cluttered. We will use some other methods to ensure these features.

## Checklist

### DIFF
--- a/crates/rspack_core/src/build_chunk_graph/code_splitter.rs
+++ b/crates/rspack_core/src/build_chunk_graph/code_splitter.rs
@@ -16,11 +16,11 @@ use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet, FxHasher};
 
 use crate::dependencies_block::AsyncDependenciesToInitialChunkError;
 use crate::{
-  add_connection_states, assign_depth, assign_depths, get_entry_runtime, merge_runtime,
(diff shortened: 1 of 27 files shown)


jerrykingxyz:
!bench

jerrykingxyz:
!bench

jerrykingxyz:
!bench

ahabhgk (approved):

PR #7940, opened 2024-09-20
feat: support `LoaderContext.loadModule`

CPunisher:
## Summary

closes https://github.com/web-infra-dev/rspack/issues/3738
closes https://github.com/web-infra-dev/rspack/issues/7723

TODOs or existing bugs to fix:
- [ ] Resolver options of referenced modules should extend their parent modules. Related to https://github.com/web-infra-dev/rspack/blob/833c058da7a9d1c2a15cd9be49065d7c75c7ed7a/tests/webpack-test/cases/loaders/issue-4959/loaders/index.js, "b" is defined in the browser field in `package.json`, but it's not resolved.
- [ ] Handle module cycle. Related to https://github.com/web-infra-dev/rspack/blob/833c058da7a9d1c2a15cd9be49065d7c75c7ed7a/tests/webpack-test/cases/errors/load-module-cycle/index.js. Currently there seems no cycle detection in rspack (https://github.com/web-infra-dev/rspack/issues/6048).
- [ ] Other options that differ from the options of `importModule`: `recursive`, `connectOrigin`, `checkCycle`

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #8096, opened 2024-10-12
refactor: Send module executor tasks to the main task loop

CPunisher:
## Summary

Module executor should not use a separate artifact. To use the artifact of the main task loop, we should send the `importModule` and `loadModule` tasks to the main task loop. This also aligns to webpack.

This could be a pre-requirement of https://github.com/web-infra-dev/rspack/pull/7940

## Checklist


jerrykingxyz:
The `ModuleExecutor` is designed as independent modules, which prevents the side effects generated by `ModuleExecutor` from affecting the module_graph. What is the problem you want to fix in this PR ?

CPunisher:
> The `ModuleExecutor` is designed as independent modules, which prevents the side effects generated by `ModuleExecutor` from affecting the module_graph. What is the problem you want to fix in this PR ?

Is this consistent with webpack? I find that the target modules of `importModule` and `loadModule` may depend on some information of their original modules. Is it possible under current design?

jerrykingxyz:
The `ImportModule` needs to be built and executed using `ModuleExecutor`, but its results should not depend on any modules generated in the "ModuleExecutor".

I think the "LoadModule" should generate a new task to the main task loop, and it should not have any relation with `ModuleExector`.

Could you provide some code example to shows the problem ? I think it should be fixed in the current `ModuleExecutor` implementation.

CPunisher:
> The `ImportModule` needs to be built and executed using `ModuleExecutor`, but its results should not depend on any modules generated in the "ModuleExecutor".
> 
> I think the `LoadModule` should generate a new task to the main task loop, and it should not have any relation with `ModuleExector`.
> 
> Could you provide some code example to shows the problem ? I think it should be fixed in the current `ModuleExecutor` implementation.

One of the problems comes from https://github.com/web-infra-dev/rspack/pull/7940:

> Resolver options of referenced modules should extend their parent modules. Related to https://github.com/web-infra-dev/rspack/blob/833c058da7a9d1c2a15cd9be49065d7c75c7ed7a/tests/webpack-test/cases/loaders/issue-4959/loaders/index.js, "b" is defined in the browser field in package.json, but it's not resolved.

Base on my current implementation, the referenced module should extend the resolve options of its original module in https://github.com/web-infra-dev/rspack/blob/aad975b2bae8b1988fbe9bf0fa269c7a0d824f03/crates/rspack_core/src/compiler/module_executor/entry.rs#L61.

I agree if there's any simpler way to achieve it, since I find this part of code is error-prone.

jerrykingxyz:
Ok, I got it. There are ways to share resolve_option from parent module, like keeping a global map to save the relation from parent module to resolve_option . But I don't think it's a good idea.

I think we should not use module executor to implement loadModule, the better way is to provide a callback when module and sub module are finish in main task loop, like `ModuleExecutor` implementation. I will try to implement it next week.

PR #8099, opened 2024-10-13
fix: avoid extra runtime for get javascript chunk filename

inottn:
## Summary

Align with webpack: https://github.com/webpack/webpack/blob/e7121fe9972b182832d401cabaf9d5ca489b1713/lib/RuntimePlugin.js#L286

## Checklist


LingyuCoder (approved):
Thanks

PR #8113, opened 2024-10-14
feat(rspack_core): use simple string parsing instead of regex

shulaoda:
## Summary

Related to #7907 

## Checklist


h-a-n-a (approved):
Thanks

PR #8119, opened 2024-10-15
feat(cli): add exports field

chenjiahan:
## Summary

- Add exports field to the `@rspack/cli` package, align with `@rspack/core` and `@rspack/dev-server`.
- Use `sort-package-json` to sort the package.json file.

## Checklist

### DIFF
--- a/packages/rspack-cli/package.json
+++ b/packages/rspack-cli/package.json
@@ -1,26 +1,7 @@
 {
   "name": "@rspack/cli",
   "version": "1.0.10",
-  "license": "MIT",
   "description": "CLI for rspack",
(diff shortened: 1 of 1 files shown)


hardfist:
what will consume @rspack/cli js API?

chenjiahan:
@hardfist Only a few public API, such as `defineConfig` and `definePlugin`:

```js
// rspack.config.ts
import { defineConfig } from '@rspack/cli';
```

I add the `exports` field to prevent the internal modules of `@rspack/cli` from being mistakenly consumed by users.

hardfist (approved):

PR #8122, opened 2024-10-15
test: not to display obsoleted snapshots when test with `-t`

LingyuCoder:
## Summary

There is no need to output all the obsoleted snapshots when test with `-t` which filters the test cases.

Before:

image

After:
image

## Checklist


chenjiahan (approved):
LGTM

PR #8136, opened 2024-10-16
fix: should not panic when passing `test` option to SourceMapDevToolPlugin

inottn:
## Summary
Passing the `test` option to `SourceMapDevToolPlugin` now results in the following error:

Message:  Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
Location: crates/rspack_binding_options/src/options/raw_devtool.rs:99

## Checklist


h-a-n-a (approved):

chenjiahan:
Thanks 😄

PR #8145, opened 2024-10-16
chore(deps): update typescript to v5.6

SoonIter:
## Summary
chore(deps): update ts 5.6
close https://github.com/web-infra-dev/rspack/pull/7211

after https://github.com/web-infra-dev/rspack/pull/8072, we don't rely on tsc to generate js files, so we can upgrade the version of typescript.

1. upgrade and unify the ts version `^5.6.3`
2. add renovate to revert https://github.com/web-infra-dev/rspack/pull/5470/files

## Checklist

### DIFF
--- a/api-extractor.json
+++ b/api-extractor.json
@@ -144,15 +144,21 @@
     "enabled": true,
 
     /**
+     * DEFAULT VALUE: "[ "complete" ]"
+     */
+    "reportVariants": ["complete"],
+
+    /**
      * The filename for the API report files.  It will be combined with "reportFolder" or "reportTempFolder" to produce
      * a full file path.
      *
      * The file extension should be ".api.md", and the string should not contain a path separator such as "\" or "/".
+     * API Extractor will automatically append an appropriate file extension such as .api.md
      *
      * SUPPORTED TOKENS: <packageName>, <unscopedPackageName>
-     * DEFAULT VALUE: "<unscopedPackageName>.api.md"
+     * DEFAULT VALUE: "<unscopedPackageName>"
      */
-    "reportFileName": "api.md",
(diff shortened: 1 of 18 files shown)


api-extractor.json
```diff
@@ -143,16 +143,22 @@
...
+     * DEFAULT VALUE: "[ "complete" ]"
+     */
+    "reportVariants": ["complete"],
+
     /**
      * The filename for the API report files.  It will be combined with "reportFolder" or "reportTempFolder" to produce
      * a full file path.
      *
      * The file extension should be ".api.md", and the string should not contain a path separator such as "\" or "/".
+     * API Extractor will automatically append an appropriate file extension such as .api.md
```
SoonIter, line 156: this a breaking change of `api-extractor`
before: `api.md`
now: `<package-name>(.beta?).api.md`

packages/rspack-test-tools/etc/test-tools.api.md
```diff
@@ -4,10 +4,6 @@
 
 ```ts
 
-/// <reference types="../jest.d.ts" />
```
SoonIter, line 7: I do not know if this is the change of `api-extractor`, it does not matter I think

SoonIter:
A noteworthy point is that `rspack-cli` is still using tsc and will preserve `await import` after this change.

We recommend using node18 and above, so I don't think there will be any problems.

@chenjiahan cc

chenjiahan (approved):

chenjiahan:
> A noteworthy point is that rspack-cli is still using tsc and will preserve await import after this change.

LGTM, and `@rspack/cli` will switch to Rslib soon

PR #8147, opened 2024-10-16
fix: perf degradation in devtool plugin

SyMind:
## Summary

Fix perf degradation in #8136

## Checklist

### DIFF
--- a/crates/rspack_plugin_devtool/src/source_map_dev_tool_plugin.rs
+++ b/crates/rspack_plugin_devtool/src/source_map_dev_tool_plugin.rs
@@ -173,15 +173,24 @@ impl SourceMapDevToolPlugin {
     let output_options = &compilation.options.output;
     let map_options = MapOptions::new(self.columns);
 
-    let futures = raw_assets
+    let matches = if let Some(test) = &self.test {
+      let features = raw_assets.iter().map(|(file, _)| test(file.to_owned()));
+      join_all(features)
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>>>()?
+    } else {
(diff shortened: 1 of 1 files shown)


SyMind:
!bench

h-a-n-a (approved):

chenjiahan:
How was this perf degradation discovered?

SyMind:
@chenjiahan The following code caused performance degradation:

```rust
raw_assets
     .into_par_iter()
     .map(|(file, asset)| async {
        // do something
    });
```

The above code uses the `into_par_iter` method, expecting to leverage `Rayon` for concurrent iteration. However, the closure passed to `map` won't execute the synchronous code concurrently because the synchronous code is wrapped within an `async` block. The `async` block is an anonymous structure that implements the `Future` trait and the closure simply returns this structure.

Rayon only handles the creation of these Future trait instances concurrently.

chenjiahan:
Get 👍

Narretz:
Can someone explain how the benchmark works? Because I see "current" here, and it has +% almost everywhere. Doesn't it mean the change is slower than before?

SyMind:
@Narretz The benchmark differs from https://github.com/web-infra-dev/rspack/commit/4ea491e84c77d42917422844b0e2ca72a73bb265.

Within this range, there is a specific performance degradation caused by rolling back a particular algorithm in rspack-sources. However, after rolling back the code, I found that the degradation was greater than expected, with a maximum degradation of 6.83%.

I reviewed other commits and discovered another unexpected degradation. This commit is intended to fix that, and then the degradation of this benchmark is as expected.

PR #8152, opened 2024-10-17
refactor: add JSDoc to improve config.target, devtool, node, cache types.

GiveMe-A-Name:
## Summary

Zod types system is not easy to understand by user.

We tend use raw ts provider intellisense and jsDoc for user in ide.

- Add type & JSDoc for config.target
- Add type & JSDoc for config.devtool
- Add type & JSDoc for config.node
- Add type & JSDoc for config.cache
- Add type & JSDoc for config.externalsPresets

See https://github.com/web-infra-dev/rspack/issues/4241 more detail.

## Checklist

### DIFF
--- a/packages/rspack/src/config/types.ts
+++ b/packages/rspack/src/config/types.ts
@@ -1220,6 +1220,48 @@ export type ModuleOptions = {
 
 //#endregion
 
+//#region Target
+type AllowTarget =
+	| "web"
+	| "webworker"
+	| "es3"
+	| "es5"
+	| "es2015"
+	| "es2016"
+	| "es2017"
+	| "es2018"
+	| "es2019"
+	| "es2020"
+	| "es2021"
+	| "es2022"
+	| "node"
+	| "async-node"
+	| `node${number}`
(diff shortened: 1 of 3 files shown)


packages/rspack/src/config/types.ts
```diff
@@ -1323,6 +1365,188 @@ export type ExternalItem =
...
+	console?: Console;
+
+	/**
+	 * Enable debug information of specified loggers such as plugins or loaders.
+	 */
+	debug?: boolean | FilterTypes;
+
+	/**
+	 * Enable infrastructure logging output.
+	 * @type {"none" | "error" | "warn" | "info" | "log" | "verbose"}
```
chenjiahan, line 1437: Do we need to use `@type` here? it is the same as the TS type

GiveMe-A-Name: We don't need it.

GiveMe-A-Name: Updated .

chenjiahan (approved):
👍

PR #8156, opened 2024-10-17
feat: add rspack_cacheable lib

jerrykingxyz:
## Summary

https://github.com/web-infra-dev/rspack/discussions/8185

## Checklist


crates/rspack_cacheable/src/context.rs
```diff
@@ -0,0 +1,65 @@
...
+  pub fn new(context: &'a dyn Any) -> Self {
+    Self { context }
+  }
+
+  pub fn add_to_sharing<S: Sharing<SerializeError>>(
+    &self,
+    sharing: &mut S,
+  ) -> Result<(), SerializeError> {
+    sharing.start_sharing(CONTEXT_ADDR);
+    sharing.finish_sharing(CONTEXT_ADDR, self as *const _ as usize)
```
quininer, line 28: Consider using [.addr](https://doc.rust-lang.org/stable/std/primitive.pointer.html#method.addr) or [sptr](https://docs.rs/sptr/0.3.2/sptr/index.html) crate when cast pointers. but I understand it is not mandatory at the moment.

LingyuCoder (approved):

PR #8157, opened 2024-10-17
fix: compatible static methods with html-webpack-plugin 5.6.0

LingyuCoder:
## Summary

Currently, the latest published version of `html-webpack-plugin` is `5.6.0` and need to get hooks by `HtmlWebpackPlugin.getHooks`.

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/HtmlRspackPlugin.ts
+++ b/packages/rspack/src/builtin-plugin/HtmlRspackPlugin.ts
@@ -378,6 +378,9 @@ const compilationOptionsMap: WeakMap<Compilation, HtmlRspackPluginOptions> =
 	new WeakMap();
 
 const HtmlRspackPlugin = HtmlRspackPluginImpl as typeof HtmlRspackPluginImpl & {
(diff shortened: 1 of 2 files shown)


packages/rspack/src/builtin-plugin/HtmlRspackPlugin.ts
```diff
@@ -378,6 +378,7 @@ const compilationOptionsMap: WeakMap<Compilation, HtmlRspackPluginOptions> =
 	new WeakMap();
 
 const HtmlRspackPlugin = HtmlRspackPluginImpl as typeof HtmlRspackPluginImpl & {
+	getHooks: (compilation: Compilation) => HtmlRspackPluginHooks;
```
chenjiahan, line 381: Should we mark `getHooks` as deprecated method? The same as https://github.com/jantimon/html-webpack-plugin/blob/41f4a7b952e80eea4878d96638b4021f4a762750/index.js#L1593

chenjiahan (approved):

Issue #8095, opened 2024-10-11
[Bug]: Default export from `@rspack/core` does not have TypeScript types


mxmason:
### System Info

  System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 71.22 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - ~/.nvm/versions/node/v20.11.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.11.1/bin/yarn
    npm: 10.2.4 - ~/.nvm/versions/node/v20.11.1/bin/npm
    pnpm: 8.15.6 - ~/.nvm/versions/node/v20.11.1/bin/pnpm
    bun: 1.1.12 - ~/.bun/bin/bun
  Browsers:
    Chrome: 129.0.6668.91
    Safari: 17.5
  npmPackages:
    @rspack/cli: ^1.0.10 => 1.0.10 
    @rspack/core: ^1.0.10 => 1.0.10 
    @rspack/dev-server: ^1.0.8 => 1.0.8 

### Details

While the following code runs as expected, the types for Compiler#run only show up if one imports `rspack` as a **named** import. Note: this is true in both ESM and CommonJS.

```js
const rspack = require('@rspack/core')

const compiler = rspack(config)

compiler.run(...)
```

![Image](https://github.com/user-attachments/assets/7264f430-10eb-458a-8ace-d040e7f2678f)
![Image](https://github.com/user-attachments/assets/8c0a7781-6e66-422a-a876-3cd6d25dfe97)

### Reproduce Steps

In an rspack project, try the following import scenarios

**ESM, has types:**
```js
import { rspack } from '@rspack/core'

const configs = require('./path/to/config.js')

const compiler = rspack(configs)
```

**ESM, no types:**
```js
import rspack from '@rspack/core'

const configs = require('./path/to/config.js')

const compiler = rspack(configs)
```

**CommonJS, has types:**
```js
const { rspack } = require('@rspack/core')

const configs = require('./path/to/config.js')

const compiler = rspack(configs)
```

**CommonJS, no types:**
```js
const rspack = require('@rspack/core')

const configs = require('./path/to/config.js')

const compiler = rspack(configs)
```

SoonIter:
In fact, rspack does not have default export in esm. It can work normally because it is converted to cjs. At present, there is only cjs dist in `@rspack/core`

[...]

PR #8169, opened 2024-10-18
fix(type): function type in cjs require

SoonIter:
## Summary

fix the function type in cjs require

close https://github.com/web-infra-dev/rspack/issues/8095

```ts
// in esm
import { rspack } from '@rspack/core';
rspack({ bail: true });

// in cjs
const rspack = require("@rspack/core");
// or const { rspack } = require("@rspack/core");
rspack({ bail: true });
```

## Checklist


SoonIter:
This is a solution with a little hack.
Maybe we need to ship both cjs and esm in the future.

packages/rspack-cli/tests/api/type/js-api-cjs.test.ts
```diff
@@ -0,0 +1,31 @@
+import rspackCjsRequire, {
```
chenjiahan, line 1: What's the difference between this cjs case and the esm case? They seem to be the same.

SoonIter: one is mts, the other is ts

chenjiahan: get👌

chenjiahan:
> Maybe we need to ship both cjs and esm in the future.

Agreed, `@rspack/core` should be a dual package.

chenjiahan (approved):

PR #8172, opened 2024-10-19
fix: escape node.__dirname

### LINKED ISSUE #8170: [Bug]: node.__dirname: true causes broken output on Windows
### System Info

  System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-7820HQ CPU @ 2.90GHz
    Memory: 5.78 GB / 15.82 GB
  Binaries:
    Node: 20.18.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.4 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.2 - ~\AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Chrome: 129.0.6668.101
    Edge: Spartan (44.19041.1266.0), Chromium (129.0.2792.89)
    Internet Explorer: 11.0.19041.1566

### Details

**Note:** This issue only reproduces when building on Windows, as Windows uses the `\` path separator.

If 

1. `node: { __dirname: true }` ([webpack documentation](https://webpack.js.org/configuration/node/#node__dirname)) is used in the RsPack config and
2. The code being built is in a subdirectory and refers to `__dirname`
3. The RsPack build is run on Windows via the Windows `cmd` command prompt

then RsPack output contains unescaped `\` characters (e.g. ``__dirname: ${'src\foo\bar\baz'}``), resulting in an unexpected string and not matching the webpack output (e.g. ``var __dirname = "src\\foo\\bar\\baz"; ... `__dirname: ${__dirname}``).

(Thank you for your awesome work on this project 🙏❤ . Hope this report proves to be useful!)

### Reproduce link

https://github.com/lauriharpf/rspack-node-dirname/tree/main

### Reproduce Steps

1. Use `node: { __dirname: true }` in the RsPack configuration.
2. Refer to `__dirname` in a subdirectory file in the code being built (for example, [like in the reproduce repo](https://github.com/lauriharpf/rspack-node-dirname/blob/main/src/foo/bar/baz/render.js#L7)). 
3. Run the build on Windows via the `cmd` command prompt

**What happens**
RsPack does not escape the `\`-separator, output is e.g. `src\foo\bar\baz`. See https://github.com/lauriharpf/rspack-node-dirname/blob/main/rspack-dist/src_foo_bar_baz_render_js.js#L15

Image

webpack escapes the `\`-separator, output is e.g. `src\\foo\\bar\\baz`. See https://github.com/lauriharpf/rspack-node-dirname/blob/main/webpack-dist/src_foo_bar_baz_render_js.js#L10

Image

**Desired result**
RsPack and webpack build results match in terms of `__dirname`: both produce escaped output, e.g. ``src\\foo\\bar\\baz``

inottn:
## Summary
fix #8170 

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/node_stuff_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/node_stuff_plugin.rs
@@ -83,7 +83,9 @@ impl JavascriptParserPlugin for NodeStuffPlugin {
           .push(Box::new(ConstDependency::new(
             ident.span.real_lo(),
             ident.span.real_hi(),
-            format!("'{dirname}'").into(),
+            serde_json::to_string(&dirname)
+              .expect("should render dirname")
+              .into(),
             None,
           )));
         return Some(true);
--- a/packages/rspack-test-tools/tests/configCases/parsing/rspack-issue-8170/dir1/dir2/file.js
+++ b/packages/rspack-test-tools/tests/configCases/parsing/rspack-issue-8170/dir1/dir2/file.js
@@ -1,0 +1,1 @@
+exports.dirname = __dirname;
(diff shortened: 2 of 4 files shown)


LingyuCoder (approved):
thanks

PR #8176, opened 2024-10-21
fix: not generate mappings for runtime code

SyMind:
## Summary

Relative https://github.com/web-infra-dev/rspack/issues/7914#issuecomment-2417893609

Rspack currently does not generate mappings for runtime modules in the source map. As a result, when using the built-in SWC minifier, the percentage of generated code without mapping information is very high.

To address this issue, the current implementation constructs and returns `RawSource` when the `code_generation` function is called for runtime modules. To improve this, we should construct and return `OriginalSource` instead, when `module.source_map_kind.enable()` is `true`.

image

In the related issue, you can see that when using the Terser plugin, the mapping information percentage is 100%. This points to another existing problem in Rspack. The loss of structured data during Rust to JavaScript communication when passing `rspack-sources` data results in this issue. When encountering `RawSource`, `rspack-sources` currently skips the generation of mappings for that portion of the code.

## Checklist


SyMind:
!bench

SyMind:
!bench

Issue #7921, opened 2024-09-18
[Bug]: Minimizing CSS using lightningcss may shuffle the CSS property order


matthiask:
### System Info

  System:
    OS: Linux 6.10 Fedora Linux 40 (Workstation Edition)
    CPU: (16) x64 11th Gen Intel(R) Core(TM) i7-11700F @ 2.50GHz
    Memory: 54.73 GB / 62.62 GB
    Container: Yes
    Shell: 5.2.26 - /bin/bash
  Binaries:
    Node: 20.12.2 - /usr/bin/node
    Yarn: 4.4.1 - /usr/bin/yarn
    npm: 10.5.0 - /usr/bin/npm
    pnpm: 9.0.6 - /usr/bin/pnpm

### Details

This is more of a heads up since it's actually a bug in lightningcss tracked here:
https://github.com/parcel-bundler/lightningcss/issues/805

The report over there also contains a reproduction. I'm opening this issue with the goal of raising awareness about this issue, and also to ask if the option exists to bring the SWC CSS minimizer back.The SWC minimizer seems to be a bit more conservative which makes breakages such as this one less likely. Of course fixing lightningcss would probably be even better... :-)

Also, having an issue in the rspack repository may help others who are running into the same problem. 

The bug was introduced between 0.7.5 and 1.0.0-alpha.0.

### Reproduce Steps

A minimal lightningcss reproduction case is mentioned in the lightning CSS issue:
https://github.com/parcel-bundler/lightningcss/issues/805

GiveMe-A-Name:
Fix by https://github.com/web-infra-dev/rspack/pull/8182

matthiask:
Nice, thanks!

PR #8182, opened 2024-10-22
fix: temporarily use `lightingcss_rs` to fix CSS property order

GiveMe-A-Name:
## Summary

Lightincss have a bug about minimizing CSS may shuffle the CSS property order.

The css code: 
```css
.foo { 
   height: auto; 
   height: calc(100vh - var(--header-height, 0rem)); 
}
```
After lightningcss minify, css code would like this
```css
.foo { 
   height: calc(100vh - var(--header-height, 0rem)); 
   height: auto; 
}
```

care the lightingcss publish cycle. we publish a new crate named `lightingcss_rs`;

So we update lightingcss to lightingcss_rs for fix this issues https://github.com/web-infra-dev/rspack/issues/7921

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2085,10 +2085,22 @@ dependencies = [
 ]
 
 [[package]]
-name = "lightningcss"
-version = "1.0.0-alpha.58"
+name = "lightningcss-derive"
+version = "1.0.0-alpha.43"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec380ca49dc7f6a1cafbdd2de5e587958eac0f67ab26b1e56727fcc60a0c3d4d"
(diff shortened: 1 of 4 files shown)


chenjiahan:
Where is the source code of `lightingcss_rs`?

GiveMe-A-Name:
> Where is the source code of `lightingcss_rs`? Is there a branch or fork of lightningcss?

From this branch https://github.com/GiveMe-A-Name/lightningcss/tree/fix/size-prop-order.

This is diff from lightningcss main branch between and this branch https://github.com/parcel-bundler/lightningcss/pull/838

chenjiahan:
I see, we should explain that this is a temporary workaround and we will switch back to the official version when lightningcss fixes the issue and releases it.

chenjiahan (approved):

PR #8186, opened 2024-10-22
docs: use ApiMeta specific label instead of Webpack specific chapter

9aoy:
## Summary

Align with the English documentation.

## Checklist


chenjiahan (commented):
`require.ensure` should be moved too?

chenjiahan (commented):
Why not use `<ApiMeta specific={['Rspack', 'Webpack']} />` as other documentation

chenjiahan (approved):

PR #8195, opened 2024-10-22
refactor: tracing

h-a-n-a:
## Summary

Supported open-telemetry as tracing layer.

Enable open-telemetry tracing with env:

```bash
RSPACK_PROFILE="TRACE=filter=trace&layer=otel" rspack build
```

You may set `OTEL_BSP_MAX_QUEUE_SIZE` if open-telemetry reports the following error:

```
OpenTelemetry trace error occurred. cannot send message to batch processor as the channel is full
```

To visualize the result, jaeger is recommended as a backend and UI. Rspack by default uses the default gRPC endpoint which is the same as jaeger all-in-one. For more detail, please refer to https://www.jaegertracing.io/docs/1.62/getting-started/#all-in-one

image

## Checklist


packages/rspack/etc/core.api.md
```diff
@@ -1967,8 +1965,8 @@ export const experiments: Experiments_2;
 interface Experiments_2 {
     // (undocumented)
     globalTrace: {
-        register: typeof registerGlobalTrace;
-        cleanup: typeof cleanupGlobalTrace;
+        register: (filter: string, layer: "chrome" | "logger" | "otel", output: string) => Promise<void>;
```
h-a-n-a, line 1968: This is a breaking change to the register API but since it's experimental, so I guess it's fine.

jerrykingxyz (approved):

PR #8197, opened 2024-10-22
fix(lightning css): align type with lightning documentation

zackarychapple:
## Summary

In the lightning css documentation there is a option to enable drafts. This option was incorrectly named as `draft` instead of `drafts`. Updated the typing to the proper name as per the [documentation](https://lightningcss.dev/transpilation.html#custom-media-queries).

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-loader/lightningcss/index.ts
+++ b/packages/rspack/src/builtin-loader/lightningcss/index.ts
@@ -254,7 +254,7 @@ export type LoaderOptions = {
 	targets?: Targets | string[] | string;
 	include?: FeatureOptions;
 	exclude?: FeatureOptions;
-	draft?: Drafts;
+	drafts?: Drafts;
 	nonStandard?: NonStandard;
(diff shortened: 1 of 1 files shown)


hardfist:
even it's a typo, this change is still a breaking change, we can try to be compatible by supporting both `draft` and `drafts` and mark `draft` deprecated and remove it in v2

chenjiahan:
@GiveMe-A-Name can you take this?

GiveMe-A-Name:
CI failed, you should run `./x ae update` to update publish api.


> even it's a typo, this change is still a breaking change, we can try to be compatible by supporting both `draft` and `drafts` and mark `draft` deprecated and remove it in v2

And like [hardfist](https://github.com/hardfist) Said, we can support both `draft` and `drafts`

GiveMe-A-Name:
This PR have long time have not progress, I fix it in another PR https://github.com/web-infra-dev/rspack/pull/8456

PR #8209, opened 2024-10-24
refactor: run async task in thread pool

hardfist:
## Summary
try run async task in thread-pool  which would bring us some benifits
### more clean profile data
* before: tokio will trigger tons of new thread which is hard to see cpu usage and also hard to debug
image
image
* after: 
image

image

### better performance 
It brings about 10% performance improvement

image

currently we still have lots of async api so it's hard to refactor all async api to sync api now, but in the future we should 

## Checklist


hardfist:
!bench

hardfist:
!bench

hardfist:
!bench

hardfist:
!bench

hardfist:
!bench

PR #8216, opened 2024-10-25
fix(ci): trigger `pkg-preview` when the needs are skipped

shulaoda:
## Summary

Trigger `pkg-preview` when the dependent jobs are skipped as well.

I noticed that sometimes `pkg-preview` gets skipped because the dependent jobs `rust_check` and `rust_test` are skipped.

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -398,5 +398,5 @@ jobs:
       - lint
       - rust_check
       - rust_test
-    if: ${{ (success() || cancelled()) && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
+    if: ${{ !failure() && github.event_name == 'push' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -398,5 +398,5 @@ jobs:
       - lint
       - rust_check
       - rust_test
-    if: ${{ (success() || cancelled()) && github.event_name == 'push' && github.ref == 'refs/heads/main' }}
```
SoonIter, line 401: why not main branch? 🤨

shulaoda: Because the branches for `push` are already specified at the top of the workflow.

```yml
  push:
    branches:
      - main
    paths-ignore:
      - "**/*.md"
      - "website/**"
    tags-ignore:
      - "**"
```

Issue #8206, opened 2024-10-24
[Bug]: Set external extension, compilation error during dependent export


lzxb:
### System Info

  System:
    OS: macOS 15.0.1
    CPU: (8) arm64 Apple M1
    Memory: 106.00 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.13.1 - ~/.nvm/versions/node/v20.13.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.13.1/bin/yarn
    npm: 10.5.2 - ~/.nvm/versions/node/v20.13.1/bin/npm
    pnpm: 8.15.8 - ~/.nvm/versions/node/v20.13.1/bin/pnpm
    bun: 1.1.27 - ~/.bun/bin/bun
  Browsers:
    Safari: 18.0.1
  npmPackages:
    @rspack/cli: 1.0.14 => 1.0.14 
    @rspack/core: 1.0.14 => 1.0.14 

### Details

Named export is normal, but all exports are incorrect
```js
// Compilation failed
export * from 'test2';
// Compilation successful
export { test2 } from 'test2'

export function test1 () {}
```
![Image](https://github.com/user-attachments/assets/1c4d9350-7117-4c13-99b7-53db9beb38e4)

### Reproduce link

https://github.com/lzxb/rspack-externals-error

### Reproduce Steps

1. rspack.config.js
```js
{
    externals: ['test2']
}
```
2. packages/test1/index.js
```js
export * from 'test2';
```
3. Run command
```bash
pnpm build
```

chenjiahan:
Hi, can you try the `modern-module` type? It should work:

![Image](https://github.com/user-attachments/assets/9eb5cc43-bf5d-493d-ac0f-faa4d2c12890)

@fi3ework When will we add documentation for `modern-module`?

lzxb:
![Image](https://github.com/user-attachments/assets/946a330c-a218-4181-bc55-833a4d8b16d1)
The exported result is missing

fi3ework:
> Hi, can you try the modern-module type? It should work:

Currently, it does not work with `modern-module` either. `export *` is still under track https://github.com/web-infra-dev/rslib/issues/45. I used to try to implement it in `modern-module` only, but seems it should be fixed as long as `output.module = true` and `module` external type used together.

> When will we add documentation for modern-module?

[...]

PR #8217, opened 2024-10-25
fix: preserve star export from external module

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rspack/issues/8206.

Overall changes:

- A module that is an externalized type of `"module"` or `"module-import"` will make `export *` classified as a new `ExportModeType::ReexportFrom`. This type will not introduce runtime.
- The `export * from 'x'` from be added in `modern-module` by a new dependency.
- **possible breaking change**: Previously, with `"module"` external type, all star export will wrongly be assigned to the default export of the entry which is totally a wrong result. In this PR, `export *` will not work anymore.

<details>
<summary>output of previous implementation with <code>export * from 'react'; export * from 'vue';</code>, react and vue are externalized with type <code>"module".</code></summary>

```js
import * as __WEBPACK_EXTERNAL_MODULE_react__ from "react";
import * as __WEBPACK_EXTERNAL_MODULE_vue__ from "vue";
var __webpack_modules__ = ({
"945": (function (module) {

module.exports = __WEBPACK_EXTERNAL_MODULE_react__;


}),
"769": (function (module) {

module.exports = __WEBPACK_EXTERNAL_MODULE_vue__;


}),

});
/************************************************************************/
// The module cache
var __webpack_module_cache__ = {};

// The require function
function __webpack_require__(moduleId) {

// Check if module is in cache
var cachedModule = __webpack_module_cache__[moduleId];
if (cachedModule !== undefined) {
return cachedModule.exports;
}
// Create a new module (and put it into the cache)
var module = (__webpack_module_cache__[moduleId] = {
exports: {}
});
// Execute the module function
__webpack_modules__[moduleId](module, module.exports, __webpack_require__);

// Return the exports of the module
return module.exports;

}

/************************************************************************/
// webpack/runtime/define_property_getters
(() => {
__webpack_require__.d = function(exports, definition) {
	for(var key in definition) {
        if(__webpack_require__.o(definition, key) && !__webpack_require__.o(exports, key)) {
            Object.defineProperty(exports, key, { enumerable: true, get: definition[key] });
        }
    }
};
})();
// webpack/runtime/has_own_property
(() => {
__webpack_require__.o = function (obj, prop) {
	return Object.prototype.hasOwnProperty.call(obj, prop);
};

})();
// webpack/runtime/make_namespace_object
(() => {
// define __esModule on exports
__webpack_require__.r = function(exports) {
	if(typeof Symbol !== 'undefined' && Symbol.toStringTag) {
		Object.defineProperty(exports, Symbol.toStringTag, { value: 'Module' });
	}
	Object.defineProperty(exports, '__esModule', { value: true });
};

})();
/************************************************************************/
var __webpack_exports__ = {};
__webpack_require__.r(__webpack_exports__);
  case1: function() { return case1; }
});
/* ESM import */var react__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__(945);

/* ESM reexport (unknown) */ var __WEBPACK_REEXPORT_OBJECT__ = {};
/* ESM reexport (unknown) */ for( var __WEBPACK_IMPORT_KEY__ in react__WEBPACK_IMPORTED_MODULE_0__) if(["case1","default"].indexOf(__WEBPACK_IMPORT_KEY__) < 0) __WEBPACK_REEXPORT_OBJECT__[__WEBPACK_IMPORT_KEY__] =function(key) { return react__WEBPACK_IMPORTED_MODULE_0__[key]; }.bind(0, __WEBPACK_IMPORT_KEY__)
/* ESM reexport (unknown) */ __webpack_require__.d(__webpack_exports__, __WEBPACK_REEXPORT_OBJECT__);
/* ESM import */var vue__WEBPACK_IMPORTED_MODULE_1__ = __webpack_require__(769);

/* ESM reexport (unknown) */ var __WEBPACK_REEXPORT_OBJECT__ = {};
/* ESM reexport (unknown) */ for( var __WEBPACK_IMPORT_KEY__ in vue__WEBPACK_IMPORTED_MODULE_1__) if(["case1","default"].indexOf(__WEBPACK_IMPORT_KEY__) < 0) __WEBPACK_REEXPORT_OBJECT__[__WEBPACK_IMPORT_KEY__] =function(key) { return vue__WEBPACK_IMPORTED_MODULE_1__[key]; }.bind(0, __WEBPACK_IMPORT_KEY__)
/* ESM reexport (unknown) */ __webpack_require__.d(__webpack_exports__, __WEBPACK_REEXPORT_OBJECT__);

const case1 = 1

var __webpack_exports__case1 = __webpack_exports__.case1;
var __webpack_exports__default = __webpack_exports__["default"];
export { __webpack_exports__case1 as case1, __webpack_exports__default as default };

```

</details>

## Checklist


crates/rspack_plugin_javascript/src/dependency/esm/esm_export_imported_specifier_dependency.rs
```diff
@@ -267,6 +280,12 @@ impl ESMExportImportedSpecifierDependency {
       export_mode.items = Some(items);
       export_mode
     } else {
+      if self.reexport_star_from_external_module(module_graph) {
```
ahabhgk, line 283: So this means that `ModernModuleReexportStarExternalDependency` is only added but not removing the original `ESMExportImportedSpecifierDependency`? what if we just replacing these `ESMExportImportedSpecifierDependency` to `ModernModuleReexportStarExternalDependency`, then we don't need to introduce a new `ExportModeType`

fi3ework: I tried to determine whether to use `ModernModuleReexportStarExternalDependency` or `ESMExportImportedSpecifierDependency` here https://github.com/web-infra-dev/rspack/blob/a901d54c78b801bb6ae4854da67bf65a3d03c0f3/crates/rspack_plugin_javascript/src/parser_plugin/esm_export_dependency_parser_plugin.rs#L81 . However, this decision depends on whether the module is externalized, and we need access to the module graph, which isn't possible here.

Therefore, IMO we can't create the correct dependency during parsing. If we attempt to replace a dependency like `finish_modules`, I previously implemented it that way, but it's too late since some logic is already in `finish_modules` and has already been executed. 🤔

ahabhgk: Which logic already executed, what if we replace it in `finish_modules stage -100`

ahabhgk (approved):

fi3ework:
cc @jerrykingxyz  PLAT as you're the code owner of some files.

jerrykingxyz (approved):

PR #8225, opened 2024-10-26
fix(experiments css): should preserve keywords decl and compose ident for self reference

### LINKED ISSUE #8211: [Bug]: css animation is removed in css module
### System Info

  System:
    OS: Linux 6.11 Arch Linux
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13980HX
    Memory: 44.75 GB / 62.56 GB
    Container: Yes
    Shell: 3.7.1 - /usr/bin/fish
  Binaries:
    Node: 22.10.0 - /usr/bin/node
    pnpm: 9.12.1 - /usr/bin/pnpm
  npmPackages:
    @rspack/cli: ^1.0.14 => 1.0.14 
    @rspack/core: ^1.0.14 => 1.0.14 

### Details

css classes is preserved, but `@keyframes` is removed.

### Reproduce link

https://github.com/Austaras/rspack-css-module

### Reproduce Steps

see dist folder

Austaras:
## Summary

Closes #8211

## Checklist


crates/rspack_plugin_css/src/dependency/local_ident.rs
```diff
@@ -95,3 +96,105 @@ impl DependencyTemplate for CssLocalIdentDependency {
...
+  }
+
+  fn get_referenced_exports(
+    &self,
+    _module_graph: &rspack_core::ModuleGraph,
+    _runtime: Option<&RuntimeSpec>,
+  ) -> Vec<rspack_core::ExtendedReferencedExport> {
+    println!("111111");
+
+    vec![ExtendedReferencedExport::Array(vec![self
```
Austaras, line 164: This doesn't seem to be called at all.

crates/rspack_plugin_css/src/parser_and_generator/mod.rs
```diff
@@ -404,11 +442,33 @@ impl ParserAndGenerator for CssParserAndGenerator {
...
+            }
+
+            idx += 1;
+          }
+
+          let unused =
+            get_unused_local_ident(exports, identifier, generate_context.runtime, &mg, deps);
+
+          context.data.insert(unused);
+
```
Austaras, line 471: As a temporary workaround, we could removed deps from unused here mannually, so `composes` would work. However it is still impossible to make keyframe work because we don't know the classes they're in. What would be the proper fix? @ahabhgk

crates/rspack_plugin_css/src/dependency/local_ident.rs
ahabhgk: This should be called, this is a bit tricky that this dependency should bind to `SelfModuleFactory`, this may require some familiarity with how rspack's tree shaking is implemented, so I will take over this PR and fix it, thanks for the contribution

ahabhgk (approved):

Issue #3169, opened 2023-05-16
Memory leak when using `@rspack/core` API


Nikaple:
### System Info

```
  System:
    OS: Linux
    CPU: (16) x64 Intel(R) Xeon(R) Platinum 8255C CPU @ 2.50GHz
    Memory: 19.11 GB / 31.10 GB
    Container: Yes
    Shell: 5.0.2 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  npmPackages:
    @rspack/cli: latest => 0.1.10 
    @rspack/core: ^0.1.10 => 0.1.10 
```

### Details

<img width="1170" alt="screenshot" src="https://github.com/web-infra-dev/rspack/assets/24775744/96d4268b-92c6-4fcf-962e-13252ccab5bb">

#### With webpack

<img width="1307" alt="screenshot-webpack" src="https://github.com/web-infra-dev/rspack/assets/24775744/16ddeb05-f528-4ba7-9044-9f3b2bb130e8">

### Reproduce Steps

1. Create a project `npm create rspack@latest`
2. Create a script to compile consecutively with `@rspack/core`
3. Run the script to get memory leak

Script content:
```js
const { rspack } = require('@rspack/core')
const config = require('./rspack.config')

function compile() {
    const compiler = rspack(config);
    return new Promise((resolve, reject) => {
        compiler.run((err, stat) => {
            if (err) {
                return reject(err)
            }
            return resolve(stat)
        })
    })
}


async function main() {
    await compile();

    setTimeout(() => {
        const memory = process.memoryUsage();
        console.log(`Total memory used: ${Math.round(memory.rss / 1024 / 1024)} MB`)
        main()
    }, 1000);
}

main().catch(console.error)
```

Boshen:
Confirmed. How did you discover this? We don't normally import it as a package 🤔

Nikaple:
I am currently in the process of migrating a complex build tool that is based on webpack. The tool has the ability to compile multiple projects simultaneously, utilizing webpack's core API. However, when attempting to migrate it to rspack, I encountered a persistent issue of the process being killed due to an OOM error.

pratyushseq:
bump!

Nikaple:
bump

[...]

PR #8268, opened 2024-10-30
fix: fix `Compiler` not getting freed after use

h-a-n-a:
## Summary

closes https://github.com/web-infra-dev/rspack/issues/3169

Fixed an issue when constantly creating compilers and the memory is not getting freed after use.

This issue was fixed with `WeakRef` and function expression which does not capture `this`, which is `Compiler` in this case. But still, I'm not able to create a MRE with both minimal Rust and JS code. So this issue might be appropriately tackled once a MRE is created.

Tests will be added in up-coming PRs. We might need to take https://github.com/napi-rs/napi-rs/tree/main/memory-testing as a reference.

## Checklist

### DIFF
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -247,6 +247,10 @@ class Compiler {
 
 		new JsLoaderRspackPlugin(this).apply(this);
 		new ExecuteModulePlugin().apply(this);
+
+		this.hooks.shutdown.tap("Compiler", () => {
+			this.#instance = undefined;
+		});
 	}
 
 	get recordsInputPath() {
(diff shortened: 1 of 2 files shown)


h-a-n-a:
[Release Canary](https://github.com/web-infra-dev/rspack/actions/runs/11588930277/job/32264457498#step:9:932)
Version: 1.0.15-canary-cc53546f-20241030081506

packages/rspack/src/Compiler.ts
```diff
@@ -247,6 +247,10 @@ class Compiler {
 
 		new JsLoaderRspackPlugin(this).apply(this);
 		new ExecuteModulePlugin().apply(this);
+
+		this.hooks.shutdown.tap("Compiler", () => {
```
h-a-n-a, line 251:
```suggestion
		this.hooks.shutdown.tap("rspack:cleanup", () => {
```

dinofx:
This approach seems to have quite a negative performance impact:

image

Above only captured the place where `const getCompiler = () => ref.deref()!` was called.  The places where `const that = new WeakRef(this)` pattern was used doesn't get measured, so there's likely additional impact not shown.

dinofx:
@h-a-n-a 
I'm guessing that `#updateNonSkippableRegisters()` is accessing most of these weak refs.  If that's the case, it looks like close to half a second is being spent dereferencing these.  That's a pretty significant percentage of our total build time.

image

Issue #8277, opened 2024-10-31
[Bug]: TypeScript errors TS2416 and TS2694 in `@rspack/core`


andersk:
### System Info

```
  System:
    OS: Linux 6.11 NixOS 24.11 (Vicuna) 24.11 (Vicuna)
    CPU: (12) x64 Intel(R) Core(TM) i7-10710U CPU @ 1.10GHz
    Memory: 6.97 GB / 15.26 GB
    Container: Yes
    Shell: 5.2.32 - /run/current-system/sw/bin/bash
  Binaries:
    Node: 20.17.0 - ~/.nix-profile/bin/node
    npm: 10.8.2 - ~/.nix-profile/bin/npm
    pnpm: 9.12.0 - ~/.nix-profile/bin/pnpm
    bun: 1.1.31 - ~/.nix-profile/bin/bun
  Browsers:
    Chromium: 130.0.6723.58
  npmPackages:
    @rspack/core: ^1.0.14 => 1.0.14 
```

### Details

I get many TypeScript errors just from doing `import "@rspack/core"` in a TypeScript file.

```
node_modules/@rspack/core/dist/Compilation.d.ts:399:5 - error TS2416: Property 'entries' in type 'Entries' is not assignable to the same property in base type 'Map<string, JsEntryData>'.
node_modules/@rspack/core/dist/Compilation.d.ts:400:5 - error TS2416: Property 'values' in type 'Entries' is not assignable to the same property in base type 'Map<string, JsEntryData>'.
node_modules/@rspack/core/dist/Compilation.d.ts:401:5 - error TS2416: Property '[Symbol.iterator]' in type 'Entries' is not assignable to the same property in base type 'Map<string, JsEntryData>'.
node_modules/@rspack/core/dist/Compilation.d.ts:407:5 - error TS2416: Property 'keys' in type 'Entries' is not assignable to the same property in base type 'Map<string, JsEntryData>'.
node_modules/@rspack/core/dist/config/zod.d.ts:2751:48 - error TS2694: Namespace 'z' has no exported member 'objectUtil'.
```

(The first four are new with TypeScript 5.6, the last is not.)

### Reproduce Steps

From an empty directory:

```console
$ npm i @rspack/core @types/graceful-fs @types/node tapable typescript webpack webpack-dev-server
$ echo 'import "@rspack/core"' > test.ts
$ npx tsc --strict --moduleResolution node --esModuleInterop --target ESNext test.ts
[...]

PR #8279, opened 2024-10-31
fix(type): Make iterators compatible with TS 5.6 `--target ESNext`

andersk:
## Summary

Use `ReturnType<>` to avoid explicitly naming the types of builtin iterators that have changed in 5.6, and convert manually implemented iterators to generators so they are compatible with builtin iterators under `--target ESNext`.

- Fixes part of #8277.

## Checklist

### DIFF
--- a/packages/rspack/src/util/ArrayQueue.ts
+++ b/packages/rspack/src/util/ArrayQueue.ts
@@ -77,35 +77,11 @@ class ArrayQueue<T> {
 		}
 	}
 
-	[Symbol.iterator]() {
-		let i = -1;
-		let reversed = false;
-		return {
-			next: () => {
-				if (!reversed) {
-					i++;
-					if (i < this._list.length) {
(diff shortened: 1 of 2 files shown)


packages/rspack/src/util/ArrayQueue.ts
```diff
@@ -77,35 +77,11 @@ class ArrayQueue<T> {
...
-				}
-				return {
-					done: false,
-					value: this._listReversed[i]
-				};
-			}
-		};
+	*[Symbol.iterator]() {
+		yield* this._list;
+		for (let i = this._listReversed.length; i >= 0; i--) {
```
andersk, line 82: Oops,

```suggestion
		for (let i = this._listReversed.length - 1; i >= 0; i--) {
```

SoonIter (approved):
LGTM, thanks❤️️

PR #8282, opened 2024-10-31
perf(core): replace `neo-async` with a minimal helper

chenjiahan:
## Summary

[neo-async](https://www.npmjs.com/package/neo-async) contains a lot of async utilities (4k LOC) and it has not been updated for more than four years.

Rspack only uses the `each()` util of `neo-async`, so we can replace `neo-async` with a minimal `each` helper to reduce bundle size and improve startup performance.

## Comparison

Compare the `@rspack/core/dist/index.js` bundle:

| -  | LOC | Size | Load time |
| --- | --- | --- | --- |
| **before** | 23383 | 772KB | 65.7 ms ±   1.7 ms    [User: 51.8 ms, System: 14.7 ms] |
| **after** | 18843 | 623KB | 63.3 ms ±   1.0 ms    [User: 50.4 ms, System: 15.1 ms] |

## Checklist


chenjiahan:
!eco-ci

LingyuCoder (approved):

Issue #8248, opened 2024-10-29
[Bug]: unclear error and too general typing on externals field


ItamarGronich:
### System Info

```
  System:
    OS: macOS 14.6.1
    CPU: (10) arm64 Apple M2 Pro
    Memory: 174.80 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
    npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
    pnpm: 9.7.1 - ~/Library/pnpm/pnpm
  Browsers:
    Chrome: 130.0.6723.70
    Safari: 17.6
  npmPackages:
    @rspack/cli: ^1.0.14 => 1.0.14 
    @rspack/core: ^1.0.14 => 1.0.14 

```

### Details
When compiling to a library with type `'umd'` and using externals, Rspack will error out if you forgot to add the `'root' | 'commonjs' | 'commonjs2'` properties to one of your externals.

for example:
Good ✅
```js
export default {
  output: { library: { type: "umd" } },
  externals: {
    react: {
      commonjs: "react",
      commonjs2: "react",
      root: "React"
    }
  }
}
```

Bad ❌
```js
export default {
  output: { library: { type: "umd" } },
  externals: {
    react: {
      commonjs: "react",
      commonjs2: "react" // Notice no 'root' property. But removing any of these three will result in an error.
    }
  }
}
```

This makes sense as critical information is missing for Rspack to be able to construct a correct UMD import of the react library.

The bad example will throw the following error:
```
[rspack-cli] [Error:   × Missing external configuration for type: root
] {
  code: 'GenericFailure'
}
```

### How Webpack treats this
When you do this in webpack - webpack compiles successfully without any error. but the resulting code has errors in it.
So i think Rspack actually does the right thing here but fails to communicate this properly.
for example if you had this code:
```js
import React from 'react'
console.log(React)
```

webpack would output something like this:
```js
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory(require("react"));
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else {
                // 🚧 notice the root[undefined] here which should have been root['React'] 
                // but because webpack doesn't have a 'root' property it just puts in undefined. 
                //                                                  This will throw in production👇🏻
[...]

PR #8286, opened 2024-10-31
fix: types and docs of umd external object

LingyuCoder:
## Summary

relate to https://github.com/web-infra-dev/rspack/issues/8248

Cause the external object value can only be used when libraryTarget or externalTarget is `umd`, and if it will throw error when `commonjs` or `commonjs2` or `root` is not set. So we can use zod to validate the external object value. 

The `amd` field is optional because it can generate `define([])` when not set, but there is a bug here in rspack that generated `define([null])`. This pr also fix it.

Not close the issue because when libraryTarget is not `umd`, the external object value is also valid but will generate a wired code. For example, `libraryTarget: "var"` and `react: { root: "React", commonjs: "react"  }` will generate `var external_react_root_react_commonjs_react = undefined` which can not actually work. 

So we may need a better validation here to check `libraryTarget` and `externals`.

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/configCases/externals/item-value-object/rspack.config.js
+++ b/packages/rspack-test-tools/tests/configCases/externals/item-value-object/rspack.config.js
@@ -3,14 +3,20 @@ const { CopyRspackPlugin } = require("@rspack/core");
 /** @type {import("@rspack/core").Configuration} */
 module.exports = {
 	entry: "./index.js",
+	output: {
+		library: {
+			type: "umd",
+		}
+	},
 	externals: {
 		lodash: {
 			root: "_",
 			commonjs: "./lodash.js",
+			commonjs2: "./lodash.js",
 			amd: "./lodash.js"
 		}
 	},
-	externalsType: "commonjs",
+	externalsType: "umd",
 	plugins: [
 		new CopyRspackPlugin({
 			patterns: ["./lodash.js"]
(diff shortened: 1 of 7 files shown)


packages/rspack-test-tools/tests/configCases/externals/item-value-object/rspack.config.js
```diff
@@ -3,14 +3,20 @@ const { CopyRspackPlugin } = require("@rspack/core");
...
 	externals: {
 		lodash: {
 			root: "_",
 			commonjs: "./lodash.js",
+			commonjs2: "./lodash.js",
 			amd: "./lodash.js"
 		}
 	},
-	externalsType: "commonjs",
+	externalsType: "umd",
```
LingyuCoder, line 19: External object value is only for umd, should not use it with other types, can just use `lodash: "./lodash.js"` as replacement

ahabhgk (approved):

packages/rspack/src/config/types.ts
```diff
@@ -1312,7 +1312,15 @@ export type ExternalItemValue =
...
 	| string[]
-	| Record<string, string | string[]>;
+	| {
+			/**
+			 * only available when libraryTarget and externalsType is 'umd'
+			 */
+			root: string | string[];
+			commonjs: string | string[];
+			commonjs2: string | string[];
+			amd?: string | string[];
```
ItamarGronich, line 1322: How come amd is optional?
Isn't it required to construct a real UMD import?

LingyuCoder: After considering it, I think you are right. The `amd` field should be handled just like `commonjs` and `commonjs2`. I will fix it.

Issue #8238, opened 2024-10-28
[Bug]: error is thrown when a value of undefined is encountered where loader is expected


kubijo:
### System Info

```
System:
    OS: Linux 6.11 Ubuntu 24.10 24.10 (Oracular Oriole)
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 41.08 GB / 60.88 GB
    Container: Yes
    Shell: 5.2.37 - /bin/bash
Binaries:
    Node: 22.8.0 - ~/.volta/tools/image/node/22.8.0/bin/node
    Yarn: 4.5.1 - ~/.volta/tools/image/yarn/4.5.0/bin/yarn
    npm: 10.8.2 - ~/.volta/tools/image/node/22.8.0/bin/npm
```

### Details

The following error is thrown when a value of `undefined` is encountered where loader is expected.
No validation error was ever thrown even though the docs mention that a strict validation is enabled by default.

Please note that I know about incompatibility with Yarn's PNP algorithm, that is not the point of this report.
The point is that the config consuming code should either filter out nil values or the validation code should shout at the user that they are supplying something fishy.

I had to go deep into the consuming code to add log-points and find out where I made a mistake (in my case, it was a forgotten reference to a loader object that I removed).

```bash
$ yarn run rspack build --config config.mjs

TypeError: Cannot read properties of undefined (reading 'type')
    at .yarn/__virtual__/@rspack-core-virtual-f10b7c4cd3/0/cache/@rspack-core-npm-1.0.14-374d04781a-3743c2dd9e.zip/node_modules/@rspack/core/dist/index.js:11275:15
    at Array.map (<anonymous>)
    at getRawModuleRule (.yarn/__virtual__/@rspack-core-virtual-f10b7c4cd3/0/cache/@rspack-core-npm-1.0.14-374d04781a-3743c2dd9e.zip/node_modules/@rspack/core/dist/index.js:11270:36)
    at .yarn/__virtual__/@rspack-core-virtual-f10b7c4cd3/0/cache/@rspack-core-npm-1.0.14-374d04781a-3743c2dd9e.zip/node_modules/@rspack/core/dist/index.js:11175:22
    at Array.map (<anonymous>)
[...]

PR #8297, opened 2024-10-31
fix: falsy rule should be filtered

LingyuCoder:
## Summary

Close https://github.com/web-infra-dev/rspack/issues/8238

## Checklist


packages/rspack/src/config/types.ts
```diff
@@ -897,10 +897,10 @@ export type RuleSetRule = {
 	enforce?: "pre" | "post";
 
 	/** A kind of Nested Rule, an array of Rules from which only the first matching Rule is used when the parent Rule matches. */
-	oneOf?: RuleSetRule[];
+	oneOf?: (RuleSetRule | Falsy)[];
```
chenjiahan, line 900: I think `false` can be allowed here, @hardfist ,what do you think?

chenjiahan (approved):

PR #8315, opened 2024-11-02
feat: throw error for outdated compilation

SyMind:
## Summary

The Rspack compilation instance will be dropped on the Rust side after the compilation stage.

Before this PR, since the memory address of the compilation instance on the Rust side didn't change across multiple compilation cycles, when properties and methods on an outdated compilation instance on the JavaScript side were called, the data from the current Rust instance was retrieved, which was not as expected.

Now, Rspack will throw an error when an outdated JavaScript side compilation instance is accessed.

The compilation instance of Rspack will be dropped on the Rust side after compilation.

## Checklist

### DIFF
--- a/crates/rspack_binding_values/src/module.rs
+++ b/crates/rspack_binding_values/src/module.rs
@@ -386,9 +386,6 @@ impl ToNapiValue for JsModuleWrapper {
             let mut instance = unassociated_ref.from_napi_value()?;
             instance.as_mut().attach(compilation_ptr.as_ptr());
 
-            if !std::ptr::addr_eq(instance.module.as_ptr(), val.module.as_ptr()) {
-              instance.module = RefCell::new(val.module);
-            }
             let napi_value = ToNapiValue::to_napi_value(env, &unassociated_ref);
             refs.insert(module.identifier(), unassociated_ref);
             napi_value
@@ -396,11 +393,6 @@ impl ToNapiValue for JsModuleWrapper {
             match refs.entry(module.identifier()) {
               std::collections::hash_map::Entry::Occupied(entry) => {
                 let r = entry.get();
-
-                let mut instance: ClassInstance<JsModule> = r.from_napi_value()?;
-                if !std::ptr::addr_eq(instance.module.as_ptr(), val.module.as_ptr()) {
-                  instance.module = RefCell::new(val.module);
-                }
                 ToNapiValue::to_napi_value(env, r)
               }
               std::collections::hash_map::Entry::Vacant(entry) => {
@@ -411,11 +403,6 @@ impl ToNapiValue for JsModuleWrapper {
                 }
                 .into_instance(Env::from_raw(env))?;
                 let r = entry.insert(OneShotRef::new(env, instance)?);
-
-                let mut instance: ClassInstance<JsModule> = r.from_napi_value()?;
-                if !std::ptr::addr_eq(instance.module.as_ptr(), val.module.as_ptr()) {
-                  instance.module = RefCell::new(val.module);
-                }
                 ToNapiValue::to_napi_value(env, r)
               }
             }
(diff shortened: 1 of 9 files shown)


SyMind:
!eco-ci

ahabhgk (approved):

crates/rspack_binding_values/src/module.rs
```diff
@@ -82,7 +82,7 @@ impl JsDependenciesBlock {
 #[napi]
 pub struct JsModule {
   identifier: ModuleIdentifier,
-  module: RefCell<NonNull<dyn Module>>,
+  module: NonNull<dyn Module>,
   compilation: Option<NonNull<Compilation>>,
```
h-a-n-a, line 86: When will this `compilation` be `None`?

SyMind: Now, in hooks like `build_module`, `still_valid_module`, `succeed_module`, I can't obtain the compilation reference.

After thinking about it, I believe this is an issue with Rust hooks. In JavaScript hooks, we can get the `compilation` through `compilation` hook closure, but on the Rust side, this can't be achieved.

Therefore, later I need to modify the hooks so that the parameters of `build_module`, `still_valid_module`, and `succeed_module` include the `compilation`.

crates/rspack_binding_values/src/module.rs
```diff
@@ -106,7 +106,7 @@ impl JsModule {
       let compilation = unsafe { compilation.as_ref() };
```
h-a-n-a, line 106: For now, as what you've been stated in the previous comment, I don't see how we're going to do the version checks between compilations for `JsModule`. Checking the `Option::Some` does not seem enough to me, right?

crates/rspack_binding_values/src/module.rs
```diff
@@ -106,7 +106,7 @@ impl JsModule {
...
       if let Some(module) = compilation.module_by_identifier(&self.identifier) {
         let module = module.as_ref();
-        *self.module.borrow_mut() = {
+        self.module = {
```
h-a-n-a, line 109: Can we reuse the module from `self.module`? What's the difference between fetching it from `Compilation` and the implementation in `JsModule::as_mut`?

crates/rspack_binding_values/src/module.rs
h-a-n-a: Thanks for the clarification.

crates/rspack_binding_values/src/module.rs
SyMind: There are two ways for this:
1. In the `as_ref` function, I will check the identifier of the `Module`.
2. In the `JsModuleWrapper::new(module, compilation)` function, if the compilation is `None`, the `Module` pointer for the `JsModule` instance will be updated.

crates/rspack_binding_values/src/module.rs
SyMind: This is a fallback strategy. The reason is that we don't actually have any internal guarantee that the address of the same module won't change during its entire life cycle. I think that in current practice, the code hardly ever reaches this point.

When we fail to assert the identifier of the module instance, we will attempt to obtain it from the compilation once.

h-a-n-a: > ...we don't actually have any internal guarantee that the address of the same module won't change during its entire life cycle....

I suppose this belongs to both `as_ref` and `as_mut`?

SyMind: We can only modify `Module` in hooks like `build_module`, `still_valid_module`, `succeed_module`. But we cannot get compilation ref in there.

crates/rspack_binding_values/src/module.rs
h-a-n-a: IIRC, `JsModule` is derived from `JsModuleWrapper` and bound to `Compilation` , right? 
The part I don't quite understand is if `JsModuleWrapper` does not contain a pointer to the underlying `Compilation` how we're going to expose `JsModule` to the JS side? Is this a case we should care about?

h-a-n-a: I see. And at the moment, we still cannot check the validity of a `JsModule`. Am I correct?

SyMind: Not really true. We can check the module identifier for the validity of a JsModule. But if the Module has moved in memory we cannot get the latest pointer for it.

h-a-n-a:
Looks good. Would you please add a test case to avoid future regression?

h-a-n-a (approved):

PR #8319, opened 2024-11-02
fix: correct usage of `pnpm --filter` in package scripts

shulaoda:
## Summary

Using single quotes in `pnpm --filter '@rspack/*'` results in `"No projects matched the filters"`. Instead, we should use double quotes for proper functionality. (This issue may only occur on `Windows`, I don't have a `Mac` to verify)

## Checklist


chenjiahan (approved):
The single quotes work in macOS, thanks for making the scripts work in Windows. 😄

PR #8331, opened 2024-11-04
perf(cli): enable Node.js compile cache

chenjiahan:
## Summary

Enable Node.js 22 new compile cache in Rspack CLI, this makes Node.js startup faster.

- before:

Screenshot 2024-11-04 at 15 22 05

- after:

Screenshot 2024-11-04 at 15 21 58

## Related Links

- https://nodejs.org/id/blog/release/v22.8.0
- https://github.com/nodejs/node/pull/54501
- https://github.com/web-infra-dev/rsbuild/pull/3627

## Checklist

### DIFF
--- a/packages/rspack-cli/bin/rspack.js
+++ b/packages/rspack-cli/bin/rspack.js
@@ -4,7 +4,7 @@ const nodeModule = require("node:module");
 // enable on-disk code caching of all modules loaded by Node.js
 // requires Nodejs >= 22.8.0
 const { enableCompileCache } = nodeModule;
-if (enableCompileCache && !process.env.NODE_DISABLE_COMPILE_CACHE) {
+if (enableCompileCache) {
 	try {
 		enableCompileCache();
 	} catch {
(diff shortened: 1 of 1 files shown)


h-a-n-a:
Are we going to apply `v8-compile-cache` as well to the lower Node.js versions? I suppose those are pretty the same thing, right?

chenjiahan:
As far as I know, they are basically the same. 

From the Node.js blog:

> This is a built-in alternative to the [v8-compile-cache](https://www.npmjs.com/package/v8-compile-cache)/[v8-compile-cache-lib ](https://www.npmjs.com/package/v8-compile-cache-lib)packages, but have [better performance](https://github.com/nodejs/node/issues/47472#issuecomment-1970331362) and supports ESM.

However, I do not want to introduce `v8-compile-cache` as it is a third-party dependency and has some limitations. Node 22 has reached LTS so more users will be able to use the native `enableCompileCache` soon.

h-a-n-a (dismissed):

PR #8363, opened 2024-11-06
fix(cli): Run config function before looking for `--config-name`

andersk:
## Summary

When the config is a function, we need to match `--config-name` on its return value, not the function itself.

## Checklist

  - `tests/webpack-cli-test/build/config-name/config-name.test.js` seems to already have tests for this (`"should work with config as a function"`, `"should work with multiple values for --config-name when the config is a function"`), but it seems we never run those tests.


chenjiahan (approved):
LGTM 👍

PR #8365, opened 2024-11-07
fix: define lib export

HigherOrderLogic:
## Summary

Add `/lib/*` export filed in `package.json`.

This exists in webpack and [there are hundred of projects import from there](https://github.com/search?q=%2F%28import%7Crequire%29.%2Bwebpack%5C%2Flib%5C%2F.%2B%2F+language%3ATypeScript&ref=opensearch&type=code).

## Checklist

### DIFF
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -17,8 +17,8 @@
     },
     "./hot/*": "./hot/*.js",
     "./hot/*.js": "./hot/*.js",
-    "./lib/*": "./lib/*.js",
-    "./lib/*.js": "./lib/*.js",
+    "./lib/*": ".dist/lib/*.js",
+    "./lib/*.js": ".dist/lib/*.js",
     "./package.json": "./package.json",
     "./module": "./module.d.ts"
   },
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -17,6 +17,8 @@
     },
     "./hot/*": "./hot/*.js",
     "./hot/*.js": "./hot/*.js",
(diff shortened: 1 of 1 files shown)


HigherOrderLogic:
Yes, I know it is recommended to use the provided webpack instance, but if then rspack is not theoretically 100% compatible with the original webpack.

I was trying out the Nuxt rspack builder, and the `fork-ts-checker-webpack-plugin` package, which import from webpack directly, throw a module not found error. I tried to fix it by adding `"webpack": "npm:@rspack/core@latest"`, and the error this time is  subpath `./lib/RequestShortener` not defined in the exports field in package.json.

chenjiahan:
Rspack does not provide `dist/lib/*` files and we have no plans to be compatible with it.

The `lib` folder contains a lot of private webpack modules that should not be used. Plugin authors should use `webpack.*` instead.

chenjiahan:
For `fork-ts-checker-webpack-plugin`:

image

https://rspack.dev/guide/migration/webpack#installing-rspack

HigherOrderLogic:
As for what you say, it is true that importing and using private modules isn't good, but there are lots of plugins working that way so fixing each one is impossible.

As for installing webpack as a workaround, it does solve the issue but it's quite inconvenient for users imo.

chenjiahan:
Although Rspack is committed to maintaining compatibility with the webpack ecosystem, it is obviously not possible to support all webpack plugins. For plugins that use internal webpack paths, we will work with the community to refactor the code for these plugins.

This PR will be closed, thank you.

PR #8369, opened 2024-11-07
fix: required amd in umd external object

LingyuCoder:
## Summary

https://github.com/web-infra-dev/rspack/pull/8286/files#r1831012797

Cause lots of projects use object item of `externals` when not `umd` and it can actually work well. For example:

```js
module.exports = {
  output: {
     library: {
       type: "commonjs"
     }
  },
  externalsType: "commonjs",
  externals: {
    // actually same as: `react: "REACT"` or `react: "commonjs REACT"`
    react: {
      commonjs: "REACT"
    }
  }
};
```

 so if we just add a strict object validation to object item of `externals`, it will be a breaking change to these projects.

The object item of `externals` should be validate according to the values of `output.library.type/output.libraryTarget` and `externalsType`. Only if `output.library.type/output.libraryTarget` is set to `umd` and `externalsType` is not set or also set to `umd`, the object item in `externals` should be handle as an umd dependency. 

We can add a new ZodType named `ZodCrossFieldsOptions` to validate the object item in `externals`. In its `_parse` method we can get the zod context, and then get its root context through its `parent` field. Then we can do some checking cross configuration options and make sure that the object item in `externals` should only be validated when it is really handle as an umd dependency.

But there is a trick in `ZodUnion`. The `parent` field is set to `null` [here](https://github.com/colinhacks/zod/blob/f487d74ecd3ae703ef8932462d14d643e31658b3/src/types.ts#L2979). Not sure if it is a bug or not. So we need to override the `ZodUnion._parse` to pass the `parent` context to children like other `ZodType` does.

## Checklist

### DIFF
--- a/packages/rspack/src/config/utils.ts
+++ b/packages/rspack/src/config/utils.ts
@@ -22,6 +22,16 @@ import {
 } from "zod";
 import type { RspackOptions } from "./types";
 
+/**
+ * The following code is modified based on
+ * https://github.com/colinhacks/zod/blob/f487d74ecd3ae703ef8932462d14d643e31658b3/src/types.ts
+ *
+ * MIT Licensed
+ * Author Colin McDonnell @colinhacks
+ * MIT License
+ * https://github.com/colinhacks/zod/blob/main/LICENSE
+ */
+
 function processCreateParams(params: RawCreateParams): ProcessedCreateParams {
 	if (!params) return {};
 	const { errorMap, invalid_type_error, required_error, description } = params;
@@ -48,7 +58,7 @@ function processCreateParams(params: RawCreateParams): ProcessedCreateParams {
 
 /**
  * Modified `z.union` for overriding its `_parse` to support `parent` field of context.
- * https://github.com/colinhacks/zod/blob/f487d74ecd3ae703ef8932462d14d643e31658b3/src/types.ts#L2935-L3034
+ *
  * We need to use `parent` field to get the root config object.
  */
 class RspackZodUnion<T extends ZodUnionOptions> extends z.ZodUnion<T> {
--- a/packages/rspack/src/config/utils.ts
+++ b/packages/rspack/src/config/utils.ts
@@ -1,0 +1,208 @@
+import {
+	type DIRTY,
+	INVALID,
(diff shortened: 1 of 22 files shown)


packages/rspack/src/config/zod.ts
```diff
@@ -804,7 +804,7 @@ const externalItemValue = z
 			root: z.string().or(z.string().array()),
 			commonjs: z.string().or(z.string().array()),
 			commonjs2: z.string().or(z.string().array()),
-			amd: z.string().or(z.string().array()).optional()
+			amd: z.string().or(z.string().array())
```
chenjiahan, line 807: I prefer to validate `amd` only when the externalType is `umd`, because other external types do not need an `amd` property.

website/docs/en/config/externals.mdx
```diff
@@ -125,10 +125,10 @@ module.exports = {
   // or
   externals: {
     lodash: {
-      root: '_', // required, indicates global variable
-      commonjs: 'lodash', // required
-      commonjs2: 'lodash', // required
-      amd: 'lodash', // optional
+      root: '_', // indicates global variable
```
chenjiahan, line 128: Should we note that these properties are all required when libraryTarget is umd?

packages/rspack/src/config/utils.ts
```diff
@@ -0,0 +1,208 @@
...
+		}
+		if (iss.code !== "invalid_type") return { message: ctx.defaultError };
+		return { message: message ?? invalid_type_error ?? ctx.defaultError };
+	};
+	return { errorMap: customMap, description };
+}
+
+/**
+ * Modified `z.union` for overriding its `_parse` to support `parent` field of context.
+ * https://github.com/colinhacks/zod/blob/f487d74ecd3ae703ef8932462d14d643e31658b3/src/types.ts#L2935-L3034
```
chenjiahan, line 51: we should add the copyright and license for copied code.

chenjiahan: And try to send a PR to zod 😄

chenjiahan (approved):

PR #8386, opened 2024-11-09
fix: variable name conflict with concatenate and runtime code

inottn:
## Summary

See https://github.com/webpack/webpack/issues/18932 for background.

## Checklist

### DIFF
--- a/tests/webpack-test/configCases/library/issue-18932/index.js
+++ b/tests/webpack-test/configCases/library/issue-18932/index.js
@@ -1,0 +1,7 @@
+it("should not have variable name conflict", function() {
+	expect(true).toBe(true);
+});
+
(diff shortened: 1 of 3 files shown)


tests/webpack-test/configCases/library/issue-18932/index.js
```diff
@@ -0,0 +1,7 @@
+it("should don't have variable name conflict", function() {
```
Narretz, line 1: This should be `should not have ...`

inottn: Nice catch, we can sync this modification back to webpack.

chenjiahan (approved):
LGTM 👍

PR #8394, opened 2024-11-11
feat: unify js dependency

SyMind:
## Summary

feat: unify js dependency

## Checklist


SyMind:
!bench

SyMind:
!eco-ci

crates/rspack_binding_values/src/dependency.rs
```diff
@@ -1,108 +1,193 @@
...
+      self.compilation = Some(compilation);
+    }
   }
 
-  #[napi(getter)]
-  pub fn category(&self) -> &str {
-    self.0.category().as_str()
-  }
+  fn as_ref(&mut self) -> napi::Result<&dyn Dependency> {
+    let dependency = unsafe { self.dependency.as_ref() };
```
JSerFeng, line 26: safety comments ?

SyMind: The JS API is now in a fixed pattern and will use ids to verify pointers, if invalid will return Err. So I think safe comments are unnecessary.

JSerFeng:
This pr depends on accessing a field within a raw pointer, for example

```
// ptr is raw pointer and might point to whatever possible
let foo = ptr.as_ref();

foo.id == last_id
```

But I wonder if access id here will panic ? And this is a solution that relies on UB ? Can we accept this @h-a-n-a

h-a-n-a:
> But I wonder if access id here will panic ? And this is a solution that relies on UB ? Can we accept this @h-a-n-a

No, this will not panic and it's more likely a undefined behavior if the memory you're accessing to is freed and it will likely to cause segmentation faults. All we need to do is to prevent user from accessing that value after the scope is dropped.

JSerFeng (approved):

PR #8418, opened 2024-11-12
chore: update error message for task loop

wtlin1228:
## Summary

`tx` is the sender of a channel for `TaskResult<Ctx>`.

## Checklist


chenjiahan (approved):
Thanks!

PR #8434, opened 2024-11-14
fix(core): remove hang risk caused by coop budget in tokio

hardfist:
## Summary
rspack may have hang risk when contains too much task in join_all and block_on, and the automatic cooperative task yielding is useless in rspack scenario(we don't care task latency), it also may bring litter performance benefits(cause no need to do automatically yield)
some users report rspack may block infinitely and I'm not sure related to it.
### Reference
* https://github.com/rust-lang/futures-rs/issues/2157
* https://tokio.rs/blog/2020-04-preemption

## Checklist


hardfist:
!bench

LingyuCoder (approved):

PR #8438, opened 2024-11-14
refactor: use asset filename new type

ahabhgk:
## Checklist


chenjiahan:
@ahabhgk should we close this PR?

PR #8456, opened 2024-11-18
fix(lightningcss): align type with lightningcss

GiveMe-A-Name:
## Summary

In lightningcss, the [draft syntax options](https://lightningcss.dev/transpilation.html#draft-syntax) is drafts. 

We should align the type with lightningcss, while have not break change. So we maintain `options.draft` and `options.drafts`, then we use drafts firstly if user provide both drafts and draft options.

It's the product of another hand, [This PR](https://github.com/web-infra-dev/rspack/pull/8197) from @zackarychapple 

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-loader/lightningcss/index.ts
+++ b/packages/rspack/src/builtin-loader/lightningcss/index.ts
@@ -254,7 +254,12 @@ export type LoaderOptions = {
 	targets?: Targets | string[] | string;
 	include?: FeatureOptions;
 	exclude?: FeatureOptions;
+	/**
+	 * @deprecated Use `drafts` instead.
+	 * This will be removed in the next major version.
+	 */
 	draft?: Drafts;
+	drafts?: Drafts;
 	nonStandard?: NonStandard;
 	pseudoClasses?: PseudoClasses;
 	unusedSymbols?: string[];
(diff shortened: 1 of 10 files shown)


packages/rspack/src/builtin-loader/lightningcss/index.ts
```diff
@@ -254,7 +254,12 @@ export type LoaderOptions = {
 	targets?: Targets | string[] | string;
 	include?: FeatureOptions;
 	exclude?: FeatureOptions;
+	/**
+	 * @deprecated Use `drasfts` instead.
+	 * This will be removed in the next minor version.
```
chenjiahan, line 259: I prefer to remove it in the next major release, as it's a breaking change and is easy to maintain compatibility.

```suggestion
	 * This will be removed in the next major version.
```

chenjiahan (commented):
Can you add the [co-author information](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors#required-co-author-information)?

Example: https://github.com/nuxt/nuxt/pull/29142

packages/rspack/src/builtin-loader/lightningcss/index.ts
GiveMe-A-Name: Done~

chenjiahan (approved):

PR #8457, opened 2024-11-18
fix(modern-module): remove extra import statement for star reexport

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/429.

Previously given a module like this:

```js
export * from "external1"
```

the diff of the output after this PR will be

```diff
- import * as __WEBPACK_EXTERNAL_MODULE_external1__ from "external1";
export * from "external1"
```

## Checklist

### DIFF
--- a/crates/rspack_plugin_library/src/modern_module_library_plugin.rs
+++ b/crates/rspack_plugin_library/src/modern_module_library_plugin.rs
@@ -1,3 +1,4 @@
+use std::collections::HashMap;
 use std::hash::Hash;
 
 use rspack_core::rspack_sources::{ConcatSource, RawSource, SourceExt};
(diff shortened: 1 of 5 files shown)


crates/rspack_plugin_library/src/modern_module_library_plugin.rs
```diff
@@ -287,10 +288,22 @@ async fn finish_modules(&self, compilation: &mut Compilation) -> Result<()> {
...
     let mut deps_to_replace = Vec::new();
+    let mut external_connections = Vec::new();
     let module = mg.module_by_identifier(module_id).expect("should have mgm");
     let connections = mg.get_outgoing_connections(module_id);
     let dep_ids = module.get_dependencies();
 
+    let mut module_id_to_connections: HashMap<
+      &rspack_collections::Identifier,
+      Vec<rspack_core::DependencyId>,
+    > = HashMap::default();
```
ahabhgk, line 299:
```suggestion
    let mut module_id_to_connections: IdentifierMap<Vec<DependencyId>> = HashMap::default();
```

PR #8366, opened 2024-11-07
refactor: rspack_fs use `Utf8Path` and `FileMetadata`

jerrykingxyz:
## Summary

* update `ReadableFileSystem` trait to use `Utf8Path` and `Utf8PathBuf`
In rspack we use `Utf8Path` and `Utf8PathBuf` instead of `Path` and `PathBuf`. more info see https://github.com/web-infra-dev/rspack/pull/7570

* define `FileMetadata` to replace `Metadata` to make Async and Sync trait return same struct

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #8467, opened 2024-11-18
fix: linux musl resolve failed

jerrykingxyz:
## Summary

Some linux musl can not get create time, and it will make resolve failed because of `metadata` return err.

This PR will specify a default value of 0 for the creation time when the platform does not support it.

## Checklist

### DIFF
--- a/crates/rspack_fs/src/file_metadata.rs
+++ b/crates/rspack_fs/src/file_metadata.rs
@@ -1,4 +1,4 @@
-use std::fs::Metadata;
+use std::{fs::Metadata, io::ErrorKind};
 
 use crate::{Error, Result};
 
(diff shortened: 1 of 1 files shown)


JSerFeng (approved):

hardfist:
why the resolve succeed before

jerrykingxyz:
> why the resolve succeed before and dont get the panic message now

Before https://github.com/web-infra-dev/rspack/pull/8366, We use `rspack_resolver::FileMetadata::from` with `std::fs::Metadata` and it does not get the `metadata.created()` values.

code:
https://github.com/web-infra-dev/rspack/pull/8366/files#diff-bf8cd6ecce0c487ed3834f3fa4e20f4bdde318d99fc6f2ff75a49a6e08307948L22

Currently we will convert `std::fs::Metadata` -> `rspack_fs::FileMetadata` -> `rspack_resolver::FileMetadata`, and this error occurs during the first conversion.

LingyuCoder (approved):

PR #8476, opened 2024-11-19
chore(ci): add rust bench ci

SoonIter:
## Summary

chore(ci): add rust bench ci

1. local: cargo bench -p rspack_benchmark
2. remote: cargo codspeed run

## Checklist


tasks/benchmark/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+description = "rspack benchmark tooling"
+edition     = "2021"
+license     = "MIT"
+name        = "rspack_benchmark"
+repository  = "https://github.com/web-infra-dev/rspack"
+version     = "0.1.0"
+publish = false
+
+[dependencies]
+criterion = { package = "codspeed-criterion-compat", version = "2.7.2" }
```
h-a-n-a, line 11: I think we need to run both remotely and locally. Would you please add a criterion package here as well?

SoonIter: done by being replaced by `criterion2`

h-a-n-a (approved):

Issue #7822, opened 2024-09-06
[Feature]: Webpack's output.clean supports the useful option to optionally keep some files around


matthiask:
### What problem does this feature solve?

Thank you for all the work y'all are doing on rspack! Not having to rethink all the frontend building configuration is really great.

Here's the documentation for the option:
https://webpack.js.org/configuration/output/#outputclean

I especially like using `clean: { keep: /\.html$/ }` to keep the generated HTML files around, so that different build or development processes do not delete each other's HTML files (generated by the `html-webpack-plugin`). 

I could use different folders for each of them, but I thought it's worth submitting since it's a difference to webpack

### What does the proposed API of configuration look like?

Webpack uses `boolean | { dry?: boolean, keep?: RegExp | string | ((filename: string) => boolean) }` according to the docs. A very useful, limited first step would be `boolean | { keep?: RegExp }`. I don't think the other options are needed except for feature parity. dry-running shouldn't matter -- if you want to clean up, clean up. If you do not, then don't. I also don't know why you'd want to use a callable here.

shulaoda:
Welcome to close #7845 at any time, I may not have enough time to support it at the moment.

shulaoda:
> cc @shulaoda @h-a-n-a ,  I'd like to take a look.

Welcome!👏

thebat93:
Bump

ClSlaid:
Sorry for the late reply, I've been busy on graduating recently, and I'm expected to be not vacant until late in March. /cc @shulaoda

matthiask:
Bump (and thanks!)

PR #8479, opened 2024-11-19
feat: implement output.clean.keep: Option<string>

ClSlaid:
## Summary

Part of #7822

## Checklist


inottn:
You can write test cases under `packages/rspack-test-tools/tests/configCases/clean/ignore-string`.

crates/node_binding/binding.d.ts
```diff
@@ -1729,7 +1740,7 @@ export interface RawOptions {
 export interface RawOutputOptions {
   path: string
   pathinfo: boolean | "verbose"
-  clean: boolean
+  clean: boolean | JsCleanFilter
```
ClSlaid, line 1743: cannot convert from
```javascript
output: {
  clean: {
    keep: "<str>",
  }
}
```
Complaining expecting a `boolean` but got an `object`.

ClSlaid (commented):
/cc @inottn help needed here.

crates/node_binding/binding.d.ts
inottn: I guess you might need to modify packages/rspack/src/config/zod.ts.

inottn:
Let's first ensure that the CI passes successfully.

crates/rspack_binding_values/src/clean_options.rs
```diff
@@ -0,0 +1,78 @@
...
+  // - support RegExp type
+  //   if path match the RegExp, keep the file
+  // - support function type
+  //    if the fn returns true on path str, keep the file
+}
+
+impl JsCleanFilter {
+  pub fn to_clean_options<T: AsRef<Path>>(&self, working_dir: T) -> CleanOptions {
+    let wd = working_dir.as_ref();
+    let keep = self.keep.as_ref().map(|p| wd.join(p));
```
inottn, line 62: It seems that we don't need to use join here; we can directly use the filename for matching.

crates/rspack_core/src/compiler/mod.rs
```diff
@@ -421,6 +396,42 @@ impl Compiler {
...
+      return Ok(());
+    }
+
+    if self.emitted_asset_versions.is_empty() {
+      self
+        .output_filesystem
+        .remove_dir_all(&self.options.output.path)
+        .await?;
+      return Ok(());
+    }
```
inottn, line 411: When I run `rspack build`, the code reaches this point, and it seems that logic related to `clean` needs to be added here as well.

ClSlaid:
Currently it keeps failing on `./x test ci` inter-op tests, working on it.

What makes debugging troublesome is that I currently have no ways to get logs (stdout, stderr, files, etc.), I'll try set up debuggers later.

inottn:
I'm not very familiar with fs and I'm unsure if it needs any changes. I'll assign this PR to someone who has a better understanding of it as a reviewer. cc @h-a-n-a

crates/rspack_binding_values/src/clean_options.rs
```diff
@@ -0,0 +1,76 @@
...
+
+impl FromNapiValue for JsCleanOptions {
+  unsafe fn from_napi_value(
+    env: napi::sys::napi_env,
+    napi_val: napi::sys::napi_value,
+  ) -> napi::Result<Self> {
+    let val = Either::from_napi_value(env, napi_val);
+    Ok(Self(val?))
+  }
+}
```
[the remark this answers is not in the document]
ClSlaid: Let's wrap it up first, since we need to extend it for RegExps and functions in the future.

ClSlaid:
I'm afraid @h-a-n-a is kind of busy now? Even this PR is for an issue he was supposed to be assigned to.

h-a-n-a:
Will take a look later ;-)

crates/rspack_fs/src/write.rs
```diff
@@ -40,3 +40,81 @@ pub trait WritableFileSystem: Debug + Send + Sync {
 
   fn stat<'a>(&'a self, file: &'a Utf8Path) -> BoxFuture<'a, Result<FileMetadata>>;
 }
+
+/// Extension trait for [`AsyncWritableFileSystem`].
+pub trait WritableFileSystemExt: WritableFileSystem {
```
LingyuCoder, line 45: Don't think the file system should be modified. Perhaps an utility function should be added, and the file system is used inside to handle this.

crates/rspack_binding_options/src/options/raw_output.rs
```diff
@@ -66,7 +66,8 @@ pub struct RawOutputOptions {
   pub path: String,
   #[napi(ts_type = "boolean | \"verbose\"")]
   pub pathinfo: Either<bool, String>,
-  pub clean: bool,
+  #[napi(ts_type = "boolean | JsCleanFilter")]
+  pub clean: JsCleanOptions,
```
LingyuCoder, line 70: Should use `RawCleanOptions` here

crates/rspack_binding_values/src/clean_options.rs
```diff
@@ -0,0 +1,76 @@
...
+/// TS Type:
+///
+/// ```typescript
+/// // in the future, we should support the following types, just like webpack
+/// // type CleanOptions = boolean | { dry?: boolean, keep?: RegExp | string | ((filename: string) => boolean) }
+///
+/// type CleanOptions = boolean | { keep?: string }
+/// ```
+#[derive(Debug)]
+pub struct JsCleanOptions(Either<bool, JsCleanFilter>);
```
LingyuCoder, line 20: Clean options should only be sent from javascript to rust, so just add `#[napi(object, object_to_js = false)]` to it. ToNapiValue trait it not necessary. And I think the conversion logic of the clean options can be handled on the js side, and rust can just receive the converted object. Then `unsafe` can be removed

h-a-n-a: +1. Does not look like we have to add the `unsafe` trait here.

crates/rspack_binding_values/src/clean_options.rs
```diff
@@ -0,0 +1,67 @@
...
+/// // in the future, we should support the following types, just like webpack
+/// // type CleanOptions = boolean | { dry?: boolean, keep?: RegExp | string | ((filename: string) => boolean) }
+///
+/// type CleanOptions = boolean | { keep?: string }
+/// ```
+#[derive(Debug)]
+pub struct RawCleanOptions(Either<bool, JsCleanOptions>);
+
+impl FromNapiValue for RawCleanOptions {
+  unsafe fn from_napi_value(
```
LingyuCoder, line 23: This `unsafe` is just using to convert `either`. You can try to handle this in `packages/rspack/src/config/normalization.ts`, just convert `clean: true/false` to a `RawCleanOptions` struct which can be used on the Rust side directly. Simple Rust API is easy for native accessors to understand.

ClSlaid: It will make the problem worse...

I think we'd better get rid of this struct and just use `Either` directly. I agree with the previous review now.

crates/rspack_binding_values/src/clean_options.rs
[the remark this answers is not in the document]
ClSlaid: > Could we directly use `Either`, instead of wrapping it in a struct?

Yes, this avoids a lot of trouble...

ClSlaid:
I think I need to learn more from the design of `JsPath`, it's sufficient for further development of `output.clean`.

crates/rspack_core/src/compiler/mod.rs
```diff
@@ -73,7 +73,7 @@ impl Compiler {
     options: CompilerOptions,
     plugins: Vec<BoxPlugin>,
     buildtime_plugins: Vec<BoxPlugin>,
-    output_filesystem: Option<Box<dyn WritableFileSystem + Send + Sync>>,
+    output_filesystem: Option<Box<dyn WritableFileSystem>>,
```
LingyuCoder, line 76: These two traits are used in a multi-thread runtime. I think they can not be removed.
I have no more problem if this is modified. Do you need a double check? @h-a-n-a

ClSlaid: https://github.com/web-infra-dev/rspack/blob/6b0e319a620873a14c6ce319a2009ba4297b8aef/crates/rspack_fs/src/write.rs#L8-L42

LingyuCoder (approved):

LingyuCoder:
Thanks

Issue #8469, opened 2024-11-19
[Feature]: There are some missing APIs for frameworks support React Server Component


yimingjfe:
### What problem does this feature solve?

In Modern.js, we want to support RSC building base on Rspack. When using webpack, the solution can work normally, but with Rspack, there are some APIs that are not exported now, so it's difficult to support RSC in the Modern.js repository.

The following APIs are currently used for support RSC, but are not seen exported in Rspack:

Server-side building:

1. Add modules to the compilation
 - `compilation.addInclude`
 - Avoiding tree-shaking of the modules
    - `webpack.util.runtime.getEntryRuntime`
    - `compilation.moduleGraph.getExportsInfo`
    - `exportsInfo.setUsedInUnknownWay`
- `compilation.hooks.needAdditionalPass`

2. Add custom dependencies
- `compilation.dependencyFactories`
- `compilation.dependencyTemplates`
- `module.addDependency`
- `NullDependency`

3. Get ModuleId
- `compilation.hooks.afterOptimizeModuleIds`
- `chunkGraph.getModuleId`

Client-side building:

1. Add client reference chunks
- `AsyncDependenciesBlock`
   - `block.addDependency`
- `Module`
   - `module.blocks`
   - `module.addBlock`

2. Get moduleInfo
- `module.nameForCondition`
- `chunkGraph.getModuleId`
- `chunkGraph.getModuleChunksIterable`
- `moduleGraph.getOutgoingConnections`

Here there are some APIs that may have alternatives , and some are difficult to have real alternatives.

I want to discuss what the recommended solution is from the perspective of Rspack, such as the `addDependency` API, in the webpack ecosystem, there are many webpack plugins that depend on it.

### What does the proposed API of configuration look like?

It is best to be consistent with webpack.

GiveMe-A-Name:
The compilation.entries.entries is exists. You can use it directly.
And other api or attributes we will add in the future.

GiveMe-A-Name:
# Tasks

- [ ] Support compilation.addInclude
- [ ] Support compilation.moduleGraph Api
- [ ] Support compilation.chunkGraph Api
- [ ] ....

SyMind:
I am supporting module graph api here: https://github.com/web-infra-dev/rspack/pull/8470

yimingjfe:
[...]

PR #8483, opened 2024-11-20
feat: support js compilation.addInclude API

GiveMe-A-Name:
### Add compilation.addInclude API

- Require by Modern RSC Plugin 
https://github.com/web-infra-dev/rspack/issues/8469#issuecomment-2487040321

## Checklist

### DIFF
--- a/crates/rspack_binding_values/src/compilation/mod.rs
+++ b/crates/rspack_binding_values/src/compilation/mod.rs
@@ -587,6 +587,8 @@ impl JsCompilation {
   ) -> Result<()> {
     let compilation = self.as_mut()?;
 
+    // SAFETY:
+    // The `dependency` is a nonNull pointer to a `Dependency` object.
     let dependency = unsafe { Box::from_raw(dependency.dependency.as_ptr()) };
 
     compilation
--- a/crates/rspack_binding_values/src/compilation/mod.rs
+++ b/crates/rspack_binding_values/src/compilation/mod.rs
@@ -23,8 +23,10 @@ use rspack_napi::OneShotRef;
 use rspack_plugin_runtime::RuntimeModuleFromJs;
 
 use super::{JsFilename, PathWithInfo};
+use crate::entry::JsEntryOptions;
 use crate::utils::callbackify;
 use crate::JsAddingRuntimeModule;
+use crate::JsDependencyWrapper;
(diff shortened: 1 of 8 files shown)


crates/rspack_binding_values/src/compilation/mod.rs
```diff
@@ -577,6 +579,25 @@ impl JsCompilation {
     Ok(())
   }
 
+  #[napi(ts_args_type = "dependency: JsDependency, options: JsEntryOptions")]
+  pub fn add_include(
+    &mut self,
+    dependency: JsDependencyWrapper,
```
SyMind, line 585: Use JsDependency instead.

crates/rspack_binding_values/src/dependency.rs
```diff
@@ -178,3 +178,20 @@ impl ToNapiValue for JsDependencyWrapper {
     })
   }
 }
+
+impl FromNapiValue for JsDependencyWrapper {
```
SyMind, line 182: JsDependencyWrapper need not impl FromNapiValue

crates/rspack_binding_values/src/compilation/mod.rs
GiveMe-A-Name: Got it.

GiveMe-A-Name: Done~

crates/rspack_binding_values/src/dependency.rs
GiveMe-A-Name: Done~

SyMind:
I'd like to know what this JavaScript API is used for in RSC?

GiveMe-A-Name:
> I'd like to know what this JavaScript API is used for in RSC?

You can see this plugin that implement in webpack 
https://github.com/web-infra-dev/modern.js/blob/feat-next-wym/packages/cli/uni-builder/src/shared/rsc/plugins/rsc-server-plugin.ts#L101-L128

SyMind:
Firstly, I haven't found the way in which `addInclude` is used in `Next.js` and `Modern.js`.

Secondly, I think that `addInclude` is supposed to receive a `Dependency` that is manually constructed by developers on the JavaScript side. However, there isn't a relevant JavaScript API available for this at present.

PR #8516, opened 2024-11-23
feat: support output.environment.nodePrefixForCoreModules

inottn:
## Summary

I found this option in the [documentation](https://rspack.dev/config/output#outputenvironment), but it is not currently supported. It is intended to control whether the node: prefix is added when importing Node.js modules in the generated runtime code, enabling the code to run on other platforms such as Deno and Bun.

## Checklist


LingyuCoder (approved):
Thanks

PR #8522, opened 2024-11-25
refactor: unify to filesystem in rspack_core

hardfist:
## Summary
It's not necessary to distinguish different filesystem trait in rust side, which only cause maintenance burden, 
we just need to distinguish input & output & intermediate filesystem in ThreadsafeNodeFS

## Checklist

### DIFF
--- a/crates/rspack_fs/src/async.rs
+++ b/crates/rspack_fs/src/async.rs
@@ -44,7 +44,7 @@ pub trait AsyncReadableFileSystem: Debug {
   /// Read the entire contents of a file into a bytes vector.
   ///
   /// Error: This function will return an error if path does not already exist.
-  fn read<'a>(&'a self, file: &'a Utf8Path) -> BoxFuture<'a, Result<Vec<u8>>>;
+  fn async_read<'a>(&'a self, file: &'a Utf8Path) -> BoxFuture<'a, Result<Vec<u8>>>;
 }
 
 /// Async readable and writable file system representation.
--- a/crates/rspack_fs/src/fs.rs
+++ b/crates/rspack_fs/src/fs.rs
@@ -1,0 +1,10 @@
+use crate::{
+  AsyncFileSystem, AsyncReadableFileSystem, AsyncWritableFileSystem, SyncFileSystem,
+  SyncReadableFileSystem, SyncWritableFileSystem,
+};
+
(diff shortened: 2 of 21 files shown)


crates/rspack_fs/src/fs.rs
```diff
@@ -0,0 +1,10 @@
+use crate::{
+  AsyncFileSystem, AsyncReadableFileSystem, AsyncWritableFileSystem, SyncFileSystem,
+  SyncReadableFileSystem, SyncWritableFileSystem,
+};
+
+pub trait FileSystem: AsyncFileSystem + SyncFileSystem {}
+
+pub trait ReadableFileSystem: AsyncReadableFileSystem + SyncReadableFileSystem {}
```
jerrykingxyz, line 8: There doesn't seem to be anywhere to use `ReadableFileSystem` and `WritableFileSystem`, maybe we can
``` rust
trait FileSystem: AsyncReadableFileSystem + SyncReadableFileSystem + AsyncWritableFileSystem + SyncWritableFileSystem {}

impl<T> FileSystem for T where T: AsyncReadableFileSystem + SyncReadableFileSystem + AsyncWritableFileSystem + SyncWritableFileSystem {}
```

crates/rspack_fs/src/async.rs
```diff
@@ -44,7 +44,7 @@ pub trait AsyncReadableFileSystem: Debug {
   /// Read the entire contents of a file into a bytes vector.
   ///
   /// Error: This function will return an error if path does not already exist.
-  fn read<'a>(&'a self, file: &'a Utf8Path) -> BoxFuture<'a, Result<Vec<u8>>>;
+  fn async_read<'a>(&'a self, file: &'a Utf8Path) -> BoxFuture<'a, Result<Vec<u8>>>;
```
jerrykingxyz, line 47: why just rename `read` to `async_read`?

jerrykingxyz (approved):

crates/rspack_fs/src/async.rs
hardfist: cause it conflict with the sync version of read

crates/rspack_fs/src/fs.rs
hardfist: I consider remove `AsyncReadableFileSystem` and `SyncReadableFileSystem` and only keep `ReadableFileSystem` in the future

PR #8528, opened 2024-11-25
perf: mem cache for chunk render source

ahabhgk:
## Summary

Although we add incremental for chunk render, since splitChunks will create new chunks if it's enabled in development mode, so these chunks created by splitChunks still take lots of time in chunk render, we introduce a simple mem cache for this case

we store the chunk source as the cache, if the chunk content_hash is not changed, we can reuse the cache, to avoid the cost of generate the chunk source and generate chunk source map (it's a `CachedSource`)

this will also improve the perf when incremental disabled, in an internal project `create_chunk_assets` cost 227ms -> 45ms

## Checklist


ahabhgk:
!bench

JSerFeng (approved):

h-a-n-a:
Will the cache be cleaned on the circumstances where they're no longer necessary?

ahabhgk:
I don't see any cleanup for old_cache, cc @jerrykingxyz, you probably need add a cleanup for new cache

jerrykingxyz:
The old_cache is outdated, it should be removed after all of stage are support incremental rebuild.

For cleanup, i think it should be implemented by incremental rebuild, like Make will cleanup the module_graph and keep only the used module.

ahabhgk:
Notice that incremental rebuild can't cover all the cases for memory cache, it only cutoff some unrelated tasks, but the rest tasks can still benefit from memory cache, even all of stage support incremental rebuild we still need memory cache.

I take Make as an example, `module a` is updated, incremental rebuild will only rebuild module a, but if `module a`'s source code is not changed, then `module::build` can reuse the cache, but since module basically always change its source code, so there is no need to add a `module::build` cache.

but this case in chunk exists a lot, chunk is updated (by splitChunks), but its actual content is not changed, incremental rebuild can't improve this case, but memory cache will improve this a lot

jerrykingxyz:
I agree, the memory cache is necessary in some step. 

But i think the memory cache should not be wrapped in `old_cache`, they should be written in target step(eg. CodeSplittingCache). Of course, memory cache can extract some common logic into tool classes, such as HashMapWithGC.

ahabhgk:
> But i think the memory cache should not be wrapped in `old_cache`, they should be written in target step(eg. CodeSplittingCache). Of course, memory cache can extract some common logic into tool classes, such as HashMapWithGC.

I'm ok with that, I'll make a refactor when time available

PR #8550, opened 2024-11-27
fix: update regex chunk filter logic to allow empty chunk names

### LINKED ISSUE #8529: [Bug]: `chunks` Regex produces different results when used in a function
### System Info

  System:
    OS: macOS 15.1.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 10.48 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.9.0 - ~/.nvm/versions/node/v22.9.0/bin/node
    Yarn: 4.5.1 - ~/.nvm/versions/node/v22.9.0/bin/yarn
    npm: 10.8.3 - ~/.nvm/versions/node/v22.9.0/bin/npm
  Browsers:
    Chrome: 131.0.6778.86
    Safari: 18.1.1
  npmPackages:
    @rspack/cli: 1.1.3 => 1.1.3 
    @rspack/core: 1.1.3 => 1.1.3 

### Details

According to the [docs](https://rspack.dev/plugins/webpack/split-chunks-plugin#splitchunkscachegroupscachegroupchunks) a `RegExp` passed as `chunks` is a short for `(chunk) => regex.test(chunk.name)` but you get a different between a regex and a function with the same regex.

After some debugging I found that I have chunks with `name: undefined`. This passes the regex check in JS but apparently not in Rust. If I change the chunks function to `(chunk) => typeof chunk.name === "string" && regex.test(chunk.name)` the result is aligned with the regex variant.

### Reproduce link

_No response_

### Reproduce Steps

1. Set any RegExp for `splitChunks.chunks`, build and check the generated assets
2. Set `splitChunks.chunks` to the same RegExp but in a function like `(chunk) => regex.test(chunk.name)`, build and see different assets

MindFreeze:
## Summary

fixes #8529 by allowing None chunks to pass the Regex like in JS . This would be a breaking change though.
If this is not ok, than the docs should be updated instead.

## Checklist

### DIFF
--- a/crates/rspack_plugin_split_chunks/src/common.rs
+++ b/crates/rspack_plugin_split_chunks/src/common.rs
@@ -41,7 +41,7 @@ pub fn create_chunk_filter_from_str(chunks: &str) -> ChunkFilter {
 }
 
 pub fn create_regex_chunk_filter_from_str(re: RspackRegex) -> ChunkFilter {
-  Arc::new(move |chunk, _| Ok(chunk.name().map_or(false, |name| re.test(name))))
+  Arc::new(move |chunk, _| Ok(chunk.name().map_or(true, |name| re.test(name))))
 }
 
 #[derive(Debug, Default, Clone)]


inottn:
The changes in this PR are not entirely correct. In JavaScript, the `regex.test` method converts the argument to a string, so `regex.test(undefined)` behaves the same as `regex.test('undefined')`.

MindFreeze:
Ah, indeed. Didn't notice because I was using to regex to exclude stuff. It would be better to just mention it in the docs then

PR #8563, opened 2024-11-28
perf: reuse swc ast struct

GiveMe-A-Name:
## Checklist


GiveMe-A-Name:
!bench

GiveMe-A-Name:
TODO: swc-loader -> xx-loader -> jsParser
Maybe It have other loader to transform js. So in this situations, we also parse twice.

Issue #8208, opened 2024-10-24
[Bug]: An error occurred when removing class in Vue `<style />`. Rerunning `npm run dev` resolved


wangzexi:
### System Info

  System:
    OS: macOS 14.6.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 92.14 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.0 - ~/.volta/tools/image/node/20.18.0/bin/node
    Yarn: 1.22.19 - ~/.volta/tools/image/yarn/1.22.19/bin/yarn
    npm: 10.8.2 - ~/.volta/tools/image/node/20.18.0/bin/npm
    bun: 1.1.20 - ~/.bun/bin/bun
  Browsers:
    Chrome: 130.0.6723.59
    Safari: 17.6

### Details

Reproduce Steps

### Reproduce Steps

Reproduce Steps

```sh
npm create rsbuild@latest # Vue/TypeScript

npm run dev
```

https://github.com/user-attachments/assets/67f936f3-45d1-47ce-9a8a-14703efa4391

CPunisher:
duplicate of https://github.com/web-infra-dev/rspack/issues/7805

Issue #7805, opened 2024-09-05
[Bug]: Strange HMR bug after removing `<style>` in Vue SFC


ZhangTaibin:
### Version

```sh
System: Window
Browser: Edge
vue: 2/3
and others node js 18/20
```

### Details

i just create a pure project with `pnpm create rsbuild@latest`. 

And then run the project with `pnpm run dev` . it's ok and i dont edit anything. 

when i delete `<style></style` element in `App.vue`. It will throw an error writting "Cannot read properties of undefined (reading 'call')". Then i stop the HMR and run the project again. It will be OK.

I think that it has some strange problem with HMR

### Reproduce link

not link

### Reproduce Steps

pnpm create rsbuild@latest

cd project

pnpm i

pnpm run dev

delete style element in `App.vue`

CPunisher:
i cant reproduce with rspack.
after debugging with rsbuild, I find vue style query triggers vue-loader before the vue file does, which is difference from the correct orders in rspack. The vue loader throws errors since there is no style block in the descriptor when handling the style query.

CPunisher:
I reproduce with the following rspack config:
```js
const rspack = require('@rspack/core');
const { VueLoaderPlugin } = require('vue-loader');

/** @type {import('@rspack/cli').Configuration} */
const config = {
  entry: './src/index.js',
  plugins: [
    new VueLoaderPlugin(),
    new rspack.HtmlRspackPlugin({}),
    new rspack.CssExtractRspackPlugin({
      runtime: false,
    }),
  ],
  // experiments: {
  //   css: true,
  // },
  module: {
    rules: [
      {
        test: /\.css$/i,
        use: [
          { loader: rspack.CssExtractRspackPlugin.loader, options: {} },
          'css-loader',
        ],
        type: 'javascript/auto',
      },
      {
        test: /\.vue$/,
        loader: 'vue-loader',
        options: {
          // Note, for the majority of features to be available, make sure this option is `true`
          experimentalInlineMatchResource: true,
        },
      },
    ],
  },
};
module.exports = config;
```

My guess of about the process:
[...]

PR #8595, opened 2024-12-02
fix: passively rebuild modules imported by `importModule`

CPunisher:
## Summary

Investigation: https://github.com/web-infra-dev/rspack/issues/7805#issuecomment-2351982202 and https://github.com/web-infra-dev/rspack/issues/7805#issuecomment-2505248648

> Do not rebuild B ahead. Instead, let A introduce B again.

**Changes:** The executor doesn't rebuilt the referenced modules but only updates the module graph. Since the referencing module also contains file dependecies of the referenced module, the `importModule` will be called again and the referenced module is rebuilt. Finally we fix the artifact after finishing all modules.

closes https://github.com/web-infra-dev/rspack/issues/7805
closes https://github.com/web-infra-dev/rspack/issues/8208

## Checklist


jerrykingxyz:
This process does not seem to be any different from the current process. I guess this error might be caused by the `finish_check` in the ctrl task. @CPunisher Are you interested in refactoring the `finish_check` logic? I can provide context and solutions.

CPunisher:
> This process does not seem to be any different from the current process. I guess this error might be caused by the `finish_check` in the ctrl task. @CPunisher Are you interested in refactoring the `finish_check` logic? I can provide context and solutions.

They are different. Let's assume one of the loaders of module A import module B by `this.importModule`. 
With original process, when the file dependencies of module B change, module B is rebuilt before module A in `hook_before_make` by the executor, and then module A is rebuilt in `make`. 
With this changed process, when the file dependencies of module B change, the executor only update the module graph(cutoff), module B is not rebuilt util the module A is rebuilding and its loader call `this.importModule(moduleB)` again.
From my test, the issue can be fixed by this pr.

If you think it's not the root cause or it's not appropriate, I'm also glad to help refactor the anything.

jerrykingxyz:
emmm...Let me review current logic.

1. we have a moduleA and use `this.importModule` to import moduleB, the moduleA will be added to compilation.module_graph, and the moduleB will be added to module_executor.module_graph.

2. modify the file which moduleB depends on. The moduleA will rebuild because its file_dependencies include all of moduleB’s file dependencies.

3. when make start, the module_executor will remove moduleB from module_executor.module_graph by `cutout`, and will rebuild moduleB by `repair`. both cutoff and repair are wrapped in `update_module_graph`. https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/compiler/module_executor/mod.rs#L68

4. make will `cutout` the moduleA and rebuild it. The moduleA will import the module again using `this.importModule`. The module_executor will run `ExecuteTask` directly, because `module_executor.hook_before_make` will ensure that module_graph is updated. 

From the above process, I think this PR just postpones the timing of rebuilding moduleB. @CPunisher It would be more helpful if you could add some test cases for this case.

jerrykingxyz:
@CPunisher Good edge case, but this PR will introduce other bugs. Lets see a example:

1. moduleA use `importModule` to add moduleB, and moduleB needs to use moduleC, we will get a `module_executor.module_graph` contains moduleB and moduleC, moduleB depends on moduleC.

2. update the file that moduleC depends on. The moduleC will be remove and never regenerate in this PR, because of only `catcut` is called and not `repair`.

I have written a test case for it. https://github.com/web-infra-dev/rspack/commit/f78268511a655c82e36efa6d58b3b2c21fd59ea3

Look back at this issue, I think we should remove the dependencies at request_dep_map from `cutout` result.

1. add a new [MakeParam](https://github.com/web-infra-dev/rspack/blob/8b0f43f00f850181d70068384275e97869d9512b/crates/rspack_core/src/compiler/make/mod.rs#L87C10-L87C19) to skip dependencies. 
``` rust
pub enum MakeParam {
   ...,
   ForceSkipDeps(HashSet<DependencyId>)
}
```

2. impl skip logic at [cutout](https://github.com/web-infra-dev/rspack/blob/8b0f43f00f850181d70068384275e97869d9512b/crates/rspack_core/src/compiler/make/cutout/mod.rs#L107)
``` rust
pub fn cutout_artifact() {
    let skip_deps = HashSet::default();
    ...
    for item in params {
      match item {
          MakeParam::ForceSkipDeps(deps) => {
              skip_deps.extend(deps);
          }
      }
    }
    ...
    force_build_deps.filter(|item| !skip_deps.includes(item))
}
```

3. module_executor call update_module_graph with MakeParam::ForceSkipDeps

``` rust
update_module_graph(vec![..., MakeParam::ForceSkipDeps(self.request_dep_map.values())])
```

CPunisher:
> @CPunisher Good edge case, but this PR will introduce other bugs. Lets see a example:
> 
> 1. moduleA use `importModule` to add moduleB, and moduleB needs to use moduleC, we will get a `module_executor.module_graph` contains moduleB and moduleC, moduleB depends on moduleC.
> 2. update the file that moduleC depends on. The moduleC will be remove and never regenerate in this PR, because of only `catcut` is called and not `repair`.
> 
> I have written a test case for it. [f782685](https://github.com/web-infra-dev/rspack/commit/f78268511a655c82e36efa6d58b3b2c21fd59ea3)
> 
> Look back to the issue that this PR will fixed, I think we can fix it by just remove the dependency at request_dep_map from the `cutout` result.
> 
> 1. add a new [MakeParam](https://github.com/web-infra-dev/rspack/blob/8b0f43f00f850181d70068384275e97869d9512b/crates/rspack_core/src/compiler/make/mod.rs#L87C10-L87C19) to skip dependencies.
> 
> ```rust
> pub enum MakeParam {
>    ...,
>    ForceSkipDeps(HashSet<DependencyId>)
> }
> ```
> 
> 2. impl skip logic at [cutout](https://github.com/web-infra-dev/rspack/blob/8b0f43f00f850181d70068384275e97869d9512b/crates/rspack_core/src/compiler/make/cutout/mod.rs#L107)
> 
> ```rust
> pub fn cutout_artifact() {
>     let skip_deps = HashSet::default();
>     ...
>     for item in params {
>       match item {
>           MakeParam::ForceSkipDeps(deps) => {
>               skip_deps.extend(deps);
>           }
>       }
>     }
>     ...
>     force_build_deps.filter(|item| !skip_deps.includes(item))
> }
> ```
> 
> 3. module_executor call update_module_graph with MakeParam::ForceSkipDeps
> 
> ```rust
> update_module_graph(vec![..., MakeParam::ForceSkipDeps(self.request_dep_map.values())])
> ```

I agree with your idea. I will try it.

crates/rspack_core/src/compiler/module_executor/mod.rs
```diff
@@ -210,12 +236,26 @@ impl ModuleExecutor {
...
+        let dep_status = v.get_mut();
+        let dep_id = dep_status.id;
+        if dep_status.should_update {
+          let dep = LoaderImportDependency::new_with_id(
+            dep_id,
+            request.clone(),
+            original_module_context.unwrap_or(Context::from("")),
+          );
+          dep_status.should_update = false;
+          (ExecuteParam::Entry(Box::new(dep), layer.clone()), dep_id)
```
jerrykingxyz, line 255: Since dep.id is already registered, we can replace it with `Execute Param::Dependency` here.
https://github.com/web-infra-dev/rspack/blob/37e17cf03489ef20b5907701f562399aac29e3e4/crates/rspack_core/src/compiler/module_executor/ctrl.rs#L162-L174

crates/rspack_core/src/compiler/module_executor/mod.rs
```diff
@@ -65,7 +75,20 @@ impl ModuleExecutor {
...
+      .iter()
+      .map(|(id, _)| *id)
+      .collect::<HashSet<_>>();
+    for mut dep_status in self.request_dep_map.iter_mut() {
+      if build_dependencies_id.contains(&dep_status.id) {
+        dep_status.should_update = true;
+        build_dependencies_id.remove(&dep_status.id);
+      }
+    }
+    build_dependencies.retain(|dep| build_dependencies_id.contains(&dep.0));
```
jerrykingxyz, line 90: We can merge the build_dependencies processing into a loop

jerrykingxyz:
For case 3, the build_dependencies returned by `cutout` does not include the dependencies for building ModuleC. Therefore, your code is the same as defining a new MakeParam.

The follow code will remove the dependencies to build ModuleC.
https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/compiler/make/cutout/mod.rs#L131-L141

CPunisher:
> For case 3, the build_dependencies returned by `cutout` does not include the dependencies for building ModuleC. Therefore, your code is the same as defining a new MakeParam.
> 
> The follow code will remove the dependencies to build ModuleC. https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/compiler/make/cutout/mod.rs#L131-L141

Only defining a new MakeParam and skip rebuilding moduleB is not enough. We also need to get `build_dependencies` to determine whether we should remove it (or add a `should_update` mark as I do now) from the `request_dep_map` when it is revoked. So the moduleA will import moduleB again.

Based on above, I give up the new MakeParam since we can directly mutate the `build_dependencies` collection.

crates/rspack_core/src/compiler/module_executor/mod.rs
CPunisher: I'm not sure how to do it. `build_dependencies` is `HashSet<(DependencyId, Option<Identifier>)>` so we can't remove the elements only by `DependencyId` in the loop.

crates/rspack_core/src/compiler/module_executor/mod.rs
CPunisher: Sorry I'm not sure what does it mean...

jerrykingxyz: My mistake, ignore it

jerrykingxyz (approved):
LGTM. This fix is ​​a good one for now. I think we'll have better options once we refactor the "finish_check" logic.

jerrykingxyz:
@CPunisher Thanks for fixing this. I'll create a new issue to provide context about the `finish_check` refactoring if you're interested.

PR #8606, opened 2024-12-03
refactor: improve dependency location

shulaoda:
## Summary

WIP

## Checklist


h-a-n-a:
What is the relationship between `DependencyLocation`, `DependencyLocation` now?

crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -7,31 +10,13 @@ use derivative::Derivative;
 pub struct DependencyRange {
```
h-a-n-a, line 10: We need some comments for these types. It might be confusing to others.

shulaoda: `DependencyRange` is the standardized span form used in `Dependency`, while `DependencyLocation` refers to the location of the `Dependency`. We can derive the `DependencyLocation` from the `source_map` and the `dependency_range`.

h-a-n-a:
Would you please resolve the conflict? I think it's good to merge ;-). Thanks

shulaoda:
> Would you please resolve the conflict? I think it's good to merge ;-). Thanks

done

h-a-n-a (approved):

PR #8614, opened 2024-12-04
feat!: align AssetGeneratorDataUrlFunction with webpack

### LINKED ISSUE #8538: [Feature]: AssetGeneratorDataUrlFunction should align with webpack

In Rspack, module.generator.dataUrl function mode signature is `({ content: string, filename: string }) => string`, it is not align with webpack.
In webpack, module.generator.dataUrl function mode signature like bellow code:
```ts
export type AssetGeneratorDataUrlFunction = (
	source: string | Buffer,
	context: {filename: string; module: Module
) => string;
```

- export `Module` API
- function signature is align with webpack

inottn:
## Summary
close #8538

This is a breaking change, but the previous behavior was inconsistent with webpack.

## Checklist


chenjiahan (commented):
How about be compatible with the previous API signature to avoid breaking the user's code?

For example, if `arg` is an object, convert it to `source + context`.

inottn:
> How about be compatible with the previous API signature to avoid breaking the user's code?
> 
> For example, if `arg` is an object, convert it to `source + context`.
> 
> > We can align type declarations with webpack, but the internal implementation is compatible with Rspack <= 1.1.5. The compatible code can be removed in Rspack v2.0.

Sorry for the delay. If the parameters were directly passed to us by the user, we could maintain compatibility with the old version. However, in this case, the user provides a dataUrl function, and we pass the parameters to it, making compatibility with the old version unfeasible. Let’s put this PR on hold for now and merge it at a more suitable time.

chenjiahan:
This is not a very common usage, and it makes more sense to align with webpack. Maybe we can include this change in Rspack v1.2.0. @hardfist what do you think

chenjiahan:
@hardfist cc

chenjiahan:
After discussion, we decided to include this PR in v1.2.0 to align with webpack.

@inottn Can you rebase the PR again~ ❤️

inottn:
> After discussion, we decided to include this PR in v1.2.0 to align with webpack.
> 
> @inottn Can you rebase the PR again~ ❤️

done~

PR #8628, opened 2024-12-05
fix: sync minify file comments from SWC

fi3ework:
## Summary

Fix https://github.com/web-infra-dev/rslib/issues/418.

The behavior of the new `minify_file_comments` function in `swc_loader` remains consistent with the current implementation:

1. `preserveAllComments: true`
2. `preserveAllComments: false`:
   1. `minify.format.comments: "all"`: true
   2. `minify.format.comments: "some"`: false
   3. `minify.format.comments: false`: false

In the future, you can directly use [`preserve_annotations`](https://github.com/swc-project/swc/blob/e6fc5327b1a309eae840fe1ec3a2367adab37430/crates/swc/src/config/mod.rs#L808) here.

## Checklist

### DIFF
--- a/crates/rspack_util/src/swc.rs
+++ b/crates/rspack_util/src/swc.rs
@@ -1,4 +1,12 @@
-use swc_core::atoms::Atom;
+use swc_config::config_types::BoolOr;
+use swc_core::{
+  atoms::Atom,
+  base::config::JsMinifyCommentOption,
+  common::{
+    comments::{Comment, CommentKind, SingleThreadedComments},
+    BytePos,
+  },
+};
 
 pub fn normalize_custom_filename(source: &str) -> &str {
   if source.starts_with('<') && source.ends_with('>') {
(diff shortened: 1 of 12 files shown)


crates/rspack_util/src/swc.rs
```diff
@@ -26,3 +34,51 @@ fn test_normalize_custom_filename() {
...
+      let preserve_excl = |_: &BytePos, vc: &mut std::vec::Vec<Comment>| -> bool {
+        // Preserve license comments.
+        //
+        // See https://github.com/terser/terser/blob/798135e04baddd94fea403cfaab4ba8b22b1b524/lib/output.js#L175-L181
+        vc.retain(|c: &Comment| {
+          c.text.contains("@lic")
+            || c.text.contains("@preserve")
+            || c.text.contains("@copyright")
+            || c.text.contains("@cc_on")
+            || (preserve_annotations
```
h-a-n-a, line 63: I may not understand this.

Isn't `preserve_annotations` alway set to `false` with `preserve_comment` set to `JsMinifyCommentOption::PreserveSomeComments`?

fi3ework: Yes, this is aligned with current implementation and `swc-loader` behavior. `preserve_annotations` is enabled only when `comments` is set to `"all"`.

h-a-n-a (approved):

PR #8634, opened 2024-12-05
fix: avoid unnamed module ids

ahabhgk:
## Summary

Try our best to avoid unnamed module ids

## Checklist

### DIFF
--- a/crates/rspack_ids/src/named_module_ids_plugin.rs
+++ b/crates/rspack_ids/src/named_module_ids_plugin.rs
@@ -34,8 +34,7 @@ fn assign_named_module_ids(
     })
     .collect();
   let mut name_to_items: HashMap<String, IdentifierIndexSet> = HashMap::default();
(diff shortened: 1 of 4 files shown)


crates/rspack_ids/src/named_module_ids_plugin.rs
```diff
@@ -34,6 +35,7 @@ pub fn assign_named_module_ids(
     .collect();
   let mut name_to_items: HashMap<String, IdentifierIndexSet> = HashMap::default();
   let mut invalid_and_repeat_names: FxHashSet<String> = Default::default();
+  invalid_and_repeat_names.insert(String::from(""));
```
JSerFeng, line 38:
```suggestion
  let mut invalid_and_repeat_names: FxHashSet<String> = std::iter::once(String::new()).collect();
```

JSerFeng (approved):

PR #8640, opened 2024-12-07
refactor: unify the logic of `span_to_location` into trait `SourcePosition`

shulaoda:
## Summary

I have unified this logic into `dependency_location.rs`.

https://github.com/web-infra-dev/rspack/blob/8646ecf26004a5c99ff09cab2b49e457d5d9ca09/crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs#L379-L380

## Checklist

### DIFF
--- a/crates/rspack_core/src/dependency/dependency_location.rs
+++ b/crates/rspack_core/src/dependency/dependency_location.rs
@@ -5,6 +5,7 @@ use std::{
 
 use derivative::Derivative;
 use rspack_cacheable::cacheable;
+use rspack_util::itoa;
 
 /// Represents a range in a dependency, typically used for tracking the span of code in a source file.
 /// It stores the start and end positions (as offsets) of the range, typically using base-0 indexing.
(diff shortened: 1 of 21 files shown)


crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -180,9 +177,52 @@ impl SourceLocation for swc_core::common::SourceMap {
         line: hi.line,
         column: hi.col_display,
       },
-    )
+    ))
+  }
+}
+
+impl SourceLocation for &str {
+  fn look_up_range_pos(&self, start: u32, end: u32) -> Option<(SourcePosition, SourcePosition)> {
```
shulaoda, line 185: I can't implement that method for dyn `rspack_sources::Source`. Perhaps we should implement a `lookup_char_pos` method on the `Source` trait, similar to the one in `swc_core::common::SourceMap`.

h-a-n-a:
I've been a little bit busy lately. Thanks for waiting. I will take a look tomorrow.

h-a-n-a (approved):

PR #8643, opened 2024-12-08
feat: consume js `inputFileSystem` on the rust side

### LINKED ISSUE #5091: Failed to resolve xxx in project root error occurred while using inputFileSystem
### System Info

  System:
    OS: macOS 14.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 387.53 MB / 64.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 19.9.0 - ~/.nvm/versions/node/v19.9.0/bin/node
    npm: 9.6.3 - ~/.nvm/versions/node/v19.9.0/bin/npm
    pnpm: 7.15.0 - ~/.nvm/versions/node/v19.9.0/bin/pnpm
  Browsers:
    Chrome: 120.0.6099.109
    Safari: 17.0
  npmPackages:
    @rspack/core: ^0.4.4 => 0.4.4 

### Details

I'm trying to bundle my project using a virtual file system.

Here is my code.
```js
import { rspack } from "@rspack/core";
import { Volume, createFsFromVolume } from "memfs";
  const files = {
    "/index.tsx": `
    import React from 'react';
    import ReactDOM from 'react-dom';
    ReactDOM.render(<div></div> , document.getElementById("root"));`,
  };
  const val = Volume.fromJSON(files);
  const fs = createFsFromVolume(val);
  const compiler = rspack({
    mode: "production",
    entry: "/index.tsx",
    output: {
      filename: "/ouput.js",
    }
  });
  compiler.inputFileSystem = fs;
  compiler.intermediateFileSystem = fs;
  //@ts-ignore
  compiler.outputFileSystem = fs;

  compiler.run((err, result) => {
    const files = val.toJSON();
    console.log(err, result, files);
  });
```

Here is error message
```shell
ERROR in × Failed to resolve /index.tsx in project root
```

### Reproduce link

_No response_

### Reproduce Steps

It can be reproduced by running the code I provided.

nilptr:
## Summary

close #5091 

- introduce a hybrid filesystem to dispatch fs read between Javascript side or Rust.
- use `inputFileSystem` to decide one file should node.js FS or rust FS.

## Checklist


nilptr:
Hi guys, I have been struggling to investigate the CI "Test Node 18" failure for 2 days. I feel I really need your help.

I believe my implementation can pass all the tests if the tests are run separately.
However, the tests will always be stuck in the `config/builtin-swc-loader/swc-plugin` case, when they are run concurrently.

My findings so far are that it's stuck in https://github.com/web-infra-dev/rspack/blob/937153100e644b068aef9a78d52844ef9324d9a0/packages/rspack-test-tools/src/runner/runner/cjs.ts#L142
even if the output is the same.

hardfist:
thanks ,it's may related to vm support
I think the biggest blocker now is  rspack_resolver doesn't use async_fs, which cause it has to block_on on fs binging call, which will introduce huge performance regression https://github.com/web-infra-dev/rspack-resolver/issues/34.

nilptr:
> thanks it maybe related to performance regression, block_on fs binding will cause huge performance regression
> 
> I think the biggest blocker now is rspack_resolver doesn't use async_fs, which cause it has to block_on on fs binging call, which will introduce huge performance regression [web-infra-dev/rspack-resolver#34](https://github.com/web-infra-dev/rspack-resolver/issues/34).
> 
> but we can still merge this pr first if we don't enable input_fs binding by default and when we solve the rspack_resolver async_fs problem, we can enable inputfs binding

yeah, I also suspect the performance regression, but cannot prove it.

I'm afraid we cannot merge this PR now, because `compiler.inputFileSystem` has been set by default.

https://github.com/web-infra-dev/rspack/blob/5ad5803a0990e4b153c2e9ece0e8933406f4d6cc/packages/rspack/src/rspack.ts#L53-L62

and

https://github.com/web-infra-dev/rspack/blob/9e1205cd17937f624dc4dc4d419641f3b58afa9c/packages/rspack/src/node/NodeEnvironmentPlugin.ts#L53

if you feel we can use the refactors in this pr, I can create a new pr and cherry pick the refactor commits.

hardfist:
> yeah, I also suspect the performance regression, but cannot prove it.
I actually tried it before and revert cause huge performance regression introduced, you can test it in project with large modules like https://github.com/hardfist/performance-compare-ng/tree/master/apps/10000

hardfist:
> I'm afraid we cannot merge this PR now, because compiler.inputFileSystem has been set by default.
that's easy, you can just ignore the inputFileSystem in the rust side, it's actually current's behavior

nilptr:
I disabled `inputFileSystem` on the js side in this PR.

I also created another branch and cherry-picked the refactor commits. maybe #8654 is exactly what you expect.

hardfist:
> I disabled `inputFileSystem` on the js side in this PR.
> 
> I also created another branch and cherry-picked the refactor commits. maybe #8654 is exactly what you expect.

yes, good job!

watsonhaw5566:
@nilptr @stormslowly  这个 pr 是不是可以使用社区的 webpack-virtual-module 的了?

stormslowly:
> @nilptr @stormslowly 你们好，这个 pr 是不是可以支持使用社区的 webpack-virtual-module 的了?

Yes, i will make a canary version; could you please give it try later?

watsonhaw5566:
> > @nilptr @stormslowly 你们好，这个 pr 是不是可以支持使用社区的 webpack-virtual-module 的了?
> 
> Yes, i will make a canary version; could you please give it try later?

好的，感谢，您是飞书群里的哪位，我能否加您一下。

watsonhaw5566:
感谢 @stormslowly @nilptr  兄的付出 ，1.3.10-canary-4a96390f-20250514074254 版本，已经可以正常运行 webpack-virtual-modules

watsonhaw5566:
发现了一个新问题，当使用虚拟模块作为入口文件的时会无法使用输出该虚拟 js 文件并引发了 rspack panic 

最小复现 demo 在这里 https://github.com/watsonhaw5566/rspack-virtual-modules-addentry-demo

packages/rspack/src/FileSystem.ts
```diff
@@ -34,6 +36,120 @@ const NOOP_FILESYSTEM: ThreadsafeNodeFS = {
...
+		// On the rust side, ReadableFileSystem only uses the readFile and stats
+		// TODO: is `memoizeFn` necessary?
+		this.readDir = memoizeFn(() => {
+			const readDirFn = util.promisify(fs.readdir.bind(fs));
+			return async (filePath: string) => {
+				const res = await readDirFn(filePath);
+				return res as string[];
+			};
+		});
+		this.readFile = memoizeFn(() => util.promisify(fs.readFile.bind(fs)));
```
nilptr, line 105: > 发现了一个新问题，当使用虚拟模块作为入口文件时，输出的虚拟 js 文件内容为空，并引发了 rspack panic

@stormslowly caused by this line. please remember revert the debug code before merge.

nilptr:
@watsonhaw5566 there are a few issues with your implementation.

1. `Error: Module not found: Can't resolve 'entry.js' in '/path/to/rspack-virtual-modules-addentry-demo'`
  your entry specifier `'entry.js'` doesn't start with `'/'`, `'./'` or `'../'`, so the [ESM resolution algorithm](https://nodejs.org/api/esm.html#resolution-and-loading-algorithm) will search for it in `node_modules`.
  changing `EntryDependency('entry.js')` to `EntryDependency('./entry.js')` or writing the virtual module to `node_modules/entry.js` can solve this.
2. `Error: build failed with unknown error`
  `entryLoader` will run for every js file, but it's not allowed to add the same entry multiple times. please reconsider your design, don't abuse loader.

watsonhaw5566:
> @watsonhaw5566 there are a few issues with your implementation.
> 
> 1. `Error: Module not found: Can't resolve 'entry.js' in '/path/to/rspack-virtual-modules-addentry-demo'`
>    your entry specifier `'entry.js'` doesn't start with `'/'`, `'./'` or `'../'`, so the [ESM resolution algorithm](https://nodejs.org/api/esm.html#resolution-and-loading-algorithm) will search for it in `node_modules`.
>    changing `EntryDependency('entry.js')` to `EntryDependency('./entry.js')` or writing the virtual module to `node_modules/entry.js` can solve this.
> 2. `Error: build failed with unknown error`
>    `entryLoader` will run for every js file, but it's not allowed to add the same entry multiple times. please reconsider your design, don't abuse loader.

Thanks , I'm modify The to EntryDependency('./entry.js') , make the Rspack Panic.
But now questions  is  the compilation.addEntry can't work on loader .

watsonhaw5566:
> > 但现在的问题是 compilation.addEntry 不能在 loader 上起作用。
> 
> **不要调用`compilation.addEntry`加载器。** 加载器用于转换源代码，而不是直接与编译交互。像这样修改编译应该在插件中完成。 如果您要注入其他模块或操纵构建过程，我建议将这些逻辑移到合适的插件中。
> 
> 您的问题似乎与此 PR 的目的无关。为了保持讨论的重点，请根据需要将此讨论移至“讨论”版块中的新主题。

OK

watsonhaw5566:
> > But now the question is the compilation.addEntry can't work on loader .
> 
> **Don't call `compilation.addEntry` in the loader.** Loaders are meant for transforming source code, not for interacting with the compilation directly. Modifying the compilation like this is something that should be done in a plugin instead. If you're trying to inject additional modules or manipulate the build process, I recommend moving that logic into a proper plugin.
> 
> It looks like your issues are unrelated to the purpose of this PR. To keep the thread focused, please move this discussion to a new thread in the Discussions section if needed.
 
OK, I wll talk it in  Discussions

crates/node_binding/src/resolver.rs
```diff
@@ -44,15 +44,21 @@ impl JsResolver {
...
-      match self.resolver.resolve(Path::new(&path), &request).await {
-        Ok(rspack_core::ResolveResult::Resource(resource)) => {
-          Ok(Either::A(ResourceData::from(resource).into()))
-        }
-        Ok(rspack_core::ResolveResult::Ignored) => Ok(Either::B(false)),
-        Err(err) => Err(napi::Error::from_reason(format!("{err:?}"))),
+    block_on(self._resolve(path, request))
+  }
+
+  pub async fn _resolve(
```
hardfist: shouldn't pub if it's not intended for public api

packages/rspack/src/Compiler.ts
```diff
@@ -845,6 +846,9 @@ class Compiler {
 			this.intermediateFileSystem
 				? ThreadsafeIntermediateNodeFS.__to_binding(this.intermediateFileSystem)
 				: undefined,
+			this.inputFileSystem && options.experiments.useInputFileSystem
```
hardfist: can skip to_binding if useInputFileSystem is empty array

hardfist: also needs add a tests case for useInputFileSystem not empty

crates/node_binding/src/resolver.rs
nilptr: This seems to be caused by a conflict with #10513. I feel it better to unify the implementation. wdyt?

packages/rspack-test-tools/tests/watchCases/build-chunk-graph/chunk-modify/test.config.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
 				});
 				break;
 			case "2":
+				break;
+			case "3":
```
hardfist: why change this?

nilptr: Because js input file system cannot pass this test case. Please refer to this commit https://github.com/web-infra-dev/rspack/pull/8643/commits/55557d3ee259e2d8c2a87b44a5d3c67e71878152 for detailed explanation.

I felt the case of multiple files being modified at the same time is too extreme, so I changed the test case.

crates/node_binding/src/resolver.rs
hardfist: I agree but it's not trivial to call _resolve in resolve since _resolve use &self which is not movable which may needs move the logic to some common util function, but it seems not neat too

hardfist: we can do some refactor in following PR

nilptr: I feel we can replace `resolve` with `_resolve` and use js to implement callback flavor api (ref: 4dccc48e8c3961d85b260dda303961f09d0504c5). There is no need to use rust implement callback api.

hardfist:
thanks

PR #8678, opened 2024-12-12
docs: clarify support for JavaScriptModulesPlugin as partial

steverep:
## Summary

Currently the plugin support table claims full support for `JavascriptModulesPlugin`, but that's not correct.  The static `getCompilationHooks` method only returns the `chunkHash` hook, but there are quite a few more.

For example, my plugin [`transform-async-modules-webpack-plugin`](https://github.com/steverep/transform-async-modules-webpack-plugin) cannot be `rspack` compatible because it uses the `renderModuleContent` tap.  Here's the full list of hooks:

https://github.com/webpack/webpack/blob/8a04c64e5e8e1b7ed1fd4b942a3d277296f14ef9/lib/javascript/JavascriptModulesPlugin.js#L175-L190

## Checklist

### DIFF
--- a/website/components/PluginSupportStatusTable.tsx
+++ b/website/components/PluginSupportStatusTable.tsx
@@ -373,7 +373,10 @@ const pluginSupportStatusList: PluginSupportStatus[] = [
   },
   {
     name: 'JavascriptModulesPlugin',
-    status: SupportStatus.FullySupported,
+    status: SupportStatus.partiallySupported,
+    notes: {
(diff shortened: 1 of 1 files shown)


chenjiahan (approved):
Thanks for correcting this information!

If you’d like Rspack to support more hooks in JavascriptModulesPlugin , feel free to file an issue. ❤️

steverep:
Thanks... opened #8714 to track support for more hooks.

PR #8682, opened 2024-12-12
Perf/reuse swc ast

GiveMe-A-Name:
## Checklist


GiveMe-A-Name:
!bench

PR #8688, opened 2024-12-12
perf(incremental): compute affected cgm incrementally

ahabhgk:
## Summary

Use `chunk_group.origins()` to help compute affected modules with chunk graph more performance way, before we need to `create_block_invalidate_key` for all modules, which is a global effect, now we `create_block_invalidate_key` incrementally, we use `Mutation::ChunkSetId { chunk }` chunk to find the chunk group, and use the `ChunkGroup::origins()` to find the module, and only `create_block_invalidate_key` for these modules.

## Checklist


tests/webpack-test/__snapshots__/StatsTestCases.basictest.js.snap
```diff
@@ -1030,7 +1030,7 @@ chunk {909} (runtime: main) main.js (main) 524 bytes (javascript) 6.82 KiB (runt
   ./modules/f.js [544] 1 bytes {909} [dependent] [built] [code generated]
 chunk {919} (runtime: main) chunk.js (chunk) 4 bytes <{909}> [rendered]
   > [10] ./index.js 3:8-4:19
-  > 9:4-10:15
+  > [10] ./index.js 9:4-10:15
```
ahabhgk, line 1033: caused by `add_origin` fix

JSerFeng (approved):

PR #8697, opened 2024-12-13
feat: export `WarnCaseSensitiveModulesPlugin` and disable it by default

ahabhgk:
## Summary

WarnCaseSensitiveModulesPlugin check case-sensitive modules, before we enable it by default in both dev mode and production mode, but since this is only a "linter" plugin so we disable it by default in both dev mode and production mode, we recommend users use `plugins: [new rspack.WarnCaseSensitiveModulesPlugin()]` to add it themselves by their own needs

## Checklist


chenjiahan (commented):
Will this change benefits the build perf?

ahabhgk:
Yes

hardfist (approved):

PR #8712, opened 2024-12-15
chore: upgrade Rust to 1.83.0 nightly

andersk:
## Summary

Upgrade Rust to nightly-2024-11-27, which was built on the same day as 1.83.0 stable.

## Checklist


andersk:
I note that ctor 0.2.3 has been obsolete for over a year, and Renovate hasn’t been updating it, presumably because it’s only a transitive dependency. According to https://github.com/renovatebot/renovate/discussions/15762, we need to enable the [`lockFileMaintenance` option](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance) to make sure transitive dependencies are updated.

PR #8725, opened 2024-12-16
docs: persistent cache

jerrykingxyz:
## Summary

Add document for `persistent_cache` feature.

## Checklist

### DIFF
--- a/website/docs/en/config/experiments.mdx
+++ b/website/docs/en/config/experiments.mdx
@@ -466,6 +466,10 @@ Specifies an array of paths in `snapshot.managedPaths` that are not managed by t
 
 #### cache.storage
 
+- **Type:** `{ type: 'filesystem', directory: string }`
+
+- **Default:** `{ type: 'filesystem', directory: 'node_modules/.cache/rspack' }`
+
 Configure cache storage. Currently only file system storage is supported. The cache directory can be set through `directory`. The default is `node_modules/.cache/rspack`.
 
 ```js title="rspack.config.js"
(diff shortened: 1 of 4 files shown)


chenjiahan (commented):
This PR contains some non-doc changes, can you split them to another PR?

website/docs/en/config/experiments.mdx
```diff
@@ -334,3 +334,150 @@ module.exports = {
...
+        directory?: string;
+      };
+    };
+```
+
+Control experimental caching behavior. This will only work if [global cache](/config/cache) is set to `true`.
+
+### Disable cache
+
+Configuring `experiment.cache` to `false` to disable cache, which is no different from configuring the [global cache](/config/cache) to `false`.
```
chenjiahan, line 372: The Info in `/config/cache` can be updated now:

image

website/docs/en/config/experiments.mdx
```diff
@@ -334,3 +334,150 @@ module.exports = {
   },
 };
 ```
+
+## experiments.cache
```
chenjiahan, line 338: We can follow the alphabetical order and put the `cache` option before `experiments.css`

website/docs/en/config/experiments.mdx
```diff
@@ -334,3 +334,150 @@ module.exports = {
...
+
+##### snapshot.unmanagedPaths
+
+- **Type:** `(RegExp | string)[]`
+
+- **Default:** `[]`
+
+Specifies an array of paths in `snapshot.managedPaths` that are not managed by the package manager
+
+#### cache.storage
```
chenjiahan, line 467: Missing `Type` and `Default` here

website/docs/en/config/experiments.mdx
jerrykingxyz: Currently they are arranged in the order of supported versions, we can modify them in another PR

chenjiahan: ok

chenjiahan (approved):

PR #8730, opened 2024-12-16
fix: module concatenation for dynamic export info

ahabhgk:
## Summary

`get_export_info` needs mutable module graph for dynamic export info, but we use immutable module graph in codegen phase for some reasons, we can't easily change it to mutable, so introduce `get_export_info_without_mut_module_graph` for these cases.

## Checklist

### DIFF
--- a/crates/rspack_core/src/exports_info.rs
+++ b/crates/rspack_core/src/exports_info.rs
@@ -1795,9 +1795,9 @@ impl ExportInfoData {
   }
 }
 
-// The reture value of `get_export_info_without_mut_module_graph`, when a module's exportType
+// The return value of `get_export_info_without_mut_module_graph`, when a module's exportType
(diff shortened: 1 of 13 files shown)


crates/rspack_core/src/exports_info.rs
```diff
@@ -1806,84 +1647,249 @@ impl ExportInfoData {
...
         }
+        Some(ResolvedExportInfoTargetWithCircular::Target(target))
       }
-      if !resolve_filter(&target, mg) {
-        return Some(ResolvedExportInfoTargetWithCircular::Target(target));
+    }
+  }
+}
+
+#[derive(Debug)]
```
hardfist, line 1795: can you add some comment about this data structure

JSerFeng (approved):

PR #8704, opened 2024-12-13
refactor: replace Regex  in consumesCommon.js

nkxrb:
What kind of change does this PR introduce?
This PR introduces a bugfix to handle compatibility issues with Safari versions below 16, which do not support zero-width assertion regular expressions. It modifies the code to avoid using features like named capture groups or other unsupported constructs.

Did you add tests for your changes?
Yes, tests have been added to ensure the fix works correctly across all affected browser versions, including Safari versions below 16. The tests also validate that the functionality remains unchanged in other environments.

Does this PR introduce a breaking change?
No, this PR does not introduce a breaking change. It ensures backward compatibility by using alternative methods for regular expression handling.

What needs to be documented once your changes are merged?
The documentation should include:

An explanation of the changes made to support older Safari versions.
A note about the limitation of zero-width assertions in Safari below version 16.
Guidance on how users can test for compatibility in their own applications.

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #8749, opened 2024-12-17
fix: mf parse range not compat with safari

ahabhgk:
## Summary

fix #8704

Avoid zero-width assertion regex for older version safari

## Checklist


chenjiahan (approved):
LGTM 👍

PR #8763, opened 2024-12-18
feat: bench js api

SyMind:
## Summary

Using codspeed to bench JS API.

## Checklist


h-a-n-a:
I was wondering if codspeed can show both JS stack and rust stack at the same time. If so, we can just port benchmark tests to JS instead of relying on porting JavaScript default value logic to the rust side.

SyMind:
@h-a-n-a My goal is to find a way that can accurately bench the JS API

h-a-n-a (approved):

PR #8775, opened 2024-12-19
feat: split big chunks based on modules path similarities

JSerFeng:
## Summary

Webpack has an heuristic algorithm that splits big chunks into small ones based on their path similarities. For example, if user's project structure is like:

```
src
 |-aaa
 |   index_1.js
 |   index_2.js
 |   index_3.js
 |-bbb
 |   index_1.js
 |   index_2.js
 |   index_3.js
```

let's say there is one chunk that contains all modules in /aaa and /bbb folders, and it's too big, we set maxSize to split it into smaller pieces, webpack will split them based on path similarities, I assume it's based on one assumption: users modify files that are closed to each other, for example you usually make some changes in some specific sub folders, and don't affect other folders, so other chunks should not change.

we can split them by their paths like, 'aaa.js' and `bbb.js`, or `aaa_1.js` `aaa_2.js` `bbb_1.js` `bbb_2.js`, instead of random splitting

## Checklist

### DIFF
--- a/crates/rspack_plugin_split_chunks/src/common.rs
+++ b/crates/rspack_plugin_split_chunks/src/common.rs
@@ -122,6 +122,10 @@ impl SplitChunkSizes {
   pub fn add_by(&mut self, other: &Self) {
     self.combine_with(other, &|a, b| a + b)
   }
+
+  pub fn subtract_by(&mut self, other: &Self) {
+    self.combine_with(other, &|a, b| a - b)
+  }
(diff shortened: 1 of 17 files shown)


LingyuCoder (approved):

h-a-n-a:
Failing CI is resolved in https://github.com/web-infra-dev/rspack/pull/8776. You may rebase and it will pass.

Issue #8735, opened 2024-12-16
[Bug]: CSS Modules [hash] can occasionally be zero length if is entirely numeric


daveskybet:
### System Info

System:
OS: macOS 14.6.1
CPU: (10) arm64 Apple M1 Pro
Memory: 454.94 MB / 16.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
Node: 20.9.0 - ~/.asdf/installs/nodejs/20.9.0/bin/node
Yarn: 4.4.0 - ~/.asdf/installs/nodejs/20.9.0/bin/yarn
npm: 10.1.0 - ~/.asdf/plugins/nodejs/shims/npm
Watchman: 2024.01.22.00 - /opt/homebrew/bin/watchman
Browsers:
Chrome: 131.0.6778.140
Safari: 17.6
npmPackages:
@rspack/cli: latest => 1.1.6
@rspack/core: latest => 1.1.6

### Details

If the hash of a module name/path is entirely numeric the resultant hashed classname has a zero length (due to the removing of numeric characters at the start of the class name), i.e.

```
generator: {
  localIdentName: "[hash]-[local]",
},
```

for a module with a path/name of:

```
./node_modules/@ppb/the-wall-web/components/walls/Badge/Badge.css
```
results in a completely numeric hash of:

```
9200104150574146
```
(default settings of `xxhash64/hex/16` - although have also had the same with `md4` on a different module)

which due to the existing logic to not have a class name starting with a numeric character will continually remove the first character until the hash has a length of zero, resulting in class names such as:

```
{
  "badgeComponent": ".-badgeComponent",
  "badgeComponentGameInfo": ".-badgeComponentGameInfo",
  "rouletteNumbers": ".-rouletteNumbers",
  "rouletteNumbersBackground": ".-rouletteNumbersBackground",
}
```

Webpack currently has a solution for checking for this edge case and providing an alternative hash here:

https://github.com/webpack/webpack/blob/964c0315df0ee86a2b4edfdf621afa19db140d4f/lib/css/CssModulesPlugin.js#L437
https://github.com/webpack/webpack/blob/964c0315df0ee86a2b4edfdf621afa19db140d4f/lib/util/nonNumericOnlyHash.js

Which doesn't appear to have an equivalent in Rspack.

### Reproduce Steps

- Any Rspack app with the localIdentName set to `[hash]-[local]`
[...]

PR #8778, opened 2024-12-19
fix: correct css modules hashing to match Webpack when leading digit #8735

daveskybet:
## Summary

As detailed in #8735 Rspack is currently out of sync with a recent change in Webpack around how css module hashes that lead with a digit are handled. 

Current Rspack functionality removes all digits from the start of the hash so that hashed classnames never start with a digit, however this produces an edge case where when the hash is entirely digits it is reduced down to an empty string.

Webpack fixed this issue recently to instead prepend hashes starting with a digit with a underscore character ("_") - https://github.com/webpack/webpack/blob/964c0315df0ee86a2b4edfdf621afa19db140d4f/lib/dependencies/CssLocalIdentifierDependency.js#L81

This commit updates the regex/replace functionality to match Webpack's implementation.  

## Checklist

### DIFF
--- a/crates/rspack_plugin_css/src/utils.rs
+++ b/crates/rspack_plugin_css/src/utils.rs
@@ -28,7 +28,7 @@ use crate::parser_and_generator::CssExport;
 
 pub const AUTO_PUBLIC_PATH_PLACEHOLDER: &str = "__RSPACK_PLUGIN_CSS_AUTO_PUBLIC_PATH__";
 pub static LEADING_DIGIT_REGEX: LazyLock<Regex> =
-  LazyLock::new(|| Regex::new(r"^\d+").expect("Invalid regexp"));
+  LazyLock::new(|| Regex::new(r"^((-?[0-9])|--)").expect("Invalid regexp"));
 pub static PREFIX_UNDERSCORE_REGEX: LazyLock<Regex> =
   LazyLock::new(|| Regex::new(r"^[0-9_-]").expect("Invalid regexp"));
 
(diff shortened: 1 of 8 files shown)


ahabhgk (approved):
LGTM, thanks! @LingyuCoder Can you help to run the CI?

LingyuCoder:
Some test cases failed. If they are fixed, this can be merged.

daveskybet:
Sorry had missed an update to a classname in a test, updated now if someone could take a look @ahabhgk @LingyuCoder

ahabhgk (approved):
Thanks

daveskybet:
Thanks for merging this @ahabhgk 🌟 Although on looking at this in `v1.2.0-alpha.0` i've unfortunately spotted an issue with the implementation 😩 

I was 100% sure i'd tested this locally with a minimal repo but the change in the logic to allow leading digits in the hash and prefixing with an underscore seems to have brought in some logic in the `escape_css` function which adds an additional underscore in the css file only (DOM does not contain this additional underscore).  

The offending line is https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_css/src/utils.rs#L125 and removing the underscore character fixes the issue, however I don't fully understand the `if` statement in this function and what potential side-effects there would be on removing this character, also whether this `if` statement is even needed at all once the underscore has been removed.

Could you provide some guidance on this? Thanks.

ahabhgk:
@daveskybet I think you are right, the `if` statement is not needed at all since LEADING_DIGIT_REGEX already add a underscore, here `escape_css` only need to do the escape job

PR #8781, opened 2024-12-19
perf: parallelize side effects optimization

ahabhgk:
## Summary

Refactor our implementation to more parallel friendly and incremental friendly, and parallelize it as much as possible.

Time cost on side effects optimization:

- internal project (FS): 180ms -> 63ms
- arco-pro: 3.5ms -> 1.9ms
- 20000 modules: 2.9ms -> 2.9ms (basically no change since there are no connections are optimizable in this case)

## Checklist


ahabhgk:
!bench

LingyuCoder (approved):

PR #8797, opened 2024-12-20
ci: disable miri, eco-ci and benchmark check

LingyuCoder:
## Summary

Because self-hosted runners are offline, the rspack team does not have enough runners. So disable the miri action temporarily.

## Checklist


h-a-n-a (approved):

h-a-n-a:
You need to disable workflow `rust_test_miri` as well.

chenjiahan:
Should we enable miri now?

Issue #8267, opened 2024-10-30
[Feature]: `Rule.with` support match LogicalCondition `not`


9aoy:
### What problem does this feature solve?

I have the configuration and script as follows. I expect that the `swc-loader` rule will be matched when without import attributes, but it actually does not. Btw, it works as expected in webpack.

```ts
import testStr from "./test.tsx" with { type: "raw" }; // => expect match `asset/source` rule
import test from "./test.tsx"; // => expect match `swc-loader` rule, actually no

```

```ts
module: {
  rules: [
    {
      test: /\.(jsx?|tsx?)$/,
      with: {
        type: {
          not: "raw"
        }
      },
      use: [
        {
          loader: "builtin:swc-loader", 
         }
      ]
    },
    {
      with: { type: "raw" },
      type: "asset/source"
    }
  ]
}
```

### What does the proposed API of configuration look like?

I think the problem is when `attributes` does not exist, it should return true if `LogicalCondition not` exists.
https://github.com/web-infra-dev/rspack/blob/d97d69f3df57b0a78ebffae25f93058779a2d6d9/crates/rspack_core/src/utils/module_rules.rs#L171-L185

Maybe we can refer to webpack's design and set a `matchWhenEmpty` flag for each condition.
https://github.com/webpack/webpack/blob/8b864dbe81bd73f0e7d101f7c9563bdedf00c392/lib/rules/RuleSetCompiler.js#L326

PR #8809, opened 2024-12-23
feat: support `match_when_empty` for rule condition 

CPunisher:
## Summary
Closes https://github.com/web-infra-dev/rspack/issues/8267
Align with webpack https://github.com/webpack/webpack/blob/main/lib/rules/RuleSetCompiler.js#L237

Considering the `match_when_empty` is only calculated once and `TryFrom` can't be async, I use `RwLock` for cache. It's better to run benchmarks to see if there's performance regression.

## Checklist

### DIFF
--- a/crates/rspack_core/src/options/module.rs
+++ b/crates/rspack_core/src/options/module.rs
@@ -636,6 +636,32 @@ impl RuleSetCondition {
       Self::Func(f) => f(data).await,
     }
   }
+
+  #[async_recursion]
+  pub async fn match_when_empty(&self) -> Result<bool> {
+    let res = match self {
+      RuleSetCondition::String(s) => s.len() == 0,
+      RuleSetCondition::Regexp(rspack_regex) => rspack_regex.test(""),
+      RuleSetCondition::Logical(logical) => logical.match_when_empty().await?,
+      RuleSetCondition::Array(arr) => {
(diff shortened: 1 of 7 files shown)


crates/rspack_core/src/options/module.rs
```diff
@@ -636,6 +637,53 @@ impl RuleSetCondition {
...
+        arr.is_empty() && try_any(arr, |c| async move { c.match_when_empty().await }).await?
+      }
+      RuleSetCondition::Func(func) => func("".into()).await?,
+    };
+    Ok(res)
+  }
+}
+
+#[derive(Debug)]
+pub struct RuleSetConditionMatch {
```
ahabhgk, line 657: I prefer `RuleSetConditionWithEmpty`

crates/rspack_core/src/options/module.rs
```diff
@@ -636,6 +637,53 @@ impl RuleSetCondition {
...
+  condition: RuleSetCondition,
+  match_when_empty: RwLock<Option<bool>>,
```
ahabhgk, line 659:
```suggestion
  match_when_empty: OnceLock<bool>,
```

Seems `OnceLock` is more suitable for this case

ahabhgk (approved):
Thanks!

PR #8820, opened 2024-12-23
fix: use cgm hash to verify rename inline modules cache

ahabhgk:
## Summary

Use ChunkGraph module hash to verify the `rename_module_cache`, `build_info().hash` represents the module's `originalSource`, but it's possible to generate different result with the same `originalSource` (checkout added test).

## Checklist


packages/rspack-test-tools/tests/statsAPICases/cache-enabled.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
 	description:
 		"should have any cache hits log of modules in incremental rebuild mode",
 	options(context) {
+		debugger;
```
fi3ework, line 8: Remove?

fi3ework (approved):

PR #8823, opened 2024-12-23
refactor: introducing the faster code-splitting algorithm

JSerFeng:
## Summary

Change the algorithm of code-splitting.

Making it more maintainable, concurrent and easy to working with incoming incremental arch.

Because of the multi-threads, now the code-splitting is more performant than before even using single thread. There can be a huge improvement on large repo.

There is a internal test project which contains about 1.5 million chunk modules, using many dynamic imports, costs about 1200 ms to complete a whole code-splitting phase, using the new algorithm on my 15 core MacBook Pro M1, only takse 600 ms to finish !

The detailed benchmark data will be present soon

## Checklist


tests/plugin-test/css-extract/cases/dependOn-multiple-files-per-entry/expected/entry1.css
```diff
@@ -1,8 +1,8 @@
...
-}
-
 .styleA {
   background: red;
 }
 
+.styleB {
+  background: blue;
+}
+
```
JSerFeng, line 8: Webpack/Rspack legacy code-splitting bug, you can see entry modules from config is A then B

hardfist:
maybe we should add a benchmark case with lots of dynamic import module first and to see the bench metrics avoid future regression

ahabhgk (approved):

ahabhgk (approved):

Issue #8788, opened 2024-12-20
[Bug]: Rspack cssParser will parse failed when url('    ')


GiveMe-A-Name:
### System Info

  System:
    OS: macOS 13.6.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 87.03 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.9.0 - ~/Library/Application Support/fnm/node-versions/v22.12.0/installation/bin/npm
    pnpm: 9.15.0 - ~/Library/pnpm/pnpm
    bun: 1.0.21 - ~/.bun/bin/bun
  Browsers:
    Chrome: 131.0.6778.205
    Safari: 16.6
  npmPackages:
    @rspack/cli: ^1.1.8 => 1.1.8 
    @rspack/core: ^1.1.8 => 1.1.8 
    @rspack/plugin-react-refresh: 1.0.0 => 1.0.0 

### Details

If i write `url('     ')` in css file.
```css
.red {
  fg: green url('    ');
}
```
In webpack, it can compile success, but compile failed in rspack.

```
ERROR in ./src/style.css
  × Empty dependency: Expected a non-empty request
   ╭─[2:16]
 1 │ .red {
 2 │   fg: green url('    ');
   ·                 ──────
 3 │ }
   ╰────
```

The webpack will ignore the url if it;s 

### Reproduce link

https://github.com/GiveMe-A-Name/rspack-css-parser-test

### Reproduce Steps

1. pnpm install 
2. npm run start
3. npm run start:webpack

PR #8824, opened 2024-12-23
fix: Rspack cssParser will parse failed when url(' ')

cbbfcd:
fix [issue 8788](https://github.com/web-infra-dev/rspack/issues/8788)
## Checklist

### DIFF
--- a/crates/rspack_plugin_css/src/parser_and_generator/mod.rs
+++ b/crates/rspack_plugin_css/src/parser_and_generator/mod.rs
@@ -164,7 +164,7 @@ impl ParserAndGenerator for CssParserAndGenerator {
           range,
           kind,
         } => {
-          if request.is_empty() {
+          if request.trim().is_empty() {
             continue;
           }
           let request = replace_module_request_prefix(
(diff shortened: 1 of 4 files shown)


crates/rspack_plugin_css/src/parser_and_generator/mod.rs
```diff
@@ -164,7 +164,7 @@ impl ParserAndGenerator for CssParserAndGenerator {
           range,
           kind,
         } => {
-          if request.is_empty() {
+          if request.trim().is_empty() {
             continue;
```
ahabhgk, line 168: Should we emit a warning/error here? Does webpack experiments.css emit a warning/error? cc @GiveMe-A-Name

GiveMe-A-Name: webpack do not emit any warning, [webpack remove the unnecessary spaces in url(" xxx ")](https://github.com/webpack/webpack/blob/main/lib/css/CssParser.js#L75-L78).
I think we no need to emit warning.

ahabhgk (approved):
Looks good to me, thanks!

GiveMe-A-Name (approved):

PR #8826, opened 2024-12-24
docs: update output.clean type declaration

cbbfcd:
```rust
// https://github1s.com/web-infra-dev/rspack/blob/main/crates/rspack_binding_values/src/clean_options.rs#L13
#[napi(object, object_to_js = false)]
#[derive(Debug)]
pub struct JsCleanOptions {
  pub keep: Option<String>,
  // todo:
  // - support RegExp type
  //   if path match the RegExp, keep the file
  // - support function type
  //    if the fn returns true on path str, keep the file
}
```

## Checklist

### DIFF
--- a/website/docs/en/config/output.mdx
+++ b/website/docs/en/config/output.mdx
@@ -165,6 +165,13 @@ module.exports = {
   output: {
     clean: true, // Clean the output directory before emit.
   },
+
+  // or
+  output: {
+    clean: {
+      keep: 'path', // keep the files under this path
+    },
+  },
 };
 ```
 
--- a/website/docs/en/config/output.mdx
+++ b/website/docs/en/config/output.mdx
@@ -169,7 +169,7 @@ module.exports = {
   // or
   output: {
     clean: {
-      keep: 'path', // keep the files under this path
+      keep: 'ignored/dir', // keep these assets under 'dist/ignored/dir'.
     },
   },
 };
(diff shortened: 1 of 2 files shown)


chenjiahan (commented):
Thanks! Could you help by adding an example using `clean.keep` ?

cbbfcd:
> Thanks! Could you help by adding an example using `clean.keep` ?

ok~

website/docs/zh/config/output.mdx
```diff
@@ -161,6 +161,13 @@ module.exports = {
   output: {
     clean: true, // Clean the output directory before emit.
   },
+
+  // or
+  output: {
+    clean: {
+      keep: 'path', // keep the files under this path
```
chenjiahan, line 168:
```suggestion
      keep: 'ignored/dir', // keep these assets under 'dist/ignored/dir'.
```

website/docs/en/config/output.mdx
```diff
@@ -165,6 +165,13 @@ module.exports = {
   output: {
     clean: true, // Clean the output directory before emit.
   },
+
+  // or
+  output: {
+    clean: {
+      keep: 'path', // keep the files under this path
```
chenjiahan, line 172:
```suggestion
      keep: 'ignored/dir', // keep these assets under 'dist/ignored/dir'.
```

chenjiahan (approved):
Thanks!

PR #8835, opened 2024-12-24
chore(modern-module): jump the asset entry module concaten for Rslib

SoonIter:
## Summary

when set `library.type = "modern-module"` and set asset as entry, there are several bugs

1. not emit asset

2. url is missing

## Checklist


SoonIter:
this is a pr from forked `SoonIter/rspack`... I cannot create canary release

PR #8837, opened 2024-12-24
 fix(modern-module): skip the asset entry module concaten for Rslib 

SoonIter:
## Summary

when set `library.type = "modern-module"` and set asset as entry, there are several bugs

1. not emit asset

2. url is missing

## Checklist


fi3ework (commented):
Need to add a test case to ensure the added code.

fi3ework (approved):

PR #8850, opened 2024-12-25
chore: remove unnecessary TODOs

shulaoda:
## Summary

I have researched and found that these TODOs are no longer necessary.

## Checklist


chenjiahan (approved):
👍

PR #8860, opened 2024-12-26
test: split incremental test cases to run in parallel

LingyuCoder:
## Summary

Since Jest can run in parallel only at the level of test files and can only execute serially within a single test file, the NewIncremental tests were split for parallel execution to reduce the test time consumption. The test on my machine was reduced from 150 seconds to 70 seconds.

image

image

## Checklist


ahabhgk (approved):
👍

PR #8862, opened 2024-12-27
fix: should recognize negative preload value

JSerFeng:
## Summary

`webpackPreload` and `webpackPrefetch` magic comments can set preload and prefetch to script tag, the value of these comments determine the priority and the order of preloaded chunks. The value can also be negative, as it is only for ordering.

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/configCases/web/fetch-priority-2/index.js
+++ b/packages/rspack-test-tools/tests/configCases/web/fetch-priority-2/index.js
@@ -1,42 +1,42 @@
 it("should set fetchPriority", () => {
 	import(/* webpackFetchPriority: "high" */ "./a");
-	expect(document.head.children).toHaveLength(3);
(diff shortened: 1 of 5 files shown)


packages/rspack-test-tools/tests/configCases/web/fetch-priority-2/index.js
```diff
@@ -1,42 +1,42 @@
 it("should set fetchPriority", () => {
 	import(/* webpackFetchPriority: "high" */ "./a");
-	expect(document.head.children).toHaveLength(3);
-	const script1 = document.head.children[1];
+	expect(document.head.children).toHaveLength(4);
```
JSerFeng, line 3: Because now we can detect prefetch correctly, negative prefetch is ignored before

LingyuCoder (approved):

PR #8863, opened 2024-12-27
test: refresh modified file mtime when run NEXT_START

jerrykingxyz:
## Summary

`NEXT_START` in cache case will generate a new compiler to rebuild, we should refresh the modified file mtime to test the snapshot module in persistent cache

## Checklist

### DIFF
--- a/packages/rspack-test-tools/src/helper/util/refreshModifyTime.ts
+++ b/packages/rspack-test-tools/src/helper/util/refreshModifyTime.ts
@@ -1,4 +1,4 @@
-import { readFile, stat, writeFile } from "fs-extra";
+import { readFile, writeFile } from "fs-extra";
 
 export async function refreshModifyTime(file: string) {
 	const data = await readFile(file);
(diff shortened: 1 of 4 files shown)


LingyuCoder (dismissed):

packages/rspack-test-tools/src/helper/util/refreshModifyTime.ts
```diff
@@ -0,0 +1,9 @@
+import { readFile, stat, writeFile } from "fs-extra";
+
+export async function refreshModifyTime(file: string) {
+	const before = await stat(file);
+	const data = await readFile(file);
+	await writeFile(file, data);
+	const after = await stat(file);
+	console.log(before.mtime, after.mtime);
```
LingyuCoder, line 8: `console.log` should be removed

LingyuCoder (dismissed):

LingyuCoder (approved):

Issue #6048, opened 2024-03-27
[Feature]: Support circular-dependency-plugin


ska-kialo:
### What problem does this feature solve?

While the `optimizeModules` hook was added in #2758 to support [circular-dependency-plugin](https://github.com/aackerman/circular-dependency-plugin). The `Module` interface does not include the `dependencies` property the plugin uses to iterate through the dependencies of the module. Additionally the `Compilation` interface does not include the `moduleGraph` property the plugin uses.

### What does the proposed API of configuration look like?

```js
compilation.hooks.optimizeModules.tap('PluginName', (modules) => {
    const seenModules = {};
    for (const module of modules) {
        seenModules[module.debugId] = true;

        for (const dependency of module.dependencies) {
            if (
                dependency.constructor &&
                dependency.constructor.name === 'CommonJsSelfReferenceDependency'
            ) {
                continue;
            }

            const depModule = compilation.moduleGraph.getModule(dependency);

            // Check things on depModule
        }
    }
});
```

The plugin uses the following properties:

- `Module.dependencies`: in Webpack this is an array of `HarmonyImportSideEffectDependency`
- `Module.debugId`: a number
- `Compilation.moduleGraph.getModule(dependency: HarmonyImportSideEffectDependency): Module`: a method on the module graph that resolves a dependency to its `Module`

h-a-n-a:
We're not intended to expose complex data structures like `ModuleGraph`, `Dependency`, `ChunkGrpah`, etc. You may port this plugin to Rust instead.

saifislamrepos:
this proved to be a very important plugin in webpack recently i faced a issue because of circular dependency which took lot of time will be good to have feature in rspack also

hardfist:
this plugin can actually be implemented using stats.json，someone want to give it a try？

ska-kialo:
I implemented a very basic version using the stats at https://github.com/kialo/rspack-circular-dependency-plugin.

For our use case it works, but note that the tests that came with the original plugin do not pass, as I haven't found the time to fix them yet. Therefore it also doesn't have an official release. I'm happy to take contributions.

saifislamrepos:
thanks @ska-kialo worked for me but was returning ids instead of readable names/path extracted modulesById[currentModule].name and resonModule.name working as expected now

ska-kialo:
@saifislamrepos ah yes, I assume you're running in production mode? I hadn't considered that when implementing it.

[...]

PR #8876, opened 2024-12-30
feat: implement CircularDependencyRspackPlugin

faultyserver:
## Summary

This PR implements `CircularDependencyRspackPlugin`, an adaptation of [`circular-dependency-plugin` for Webpack](https://github.com/aackerman/circular-dependency-plugin) implemented natively in Rust to work with the Rspack module graph.

`circular-dependency-plugin` is a moderately-used plugin (~500k weekly downloads) to detect circular import dependencies in bundled code. It's not perfect, and the output isn't always meaningful (i.e., import cycles aren't forbidden in JS, so long as initialization can occur without creating a cycle. This plugin cannot detect that, as is), but identifying cycles is helpful for debugging and resolving bugs and errors _caused_ by circular dependencies.

There's been [an issue requesting support for this plugin](https://github.com/web-infra-dev/rspack/issues/6048) for most of this year, but the major blocker for being able to run the existing plugin is exposing the module graph and dependencies on the JS side, which incurs a major serialization cost. One suggestion was to use the existing `stats.json` object to reverse-build the module graph, but that is also error-prone and not easy to parse, alongside still requiring expensive up-front serialization. So, this PR rewrites the plugin natively in Rust, specifically for Rspack's module graph.

At the same time, the existing plugin is known to be very slow, and only really feasible to run on production builds rather than in a hot-reloading development load. Instead, the implementation for detecting cycles here has been updated to be far more efficient, requiring only a single traversal of the module graph rather than one traversal per module. With that change, and the memory efficiency of reusing the existing module graph, performance is imperceptible, even for projects with thousands of modules and hundreds of thousands of connections between them (in testing, our codebase has ~20k modules and ~230k connections, and the plugin executes with the JS callback hooks in under 50ms). This implementation means it can easily be used in development mode and hot-reloading environments and provide better feedback in the moment.

**Divergence:**

This implementation is _effectively_ the same as the webpack version, but it has some differences:

- It works with module _identifiers_ rather than _relative paths_ (meaning the `cwd` option also isn't implemented). There isn't a big reason for this other than it being expensive to copy Strings all over the place for comparing with the exclude/connection patterns, and the borrow checker really doesn't like keeping Cow references around when the JS callbacks are invoked with a mutable compilation.
- The `include` option is not present, because the original plugin only used this to filter the list of modules it traversed. It can be added without much issue, but doesn't _really_ do much meaningfully.
- Added `ignoredConnections` to give more granular control over ignoring specific cycles, since ignoring an entire module is often too heavy-handed and would prevent detecting any cycles through that module, which seems antithetical to the intent of the plugin (you can never know _absolutely_ that a module isn't problematically cyclical, you can only guess, at which point the plugin gives you nothing).
- The plugin detects _all_ cycles and then filters them out, allowing for an additional `onIgnored` callback to handle them if desired.
- The `onDetected` callback also includes an `entrypoint` parameter first indicating which entrypoint the cycle occurred in (the same cycle will be reported for each entrypoint it occurs in).

## Checklist


crates/rspack_plugin_circular_dependencies/src/lib.rs
```diff
@@ -0,0 +1,385 @@
...
+  /// Returns all dependency cycles contained in the dependency graph starting at
+  /// `initial_module_id`.
+  fn find_cycles_from(
+    &mut self,
+    initial_module_id: ModuleIdentifier,
+  ) -> Vec<Vec<ModuleIdentifier>> {
+    let mut cycles = vec![];
+    self.recurse_dependencies(
+      initial_module_id,
+      &mut IdentifierSet::default(),
```
JSerFeng, line 41: Should we use `&mut Identifier::default()` here ?

This is used for reduce dup visit I suppose, should we move this up to the upper call site ?

Assume entry module a and entry module b both depend on module shared, shared is the root of a cycle, if we make seen_set a new HashSet everytime we will end up 2 same cycle

faultyserver: I think re-reporting each cycle for each entrypoint is _somewhat_ useful, purely for the sake of completeness (i.e., you can know that a cycle is encountered in 2 of 5 entrypoints, but not in the other 3, and use the `onDetected` hook to report them separately). I don't feel super strongly about that, though.

JSerFeng:
LGTM, I can approve once CI succeed

honzahruby:
Hey @faultyserver, any progress on this one? 🙏

faultyserver:
I think the goal is to add integration tests matching the ones that exist in the webpack version of the plugin: https://github.com/aackerman/circular-dependency-plugin/tree/master/__tests__.

I don't currently have time to get those implemented and passing, but happy to look if someone else wants to pick that part up. Alternatively it could be simpler to merge and add tests later rather than trying to work through my forked branch.

JSerFeng:
I'll complete this next week, If somebody want to contribute please let me know

illogic-al:
> I'll complete this next week, If somebody want to contribute please let me know

I mean, I don't mind taking a stab at writing these tests. But the most work I've done with rspack is fixing a few documentation typos so, y'know, I would probably need some help.

hardfist:
> I think the goal is to add integration tests before merging, matching the ones that exist in the webpack version of the plugin: [aackerman/circular-dependency-plugin@`master`/__tests](https://github.com/aackerman/circular-dependency-plugin/tree/master/__tests?rgh-link-date=2025-03-05T21%3A14%3A37Z)__.
> 
> I don't currently have time to get those implemented and passing, but happy to look if someone else wants to pick that part up. Alternatively it could be simpler to merge and add tests later rather than trying to work through my forked branch.

@faultyserver we can continue to add tests case to your pr, if you don't mind

faultyserver:
please do, yes! i unfortunately still don't have much time to continue development here myself, but if others are willing and able to make the needed changes that sounds great to me

fireairforce:
Implement of @faultyserver in the pr description is very detailed, i just so something to follow the job:
- [x] complete the CN doc
- [x] add test case for the plugin, test case can be found at `packages/rspack-test-tools/tests/diagnosticsCases/plugins/circular-dependency-plugin` , i add some warning case to get the `circular_dependency_plugin` diagnostic output.

crates/node_binding/src/raw_options/raw_builtins/raw_circular_dependency.rs
```diff
@@ -0,0 +1,107 @@
...
+  pub on_end: Option<ThreadsafeFunction<JsCompilationWrapper, JsUnknown>>,
+}
+
+impl From<RawCircularDependencyRspackPluginOptions> for CircularDependencyRspackPluginOptions {
+  fn from(value: RawCircularDependencyRspackPluginOptions) -> Self {
+    // This explicit cast is needed because Rust otherwise infers an incompatible type
+    // for the closure compared to the field in the options object.
+    let on_detected: Option<CycleHandlerFn> = match value.on_detected {
+      Some(callback) => Some(Arc::new(move |entrypoint, modules, compilation| {
+        callback.blocking_call_with_sync((
```
LingyuCoder, line 51: tsfn::blocking_call_with_sync may lead to deadlock, perhaps we can use async functions instead.

fireairforce: i am blocked by this, i will disscuss with you later~

fireairforce: done~ i update here from `blocking_call_with_sync` to `.call().await`

PR #8880, opened 2024-12-30
fix: rspack_version macro support match alpha version

jerrykingxyz:
## Summary

`rspack_version!()` macro support match alpha version like `1.2.0-alpha.0`

## Checklist

### DIFF
--- a/crates/rspack_macros/src/rspack_version.rs
+++ b/crates/rspack_macros/src/rspack_version.rs
@@ -1,5 +1,5 @@
 pub fn rspack_version() -> String {
-  let re = regex::Regex::new(r#""version": ?"([0-9\.]+)""#).expect("should create regex");
+  let re = regex::Regex::new(r#""version": ?"([0-9a-zA-Z\.-]+)""#).expect("should create regex");
   // package.json in project root directory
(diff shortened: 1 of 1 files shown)


LingyuCoder (approved):

crates/rspack_macros/src/rspack_version.rs
```diff
@@ -1,5 +1,5 @@
 pub fn rspack_version() -> String {
-  let re = regex::Regex::new(r#""version": ?"([0-9\.]+)""#).expect("should create regex");
+  let re = regex::Regex::new(r#""version": ?"([0-9a-zA-Z\.-]+)""#).expect("should create regex");
```
hardfist, line 2: this is kind of too hacky, even formatter will kill this

hardfist (approved):

crates/rspack_macros/src/rspack_version.rs
hardfist: why not using json parse?

jerrykingxyz: Our package.json is in a standard json format, so there should be no problem in theory.

Issue #5658, opened 2024-02-18
[Feature]: Persistent Cache support


pushkargupta9807:
### What problem does this feature solve?

Faster warm starts for applications. 

Currently the warm starts with webpack are slightly better than the cold starts with rspack. This is preventing us to move to rspack for local development. This support of persistent cache is a must have to onboard developers for their local development.

### What does the proposed API of configuration look like?

https://webpack.js.org/configuration/cache/

NetherGranite:
Bump 🙂

pushkargupta9807:
Bump :)

woshirubbish:
Bump

huhuaaa:
Bump

woshirubbish:
收到！

childrentime:
@jerrykingxyz I think when solving this problem, we can conveniently add remote caching functionality. https://turbo.build/repo/docs/core-concepts/remote-caching

hardfist:
> @jerrykingxyz I think when solving this problem, we can conveniently add remote caching functionality. [turbo.build/repo/docs/core-concepts/remote-caching](https://turbo.build/repo/docs/core-concepts/remote-caching)

this may depend on portable persistent cache which is another feature

woshirubbish:
收到！

woshirubbish:
收到！

jerrykingxyz:
@slorber The current design of rspack persistent cache does not support shared cache with different configurations. For your needs, the configuration and conclusions on rspack persistent cache are consistent with those of webpack.

chenjiahan:
See https://rspack.dev/config/experiments#persistent-cache

PR #8920, opened 2025-01-02
feat: support invalidate persistent cache using `config.mode` and `config.name`

jerrykingxyz:
## Summary

Support invalidate persistent cache using config.mode and config.name.

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -1752,6 +1752,7 @@ export interface RawOptimizationOptions {
 }
 
 export interface RawOptions {
+  name?: string
   mode?: undefined | 'production' | 'development' | 'none'
   context: string
   output: RawOutputOptions
--- a/crates/rspack_binding_values/src/raw_options/mod.rs
+++ b/crates/rspack_binding_values/src/raw_options/mod.rs
@@ -41,6 +41,7 @@ pub use crate::raw_resolve::*;
 #[derive(Debug)]
 #[napi(object, object_to_js = false)]
 pub struct RawOptions {
+  pub name: Option<String>,
   #[napi(ts_type = "undefined | 'production' | 'development' | 'none'")]
   pub mode: Option<RawMode>,
   pub context: String,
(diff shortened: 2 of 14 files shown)


LingyuCoder (approved):

slorber:
Hey 👋 

Just wondering, do you also plan to support `cache.name` like webpack?

On Docusaurus we have to manage multiple caches with 

```js
name: `${name}-${mode}-${props.i18n.currentLocale}`
```

More details here:
- https://github.com/facebook/docusaurus/issues/10823#issuecomment-2574916300
- https://github.com/web-infra-dev/rspack/issues/5658#issuecomment-2485266331

jerrykingxyz:
@slorber We currently have no plans to support "cache.name" since this field provides the same features as "cache.version", just use "cache.version".

``` javascript
cache.version: props.i18n.currentLocale + otherVersion
```

chenjiahan:
@jerrykingxyz We can add some guides for migrating from the webpack cache configuration to Rspack experiments cache.

slorber:
Hi @jerrykingxyz 


> @slorber We currently have no plans to support `cache.name` since this field provides the same features as `cache.version`, just use `cache.version`.

I tried that and it works, but I noticed a difference, explained in https://github.com/facebook/docusaurus/pull/10931#discussion_r1955908019

In Webpack, when the version changes, the cache dir remains clean over time:

image

In Rspack, when the version changes, the cache dir continues to grow, with new subfolders being added, and the former ones not being removed:

image

It's not a big deal to me, but I thought I should report this issue and it could be better to avoid a `.cache/rspack` dir that grows indefinitely over time

jerrykingxyz:
> In Rspack, when the version changes, the cache dir continues to grow, with new subfolders being added, and the former ones not being removed
Yes, Rspack will try to clean up cache dir that have not been used for a long time every time the program is started.

The current directory structure is not necessarily a perfect solution, but we hope to expose as few configuration items as possible so that the subsequent remote cache can have more flexibility.

PR #8934, opened 2025-01-04
fix: allow custom `lazyCompilation.backend.client`

colinaaa:
## Summary

There is an option [`experiments.lazyCompilation.backend.client`](https://rspack.dev/config/experiments#experimentslazycompilation) that can specify custom client of LazyCompilation.

But it doesn't seem to work since it will always be overwritten by `hot/lazy-compilation-web.js` or `hot/lazy-compilation-node.js`

## Checklist


JSerFeng (approved):

JSerFeng (approved):

JSerFeng:
Thank u !

PR #8939, opened 2025-01-06
fix: resolve the inconsistent type issue in `JsModule`

### LINKED ISSUE #8937: [Bug]: TS Errors When Running With `skipLibCheck: false`
### System Info

  System:
    OS: macOS 12.7.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 2.82 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.18.1 - /usr/local/bin/node
    Yarn: 3.6.4 - /opt/homebrew/bin/yarn
    npm: 10.2.5 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 131.1.73.101
    Safari: 15.6.1

### Details

We're building our project which has `skipLibCheck:false` in one of our build steps.

In that process, build fails with the following TS errors:

```
webiny error: ../../node_modules/@rspack/binding/binding.d.ts(172,7): error TS2380: The return type of a 'get' accessor must be assignable to its 'set' accessor type
  Type 'undefined' is not assignable to type 'string'.
../../node_modules/@rspack/core/dist/config/index.d.ts(5,1): error TS1383: Only named exports may use 'export type'.
../../node_modules/@rspack/core/dist/exports.d.ts(36,1): error TS1383: Only named exports may use 'export type'.
```

Dug into the code a bit, and for example, for the first TS2380 error, I can see it:
Image

### Reproduce link

_No response_

### Reproduce Steps

In our repo, we've run our build process with the `skipLibCheck: false` in our TS config. With this setting, we can see the above errors.

shulaoda:
## Summary

closes #8937 

## Checklist

### DIFF
--- a/packages/rspack/src/Module.ts
+++ b/packages/rspack/src/Module.ts
@@ -271,8 +271,10 @@ export class Module {
 				get(): string | undefined {
 					return module.userRequest;
 				},
-				set(val: string) {
-					module.userRequest = val;
+				set(val: string | undefined) {
+					if (val) {
+						module.userRequest = val;
+					}
 				}
 			},
 			rawRequest: {
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -222,7 +222,7 @@ export declare class JsModule {
   get nameForCondition(): string | undefined
   get request(): string | undefined
   get userRequest(): string | undefined
-  set userRequest(val: string)
+  set userRequest(val: string | undefined)
   get rawRequest(): string | undefined
   get factoryMeta(): JsFactoryMeta | undefined
   get type(): string
--- a/crates/rspack_binding_values/src/module.rs
+++ b/crates/rspack_binding_values/src/module.rs
@@ -135,13 +135,16 @@ impl JsModule {
   }
 
   #[napi(setter)]
-  pub fn set_user_request(&mut self, val: String) -> napi::Result<()> {
-    let module: &mut dyn Module = self.as_mut()?;
-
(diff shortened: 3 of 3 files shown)


SoonIter (approved):

packages/rspack/src/Module.ts
```diff
@@ -271,8 +271,10 @@ export class Module {
 				get(): string | undefined {
 					return module.userRequest;
 				},
-				set(val: string) {
-					module.userRequest = val;
+				set(val: string | undefined) {
+					if (val) {
```
SoonIter, line 275: what if val is an empty string `''`?

SoonIter (dismissed):

packages/rspack/src/Module.ts
shulaoda: The `userRequest` property exists only when the module is a `NormalModule`; it does not exist in other cases. When the `userRequest` of a `NormalModule` is an empty string (''), it might overlap in meaning with cases where the `userRequest` is absent in non-NormalModule instances. I am uncertain whether this change is appropriate.

chenjiahan: @SyMind can you confirm this?

SyMind: currently, Module does not have class inheritance informations on the JS side, so I think it is ok in order to solve the TS error

SyMind (approved):

adrians5j:
Thanks! 😊

PR #8967, opened 2025-01-08
feat(core): bump swc_core to 10.0.1

GiveMe-A-Name:
## Summary

Swc v10.0.1 make performance optimize, about minify specially.
When compile large project that has many modules, it would be fastest in production mode.
So we upgrade Swc to v10.0.1

## Checklist


chenjiahan (approved):
🚀

PR #9005, opened 2025-01-13
docs: prefer to use named import and destructuring require

chenjiahan:
## Summary

Although Rspack exports the `rspack` object via both the default export and the named export, we recommend using the named import instead of the default import because it provides better traceability and clarity.

## Checklist


packages/rspack-test-tools/tests/Validation.test.js
```diff
@@ -8,7 +8,7 @@ describe("Validation", () => {
 				errors.push(...args);
 			};
 			try {
-				const rspack = require("@rspack/core");
```
9aoy, line 11: Do we keep a case to verify the default export?

chenjiahan: Yes, see https://github.com/web-infra-dev/rspack/blob/main/packages/rspack-cli/tests/api/type/js-api-cjs.test.ts#L1

9aoy (approved):

PR #9011, opened 2025-01-14
fix: align `amd` option behavior with webpack

### LINKED ISSUE #8987: [Tracking]: Lack of documentation for AMD configuration
### What subject does this issue tracking?

The official website lacks relevant documentation on AMD configuration.

related #8389 

Also, could keep it consistent with webpack and inject it by default when `config.amd !== false`

nilptr:
## Summary

closes #8987

- fix: align `amd` option behavior with webpack
  amd plugins should be applied unless `amd` option is set to `false`
- doc: add doc for `amd` option

other fixes:

- fix: brackets are missing in the generated code for `require(["deps"])` 4ec6bfcd381be4478e804ff0e84e0bc883f70af3
- `define`/`define()`/`typeof define` processing should be skipped if the module is esm 7954d46d9072aa9b498da582bad860558459f7bf
- add `byDependency.amd` to `getResolveDefaults`. Otherwise, amd `define`/`require` cannot resolve modules without extensions  b61a6f14ed8ce19163e84ed0b4c55d466c9ecb26
- update defaults tests snapshots 3eff58c126d85b45dae1596535589c9113b36516
- update webpack tests snapshots 743166516b17942b64d3a98dd3aa27cc0f49e0b3

## Checklist


packages/rspack/src/config/normalization.ts
```diff
@@ -372,7 +372,7 @@ export const getNormalizedRspackOptions = (
 		watchOptions: cloneObject(config.watchOptions),
 		devServer: config.devServer,
 		profile: config.profile,
-		amd: config.amd ? JSON.stringify(config.amd) : undefined,
+		amd: config.amd !== false ? JSON.stringify(config.amd || {}) : undefined,
```
nilptr, line 375: webpack does not explicitly specify a default value for `amd` option. however, `undefined` is equivalent to `{}` in the implement. [ref](https://github.com/webpack/webpack/blob/3919c844eca394d73ca930e4fc5506fb86e2b094/lib/WebpackOptionsApply.js#L394)

this change ensures the behavior is consistent with webpack. but i'm wondering if we should explicitly state the default value is `{}`. wdyt?

crates/rspack_plugin_javascript/src/parser_plugin/esm_detection_parser_plugin.rs
```diff
@@ -97,17 +102,35 @@ impl JavascriptParserPlugin for ESMDetectionParserPlugin {
...
+    (parser.is_esm && is_non_esm_identifier(for_name)).then_some(true)
+  }
+
+  fn call(
+    &self,
+    parser: &mut JavascriptParser,
+    _expr: &swc_core::ecma::ast::CallExpr,
+    for_name: &str,
+  ) -> Option<bool> {
+    (parser.is_esm && is_non_esm_identifier(for_name)).then_some(true)
```
nilptr, line 133: if the module is an esm, `define`/`typeof define`/`define(...)` processing will be skipped by [HarmonyDetectionParserPlugin](https://github.com/webpack/webpack/blob/3919c844eca394d73ca930e4fc5506fb86e2b094/lib/dependencies/HarmonyDetectionParserPlugin.js#L104-L121).

website/components/PluginSupportStatusTable.tsx
```diff
@@ -204,7 +204,7 @@ const pluginSupportStatusList: PluginSupportStatus[] = [
   },
   {
     name: 'RequireJsStuffPlugin',
-    status: SupportStatus.NotSupported,
```
ahabhgk, line 207: Shouldn't change this, we don't have this export, so users can't use this by plugin

website/components/PluginSupportStatusTable.tsx
```diff
@@ -217,7 +217,7 @@ const pluginSupportStatusList: PluginSupportStatus[] = [
   },
   {
     name: 'AMDPlugin',
-    status: SupportStatus.NotSupported,
```
ahabhgk, line 220: Same here

ahabhgk (approved):
Thanks

packages/rspack/src/config/normalization.ts
ahabhgk: This is fine

Issue #8976, opened 2025-01-08
[Bug]: Memory Leak in Dev Server Causing System to Run Out of Application Memory


Fausto95:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (11) arm64 Apple M3 Pro
    Memory: 119.66 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Safari: 17.6
  npmPackages:
    @rsbuild/core@1.11.13
    @rsbuild/plugin-react@1.1.0
```

### Details

While using `rsbuild dev`, I encountered a significant memory leak that caused the system to run out of application memory. This issue makes the system unstable and forces the user to quit applications to free up resources.

![Image](https://github.com/user-attachments/assets/8fc43193-f763-452a-b93c-35a45785c1f8)

CodeSandbox: https://codesandbox.io/p/github/Fausto95/react-native-web-rsbuild/main?import=true

### Reproduce link

https://github.com/Fausto95/react-native-web-rsbuild

### Reproduce Steps

1. `pnpm install`
2. `pnpm start:web`
3. Keep on applying changes to `/apps/rnw/src/App.js`
4. Use `Activity Monitor` and observe the continuous increase in memory usage for the `rsbuild-node` process

uladzimirdev:
I can report the same issue with rspack, `rspack-node` process can take ~30Gb of RAM.

I use `@rspack/core@1.1.6`, rspack config can be found [here](https://github.com/metabase/metabase/blob/master/rspack.config.js)

h-a-n-a:
Not sure how to reproduce this mount of memory consumption, but I found there might be something related to memory reseting in mimalloc. I changed memory semantic from `MADV_FREE` to `MADV_DONTNEED` to reflect memory consumption accurately in statistics like `RSS`.

uladzimirdev:
@h-a-n-a I'm ready to test possible fixes as just running dev server for a while increases memory consumption for me, it wasn't a problem in v1

h-a-n-a:
> @h-a-n-a I'm ready to test possible fixes as just running dev server for a while increases memory consumption for me, it wasn't a problem in v1

@uladzimirdev Do you mind adding a reproducible demo? What version of rapack do you that doesn't have this issue?

uladzimirdev:
@h-a-n-a 

this is what I got on `rspack server`

```
[...]

PR #9037, opened 2025-01-17
fix: patched mimalloc to use `MADV_DONTNEED` than `MADV_FREE` on Linux

h-a-n-a:
## Summary

Use patched mimalloc to use `MADV_DONTNEED` instead of `MADV_FREE` on reseting memory.
This also takes effect on macOS.

While `MADV_FREE` is faster than `MADV_DONTNEED`, it does not affect the `RSS` as `MADV_DONTNEED` does.
To reiterate, using `MADV_DONTNEED` means process-level memory statistics like RSS will be more 
accurately reflect the amount of physical memory being used by rspack.

Even this does not show any evidence that it would cause OOM issues, the fact seems stay the opposite 
with [reported incidents in go](https://github.com/golang/go/issues/42354).  GoLang team and many have encountered this issue and have reverted the option to `MADV_DONTNEED` in https://github.com/golang/go/issues/42330.

Rspack uses mimalloc for memory management.
The default strategy of mimalloc@2 is to **purge** the memory instead of reseting them. 
However, in some cases, it would still reset memory (mi_free_block_mt, freeing blocks multi-thread): 
image

With this patch, `MADV_FREE` is replaced with `MADV_DONTNEED`:
image

Related: https://github.com/microsoft/mimalloc/issues/776
Rspack related: https://github.com/web-infra-dev/rspack/issues/8976

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2671,9 +2671,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "libmimalloc-sys"
-version = "0.1.39"
-source = "git+https://github.com/h-a-n-a/mimalloc_rust.git?rev=fc360a444dd6a4d8de6eb73fea7e964c0c16405f#fc360a444dd6a4d8de6eb73fea7e964c0c16405f"
+name = "libmimalloc-sys-rspack"
+version = "0.1.40"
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
👍

PR #9046, opened 2025-01-20
perf: disable js tracing in non-profile mode

hardfist:
## Summary
when dynamic import failed it won't create cache, so it will import @rspack/tracing every time if user don't install @rspack/tracing which may bring huge cost
image
It's also no need to import @rspack/tracing in non profile mode which can save some time

## Checklist


packages/rspack/src/loader-runner/index.ts
```diff
@@ -339,8 +339,12 @@ function getCurrentLoader(
 	}
 	return null;
 }
-
+let tracingCache!: { tracer: any; activeContext: any };
```
chenjiahan, line 342: Can we avoid using any here

hardfist: fixed

packages/rspack-tracing/types/package.json
```diff
@@ -0,0 +1,4 @@
+{
+  "//": "this file is here to make typescript happy when moduleResolution=node16+",
```
chenjiahan, line 2:
```suggestion
  "//": "This file is for making TypeScript work with moduleResolution node16+.",
```

chenjiahan (approved):

Issue #9016, opened 2025-01-15
[Bug]: DLL Reference plugin results in ESModulesLinkingWarning for CJS files


robrichard:
### System Info

```
  System:
    OS: macOS 15.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 100.17 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.11.1 - /usr/local/bin/node
    Yarn: 3.3.0 - ~/.npm-global/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    Watchman: 2023.09.04.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 131.0.6778.265
    Firefox: 133.0.3
    Safari: 18.1
```

### Details

Rspack warns (or errors with `strictExportPresence: true`) when referencing a cjs module from a DLL manifest.

### Reproduce link

https://github.com/robrichard/rspack-dll-test

### Reproduce Steps

1. Create a DLL manifest containing cjs modules, example `relay-runtime`
1. Create a bundle using `strictExportPresence: true` and DLLReferencePlugin using the manifest created in step 1
1. Build fails with the below error

```
$ npm run build                                                      

> rspack-dll-test@1.0.0 build
> rspack build -c ./config/rspack.dll.config.js && rspack build -c ./config/rspack.dllRef.config.js

Rspack compiled successfully in 141 ms
ERROR in ./src/entry.js
  × ESModulesLinkingError: export 'readInlineData' (imported as 'readInlineData') was not found in 'relay-runtime' (module has no exports)
   ╭─[3:12]
 1 │ import { readInlineData } from 'relay-runtime';
 2 │ 
 3 │ console.log(readInlineData)
   ·             ──────────────
   ╰────


Rspack compiled with 1 error in 13 ms
```

PR #9050, opened 2025-01-20
fix: 🐛 align get provided exports with webpack

stormslowly:
## Summary
close https://github.com/web-infra-dev/rspack/issues/9016
the issue is caused by rspack's dll's manifest file contains an extract `exports: []` in delegated modules's description.
the root of  `exports: []` is that rspack always return vec in get provided exports function.

### solution
align the implementation with [webpack ](https://github.com/webpack/webpack/blob/3919c844eca394d73ca930e4fc5506fb86e2b094/lib/ExportsInfo.js#L538)

## Checklist


hardfist (approved):

stormslowly:
related case snapshot updated；cjs‘s exports which were marked as empty within previous implementation are change to `null` (as determined in runtime)

hardfist (dismissed):

hardfist (approved):

ahabhgk:
I don't think this is an appropriate fix. I'm more inclined to think that there's a mistake in webpack. According to the logic here, our original way of writing was correct.

ahabhgk:
Ok, our original way of writing is also wrong, should be `if info.redirect_to.is_none()`

PR #9059, opened 2025-01-21
feat: enable parallel code splitting by default

JSerFeng:
## Checklist


JSerFeng:
!bench

JSerFeng:
Because of removeAvailableModules is false by default, new code splitting won't remove, so there are many dup modules in the end, costs much time in emitFile

SyMind (approved):

PR #9079, opened 2025-01-22
feat: integrate rspack_dojang for enhanced template rendering

GiveMe-A-Name:
## Summary

Use [dojang](https://github.com/rspack-contrib/rspack-dojang) to render runtime modules, related to https://github.com/web-infra-dev/rspack/issues/5316

Use dojang because it supports custom function, and we can use the same code in template to generate different result by inject functions according to output.environment, and also html plugin is using it so the package size will not increase.
Inject runtime globals as parameters by default.
Dojang only support string literal so array of strings need to be joined with ,.
relate to https://github.com/rspack-contrib/rspack-dojang/pull/3

for example, the on chunk loaded runtime module can be rendered by template blow:
```js
var deferred = [];
<%= ON_CHUNKS_LOADED %> = <%= basicFunction("result, chunkIds, fn, priority") %> {
	if (chunkIds) {
		priority = priority || 0;
		for (var i = deferred.length; i > 0 && deferred[i - 1][2] > priority; i--)
			deferred[i] = deferred[i - 1];
		deferred[i] = [chunkIds, fn, priority];
		return;
	}
	var notFulfilled = Infinity;
	for (var i = 0; i < deferred.length; i++) {
		<%= destructureArray("chunkIds, fn, priority", "deferred[i]") %>
		var fulfilled = true;
		for (var j = 0; j < chunkIds.length; j++) {
			if (
				(priority & (1 === 0) || notFulfilled >= priority) &&
				Object.keys(<%= ON_CHUNKS_LOADED %>).every(<%= returningFunction(ON_CHUNKS_LOADED + "[key](chunkIds[j])", "key") %>)
			) {
				chunkIds.splice(j--, 1);
			} else {
				fulfilled = false;
				if (priority < notFulfilled) notFulfilled = priority;
			}
		}
		if (fulfilled) {
			deferred.splice(i--, 1);
			var r = fn();
			if (r !== undefined) result = r;
		}
	}
	return result;
};
```

Fork from https://github.com/web-infra-dev/rspack/pull/7764

## Checklist


crates/rspack_plugin_runtime/src/runtime_module/async_module.rs
```diff
@@ -17,11 +17,20 @@ impl Default for AsyncRuntimeModule {
 }
 
 impl RuntimeModule for AsyncRuntimeModule {
-  fn generate(&self, _compilation: &Compilation) -> rspack_error::Result<BoxSource> {
-    Ok(RawStringSource::from_static(include_str!("runtime/async_module.js")).boxed())
+  fn generate(&self, compilation: &Compilation) -> rspack_error::Result<BoxSource> {
```
GiveMe-A-Name, line 20: TODO: add tests open all output.environment

GiveMe-A-Name: Not all runtime-template ready, so we don't need tests open all output.environment.

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -2149,6 +2149,7 @@ impl Compilation {
     module.attach(*chunk_ukey);
 
     self.chunk_graph.add_module(runtime_module_identifier);
+    self.runtime_template.add_templates(module.template());
```
LingyuCoder, line 2152: If `module.template()` is called every time when `addRuntimeModule`, will it cause the runtime module to read the template files multiple times?

GiveMe-A-Name: We should using `include_str!` api  that read template file in rust compile-time,  when we implement `module.template` fn.
So runtime_module will not read template file multiple times, it will return a const variable, even though we call `module.template()` multi times.

LingyuCoder (approved):

Issue #9078, opened 2025-01-22
[Bug]: In the AMD module, the define in the function's formal parameters will be compiled into __webpack_require__.amdD.


ununian:
### System Info

System:
    OS: macOS 15.1.1
    CPU: (16) arm64 Apple M3 Max
    Memory: 612.94 MB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.10.0 - ~/Library/Caches/fnm_multishells/46568_1737513178570/bin/node
    Yarn: 1.22.22 - ~/Library/Caches/fnm_multishells/46568_1737513178570/bin/yarn
    npm: 10.2.5 - ~/Library/Caches/fnm_multishells/46568_1737513178570/bin/npm
    pnpm: 9.14.2 - ~/Library/Caches/fnm_multishells/46568_1737513178570/bin/pnpm
    bun: 1.1.25 - ~/.bun/bin/bun
    Watchman: 2024.12.02.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 132.0.6834.83
    Edge: 132.0.2957.115
    Safari: 18.1.1

### Details

When there is an AMD module in the dependencies (seedrandom in this example), the `define` in the function's formal parameters will be compiled into `__webpack_require__.amdD`, causing an error.

<img width="489" alt="Image" src="https://github.com/user-attachments/assets/5e36097a-f575-43a1-bc0f-c8c8ecadae09" />

=====>

<img width="511" alt="Image" src="https://github.com/user-attachments/assets/21006e74-e554-4a2a-b712-ff77f08f5f62" />

### Reproduce link

https://github.com/ununian/rsbuild-bug-define

### Reproduce Steps

 1. npm run dev
2. Open the browser

====
1. npm run build

ununian:
I want to try to fix this issue.

etripier:
Seems to have changed with https://github.com/web-infra-dev/rspack/pull/9011/files#diff-794c8f9fdb152e4e7cd98d168aa635d15009b24239ea9e519f52897ef197a89f.

If you're like me and you want things to work while the amazing maintainers fix the underlying issue, just set `amd: false` in your `rspack.config.js` to get the old behavior.

GiveMe-A-Name:
> I want to try to fix this issue.

Welcome PR

PR #9083, opened 2025-01-22
fix:amd-define-in-parameter (#9078)

ununian:
## Summary

to fix: #9078

I couldn't find a way to get the parent of Expr. I can only add a parameter to describe if the ident is a function's formal parameter. I'm not sure if this is appropriate.

## Checklist


ahabhgk:
Sorry, wrong solution, prefer https://github.com/web-infra-dev/rspack/pull/9088, but still thank you

PR #9086, opened 2025-01-22
fix: should not cache tracing instance

h-a-n-a:
## Summary

Only cache `@rspack/tracing` instead of tracing instance to avoid tracing being written to the incorrect context.

After:

image

## Checklist

### DIFF
--- a/packages/rspack/src/loader-runner/index.ts
+++ b/packages/rspack/src/loader-runner/index.ts
@@ -380,7 +380,7 @@ async function getCachedTracing() {
 }
 
 async function tryTrace(context: JsLoaderContext) {
-	let cachedTracing = await getCachedTracing();
+	const cachedTracing = await getCachedTracing();
 	if (cachedTracing) {
 		const { trace, propagation, context: tracingContext } = cachedTracing;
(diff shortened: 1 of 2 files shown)


packages/rspack/src/loader-runner/index.ts
```diff
@@ -340,36 +340,64 @@ function getCurrentLoader(
...
-	try {
-		const {
-			trace,
-			propagation,
-			context: tracingContext
-		} = await import("@rspack/tracing");
+	if (cachedTracing) {
+		return cachedTracing;
+	}
+	if (cachedTracing === undefined) {
```
hardfist, line 363: this seems has performance issue when user dont install @rspack/tracing，it will try load many times

h-a-n-a: `cachedTracing` will be set to `null` on following calls. Only the first call would match `try catch`

h-a-n-a: cachedTracing will be set to null on following calls. Only the first call would match try catch

hardfist (approved):

PR #9103, opened 2025-01-23
revert: fix: align `amd` option behavior with webpack

ahabhgk:
## Summary

revert https://github.com/web-infra-dev/rspack/pull/9011

In https://github.com/web-infra-dev/rspack/pull/9011 we have basic AMD support, but there are still some edge cases that will broken by AMD, so we disable it for now, we will re-enable it when our AMD support becomes more stable

## Checklist


hardfist (approved):

nanianlisao:
Revert this commit is indeed a bad decision, so can I safely set amd to `{}` ?

ahabhgk:
Back then this feature has bugs, so I think it's fairly reasonable to revert this commit, now this feature is stabled in >=v1.3.0-beta.0 and we've updated our docs and added the reason about why we decided to keep the defaults as is https://rspack.dev/config/other-options#amd

Issue #4550, opened 2023-11-07
[Tracking]: Wasm support and support run on stackblitz and WebContainer


hardfist:
### What subject does this issue tracking?

# [Subject]

```[tasklist]
### Tasks
- [ ] napi-rs wasm support
- [ ] build Rspack wasm target
- [ ] integrate Rspack into Stackblitz
- [ ] run stackblitz example on rspack website
```

hardfist:
this is for someone who is interested in wasm integration

hussain-s6:
Any plans for this support? I believe it is required for Rspack/Rsbuild to work inside Stackblitz (as you have already mentioned inside this ticket: #2986 )
I will try to run Rsbuild inside codesandbox but I much prefer Stackblitz as it's much faster and has much better "free" version.

hardfist:
we will support wasm but it's low priority

busy-mango:
> we will support wasm but it's low priority

I hope this feature comes sooner rather than later

chenjiahan:
Update: https://x.com/ScriptedAlchemy/status/1805779131365490778

![image](https://github.com/web-infra-dev/rspack/assets/7237365/3edc1d3e-38a5-4e68-a585-585ec14324f1)

hardfist:
close in favor of #9134

PR #9134, opened 2025-01-27
feat: support wasm

CPunisher:
## Description

Related: https://github.com/web-infra-dev/rspack/issues/4550

By following https://github.com/rolldown/rolldown/pull/467, I have made an ugly [MVP](https://x.com/c_punisher/status/1882412147965382689) to run rspack in browser environment.

Since I find that support for wasm is a feature that involves lots of small changes in current codebase, I'd like to list all the known steps in the following tasklist and divide the process into stages:

1. Compile rspack to wasm target, i.e. we can get `.wasm` files by `napi build`. In this stage we mostly add `#[cfg(not(target_family = "wasm"))]` to distinguish platforms.
2. Make rspack wasi available in node.
3. Make rspack wasi available in browser and provide an easy way for users.

## Roadmap

- [x] Compile rspack to wasm target: https://github.com/web-infra-dev/rspack/pull/9585
- Run rspack wasi in node
  - [x] fix: pass regex error https://github.com/web-infra-dev/rspack/pull/9872
  - [x] feat: bump napi to support multi-thread https://github.com/web-infra-dev/rspack/pull/9855
  - [x] fix: too much exports: https://github.com/napi-rs/napi-rs/pull/2523
  - [x] fix: rayon thread pool stucks workers. https://github.com/web-infra-dev/rspack/pull/10478
  - [x] fix: version & platform check: https://github.com/web-infra-dev/rspack/pull/10097
  - [x] fix: `module_export` in `napi` doesn't support wasm target. https://github.com/web-infra-dev/rspack/pull/10437
  - [x] fix: check test cases. https://github.com/web-infra-dev/rspack/pull/10757

## Debugging Hint (Browser)

1. Install https://chromewebstore.google.com/detail/cc++-devtools-support-dwa/pdcpmagijalfljmkmjngeonclgbbannb
2. Set `profile.release.debug = true` in `Cargo.toml` and ensure the wasm file is smaller than 1GB, which is the limitation of the browser
3. Add path mappings by following https://users.rust-lang.org/t/getting-raw-wasm-debugging-working-nicely-in-chrome-devtools/94646#getting-debugging-working-4
4. Add another mappings by following https://issues.chromium.org/issues/401522579#comment5 due to the bugs of the extension in current version.

### DIFF
--- a/crates/rspack_allocator/Cargo.toml
+++ b/crates/rspack_allocator/Cargo.toml
@@ -11,5 +11,5 @@ version     = "0.2.0"
 # Turned on `local_dynamic_tls` to avoid issue: https://github.com/microsoft/mimalloc/issues/147
 mimalloc = { workspace = true, features = ["local_dynamic_tls"] }
 
-[target.'cfg(not(target_os = "linux"))'.dependencies]
+[target.'cfg(not(any(target_os = "linux", target_family = "wasm")))'.dependencies]
 mimalloc = { workspace = true }
--- a/crates/rspack_allocator/src/lib.rs
+++ b/crates/rspack_allocator/src/lib.rs
@@ -1,3 +1,3 @@
 #[global_allocator]
-#[cfg(not(miri))]
+#[cfg(not(any(miri, target_family = "wasm")))]
 static GLOBAL: mimalloc::MiMalloc = mimalloc::MiMalloc;


hardfist:
> Run @rspack/core in the browser. Since rspack interface exposed by node_binding is missing lots of features, such as option nomalization, essential plugins, loader context, which make it hard to directly use, we also need to make this part avaliable. Actually I have not explored this stage in the MVP.
If Rust API is good enough, maybe we don't need to ship the js binding parts to browser, @h-a-n-a what do you think?

hardfist:
>  tokio: async tokio::fs => sync std::fs 
seems unnecessary cause the ntaivefs is used for native OS, we can just ship a browser_fs for browser side which even don't need std::fs

CPunisher:
> > tokio: async tokio::fs => sync std::fs
> 
> seems unnecessary cause the ntaivefs is used for native OS, we can just ship a browser_fs for browser side which even don't need std::fs

This is necessary because:
1. tokio fs feature is not compilable in wasm: https://docs.rs/tokio_wasi/latest/tokio/#wasm-support.
2. with the help of napi, `std::fs` is easily linked to an external file system provided in js side. See: https://github.com/rolldown/rolldown/pull/467/files#diff-603a6b705934598e015625888d735ef43ae35ffc4cb19aba185610fd7c9eb0dfR18

So I just call `std::fs` in wasm with `#[cfg]` in the MVP.

hardfist:
I mean `tokio::fs` should only be used in NativeFileSystem https://github.com/web-infra-dev/rspack/blob/5a6ed3137e20fad8f9ca3d6b1842178fa22335fa/crates/rspack_fs/src/native_fs.rs#L19 which maybe not necessary for browser, so you could choose either ways
* change tokio::fs to std::fs in NativeFileSystem when target to wasm 
* don't use NativeFileSystem and passing fs instance to compiler in browser https://github.com/web-infra-dev/rspack/blob/5a6ed3137e20fad8f9ca3d6b1842178fa22335fa/crates/rspack_core/src/compiler/mod.rs#L80

hardfist:
> rspack_futures: its dependency async-scoped depends on tokio multi-thread rt.


not sure rspack_futures can be replaced with with tokio_stream and wether tokio_stream can be compatible with wasm

hardfist:
@CPunisher do you plan to finish all tasks in this PR all separated into multi PRs

CPunisher:
> > rspack_futures: its dependency async-scoped depends on tokio multi-thread rt.
> 
> not sure rspack_futures can be replaced with with tokio_stream and wether tokio_stream can be compatible with wasm

I'm not sure whether `rspack_futures` is used for performance. In the MVP I just replace all the usages with `join_all`. See: https://github.com/CPunisher/rspack/commit/7123f19b2d3ee6ea8c3c6a9eaff1dd5d1bce81ad#diff-5eb759d814281903aa0355174051f134df9444a46e9a1ce332cad27eed896215R1051

CPunisher:
> @CPunisher do you plan to finish all tasks in this PR or separated into multi PRs

I'm also considering, but I think both are ok. Because this feature consist of many small changes, it's very easy to separate it. But it seems that incomplete changes are useless. So what's your suggestion?

hardfist:
that depends on how long would it take, I suggest when you can successfully compile rspack to wasm target we can merge it first and introduce wasm target in CI so we wouldn't introduce wasm incompatible crate in the future.

chenjiahan:
> I'm not sure whether we need a light weight implementation or a full implementation

It would be great if a full implementation could be provided, which allows us to run Rsbuild on Stackblitz.

CPunisher:
Close as the wasm target is avaliable in [Rspack 1.4](https://rspack.rs/blog/announcing-1-4)

PR #9152, opened 2025-02-03
test(webpack-test): add back two test case

stormslowly:
## Summary

- add back two test cases(remove `skip`)
- refectory the test utils to fail fast

## Checklist


tests/webpack-test/fixtures/node_modules/complexm/step1.js
```diff
@@ -0,0 +1 @@
+module.exports = require("m1/a") + require("m1");
```
GiveMe-A-Name, line 1: Why we need add `step1.js`, `step2.js`, `m1/a.js`, `m1/b.js`.... files? 
Where are them using?

stormslowly: all files in node_modules/complexm are dead files.

deldeted.

GiveMe-A-Name:
Thanks your PR.

PR #9166, opened 2025-02-06
fix: lazy compilation request handler should filter lazy-compilation request

JSerFeng:
## Summary

The requestHandler of lazyCompilation should only handle request from lazy-compilation client.

If user defines `lazyCompilation.backend.server`, setup of the server should be handled by user who provide the server, so the listen method should be called by user as well.
But if user specify the listen behavior explicitly, lazyCompilation will use that listen

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/lazy-compilation/backend.ts
+++ b/packages/rspack/src/builtin-plugin/lazy-compilation/backend.ts
@@ -2,16 +2,33 @@
  * MIT License http://www.opensource.org/licenses/mit-license.php
  * Author Tobias Koppers @sokra
  */
-
 import type {
 	IncomingMessage,
 	ServerOptions as ServerOptionsImport,
 	ServerResponse
 } from "node:http";
-import type { AddressInfo, ListenOptions, Server, Socket } from "node:net";
+import type { AddressInfo, ListenOptions, Socket } from "node:net";
 import type { SecureContextOptions, TlsOptions } from "node:tls";
 
 import type { Compiler } from "../..";
+
+interface Server {
+	on(
+		event: "request",
+		callback: (req: IncomingMessage, res: ServerResponse) => void
+	): void;
+	on(event: "connection", callback: (socket: Socket) => void): void;
+	on(event: "listening", callback: (err?: Error) => void): void;
+
+	off(
+		event: "request",
+		callback: (req: IncomingMessage, res: ServerResponse) => void
+	): void;
+
+	address(): AddressInfo;
+	close(callback: (err?: any) => void): void;
+	listen(listenOptions?: number | ListenOptions): void;
+}
 
 export interface LazyCompilationDefaultBackendOptions {
 	/**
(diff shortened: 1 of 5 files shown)


packages/rspack/src/builtin-plugin/lazy-compilation/backend.ts
```diff
@@ -82,16 +82,22 @@ const getBackend =
...
 			typeof options.listen === "function"
 				? options.listen
-				: (server: Server) => {
-						let listen = options.listen;
-						if (typeof listen === "object" && !("port" in listen))
-							listen = { ...listen, port: undefined };
-						server.listen(listen);
-					};
+				: typeof options.server === "function"
+					? // if user offers custom server, no need to listen
```
chenjiahan, line 86: Can you update the documentation to tell this behavior?

chenjiahan: I think we are missing a lot of documentation on custom lazy compilation backends

website/docs/en/config/experiments.mdx
```diff
@@ -168,26 +168,108 @@ In addition, you can also configure a `test` parameter for more fine-grained con
...
+  close(callback: (err?: any) => void): void;
+
+  listen(listenOption: number | ListenOption): void;
+}
+```
+
 ### lazyCompilation.backend.listen
 
-- **Type:** `number | ListenOptions`
+- **Type:** `number|ListenOption`
```
chenjiahan, line 254:
```suggestion
- **Type:** `number | ListenOptions | ((server: Server) => void)`
```

website/docs/en/config/experiments.mdx
```diff
@@ -168,26 +168,108 @@ In addition, you can also configure a `test` parameter for more fine-grained con
 The current lazy compilation aligns with the webpack implementation, **and is still in the experimental stage**. In some scenarios, lazy compilation might not work as expected, or the performance improvement may be insignificant.
 :::
 
+## experiments.lazyCompilation.backend
+
+- **Type:** `LazyCompilationBackendOptions`
+
+```ts
+type LazyCompilationBackendOptions = {
```
chenjiahan, line 176: We can change L114 from `backend?: { ... }` to `LazyCompilationBackendOptions` to reduce duplication of types in the document

website/docs/en/config/experiments.mdx
```diff
@@ -168,26 +168,108 @@ In addition, you can also configure a `test` parameter for more fine-grained con
...
-   * @default false
-   */
-  ipv6Only?: boolean | undefined;
+  readableAll?: boolean | null;
+  writableAll?: boolean | null;
+  ip6Only?: boolean | null;
 };
 ```
 
+Specify how the sever listens ,you can pass in object form listening configuration,such as specifying listening ports,you can also pass in functions control how listen is done .
```
chenjiahan, line 269:
```suggestion
Specify how the sever listens ,you can pass in object form listening configuration,such as specifying listening ports,you can also pass in functions control how listen is done.
```

website/docs/en/config/experiments.mdx
```diff
@@ -168,26 +168,108 @@ In addition, you can also configure a `test` parameter for more fine-grained con
...
+
+If you provide your own service instance,the service's listen is called by provider unless explicitly configured with functional forms of listening.
```
chenjiahan, line 271:
```suggestion
When providing a custom server instance, you are responsible for calling the server's `listen` method to start the service, unless you explicitly configure a function forms` listen` option.
```

chenjiahan (approved):
👍

chenjiahan (approved):

Issue #9154, opened 2025-02-04
[Bug]: TS Errors When Running With `skipLibCheck: false`


adrians5j:
### System Info

  System:
    OS: macOS 12.7.6
    CPU: (10) arm64 Apple M1 Pro
    Memory: 3.70 GB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 20.18.2 - /usr/local/bin/node
    Yarn: 4.6.0 - /opt/homebrew/bin/yarn
    npm: 10.2.5 - /opt/homebrew/bin/npm
  Browsers:
    Brave Browser: 132.1.74.48
    Safari: 15.6.1

### Details

> [!NOTE]  
> This issue is similar to the one I described [here](https://github.com/web-infra-dev/rspack/issues/8937). This time just have a different TS error that's occurring.

We're building our project which has `skipLibCheck:false` in one of our build steps.

In that process, build fails with the following TS errors:

```
Error: [FAILED] ../../node_modules/@rspack/core/compiled/@swc/types/index.d.ts(1,29): error TS2307: Cannot find module './assumptions' or its corresponding type declarations.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2378,57): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2378,92): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2398,31): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2409,31): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2425,57): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2425,92): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2439,31): error TS2315: Type 'Buffer' is not generic.
Error: [FAILED] ../../node_modules/@rspack/core/dist/config/zod.d.ts(2447,31): error TS2315: Type 'Buffer' is not generic.
[...]

PR #9173, opened 2025-02-06
fix(types): prebundle `@swc/types` to fix missing dts files

inottn:
## Summary
relate #9154

## Checklist

### DIFF
--- a/packages/rspack/prebundle.config.mjs
+++ b/packages/rspack/prebundle.config.mjs
@@ -79,14 +79,7 @@ export default {
 			}
 		},
 		{
-			name: "@swc/types",
-			ignoreDts: true,
-			afterBundle(task) {
-				copyFileSync(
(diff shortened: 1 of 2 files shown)


packages/rspack/prebundle.config.mjs
```diff
@@ -80,13 +80,7 @@ export default {
...
 		{
 			name: "@swc/types",
-			ignoreDts: true,
-			afterBundle(task) {
-				copyFileSync(
-					join(task.depPath, "index.d.ts"),
-					join(task.distPath, "index.d.ts")
-				);
-			}
+			ignoreDts: false
```
chenjiahan, line 83: We can remove `ignoreDts: false` as it is the default value

inottn: done~

chenjiahan (approved):

PR #9179, opened 2025-02-06
docs(typo): Fix rspack.config typo in nestjs

Codfisher:
## Summary

After executing `cross-env BUILD=true rspack build`, it will appear

```
Invalid configuration object. Rspack has been initialized using a configuration object that does not match the API schema.
- Unrecognized key(s) in object: 'rule' at "module"
```

After comparing the [example content](https://github.com/rspack-contrib/rspack-examples/blob/03aa9913fe1208173111488860a56bfd56d15aaf/rspack/nestjs/rspack.config.js#L15), I found that `rule` should be `rules`.

## Checklist


chenjiahan (approved):
Thanks!

Issue #9176, opened 2025-02-06
[Feature]: Provide a CLI flag to disable interpret


krutoo:
### What problem does this feature solve?

In Webpack there is a CLI option `--disable-interpret` to disable interpret

This is usual when you want to use custom loader for TS, for example I use `tsimp`

Can we add this flag to `@rspack/cli`?

### What does the proposed API of configuration look like?

Here is an example of usage 

```bash
npx rspack --disable-interpret
```

hardfist:
yeah, can you make a pr to support it

krutoo:
@hardfist Yes, will trying soon

PR #9210, opened 2025-02-08
feat: add `--config-loader` CLI option

krutoo:
## Summary

According to my issue #9176 I prepare PR with implementation of `--disable-interpret` CLI option. 

This CLI option is analogue of same option in Webpack.

For myself this option is wery helpful because in my projects I use tsimp as main loader for TS.

## Checklist

### DIFF
--- a/packages/rspack-cli/src/utils/options.ts
+++ b/packages/rspack-cli/src/utils/options.ts
@@ -47,8 +47,8 @@ export const commonOptions = (yargs: yargs.Argv) => {
 			},
 			'config-loader': {
 				type: "string",
-				default: "native",
-				describe: "Specify the loader to load the config file, can be `none` or `native`.",
+				default: "register",
+				describe: "Specify the loader to load the config file, can be `native` or `register`.",
 			},
 		})
 		.alias({ v: "version", h: "help" });
--- a/packages/rspack-cli/src/utils/options.ts
+++ b/packages/rspack-cli/src/utils/options.ts
@@ -45,11 +45,12 @@ export const commonOptions = (yargs: yargs.Argv) => {
 				string: true,
 				describe: "Name of the configuration to use."
 			},
-			'config-loader': {
+			"config-loader": {
 				type: "string",
 				default: "register",
-				describe: "Specify the loader to load the config file, can be `native` or `register`.",
-			},
+				describe:
+					"Specify the loader to load the config file, can be `native` or `register`."
+			}
 		})
 		.alias({ v: "version", h: "help" });
 };
--- a/packages/rspack-cli/src/utils/options.ts
+++ b/packages/rspack-cli/src/utils/options.ts
@@ -45,11 +45,10 @@ export const commonOptions = (yargs: yargs.Argv) => {
 				string: true,
 				describe: "Name of the configuration to use."
 			},
-			'disable-interpret': {
-				type: "boolean",
-				default: false,
-				describe: "Disable interpret for loading the config file.",
-				alias: "d"
+			'config-loader': {
+				type: "string",
+				default: "native",
+				describe: "Specify the loader to load the config file, can be `none` or `native`.",
 			},
 		})
 		.alias({ v: "version", h: "help" });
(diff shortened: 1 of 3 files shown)


chenjiahan (commented):
Thank you for your contribution!

Since [interpret](https://www.npmjs.com/package/interpret) is no longer maintained, Rspack will most likely replace `interpret` with other solutions in the future. Therefore we prefer a more general flag name.

How about `--config-loader native`? Both Rsbuild and Vite provides `--config-loader native` flag.

See:

- https://github.com/web-infra-dev/rsbuild/pull/4486
- https://github.com/vitejs/vite/pull/19178

krutoo:
@chenjiahan sounds good, do I need to wait for someone else's opinion (from maintainers of this repo) or can I remake the option to `--loader-config`?

chenjiahan:
You can remake the options first. 
I'll invite other maintainers if needed.

krutoo:
@chenjiahan would it be more correct to use `--config-loader interpret` instead of `--config-loader native`?

As I understand "native" means using `--strip-types` (native Node.js feature to partially use TS)

packages/rspack-cli/src/utils/options.ts
```diff
@@ -45,11 +45,10 @@ export const commonOptions = (yargs: yargs.Argv) => {
...
 			},
-			'disable-interpret': {
-				type: "boolean",
-				default: false,
-				describe: "Disable interpret for loading the config file.",
-				alias: "d"
+			'config-loader': {
+				type: "string",
+				default: "native",
+				describe: "Specify the loader to load the config file, can be `none` or `native`.",
```
chenjiahan, line 51: My recommendation:

- `native` - use the native Node.js behaviour
- `register` - use any register installed in the user project

Defaults to `register` as before.

krutoo: By register you mean interpret?

chenjiahan: yep, `interpret` is the library name and `register` could be a more accurate term for these registers

krutoo: @chenjiahan how i use this CLI option for disable any loaders?

chenjiahan: `--config-loader native` to disable the loaders and use Node.js native behavior

krutoo: What dou you mean by native? strip types? I think this value doesn't make it clear what exactly will happen

krutoo: Hmm, I checked it again now and realized they are a bit confusing:

The option is named as "config-loader" and as I understand it means what loader will be used

And, "native" sounds like "native Node.js loader will be used"

But in Node.js there is no default loaders for modules, there is only API for register custom loaders

`--config-loader native` sound like there is a some native loader for typescript for example

We don't say that importing regular JS file is "using native Node.js loader"

krutoo: Unfortunately, it is already done in Vite and Rsbuild so I updated PR according this tools

chenjiahan: > But in Node.js there is no default loaders for modules

When we say "native loaders", we mean the ESM loaders or CJS loaders inside Node.js:

- https://github.com/nodejs/node/blob/3186468bed7c40b9a76a69643468d8e5a16c8771/lib/internal/modules/cjs/loader.js
- https://github.com/nodejs/node/blob/3186468bed7c40b9a76a69643468d8e5a16c8771/lib/internal/modules/esm/loader.js

I think the "native" here does not only refer to Node.js. Bun and Deno also provide native support for TypeScript.

krutoo: Just want to say that it sound strange: 
When I pass "register" value instead "native" to "--config-loader" option it looks like i decline "native" and disable native loader

But in fact it doesn't work that way: under the hood there is no way to disable native runtime feature to importing some modules, there is only ability to override some default behevior

chenjiahan: I completely understand your confusion. This option should be removed in the future, when all Node.js versions used by Rspack users support TypeScript. It will be much easier for users to understand then.

chenjiahan (approved):

krutoo:
@chenjiahan should I also update docs?

chenjiahan:
That would be helpful 👍

krutoo:
@chenjiahan here is PR with some docs: https://github.com/web-infra-dev/rspack/pull/9291

Issue #8312, opened 2024-11-01
[Tracking]: migrate all blocking call to async call


hardfist:
### What subject does this issue tracking?

# migrate all blocking call to async call
we already migrate the whole architecture to async thread pool in https://github.com/web-infra-dev/rspack/pull/8226
so blocking call will block the whole thread and will hurt performance, so we should migrate all blocking call to async call for performance reason, which includes
```[tasklist]
### Tasks
- [x] migrate all block_on to async call
- [x] migrate all blocking_call in tsfn to call_async
- [x] migrate rspack_resolver to use async fs(this may need benchmark)
- [x] migrate all rust side sync hook to async hook
```

PR #9217, opened 2025-02-10
perf: migrate all blocking call to async call

CPunisher:
## Summary

Related: https://github.com/web-infra-dev/rspack/issues/8312

Help me start benchmarking. **I will try to polish the code further if performance improves** (and split the pr).

There are still undo tasks:
1. Many `block_on`s and `blocking_call`s are called in `impl From<RawXX> for XX`s, which is hard to be converted into `async`. Maybe we should create our own `AsyncFrom`.
2. Some hooks are hard to migrate:
    - `NormalModuleLoaderShouldYield`: `module: NonNull<dyn Module>` is not `Send` in RunnerContext: https://github.com/CPunisher/rspack/blob/87945e092d42741490ae4b36d405ce400a375149/crates/rspack_core/src/loader/rspack_loader.rs#L49
    - `CompilationRuntimeRequirementInModule`, `CompilationRuntimeRequirementInChunk`, `CompilationRuntimeRequirementInTree`: It's hard to migrate the related code in https://github.com/CPunisher/rspack/blob/87945e092d42741490ae4b36d405ce400a375149/crates/rspack_core/src/compiler/compilation.rs#L1700
3. `block_on` in Module trait functions: https://github.com/CPunisher/rspack/blob/87945e092d42741490ae4b36d405ce400a375149/crates/rspack_macros/src/runtime_module.rs#L189, https://github.com/CPunisher/rspack/blob/87945e092d42741490ae4b36d405ce400a375149/crates/rspack_macros/src/runtime_module.rs#L131

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -9,6 +9,7 @@ use std::{
 };
 
 use dashmap::DashSet;
+use futures::{future::join_all, FutureExt};
 use indexmap::{IndexMap, IndexSet};
 use itertools::Itertools;
 use rayon::prelude::*;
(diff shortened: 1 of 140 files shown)


chenjiahan:
benchmark triggered 😄

CPunisher:
🤣Strange performance regression since my local tests are positive. I will continue to investigate.

hardfist:
@CPunisher  notice that async call have performance gain when have lots of async js call but may not have performance gain(may even bring performance regression) if not have any or little js call
but 19% performance regression seems too big

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -937,33 +941,42 @@ impl Compilation {
...
-                    &self.options.output.hash_salt,
-                  );
-                  codegen_res
-                })
-            }),
-        )
-      })
-      .collect::<Vec<_>>();
+    let compilation = &*self;
+    let results = join_all(jobs.into_iter().map(|job| async move {
```
hardfist, line 945: this seems run in single thread while  rayon runs in multi thread, so you may need to use spawn to run task in multi thread, this may cause performance regression now

CPunisher: yes, i realize this is not a good migration pattern

hardfist: actually I don't know what's the perfect way to migrate these, async parallel is kind of hard in rust now

hardfist: https://github.com/web-infra-dev/rspack/pull/9587/files#diff-5eb759d814281903aa0355174051f134df9444a46e9a1ce332cad27eed896215R1059 async-scoped can solve the performance problem using unsafe function

PR #9259, opened 2025-02-12
fix(deps): govern graceful-fs sourcemap phantom dependencies

SoonIter:
## Summary

1. patch is not patched when upgrading graceful-fs

https://github.com/web-infra-dev/rspack/pull/7803

2. `graceful-fs` and `source-map` are phantom dependencies 

The CI isn't throwing errors because we're using `pnpm run`. The key difference between `pnpm run` and `npm run` is that pnpm run overrides the `NODE_PATH` environment variable to point to `.pnpm/node_modules`.

image

## Checklist


SoonIter:
> I'm not sure whether the patch is needed anymore, you can try remove it

pnpm does not remove this patch, so do we

stormslowly:
wait there are some more

packages/rspack-test-tools/package.json
```diff
@@ -110,6 +111,7 @@
     "react-refresh": "^0.16.0",
     "sass-embedded": "^1.77.8",
     "sass-loader": "^16.0.0",
+    "source-map": "^0.7.4",
```
hardfist, line 114: who relys on this?

SoonIter: image

stormslowly (approved):

PR #9281, opened 2025-02-13
refactor: NormalModule.source remove diagnostic data

jerrykingxyz:
## Summary

1. `Diagnosable` trait remove `has_error` and add `first_error` method.
``` diff
trait Diagnosable {
-  fn has_error(&self) -> bool;
+  fn first_error(&self) -> Option<Cow<Diagnostic>>;
}
```
2. update NormalModule.source to Option<BoxSource>, we should use Module.first_error to get the error information instead of getting it from Module.source.
``` diff
struct NormalModule {
-  source: NormalModuleSource,
+  source: Option<BoxSource>
}
```
3. remove NormalModule.original_source since it will always have the same value as NormalModule.source.
``` diff
struct NormalModule {
-  original_source: Option<BoxSource>
}
```

## Checklist

### DIFF
--- a/crates/rspack_core/src/context_module.rs
+++ b/crates/rspack_core/src/context_module.rs
@@ -847,7 +847,7 @@ impl Module for ContextModule {
     &[SourceType::JavaScript]
   }
 
-  fn original_source(&self) -> Option<&dyn rspack_sources::Source> {
+  fn source(&self) -> Option<&rspack_sources::BoxSource> {
     None
   }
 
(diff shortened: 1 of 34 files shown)


crates/rspack_core/src/context_module.rs
```diff
@@ -847,7 +847,7 @@ impl Module for ContextModule {
     &[SourceType::JavaScript]
   }
 
-  fn original_source(&self) -> Option<&dyn rspack_sources::Source> {
+  fn source(&self) -> Option<&rspack_sources::BoxSource> {
```
ahabhgk, line 850: prefer original_source, imply it is the input from user land

ahabhgk (commented):
Basically looks good

crates/rspack_core/src/context_module.rs
jerrykingxyz: The reason why original_source is merged into source is that source will not be modified after build is assigned, so I think it is OK to use source directly.

ahabhgk (approved):

PR #9285, opened 2025-02-13
fix(splitChunks): splitChunks maxSize not work as expected if size of some type is small

JSerFeng:
## Summary

Align: https://github.com/webpack/webpack/blob/3919c844eca394d73ca930e4fc5506fb86e2b094/lib/util/deterministicGrouping.js#L326.

there is a chunk contains following modules (use 0 represents module)

```
[0 0 0 0 0 0 0]
```

User set maxSize, and total size of these modules is greater than maxSize, so we need to split it

From left, we find a position which can just fulfill the minSize

From right, we also find a position which can just fulfill the minSize

Case 1:

Perfact splitting

```bash
[0 0 0 0 0 0 0]
[0 0 0]
 left^
      [0 0 0 0]
       ^right
```

Split it into 2 chunks, left and right are all greater than minSize

Case 2:

```bash
[0 0 0 0 0 0 0]
[0 0 0]
 left^
          [0 0]
           ^right
```

Split it based on some special algorithm, still, the 2 chunks are all greater than minSize, so it's fine

Case 3:

```bash
[0 0 0 0 0 0 0]
[0 0 0 0 0]
     left^
        [0 0 0]
         ^right
```

Left and Right has overlaps, if split left, right part doesn't satisfy minSize, for current version before this pr, we just ignore this, and do not split this chunk at all.

But when size is smaller than minSize, there is a possibility that not all module type's size is smaller than minSize,

eg. the right 3 modules has only 1 module that contains css, the total css size is smaller than minSize but the total js size is larger than minSize.

In this case move the module that contains that css into a existing result chunk, and re-check the maxSize

Align this logic in this pr

## Checklist


bigbossx:
LGTM~ nice job!

chenjiahan (approved):

PR #9345, opened 2025-02-17
chore(benchmark): add rspack benchcases 

h-a-n-a:
## Summary

With Rust Builder API, we can migrate some pure rust benchmarks in https://github.com/web-infra-dev/rspack-ecosystem-benchmark to codspeed. 

Since codspeed calculates cpu cycles, there's no need to scale up the benchmark. So we only preserve a little modules in the `1000` benchmark to speed up ci.

## Checklist


.github/workflows/reusable-build-bench.yml
```diff
@@ -24,7 +24,7 @@ permissions:
 jobs:
   bench:
     name: Bench
-    runs-on: ${{ fromJSON(inputs.runner) }}
+    runs-on: ${{ fromJSON(vars.LINUX_SELF_HOSTED_RUNNER_LABELS || '"ubuntu-latest"') }}
```
hardfist: it seems this shouldn't be hardcoded here

stormslowly (approved):

tasks/benchmark/benches/groups/bundle.rs
```diff
@@ -0,0 +1,39 @@
...
+
+  #[cfg(feature = "codspeed")]
+  group.sample_size(10);
+
+  let projects: Vec<(&'static str, CompilerBuilderGenerator)> = vec![
+    ("basic-react", Arc::new(basic_react::compiler)),
+    ("threejs", Arc::new(threejs::compiler)),
+  ];
+
+  // Codspeed can only handle to up to 500 threads by default
```
hardfist, line 25: any issue link or documentation about this limitation

hardfist (approved):

tasks/benchmark/benches/groups/bundle.rs
CPunisher: 1. codspeed runner uses valgrind: https://github.com/CodSpeedHQ/runner/blob/d26117ec30d474d50d719a7ae1fc512374b4567f/src/run/runner/valgrind/measure.rs#L109
2. search `--max-threads` in valgrind docs: https://valgrind.org/docs/manual/manual-core.html
3. It seems that there's no way to pass args from codspeed to valgrind (maybe we can submit a pr)

Issue #9336, opened 2025-02-17
[Bug]: When the favicon path contains Chinese characters, HtmlRspackPlugin may fail to load the file!


nuintun:
### System Info

System: Win11
Version: 1.2.3

### Details

When the favicon path contains Chinese characters, HtmlRspackPlugin may fail to load the file!

![Image](https://github.com/user-attachments/assets/f568498f-ad05-40dc-9cfc-af54c19273bf)

### Reproduce Steps

Set favicon path contains Chinese characters and run build.

PR #9348, opened 2025-02-17
fix: remove unnecessary path-to-URL conversion in html plugin

inottn:
## Summary
Fixes: #9336

I'm not sure why the path is being converted to a URL here. It doesn't appear to be required, and it could introduce issues.

## Checklist


LingyuCoder (approved):
LGTM, thanks

PR #9403, opened 2025-02-20
perf(loader-runner): memoize resolver creation for improved performance

### LINKED ISSUE #9399: [Bug]: performance bottleneck in loader-runner `resolver` creation
### System Info

macOS

### Details

I have a build that compiled a few thousand modules in 9 seconds.  About 5% of the build time is wasted in creating the loaderContext's resolve function (which is never called in my case).  The slow code is the first line of:

https://github.com/web-infra-dev/rspack/blob/b271e1bfedd9e2e0161936a289d1b9b6897bb7a7/packages/rspack/src/loader-runner/index.ts#L633-L636

Image

To confirm the issue, I tried hacking my installed index.js file (version 1.2.5) so that `resolver` was only created if called:

```js
  let resolver;
  loaderContext.resolve = function resolve2(context3, request, callback) {
    if (!resolver)
      resolver = compiler._lastCompilation.resolverFactory.get("normal");
    resolver.resolve({}, context3, request, getResolveContext(), callback);
  };
```

With the above change, the overall time for the `runLoaders` function decreased by roughly the 438ms shown above.

### Reproduce link

_No response_

### Reproduce Steps

Run a build with `RSPACK_PROFILE=JSCPU`, and then look in the timing information for the `runLoaders` function.

Currently:

Image

With the "fix" above:

Image

inottn:
## Summary
close #9399

## Checklist

### DIFF
--- a/packages/rspack/src/loader-runner/index.ts
+++ b/packages/rspack/src/loader-runner/index.ts
@@ -630,12 +630,16 @@ export async function runLoaders(
 		};
 	};
 
-	const resolver = compiler._lastCompilation!.resolverFactory.get("normal");
+	const getResolver = memoize(() => {
+		return compiler._lastCompilation!.resolverFactory.get("normal");
+	});
+
 	loaderContext.resolve = function resolve(context, request, callback) {
-		resolver.resolve({}, context, request, getResolveContext(), callback);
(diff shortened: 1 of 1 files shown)


h-a-n-a (commented):
Thanks!

h-a-n-a (approved):

PR #9419, opened 2025-02-21
fix(json): correctly escape unicode characters

### LINKED ISSUE #9415: [Bug]: Incorrect escape Unicode chars in json
### System Info

Rspack 1.2.4

### Details

If the imported json contains a unicode character, it is escaped twice, which leads to incorrect display in the application.
I think the problem is with one of the loaders.

Example json file:
```
{
  "test_key": "Some test message with unicode char" //  U+2028 before 'char'
}
```

```
// Production bundle built with rspack
// i=JSON.parse('{"test_key":"Some test message with unicode \\\\u2028char"}');

// Production bundle built with webpack
// i=JSON.parse('{"test_key":"Some test message with unicode \\u2028char"}');
```
Image

Image

### Reproduce link

https://codesandbox.io/p/github/Fausto95/react-native-web-rsbuild/csb-9n6wzh/draft/eloquent-waterfall

### Reproduce Steps

1. Open https://codesandbox.io/p/github/Fausto95/react-native-web-rsbuild/csb-9n6wzh/draft/eloquent-waterfall
2. `npm run start:web`
3. An example of the problematic text will be on the screen.

inottn:
## Summary
fix #9415

## Checklist


h-a-n-a (approved):
Thanks

dgaponov:
Thank you

PR #9426, opened 2025-02-21
chore: move modern module library finish_modules to finish_make

ahabhgk:
## Summary

finish_modules means the module graph (modules, connections, dependencies) are frozen and start to optimize (provided exports, infer async, etc.) based on the module graph, so any kind of change that affect these should be done before the finish_modules.

## Checklist


ahabhgk:
!eco-ci

ahabhgk:
Don't know why rsbuild hang in eco-ci, but it passes in my local

fi3ework (approved):

PR #9453, opened 2025-02-25
fix: detect module graph change right before build_chunk_graph

JSerFeng:
### Background

We have a way to skip re-building chunk graph, that is, when module graph has no change, no incoming outgoing changes, we can skip creating chunks for current compilation.

We detect module graph change in make phase before, which is not exactly accurate.

### Issue

Many optimizations are applied **after** make phase, for example, `innerGraph`, `sideEffects` and `usedExports`, they can change module graph as well. To solve this, we did a lot of hack to detect change according to many aspects, even `dependency.ids`, as it can affect how `sideEffects` optimized.

### Current solution

Now I move this detection later right before `build_chunk_graph`, and compare current outgoings with last compilation, in this stage, all optimizations are already applied.

### Long term solution

When incremental for `build_chunk_graph` is stable, we no longer need to detect the changes and skip creating chunks, as incremental build chunks is performant enough

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/watchCases/build-chunk-graph/skip-building-chunk-graph/test.config.js
+++ b/packages/rspack-test-tools/tests/watchCases/build-chunk-graph/skip-building-chunk-graph/test.config.js
@@ -1,0 +1,35 @@
+function assert(condition) {
+	if (!condition) {
+		throw new Error("Assertion failed for");
+	}
+}
+
+function should_rebuild(stats) {
+	return stats.includes("<t> rebuild chunk graph");
+}
+
+module.exports = {
+	checkStats(stepName, _, stats) {
+		switch (stepName) {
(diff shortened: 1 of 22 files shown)


packages/rspack-test-tools/tests/watchCases/build-chunk-graph/skip-building-chunk-graph/test.config.js
```diff
@@ -0,0 +1,33 @@
+// const { assert } = require("node:assert");
+function assert(condition) {
+	console.log(condition);
```
ahabhgk, line 3: Change to node:assert?

ahabhgk (commented):
LGTM, cc @jerrykingxyz

jerrykingxyz (approved):

PR #9454, opened 2025-02-25
docs: compiler.modifiedFiles and removedFiles

jerrykingxyz:
## Summary

This PR will add documentation for 'compiler.modifiedFiles' and 'compiler.removedFiles', and add warnings to 'compiler.run' to inform users to add them manually.

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #9455, opened 2025-02-25
feat: rebuild without modifiedFiles and removedFiles show warning

jerrykingxyz:
## Summary

User can trigger rspack rebuild via `compiler.watch` or manually call `compiler.run`. 

The watcher automatically adds `compiler.modifiedFiles` and `compiler.removedFiles`, but manually call `compiler.run` can sometimes forget to set them, which causes the rspack memory cache to not be invalidated correctly. This PR will add a warning for this situation.

## Checklist

### DIFF
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -702,16 +702,21 @@ class Compiler {
 				return callback?.(error);
 			}
 			if (!this.#initial) {
-				instance!.rebuild(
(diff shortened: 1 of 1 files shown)


h-a-n-a (approved):

chenjiahan (changes requested):
Should wait for https://github.com/web-infra-dev/rspack/pull/9454/files#r1970800468

PR #9464, opened 2025-02-26
fix: unify js assets info type

SyMind:
## Summary

fix: unit js assets info type

KnownAssetInfo and AssetInfo currently defined by napi.

## Checklist


packages/rspack/src/util/AssetInfo.ts
```diff
@@ -1,71 +1 @@
...
-			extras,
-			javascriptModule,
-			sourceFilename,
-			copied
-		};
-	}
-}
-
-export { JsAssetInfo };
+export type { AssetInfo } from "@rspack/binding";
```
chenjiahan, line 1: Should we keep this module? It only contains one line of code.

chenjiahan (commented):
Related documentation should be updated too:

image

crates/node_binding/binding.d.ts
```diff
@@ -1305,6 +1266,40 @@ export interface JsTap {
...
+  /** the value(s) of the full hash used for this asset */
+  fullhash?: string | Array<string>
+  /** the value(s) of the chunk hash used for this asset */
+  chunkhash?: string | Array<string>
+  /**
+   * the value(s) of the module hash used for this asset
+   * the value(s) of the content hash used for this asset
+   */
+  contenthash?: string | Array<string>
+  sourceFilename?: string
```
chenjiahan, line 1283: Why the comments for `sourceFilename` is not generated here

SyMind: should use `///`, i will fix it.

chenjiahan (approved):

PR #9471, opened 2025-02-26
feat: add `module._constructorName`

ahabhgk:
## Summary

Add `_constructorName` for module, some plugin need to have different behavior for different kind of module, add `_constructorName` to support that

PS: I'm not sure is there a better solution for this, this PR maybe a temp solution or maybe not, so I use `_constructorName` with `@internal` JSDoc for now

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -231,6 +231,7 @@ export declare class JsExportsInfo {
 }
 
 export declare class JsModule {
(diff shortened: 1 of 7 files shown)


hardfist (approved):

jerrykingxyz (approved):

ahabhgk:
This need more discussion before merge main, for now keep it in more-js-api-v2 branch only

PR #9475, opened 2025-02-26
feat: compiler.run add options to set modifiedFiles and removedFiles

jerrykingxyz:
## Summary

Rspack `compiler.run` has the same functionality as webpack, the first call will cold start the build, subsequent calls will rebuild using the memory cache. The difference is that Rspack relies on `compiler.modifiedFiles` and `compiler.removedFiles` to invalidate the memory cache.

This PR will add a params to allow user to set `compiler.modifiedFiles` and `compiler.removedFiles` when call run.

## Checklist

### DIFF
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -424,6 +424,7 @@ class Compiler {
 
 	/**
 	 * @param callback - signals when the call finishes
+	 * @param options - additional data like modifiedFiles, removedFiles
 	 */
 	run(
 		callback: liteTapable.Callback<Error, Stats>,
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -425,10 +425,19 @@ class Compiler {
 	/**
 	 * @param callback - signals when the call finishes
 	 */
-	run(callback: liteTapable.Callback<Error, Stats>) {
+	run(
+		callback: liteTapable.Callback<Error, Stats>,
+		options: {
+			modifiedFiles?: ReadonlySet<string>;
+			removedFiles?: ReadonlySet<string>;
+		} = {}
+	) {
 		if (this.running) {
 			return callback(new ConcurrentCompilationError());
 		}
+
+		this.modifiedFiles = options.modifiedFiles;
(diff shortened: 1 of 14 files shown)


packages/rspack/src/Watching.ts
```diff
@@ -261,20 +262,15 @@ export class Watching {
 			this.compiler.contextTimestamps = contextTimeInfoEntries;
 		}
 
-		const modifiedFiles = (this.compiler.modifiedFiles =
-			this.#collectedChangedFiles);
-		const deleteFiles = (this.compiler.removedFiles =
-			this.#collectedRemovedFiles);
+		this.compiler.modifiedFiles = this.#collectedChangedFiles;
```
chenjiahan, line 265: Should we remove `this.compiler.modifiedFiles = this.#collectedChangedFiles`? I guess this is still required for `compiler.watch`

packages/rspack/src/Compiler.ts
```diff
@@ -425,10 +425,19 @@ class Compiler {
...
+		options: {
+			modifiedFiles?: ReadonlySet<string>;
+			removedFiles?: ReadonlySet<string>;
+		} = {}
+	) {
 		if (this.running) {
 			return callback(new ConcurrentCompilationError());
 		}
+
+		this.modifiedFiles = options.modifiedFiles;
```
chenjiahan, line 439: Should we assign `undefined` to `this.modifiedFiles`? 

```suggestion
		if (options.modifiedFiles) {
				this.modifiedFiles = options.modifiedFiles;
		}
```

packages/rspack/src/Compiler.ts
```diff
@@ -425,10 +425,19 @@ class Compiler {
 	/**
 	 * @param callback - signals when the call finishes
 	 */
-	run(callback: liteTapable.Callback<Error, Stats>) {
+	run(
+		callback: liteTapable.Callback<Error, Stats>,
+		options: {
+			modifiedFiles?: ReadonlySet<string>;
```
chenjiahan, line 431: Can we add some JSDoc like `* @param callback - signals when the call finishes`

chenjiahan (approved):

PR #9476, opened 2025-02-26
fix: entry dependency passed to addInclude should be used in module graph

SyMind:
## Summary

fix: entry dependency passed to addInclude should be used in module graph

```
const dependency = EntryPlugin.createDependency(path.resolve(__dirname, "foo.js"));
compilation.addInclude(compiler.context, dependency, {}, () => {
    const module = compilation.moduleGraph.getModule(dependency); // should get the module
});
```

rewrite `Symbol.hasInstance` for `Dependency` class to impl `entryDependency instanceof Dependency`.

## Checklist


crates/node_binding/src/dependencies/entry_dependency.rs
```diff
@@ -0,0 +1,66 @@
+use std::any::TypeId;
+
+use napi::{Either, Env, JsString};
+
+#[napi]
+#[repr(C)]
+pub struct EntryDependency {
+  #[allow(dead_code)]
+  type_id: TypeId,
```
hardfist, line 9: what's this for?

SyMind: napi-rs uses `ValidateNapiValue::validate` to check if the value `napi_unwrap` from JavaScript object is of a specific type. In `ValidateNapiValue::validate`, it calls `napi_get_reference_value` to get the class constructor and `napi_instanceof` to check the type.

This method is slow and can degrade performance. Therefore, we use `#[repr(C)]` to ensure the order of struct fields, and all dependencies include `type_id` as the first field. This way, we can determine the type using a raw pointer.

crates/node_binding/src/dependencies/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    ))?;
+
+    let type_id = unknown_tagged_object as *const TypeId;
+    if *type_id == TypeId::of::<Dependency>() {
+      let tagged_object = &*(unknown_tagged_object as *mut Dependency);
+      Ok(DependencyObject(Some(tagged_object.dependency_id)))
+    } else if *type_id == TypeId::of::<EntryDependency>() {
+      let tagged_object = &*(unknown_tagged_object as *mut EntryDependency);
+      Ok(DependencyObject(tagged_object.dependency_id))
+    } else {
```
hardfist, line 49: I'm not sure whether we can ensure safety here

hardfist: since we already know this has performance cost, maybe we shouldn't include this optimization in this pr

crates/node_binding/src/dependencies/mod.rs
```diff
@@ -0,0 +1,60 @@
...
+    let js_object = Object::from_napi_value(env, napi_val)?;
+
+    let mut unknown_tagged_object: *mut c_void = ptr::null_mut();
+    check_status!(sys::napi_unwrap(
+      env,
+      js_object.raw(),
+      &mut unknown_tagged_object,
+    ))?;
+
+    let type_id = unknown_tagged_object as *const TypeId;
```
h-a-n-a, line 46: What is the type of `unknown_tagged_object`? Is this safe to cast to `TypeId`?

SyMind: EntryDependency and Dependency all first field is TypeId, and use repr(C)

SyMind: There are comments to explain this.

h-a-n-a: I thought it was `TaggedObject`. Didn't see where it was wrapped with `napi_wrap`.

SyMind: @h-a-n-a In ClassInstrance::to_instance

crates/node_binding/src/dependencies/entry_dependency.rs
hardfist, line 9: what happens if we forget to use repr(c) and add type_id for some dependency? can it be checked in compile time or will cause segfault during runtime?

SyMind: this is a UB. napi-rs currently has same problem. I will fix this problem in napi-rs.

https://github.com/napi-rs/napi-rs/issues/2484

hardfist (approved):

PR #9485, opened 2025-02-27
docs: fix the reference to process asset stages in examples

xbzhang2020:
## Summary

When using process assets examples provided by the document, I encountered the following TS error: **Property 'PROCESS_ASSETS_STAGE_ADDITIONS' does not exist on type 'Compilation'.** 

Actually, asset processing stages are exported by `compiler.webpack.Compilation` instead of `compilation`. See [Webpack docs](https://webpack.js.org/api/compilation-hooks/#processassets).

Screenshot 2025-02-27 at 09 40 21

## Checklist


chenjiahan (approved):
Thank you!

PR #9492, opened 2025-02-27
perf: try reducec jmalloc arena number to 1

stormslowly:
## Checklist


stormslowly:
not so useful

PR #9511, opened 2025-02-28
fix: enhance swc error message

GiveMe-A-Name:
## Summary

With swc_core version upgrades, Swc Wasm plugin maybe have a break change. We should add more detail message help user to select corret plugin version. 

## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/compiler.rs
+++ b/crates/rspack_loader_swc/src/compiler.rs
@@ -374,7 +374,7 @@ impl SwcCompiler {
         let help_msg = formatdoc!{"
           The version of the SWC Wasm plugin you're using might not be compatible with `builtin:swc-loader`.
           The `swc_core` version of the current `rspack_core` is {swc_core_version}. 
-          Please check the `swc_core` version of SWC Wasm plugin to make sure these versions are within the compatible range.
+          Please visit https://plugins.swc.rs/versions/from-core/{swc_core_version} to verify the swc_core version of the SWC Wasm plugin and ensure it falls within the compatible range.
           See this guide as a reference for selecting SWC Wasm plugin versions: https://rspack.dev/errors/swc-plugin-version"};
         let report: Report = MietteDiagnostic::new(format!("{}{}",error_msg,help_msg)).with_code("Builtin swc-loader error").into();
         Err(report)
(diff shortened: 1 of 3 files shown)


crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -374,7 +374,7 @@ impl SwcCompiler {
         let help_msg = formatdoc!{"
           The version of the SWC Wasm plugin you're using might not be compatible with `builtin:swc-loader`.
           The `swc_core` version of the current `rspack_core` is {swc_core_version}. 
-          Please check the `swc_core` version of SWC Wasm plugin to make sure these versions are within the compatible range.
+          Please visit https://plugins.swc.rs/versions/from-core/{swc_core_version} to verify the `swc_core` version of the SWC Wasm plugin and ensure it falls within the compatible range.
```
h-a-n-a, line 377: Should strip the prefix `=` if available.

GiveMe-A-Name: Done~

crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -370,11 +370,12 @@ impl SwcCompiler {
           None => "unknown error"
         };
         let swc_core_version = env!("RSPACK_SWC_CORE_VERSION");
+        let swc_core_version_strip_prefix = swc_core_version.replace("=", "");
         // FIXME: with_help has bugs, use with_help when diagnostic print is fixed
         let help_msg = formatdoc!{"
           The version of the SWC Wasm plugin you're using might not be compatible with `builtin:swc-loader`.
           The `swc_core` version of the current `rspack_core` is {swc_core_version}. 
-          Please check the `swc_core` version of SWC Wasm plugin to make sure these versions are within the compatible range.
+          Please visit https://plugins.swc.rs/versions/from-core/{swc_core_version_strip_prefix} to verify the `swc_core` version of the SWC Wasm plugin and ensure it falls within the compatible range.
```
chenjiahan, line 378: We have already provided the https://rspack.dev/errors/swc-plugin-version link, and we should avoid adding third-party links to our logs.

GiveMe-A-Name: It's not visualized. User need visit https://rspack.dev/errors/swc-plugin-version then open another link to see this page.

SWC is a third-part lib in rspack, but we are work cooperate closely.

chenjiahan: 1. We cannot guarantee that `https://plugins.swc.rs/versions/from-core/` will still be accessible in several years
2. Not all SWC plugins are listed in the `https://plugins.swc.rs`, when user see the SWC error message, it is possible that he is using an internal plugin or a plugin that is not included.

GiveMe-A-Name: Got it

chenjiahan: If necessary we can add more details or guide to https://rspack.dev/errors/swc-plugin-version 😄

PR #9526, opened 2025-03-03
fix(html): include `innerHTML` content in rendered asset tags

faultyserver:
## Summary

I was trying to implement an `HtmlRspackPlugin` addition using the `alterAssetTags` hook, and for my use case I needed to set `innerHTML` on the generated tag (to get something like `<script>console.log("Hello")</script>`). After a lot of confusion about why things weren't showing up as I expected and instead got empty tags, I went to look at the implementation and found that [`HtmlPluginTag` was always setting `children` to an empty vector](https://github.com/web-infra-dev/rspack/blob/39f2bfba38266f72ea9e2dcdaacffe5a23cc322f/crates/rspack_plugin_html/src/tag.rs#L282).

This PR fixes that to render the `innerHTML` property of the tag there as a plain `Text` node if it's present. I also tried to add an end to end test for this, but wasn't able to get them running locally immediately, so hopefully CI will indicate if it's correct, otherwise I'll get them set up and running to try it out.

## Checklist

### DIFF
--- a/crates/rspack_plugin_html/src/tag.rs
+++ b/crates/rspack_plugin_html/src/tag.rs
@@ -8,7 +8,7 @@ use serde::{
   Deserialize, Deserializer, Serialize, Serializer,
 };
 use swc_core::{atoms::Atom, common::DUMMY_SP};
-use swc_html::ast::{Attribute, Element, Namespace};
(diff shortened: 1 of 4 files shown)


LingyuCoder (approved):
LGTM, thanks

PR #9533, opened 2025-03-03
feat(memory-allocator): use mimalloc-v3 on macOS and bump mimalloc to v2.1.7 on other platforms

h-a-n-a:
## Summary

1. Bumped mimalloc to v3 for macOS. 

According to my [investigation](https://github.com/microsoft/mimalloc/issues/1025#issuecomment-2693504806), although memory is still growing, upgrading mimalloc to v3 will greatly mitigate the memory bloating issue in development mode. That is to say, memory growing is slower than it was on mimalloc-v2 for each time the compiler rebuilds:

**With mimalloc dev3-bin:**
```
hyperfine --warmup 2 --runs 5 './target/release/mimalloc-test'
Benchmark 1: ./target/release/mimalloc-test
  Time (mean ± σ):      9.230 s ±  0.157 s    [User: 28.923 s, System: 13.791 s]
  Range (min … max):    9.052 s …  9.450 s    5 runs
```

**With mimalloc dev2-bin (rspack main):**
```
hyperfine --warmup 2 --runs 5 './target/release/mimalloc-test'
Benchmark 1: ./target/release/mimalloc-test
  Time (mean ± σ):      9.360 s ±  0.126 s    [User: 29.537 s, System: 14.543 s]
  Range (min … max):    9.216 s …  9.525 s    5 runs
```

Case: building 10000 modules and rebuild 10 times, for each time a new `console.log` will be written to the entry module.

2. Bumped mimalloc to v2.1.7 for non-macOS platforms.

This is the latest version of mimalloc-v2. We don't want to rollout too much change at once. So we're staying at v2 at least for now.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2683,16 +2683,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "libmimalloc-sys-rspack"
-version = "0.1.40"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a7997f74815cc4b73f7b1897f13aed74d8ea49e7d86192d9c08779edfc9e1d9f"
-dependencies = [
- "cc",
- "libc",
-]
-
-[[package]]
 name = "libredox"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


chenjiahan (dismissed):

h-a-n-a:
Downgrade mimalloc v2.1.9 to v2.1.7 as windows not being able to compile.

chenjiahan (approved):

PR #9535, opened 2025-03-03
fix: should give js entry dependency id when hit cache

SyMind:
## Summary

When hit the cache for js dependency, should give the dependency id to js object.

## Checklist

### DIFF
--- a/crates/node_binding/src/compilation/mod.rs
+++ b/crates/node_binding/src/compilation/mod.rs
@@ -736,6 +736,7 @@ impl JsCompilation {
         let dependency = if let Some(map) = include_dependencies_map.get(&js_dependency.request)
           && let Some(dependency) = map.get(&options)
         {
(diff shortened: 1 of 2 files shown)


ahabhgk (approved):

crates/node_binding/src/compilation/mod.rs
```diff
@@ -736,6 +736,7 @@ impl JsCompilation {
         let dependency = if let Some(map) = include_dependencies_map.get(&js_dependency.request)
           && let Some(dependency) = map.get(&options)
         {
+          js_dependency.dependency_id = Some(*dependency.id());
```
hardfist, line 739: what will this affect?

SyMind: cause bug when pass to method for module graph

SyMind: you can see the test case

PR #9571, opened 2025-03-06
test: pass more webpack test after ModuleInfoHeaderPlugin finished

stormslowly:
## Summary
enable 3 webpack  test cases , because we add module info header plugin.

## Checklist


chenjiahan (commented):
Please add a description for this change 😄

PR #9587, opened 2025-03-07
fix(core): remove blocking to avoid deadlock for multi compiler

hardfist:
## Summary
FutureResults will call blocking_in_place to support non `'static` future parallel, but if there're multi compilers call it multi times, it will cause all tokio runtime working blocking and other  spawned task can't be scheduled, which will cause deadlock.

Unfortunately there're no safe way to implement parallel future which is proved in https://tmandry.gitlab.io/blog/posts/2023-03-01-scoped-tasks/

so we need choose use unsafe code or sacrifice parallel(which means sacrifice performance), for us performance is important so I choose use unsafe code.

## Checklist

### DIFF
--- a/crates/rspack_core/Cargo.toml
+++ b/crates/rspack_core/Cargo.toml
@@ -8,6 +8,7 @@ version     = "0.2.0"
 [dependencies]
 anymap = { workspace = true }
 async-recursion = { workspace = true }
+async-scoped = { workspace = true }
 async-trait = { workspace = true }
 bitflags = { workspace = true }
 cow-utils = { workspace = true }
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -1055,6 +1055,7 @@ impl Compilation {
     } else {
       self.chunk_by_ukey.keys().copied().collect()
     };
+    // SAFETY: await immediately and trust caller to poll future entirely
     let (_, results) = unsafe {
       async_scoped::TokioScope::scope_and_collect(
         |s: &mut TokioScope<'_, Result<(ChunkUkey, ChunkRenderResult)>>| {
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -1083,7 +1083,8 @@ impl Compilation {
     .await;
     let mut chunk_render_results: UkeyMap<ChunkUkey, ChunkRenderResult> = Default::default();
     for result in results {
-      let item: std::result::Result<(ChunkUkey, ChunkRenderResult), _> = result.unwrap();
+      let item: std::result::Result<(ChunkUkey, ChunkRenderResult), _> =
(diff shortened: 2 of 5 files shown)


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1054,31 +1055,39 @@ impl Compilation {
...
-          },
-        ))
-      })
-      .collect::<FuturesResults<Result<_>>>();
-    let chunk_render_results = chunk_render_results
-      .into_inner()
-      .into_iter()
-      .collect::<Result<UkeyMap<_, _>>>()?;
-
+    let (_, results) = unsafe {
```
h-a-n-a, line 1058: Missing `SAFETY` comment here

crates/rspack_core/src/compiler/mod.rs
```diff
@@ -343,36 +344,36 @@ impl Compiler {
...
-        // collect version info to new_emitted_asset_versions
-        if self
-          .options
-          .experiments
-          .incremental
-          .contains(IncrementalPasses::EMIT_ASSETS)
-        {
-          new_emitted_asset_versions.insert(filename.to_string(), asset.info.version.clone());
-        }
+    let _ = unsafe {
```
h-a-n-a, line 347: Here as well

crates/rspack_core/src/compiler/mod.rs
```diff
@@ -343,36 +344,37 @@ impl Compiler {
...
-          .options
-          .experiments
-          .incremental
-          .contains(IncrementalPasses::EMIT_ASSETS)
-        {
-          new_emitted_asset_versions.insert(filename.to_string(), asset.info.version.clone());
-        }
+    // SAFETY: await immediately, trust caller to handle the future correctly
+    let _ = unsafe {
+      TokioScope::scope_and_collect(|s| {
```
h-a-n-a, line 349: Are we supposed to switch over to our custom scoped future impl?

crates/rspack_core/Cargo.toml
```diff
@@ -8,6 +8,7 @@ version     = "0.2.0"
 [dependencies]
 anymap = { workspace = true }
 async-recursion = { workspace = true }
+async-scoped = { workspace = true }
```
h-a-n-a, line 11: Supposed to be removed?

hardfist: not yet, still a case use this and i need to discuss further about @quininer how to migrate all  exsiting code to future_scope

crates/rspack_core/src/compiler/mod.rs
hardfist: yeah and async-scoped doesn't support wasm 😭

crates/rspack_core/Cargo.toml
hardfist: i need to remove this to support wasm

hardfist: removed

SyMind (approved):

PR #9608, opened 2025-03-10
fix: incorrect type signatures on Hash's digest and update methods

### LINKED ISSUE #9484: [Bug]: incorrect type signatures on Hash's digest and update methods
### System Info

macOS

### Details

The `Hash` class' digest and update methods are incorrect.  They are ambiguous and should use overloads like the node type declarations do.

The existing signatures are:

https://github.com/web-infra-dev/rspack/blob/2d10e9a0e21e2e56a9f83fe9205a838936fe60de/packages/rspack/src/util/hash/index.ts#L21-L23

https://github.com/web-infra-dev/rspack/blob/2d10e9a0e21e2e56a9f83fe9205a838936fe60de/packages/rspack/src/util/hash/index.ts#L32-L34

Using overloaded signatures makes it clear when the `inputEncoding` is required, and when the return value will be a string vs. buffer.  The correct signatures would be something like:

```ts
update(data: string, inputEncoding: string): this;
update(data: Buffer): this;
update(data: string | Buffer, inputEncoding?: string): this {
...

digest(): Buffer;
digest(encoding: string): string;
digest(encoding?: string): string | Buffer {
...
```

### Reproduce link

_No response_

### Reproduce Steps

In a loader, do something like:

```ts
const hash = this.utils.createHash('md5').update(buffer).digest('hex');
```

The call to update suggests that a second param can be passed, but this isn't true for buffers.  And the type of `hash` above is ambiguously `string | Buffer`, when it is guaranteed to be `string` since an encoding was provided.

hulin32:
## Summary

fix #9484

## Checklist


LingyuCoder:
LGTM, cc @SyMind

LingyuCoder (approved):
I will merge this PR and refactor the hashing to caluldate hash in Rust to improve performance in the future. Thanks for your contribution.

PR #9610, opened 2025-03-10
docs: deprecate `sri` option of HTML plugin

LingyuCoder:
## Summary

The `sri` option of html rspack plugin only support html tags and not support dynamic import chunk loading. The full-featured SRI has been supported by `SubresourceIntegrityPlugin`. So the `sri` option of html rspack should be deprecated.

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/html-plugin/options.ts
+++ b/packages/rspack/src/builtin-plugin/html-plugin/options.ts
@@ -90,7 +90,7 @@ export type HtmlRspackPluginOptions = {
 	chunksSortMode?: "auto" | "manual";
 
 	/**
-	 * The SRI hash algorithm, disabled by default.
+	 * Configure the SRI hash algorithm, which is disabled by default.
 	 * @deprecated Use `experiments.SubresourceIntegrityPlugin` instead.
 	 */
 	sri?: "sha256" | "sha384" | "sha512";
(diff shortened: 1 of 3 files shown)


website/docs/en/plugins/rspack/html-rspack-plugin.mdx
```diff
@@ -274,7 +274,8 @@ type HtmlRspackPluginOptions = {
       name: '`sri`',
       type: '`"sha256"" | "sha384"" | "sha512"" | undefined`',
       default: '`undefined`',
-      description: 'The SRI hash algorithm, disabled by default.',
+      description:
+        '<p>**Deprecated**: use [`SubresourceIntegrityPlugin`](./subresource-integrity-plugin) instead.</p><p>The SRI hash algorithm, disabled by default. </p>',
```
chenjiahan, line 278:
```suggestion
        '<p>**Deprecated**: use [`SubresourceIntegrityPlugin`](./subresource-integrity-plugin) instead.</p><p>Configure the SRI hash algorithm, which is disabled by default. </p>',
```

packages/rspack/src/builtin-plugin/html-plugin/options.ts
```diff
@@ -89,7 +89,10 @@ export type HtmlRspackPluginOptions = {
 	 */
 	chunksSortMode?: "auto" | "manual";
 
-	/** The SRI hash algorithm, disabled by default. */
+	/**
+	 * The SRI hash algorithm, disabled by default.
```
chenjiahan, line 93:
```suggestion
	 * Configure the SRI hash algorithm, which is disabled by default.
```

chenjiahan (approved):

Issue #8816, opened 2024-12-23
[Feature]: Implement webpack extends property


KuSh:
### What problem does this feature solve?

Migrating a project that currently use an extends property requires a rewrite of the configuration using webpack-merge (as documented [here](https://rspack.dev/config/index#merge-configurations))

Implementing the extends property will make migration transparent
It can also be an opportunity to rewrite a performance oriented builtin webpack-merge in rust

### What does the proposed API of configuration look like?

Webpack documentation: https://webpack.js.org/configuration/extending-configurations/

```js
module.exports = {
  extends: path.resolve(__dirname, './base.rspack.config.js'),
  entry: './src/index.js',
  output: {
    path: path.resolve(__dirname, 'dist'),
    filename: 'bundle.js',
  },
};
```

cbbfcd:
Let me try.

KuSh:
bump

hardfist:
close since it's landed in #9617

PR #9617, opened 2025-03-10
feat: add extends property in config options

hulin32:
## Summary

[#8816](https://github.com/web-infra-dev/rspack/issues/8816) implemented webpack config extend feature.
ref: https://webpack.js.org/configuration/extending-configurations/

## Checklist


LingyuCoder:
Need to add a hint in the type annotation and documentation that this configuration is only available in the CLI. Just like webpack:

image

packages/rspack/src/config/types.ts
```diff
@@ -2730,6 +2730,11 @@ export type RspackOptions = {
 	 * An array of dependencies required by the project.
 	 */
 	dependencies?: Dependencies;
+	/**
+	 * Configuration files to extend from. The configurations are merged from right to left,
+	 * with the rightmost configuration taking precedence.
```
LingyuCoder, line 2735:
```suggestion
	 * with the rightmost configuration taking precedence (only works when using @rspack/cli)..
```

hulin32: thx for reviewing

PR #9626, opened 2025-03-11
refactor: modify `module.code_generation` to async fn

LingyuCoder:
## Summary

Modify `module.code_generation` to async function to prevent using `block_on` during code generation.

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -870,7 +870,7 @@ impl Compilation {
   }
 
   #[instrument("Compilation:code_generation", skip_all)]
-  fn code_generation(&mut self, modules: IdentifierSet) -> Result<()> {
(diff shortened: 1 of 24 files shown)


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -930,34 +935,36 @@ impl Compilation {
...
-                  );
-                  codegen_res
-                })
-            }),
-        )
-      })
-      .collect::<Vec<_>>();
-    for (module, (codegen_res, runtimes, from_cache)) in results {
+
+    let results = join_all(jobs.into_iter().map(|job| async {
```
hardfist, line 939: this change parallel to concurrency which seems will cause performance regression consider code_gen is cpu heavy task

LingyuCoder:
Duplicated with https://github.com/web-infra-dev/rspack/pull/9217, so this pr will be closed

PR #9564, opened 2025-03-05
test(webpack): Sync new webpack tests

KuSh:
## Summary

Sync missing tests or missing files in existing webpack tests.
Every failing tests above obvious corrections are disabled. It still adds a meaningful coverage.

My aim is to also sync existing tests but that was a first good batch

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #9639, opened 2025-03-12
test(webpack): Sync new webpack cases tests

KuSh:
## Summary

Sync missing tests or missing files in existing webpack cases tests.
Every failing tests above obvious corrections are disabled. It still adds a meaningful coverage.

My aim is to also sync existing tests but that was a first good batch

This is split version of #9564 

## Checklist


chenjiahan (approved):
Thank you!

PR #9666, opened 2025-03-13
feat: implement `module.generator.json.JSONParse`

colinaaa:
## Summary

For large `.json` modules, webpack will generate `JSON.parse` by default for
better performance. But there are some circumstances that `JSON.parse` is not a
good choice (e.g.: when doing AOT compilation).

Thus, a new generator option `module.generator.json.JSONParse` is added to
disable generating `JSON.parse` for `.json` module.

The default value is kept as `true` and can be opt-out by custom rules.

This is a port of https://github.com/webpack/webpack/pull/19320.

## Checklist

- [x] **Tests updated** (or not required).
- [x] **Documentation updated** (or not required).


KuSh (commented):
Why not integrating associated webpack tests to the suite?

ahabhgk (dismissed):

packages/rspack-test-tools/tests/configCases/json/generator-json-parse-false/index.js
```diff
@@ -0,0 +1,24 @@
+it("should avoid JSON.parse", () => {
+	const JSONParse = jest.spyOn(JSON, 'parse');
+	JSONParse.mockClear();
+
+	const data = require('./data.json');
```
ahabhgk, line 5: You can move this to webpack-test since the webpack PR is already merged

ahabhgk (approved):

PR #9685, opened 2025-03-14
fix: compiler js object cannot be gc

SyMind:
## Summary

Compiler is unable to GC due to some `ThreadsafeFunctions`, which include
1. hooks callback
2. run_loaders

The solution is to strictly drop these `ThreadsafeFunctions` at the end of the build.

## Checklist


crates/rspack_util/src/defer.rs
```diff
@@ -0,0 +1,29 @@
...
+    f();
+  }
+}
+
+#[inline(always)]
+pub fn defer(f: impl FnOnce() + 'static) -> DeferGuard {
+  DeferGuard(ManuallyDrop::new(Box::new(f)))
+}
+
+pub struct AsyncDeferGuard(ManuallyDrop<Box<dyn FnOnce() + Send>>);
```
quininer, line 17: Maybe `SendableDeferGuard` would be a better name.

crates/rspack_util/src/defer.rs
```diff
@@ -0,0 +1,29 @@
+use std::mem::ManuallyDrop;
+
+pub struct DeferGuard(ManuallyDrop<Box<dyn FnOnce()>>);
```
quininer, line 3: Box doesn't seem necessary?

crates/node_binding/src/compilation/mod.rs
```diff
@@ -595,37 +596,44 @@ impl JsCompilation {
...
+                .map(|module| ModuleObject::with_ref(module.as_ref(), compiler_id))
+                .collect::<Vec<_>>()
+            },
+          )
+          .await
+          .to_napi_result()?;
+
+        Ok(modules)
+      },
+      || drop(defer_guard),
```
quininer, line 628: Can we write `drop(reference)` directly here without going through `defer_guard`?

quininer (changes requested):

crates/rspack_util/src/defer.rs
SyMind: image

`FnOnce` should use `ManuallyDrop::take` move ownership.

quininer: we can keep the generics here

quininer: see https://docs.rs/scopeguard/latest/scopeguard/

crates/rspack_util/src/defer.rs
```diff
@@ -0,0 +1,29 @@
...
+
+impl Drop for SendableDeferGuard {
+  fn drop(&mut self) {
+    let f = unsafe { ManuallyDrop::take(&mut self.0) };
+    f();
+  }
+}
+
+#[inline(always)]
+pub fn async_defer(f: impl FnOnce() + Send + 'static) -> SendableDeferGuard {
```
quininer, line 27: same name

quininer (approved):

PR #9692, opened 2025-03-17
chore: use ci profile to run rust bench

hardfist:
## Summary
run 3 release build for rust bench is too slow and unnecessary which makes bench job is the most time-consuming job in CI(takes 13min to build),
change profile to ci to speedup which takes only 6min to build

## Checklist


LingyuCoder (approved):

hardfist:
not  necessary now, since it's not the bottleneck now

PR #9701, opened 2025-03-17
refactor: async `runtime_module.generate`

LingyuCoder:
## Summary

Modify `runtime_module.generate` and `hooks.render_chunk` to async

## Checklist


crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1940,7 +1940,7 @@ impl Compilation {
       for runtime_module_identifier in self.chunk_graph.get_chunk_runtime_modules_iterable(chunk) {
         let runtime_module = &self.runtime_modules[runtime_module_identifier];
         let mut hasher = RspackHash::from(&self.options.output);
-        runtime_module.update_hash(&mut hasher, self, None)?;
+        runtime_module.get_hash_async(self).await?.hash(&mut hasher);
```
LingyuCoder, line 1943: Need to generate final code in `update_hash` of `runtime_module` if not exists, but `update_hash` of `module` trait is sync, so we need another method to be async to handle this.

packages/rspack-test-tools/tests/__snapshots__/NewCodeSplitting-stats-output.test.js.snap
```diff
@@ -58,7 +58,7 @@ runtime modules xx KiB
     [no exports]
     [used exports unknown]
   
-Rspack compiled successfully (27eddb11e03f28d2)
+Rspack compiled successfully (b647003278ae1d01)
```
LingyuCoder, line 61: The fullhash changes because we use new method to calculate a string hash result of runtime module instead of `update_hash`.

hardfist (approved):

Issue #8957, opened 2025-01-07
API mismatch between CopyRspackPlugin transform and CopyWebpackPlugin transform


patricksevat:
Hi all,

First, thanks for creating Rspack!

I'm trying to from Webpack to Rspack and encountered an issue with CopyRspackPlugin's compatibility with CopyWebpackPlugin.

The gist is that in CopyWebpackPlugin, [`pattern.transform`](https://github.com/webpack-contrib/copy-webpack-plugin?tab=readme-ov-file#transform) accepts an object whereas CopyRspackPlugin does not (Note: this [has been introduced in CopyWebpackPlugin@7](https://github.com/webpack-contrib/copy-webpack-plugin/releases/tag/v7.0.0))

```typescript
// CopyWebpackPlugin
type transform =
  | {
      transformer: (input: string, absoluteFilename: string) => string | Buffer;
      cache?: boolean | TransformerCacheObject | undefined;
    }
  | ((input: string, absoluteFilename: string) => string | Buffer);
```

vs

```typescript
// CopyRspackPlugin
type Transform = (
  input: Buffer,
  absoluteFilename: string,
) => string | Buffer | Promise<string> | Promise<Buffer>;
```

This lead to a cryptic error:

```
 #  rspack-node[77491]: napi_status napi_release_threadsafe_function(napi_threadsafe_function, napi_threadsafe_function_release_mode) at ../src/node_api.cc:1398
  #  Assertion failed: (func) != nullptr
```

I see two possible solutions:

1. Accept the API difference, and update the [migration docs from Webpack to Rspack relating CopyWebpackPlugin](https://rspack.dev/guide/migration/webpack#copy-webpack-plugin)
  a. possibly extended with some configuration validation which would throw a more informative error
2. Update CopyRspackPlugin to align the APIs

Minimal reproduction example: https://github.com/rspack-contrib/rspack-examples/compare/main...patricksevat:rspack-examples:feat/copy-webpack-plugin-transform-obj-config

hardfist:
I think we can align the copyWebpackPlugin api and support transformer

PR #9714, opened 2025-03-18
feat(copyRspackPlugin): align transform api with webpack plugin

fireairforce:
## Summary

closes: #8957 

Api align need something more to support:

- [x] transform api align with webpack
- [ ] copyRspackPlugin support cache feature
   - [ ]  Rspack compilation support `get_cache()` function

I will support cache feature after `compilation` support `compilation.get_cache()` function, so this pr will not execute it.

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -1558,7 +1558,7 @@ export interface RawCopyPattern {
    * @default false
    */
   copyPermissions?: boolean
-  transform?: (input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>
+  transform?: { transformer: (input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  } | ((input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>)
 }
 
 export interface RawCopyRspackPluginOptions {
@@ -2360,7 +2360,11 @@ export interface RawSwcJsMinimizerRspackPluginOptions {
 
 export interface RawToOptions {
   context: string
-  absoluteFilename: string
+  absoluteFilename?: string
+}
+
+export interface RawTransformWithCacheOptions {
+  transformer: TransformerFn
 }
 
 export interface RawTrustedTypes {
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -2363,7 +2363,7 @@ export interface RawToOptions {
   absoluteFilename?: string
 }
 
-export interface RawTransformWithCacheOptions {
+export interface RawTransformOptions {
   transformer: TransformerFn
 }
 
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -2364,7 +2364,7 @@ export interface RawToOptions {
 }
 
 export interface RawTransformOptions {
-  transformer: TransformerFn
+transformer: { transformer: (input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  }
 }
 
 export interface RawTrustedTypes {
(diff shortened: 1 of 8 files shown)


crates/node_binding/binding.d.ts
```diff
@@ -13,33 +13,33 @@ export type RawLazyCompilationTest = RegExp | ((module: Module) => boolean);
...
 export interface Module {
-	readonly type: string;
-	get context(): string | undefined;
-	get layer(): string | undefined;
-	get factoryMeta(): JsFactoryMeta | undefined
-	get useSourceMap(): boolean;
-	get useSimpleSourceMap(): boolean;
-	buildInfo: Record<string, any>;
-	buildMeta: Record<string, any>;
+  readonly type: string;
```
hardfist, line 16: seems your format is not same

fireairforce: i will update this later, i finish the api align and will add some test case for this~

hardfist:
cache is tricky and not compatible with rspack's current persistent cache, so I think we can not support cache now

crates/node_binding/src/raw_options/raw_builtins/raw_copy.rs
```diff
@@ -9,17 +9,24 @@ use rspack_plugin_copy::{
...
+type TransformerFn = ThreadsafeFunction<(Buffer, String), Either<String, Buffer>>;
+type RawTransformer = Either<RawTransformWithCacheOptions, TransformerFn>;
 
 type RawToFn = ThreadsafeFunction<RawToOptions, String>;
 
 type RawTo = Either<String, RawToFn>;
 
+#[derive(Debug, Clone)]
+#[napi(object, object_to_js = false)]
+pub struct RawTransformWithCacheOptions {
```
hardfist, line 21: `RawTransformOptions` seems better?

crates/rspack_plugin_copy/src/lib.rs
```diff
@@ -291,24 +294,50 @@ impl CopyRspackPlugin {
...
+          handle_transform(
+            transformer,
+            source_vec,
+            absolute_filename.clone(),
+            &mut source,
+            diagnostics,
+          )
+          .await
         }
+        Transformer::Opt((transformer, cache)) => match cache {
```
hardfist, line 312: I think we should all cache related logic if not support now

fireairforce:
> cache is tricky and not compatible with rspack's current persistent cache, so I think we can not support cache now, can you also add params validation logic so unnecessary param type is not passed to napi options directly

For the reason we don't support caches config now, i just remove the config now to avoid the user to passed a useless config to napi, if we support in the feature, i will add it back and update the document, i think this is more reasonable.

crates/rspack_plugin_copy/src/lib.rs
fireairforce: it will not exposed to our users, cache is just a `None` here.

crates/rspack_plugin_copy/src/lib.rs
```diff
@@ -789,6 +818,29 @@ fn set_info(target: &mut AssetInfo, info: Info) {
   }
 }
 
+async fn handle_transform(
+  transformer: &TransformerFn,
+  source_vec: Vec<u8>,
+  absolute_filename: Utf8PathBuf,
+  source: &mut RawSource,
+  diagnostics: &Mutex<Vec<Diagnostic>>,
```
hardfist, line 826: `&Mutex<Vec<Diagnostics>>` is super weird in async context, `Arc<Mutex` seems better, can you help fix this

crates/rspack_plugin_copy/src/lib.rs
fireairforce: i remove the logic here and add a comment here to wait cache support in the furture

hardfist (approved):

PR #9737, opened 2025-03-19
fix: jsonp_chunk_loading_with_callback format problem

1714080902120:
## Summary

I have a custom plugin that needs to add some code to the runtime, and during the migration process I find that the code replaced fail, and after debugging, I find that the code should have been replaced:
```js
chunkLoadingGlobal.push = webpackJsonpCallback.bind(null, chunkLoadingGlobal.push.bind(chunkLoadingGlobal));
```
 It becomes as follows: 
```js
chunkLoadingGlobal.push = webpackJsonpCallback.bind( null,   chunkLoadingGlobal.push.bind(chunkLoadingGlobal));
```

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5012,6 +5012,7 @@ name = "rspack_plugin_library"
 version = "0.2.0"
 dependencies = [
  "async-trait",
+ "futures",
  "regex",
  "rspack_cacheable",
  "rspack_collections",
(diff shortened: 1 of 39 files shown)


GiveMe-A-Name:
You need update test snapshot.
`cd packages/rspack-test-tools` and run `npm run testu`

GiveMe-A-Name:
There have some snapshot was not updated.
You can cd `tests/webpack-test`, then run pnpm run testu to update snapshots

1714080902120:
Well, it looks like I can see how to run based on a failed CI build.

PR #9741, opened 2025-03-20
feat: support experiments build http

ScriptedAlchemy:
## Summary

Add buildHttp experiment bindings for rust plugin

## Checklist


tests/webpack-test/configCases/asset-modules/build-http/webpack.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
 	mode: "development",
 	experiments: {
 		buildHttp: {
-			allowedUris: [() => true],
+			allowedUris: ['https://'],
```
ScriptedAlchemy, line 8: do not currently support function callback, only regex and string pattern

ahabhgk: Let's add a additional test in rspack-test-tools, usually we won't change test inside webpack-test

crates/rspack_plugin_schemes/src/http_cache.rs
```diff
@@ -132,9 +140,11 @@ impl HttpCache {
 
     let (store_lock, store_cache, valid_until) = parse_cache_control(&cache_control, request_time);
 
+    // Handle 304 Not Modified (similar to webpack)
     if status == 304 {
       if let Some(cached) = cached_result {
         let new_valid_until = valid_until.max(cached.meta.valid_until);
+        println!("GET {} [{}] (unchanged)", url, status);
```
ahabhgk, line 147: `println!` can be replaced to `logger`, there is a `get_logger` on compilation

ahabhgk:
Let's merge it first, we can resolve the remaining things in the future PR since it's still experimental

ahabhgk (approved):

PR #9773, opened 2025-03-24
chore(deps): bump swc to 16.10.0

GiveMe-A-Name:
## Summary

Swc lastest version has a performance improvement in swc_ecma_minifier and fix some bugs in es/minifier and ts/fast-strip. So we bump swc version.

- bump `swc_core` to 16.10.0
- bump `swc_ecma_minifier` to 12.4.0
- bump `swc_error_reporters` to 9.1.1

By default, Swc using chili when open concurrent features, but chili has high Idle CPU usage. So swc recommends open concurrent features using rayon.
- Open `rayon` features when concurrent is open.

More detail see: https://github.com/swc-project/swc/issues/10243#issuecomment-2742856826

## Checklist


chenjiahan:
🚀

PR #9807, opened 2025-03-26
feat: support experimental loader parallelism

h-a-n-a:
## Summary

This PR added experimental support for loader parallelism. To enable parallelism, set either `Rule.use.parallel = true` or `Rule.parallel = true` when using loader without `Rule.use` and `experiments.parallelLoader = true`

### Worker threads

Unlike `thread-loader`, rspack uses worker threads to drive loader tasks, using `tinypool` under the hood. That said, loader options need to be serializable in order to be sent to the worker thread and not all methods are available in `LoaderContext`. Currently, `importModule` and most of the methods of `_compilation`, `_compiler`, `_module` are missing.

### Loader runner

`thread-loader` dispatches remaining loaders by default. If I write `use: ["thread-loader", "css-loader", "less-loader"]`, `css-loader` and `less-loader` would end up in the worker thread. This is not the same in rspack. You may consider the worker as a loader runner but in worker. If rspack encounters a loader that is set `parallel: true`, loader runner on the main thread would send the execution state to worker and yield if the next loader is not `parallel` or it's a rust builtin loader. This chains loaders with `parallel` enabled automatically, achieving the best performance.

### Handling synchronous calls

Worker threads are spawned on demand and, by default, is using the max thread available on the OS. In each worker, a channel is created to pass requests to the main thread for on-demand requesting, such as `LoaderContext["getResolve"]`. That is to say, each request to the main thread is running asynchronously. This also requires most of the method of `LoaderContext` to be `async`. To make it compatible with synchronous loader context calls to the main thread, rspack uses an internal wrapper to wrap an `async` method into sync, by exposing `wait()` to each request. For example, `addDependency` does not resolve until the main thread yield, but `getDependencies` will wait until all previous adding operation to finish and get the latest dependencies in a synchronous call. This was achieved by putting the thread to sleep and waking it up when it's necessary.

### Caveats

- Missing implementations to `importModule`, `_compilation`, `_compiler` and `_module` methods.
- Options passed to loader should be serializable with regard to the [specification](https://nodejs.org/api/worker_threads.html#portpostmessagevalue-transferlist).

### Benchmark

**Building 100 copies of `antd.less` with `css-loader`, `rspack-plugin-extract-css` and `less-loader`:**

```
# With parallelism enabled on `css-loader` and `less-loader`
hyperfine --runs 5 "rspack build"
Benchmark 1: rspack build
  Time (mean ± σ):      6.972 s ±  0.127 s    [User: 49.337 s, System: 5.876 s]
  Range (min … max):    6.818 s …  7.127 s    5 runs
```

```
# Without parallelism
hyperfine --runs 5 "rspack build"
Benchmark 1: rspack build
  Time (mean ± σ):     15.769 s ±  0.058 s    [User: 24.518 s, System: 1.793 s]
  Range (min … max):   15.692 s … 15.836 s    5 runs
```

**Building 100 copies of `antd.less` with `experiments.css` and `less-loader`:**

```
# With parallelism enabled on `less-loader`
hyperfine --runs 5 "rspack build"
Benchmark 1: rspack build
  Time (mean ± σ):      3.960 s ±  0.158 s    [User: 32.401 s, System: 3.055 s]
  Range (min … max):    3.823 s …  4.218 s    5 runs
```

```
# Without parallelism enabled on `less-loader`
hyperfine --runs 5 "rspack build"
Benchmark 1: rspack build
  Time (mean ± σ):      8.526 s ±  0.097 s    [User: 14.033 s, System: 1.005 s]
  Range (min … max):    8.425 s …  8.679 s    5 runs
```

```
# With `thread-loader`
hyperfine --runs 5 "rspack build"  
Benchmark 1: rspack build
  Time (mean ± σ):      3.814 s ±  0.099 s    [User: 31.751 s, System: 2.394 s]
  Range (min … max):    3.755 s …  3.989 s    5 runs
```

_Benchmark was running on Apple M2 Max, 64G. `antd.less` is referenced from [this](https://github.com/umijs/benchmark/blob/610a858026b1212a3bc4866a98038eec80ed7f23/projects/lots-of-less/src/less/antd-1.less)._

RFC: https://github.com/web-infra-dev/rspack/discussions/9815

## Checklist


lint-staged.config.mjs
```diff
@@ -5,7 +5,8 @@ export default {
 	"*.{ts,tsx,js,cts,cjs,mts,mjs}": [
 		"pnpm run lint:js",
 		// ignore staged file list; execute x without extra args
-		() => "pnpm run x ae ci"
+		() => "pnpm run x ae update",
+		"git add ."
```
chenjiahan, line 9: This may add unexpected files

h-a-n-a: Going to revert this. It's used for local pushing, which should be removed at the end. 🥲

ahabhgk (approved):

PR #9810, opened 2025-03-26
perf: run resolve in spawn_blocking

hardfist:
## Summary
resolve is IO bound, so try run it in spawn_blocking to avoid blocking CPU work

## Checklist


hardfist:
sadly it cause performance regression

PR #9818, opened 2025-03-28
chore: Upgrade Rust to nightly of 1.86.0 release date

andersk:
## Summary

[Rust 1.86.0 stable](https://blog.rust-lang.org/2025/04/03/Rust-1.86.0.html) was released on 2025-04-03. Upgrade to the 2025-04-03 nightly, fixing all build errors and Clippy warnings.

(I’m not sure there’s advantage to aligning the date this way, given that stable is branched from nightly well before the release, but this matches what we’ve done before.)

I carefully made one commit per warning type, so ideally this should be merged using either **Merge** or **Rebase and merge** to preserve the individual commits, rather than **Squash and merge** which does not.

## Checklist

### DIFF
--- a/crates/node_binding/src/options/entry.rs
+++ b/crates/node_binding/src/options/entry.rs
@@ -57,7 +57,7 @@ impl From<JsEntryOptions> for EntryOptions {
       base_uri: value.base_uri,
       filename: value.filename.map(Into::into),
       library: value.library.map(Into::into),
-      depend_on: value.depend_on.map(Into::into),
+      depend_on: value.depend_on,
       layer: value.layer,
     }
   }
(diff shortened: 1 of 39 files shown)


andersk:
Rebased and upgraded to today’s nightly to correspond with the 1.86.0 stable release.

hardfist:
thanks, I agree that nightly date align with stable date is not meaningful and we may switch to stable version in the future when we remove all nightly feature in codebase.

hardfist (approved):

chenjiahan:
This PR caused the Rspack binary size to increase by 3+MB and we need to investigate how to optimize it.
https://web-infra-qos.netlify.app/?product=rspack&metrics=binary-size

image

PR #9820, opened 2025-03-28
fix(diagnostics): enhance transform method to return diagnostics alongside AST

GiveMe-A-Name:
## Summary

This pull request includes several changes to the `SwcCompiler` and `SwcLoader` implementations in the `crates/rspack_loader_swc` module. The changes primarily focus on improving the handling of diagnostics and the structure of the transformation output.

The Swc plugin emit warn diagnostics will ignore by swc before, according to this PR [#10027](https://github.com/swc-project/swc/pull/10027), swc will return a diagnostics. So we can emit diagnostics to loaderContext.

Improvements to diagnostics handling and transformation output:

* [`crates/rspack_loader_swc/src/compiler.rs`](diffhunk://#diff-2af37dc1c677bb8f7d6cc3d1c81840dc61458482fa18cd002ee21d661c9c352fL349-R361): Modified the `transform` method to return a new `SwcTransformOutput` struct that includes both the AST and diagnostics. This change enhances the ability to capture and handle diagnostics during the transformation process. [[1]](diffhunk://#diff-2af37dc1c677bb8f7d6cc3d1c81840dc61458482fa18cd002ee21d661c9c352fL349-R361) [[2]](diffhunk://#diff-2af37dc1c677bb8f7d6cc3d1c81840dc61458482fa18cd002ee21d661c9c352fL380-R385) [[3]](diffhunk://#diff-2af37dc1c677bb8f7d6cc3d1c81840dc61458482fa18cd002ee21d661c9c352fL393-R401)
* [`crates/rspack_loader_swc/src/compiler.rs`](diffhunk://#diff-2af37dc1c677bb8f7d6cc3d1c81840dc61458482fa18cd002ee21d661c9c352fR554-R558): Added a new `SwcTransformOutput` struct to encapsulate the AST and diagnostics.
## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/compiler.rs
+++ b/crates/rspack_loader_swc/src/compiler.rs
@@ -346,14 +346,19 @@ impl SwcCompiler {
     }
   }
 
-  pub fn transform(&self, config: BuiltInput<impl Pass>) -> Result<Program, miette::Report> {
+  pub fn transform(
+    &self,
+    config: BuiltInput<impl Pass>,
+  ) -> Result<SwcTransformOutput, miette::Report> {
     let program = config.program;
     let mut pass = config.pass;
 
+    let mut diagnostics = vec![];
     let program = self.run(|| {
       helpers::HELPERS.set(&self.helpers, || {
-      let result = try_with_handler(self.cm.clone(), Default::default(), |_handler| {
+      let result = try_with_handler(self.cm.clone(), Default::default(), |handler| {
         let result = program.apply(&mut pass);
+        diagnostics.extend(handler.take_diagnostics());
         Ok(result)
       });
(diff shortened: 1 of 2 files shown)


h-a-n-a:
Will diagnostics be considered as a error or warning? I suppose this is for emitting wasm plugin version error, right?

GiveMe-A-Name:
> Will diagnostics be considered as a error or warning? I suppose this is for emitting wasm plugin version error, right?

Actually, diagnostics should be considered a warning. If swc_parser or wasm_plugin emit a error , swc api will return a Err(anyhow::Error) as Result<Ret, E> returns value.
So if we can get diagnostics from handler, it means swc emits warning.

PR #9828, opened 2025-03-28
refactor!: lazy compilation middleware supports multiCompiler and use config from compiler instance

JSerFeng:
## Summary

Lazy compilation middleware can receive multiCompiler as parameter

When there is a module triggers update, we notify the multiWatching, then multiWatching will invalidate its sub watching.

Every compiler can has its own lazy-compilation config, for example you can make compiler_1 only lazy compiles entries, while compiler_2 only lazy compiles imports

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
+++ b/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
@@ -1,6 +1,6 @@
 import type { IncomingMessage, ServerResponse } from "node:http";
-import type { Compiler, LazyCompilationOptions } from "../..";
-import type { Module } from "../../Module";
+import { type Compiler, MultiCompiler } from "../..";
+import type { LazyCompilationOptions } from "../../config";
 import type { Middleware } from "../../config/devServer";
 import { BuiltinLazyCompilationPlugin } from "./lazyCompilation";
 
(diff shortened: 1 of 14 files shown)


chenjiahan:
@JSerFeng Can we fix the CI and get this PR merged? ❤️

packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
```diff
@@ -37,19 +37,104 @@ const getFullServerUrl = ({ serverUrl, prefix }: LazyCompilationOptions) => {
...
+			);
+
+			const compilerConfig = c.options.experiments.lazyCompilation
+				? c.options.experiments.lazyCompilation
+				: {};
+
+			const testConfig = compilerConfig.test ?? options.test;
+
+			const plugin = new BuiltinLazyCompilationPlugin(
+				({ module, path }) => {
```
chenjiahan, line 88: Can we extract a function to reduce duplicated code:

- https://github.com/web-infra-dev/rspack/blob/83046475af1e94b20dbfb79458ef104f150c76f0/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts#L88-L103
- https://github.com/web-infra-dev/rspack/blob/83046475af1e94b20dbfb79458ef104f150c76f0/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts#L138-L153

packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
```diff
@@ -37,19 +37,104 @@ const getFullServerUrl = ({ serverUrl, prefix }: LazyCompilationOptions) => {
...
+	if (userOptions === true) {
+		userOptions = {};
+	}
+
+	if (compiler instanceof MultiCompiler) {
+		const middlewareByCompiler: Map<string, Middleware> = new Map();
+
+		let i = 0;
+		for (const c of compiler.compilers) {
+			if (!c.options.experiments.lazyCompilation) {
```
chenjiahan, line 57: Where is `c.options.experiments.lazyCompilation` been set

chenjiahan: We should force users to set `experiments.lazyCompilation` instead of passing `rawOptions` to the middlewware

chenjiahan (approved):

PR #9842, opened 2025-03-30
fix: return dynamic import script externals

Leonewu:
## Summary

As the [document](https://rspack.dev/zh/config/externals#externalstypescript) says, rspack supports `script` externalsType .Also, externals could be configured like `script global@your-cdn-link`.

When dynamically importing an externalscript module, the compiled code was missing a return statement , causing the import to resolve to undefined.

**rspack.config:**
```javascript
module.exports ={
  ...,
  externals: {
    example: 'script example@https://unpkg.com/example@1.0.0/index.min.js'
  }
}
```

**source code:**

```javascript
const example = import('example').then(mod => {
  console.log('example: ' mod);   // output: example: undefined
  return mod;
})
```

**compiled code:**

- Before:

  ```javascript
  .then(function(m) {
    __webpack_require__.t(m, 22)
  })
  ```

- After:
  ```javascript
  .then(function(m) {
    return __webpack_require__.t(m, 22)
  })
  ```

This fixes the issue where `import()` of external script modules returns undefined.

Built rspack locally and verified:
1. The generated code now contains the return statement
2. Dynamic imports of external modules properly resolve to the module content.
3. Built a new binding file then replaced it in my project.  It worked.

## Checklist

### DIFF
--- a/crates/rspack_core/src/dependency/runtime_template.rs
+++ b/crates/rspack_core/src/dependency/runtime_template.rs
@@ -497,7 +497,7 @@ pub fn module_namespace_promise(
         }
         appending.push_str(
           format!(
-            ".then(function(m){{\n {}(m, {fake_type}) \n}})",
+            ".then(function(m){{\n return {}(m, {fake_type}) \n}})",
             RuntimeGlobals::CREATE_FAKE_NAMESPACE_OBJECT
           )
           .as_str(),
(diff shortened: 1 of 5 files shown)


LingyuCoder:
cc @GiveMe-A-Name

GiveMe-A-Name (approved):

PR #9846, opened 2025-03-31
test(loader-parallel): disable cases that rely on `import()` in node 16

h-a-n-a:
## Summary

In this PR, tests relying on `import()` are disabled. It causes segmentation fault in `vm` with `--experimental-vm-modules`, which is exactly what jest has been relying on and the issue has not been resolved in node 16 yet according to my local test. 

## Checklist


chenjiahan (approved):

chenjiahan:
LGTM 👍 We can drop support for Node 16 in the next major release

PR #9848, opened 2025-03-31
feat: support passing `PathBuf` and `&Path` to `Context`

h-a-n-a:
## Summary

Polish rspack crate document and add `PathBuf` and `&Path` support to `Context`. This cuts a dependency that users need to rely on to run rspack in Rust.

## Checklist

### DIFF
--- a/crates/rspack/src/lib.rs
+++ b/crates/rspack/src/lib.rs
@@ -1,14 +1,6 @@
 #![forbid(missing_docs)]
 
-//! Rspack is a high performance JavaScript bundler written in Rust. It offers strong compatibility with the webpack ecosystem, allowing for seamless replacement of webpack, and provides lightning fast build speeds.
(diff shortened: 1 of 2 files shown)


crates/rspack/src/lib.rs
```diff
@@ -1,5 +1,64 @@
 #![forbid(missing_docs)]
 
-//! The Rspack compiler.
-
+//! Rspack is a high performance JavaScript bundler written in Rust. It offers strong compatibility with the webpack ecosystem, allowing for seamless replacement of webpack, and provides lightning fast build speeds.
+//!
+//! ## Why rspack?
```
chenjiahan, line 5: Can we link these introduction to https://rspack.dev/ to reduce maintenance overhead.

chenjiahan (approved):

PR #9851, opened 2025-03-31
chore: add skipLibCheck:false for template

hardfist:
## Summary
enable `skipLibCheck:false` for template cause `skipLibCheck:true` will hide rspack's type error 

## Checklist


fireairforce (approved):

chenjiahan (changes requested):
Or we can add a test to check instead of turning off skipLibCheck in the templates.

PR #9871, opened 2025-04-01
feat: asset info support symbol key

SyMind:
## Checklist


crates/node_binding/src/compilation/mod.rs
```diff
@@ -295,20 +309,36 @@ impl JsCompilation {
...
     filename: String,
     source: JsCompatSource,
-    js_asset_info: Option<AssetInfo>,
+    object: Option<Object>,
   ) -> Result<()> {
     let compilation = self.as_mut()?;
 
-    let asset_info: rspack_core::AssetInfo = js_asset_info.map(Into::into).unwrap_or_default();
+    let asset_info = if let Some(object) = object {
+      let js_asset_info: AssetInfo =
```
hardfist, line 325: need add safety comments and is it right that needs write unsafe code here?

hardfist: it seems all napi function now needs to manually write ts signature and unsafe code now and it seems it cause DX regression

SyMind: Not all situations require it, only when it is necessary to bind JS Objects when initializing.

I do not think need safety comments here, because  `FromNapiValue::from_napi_value()` is very common in napi.

quininer: What is the safety condition of `from_napi_value`, and can we construct a `safe_from_napi_value`?

SyMind: @quininer ok, i declare from_jsobject in AssetInfo struct to prevent directly using unsafe block.

quininer (approved):

PR #9872, opened 2025-04-01
feat: support rspack wasi with node.js

CPunisher:
## Summary

Here are some final works to make rspack wasi available in node.js. Now you can run rspack wasi on rspack/rolldown benchmark (remember to use `NAPI_RS_FORCE_WASI=1`):

1. Fix regex passing from js to rs. Although I don't know why `globalThis.Object.prototype` is an `External` other than `JsObject`.
2. Copy `canonicalize` from `rspack_resolver` because `fs::canonicalize` doesn't support wasm.
3. Enable tokio multi thread feature to avoid `tokio::spawn` blocking the program.

**BUT PAY ATTENTION**:

1. Only release mode is avaliable. If you build rspack with `build:cli:dev:wasm`, you will get random memory errors. Currently I don't figure it out yet.
2. Rspack doesn't exit when it finishes works. Currently I don't call `shutdownAsyncRuntime` in js side because rayon thread pool also blocks the program. I'd like to handle these two cases in next pr.
3. You may get memory allocation failure in very big project.

## Checklist


h-a-n-a (approved):

h-a-n-a:
Thanks!

crates/rspack_fs/src/native_fs.rs
```diff
@@ -224,8 +224,31 @@ impl ReadableFileSystem for NativeFileSystem {
   }
 
   fn canonicalize(&self, path: &Utf8Path) -> Result<Utf8PathBuf> {
-    let path = dunce::canonicalize(path)?;
-    Ok(path.assert_utf8())
+    // Comes from rspack_resolver
+    use std::path::Component;
```
hardfist, line 228: is this wasi only and should be put into cfg_if!?

crates/rspack_regex/src/napi.rs
```diff
@@ -28,8 +28,8 @@ impl FromNapiValue for RspackRegex {
     let global = env.get_global()?;
     let object_prototype_to_string = global
       .get_named_property_unchecked::<JsObject>("Object")?
-      .get_named_property::<JsObject>("prototype")?
-      .get_named_property::<Function>("toString")?;
+      .get_named_property_unchecked::<JsObject>("prototype")?
+      .get_named_property_unchecked::<Function>("toString")?;
```
hardfist, line 32: why this changed?

crates/rspack_fs/src/native_fs.rs
CPunisher: Here is the cfg: https://github.com/web-infra-dev/rspack/pull/9872/files/1350ac606b87cab251a08c0d77780d130674ed4f#diff-98c48758ea9967701abd9a996d1ba7ae2a0fb3afbade319da1f628f457332aa4R209

crates/rspack_regex/src/napi.rs
CPunisher: I'm not sure but from my experiments:

1. `.get_named_property_unchecked::<JsObject>("prototype")` returns an `External` in wasm.
2. But `_unchecked` in `.get_named_property_unchecked::<Function>("toString")?` is actually needless.

PR #9885, opened 2025-04-02
feat: resolver restriction support regex

SyMind:
## Summary

Resolver restriction option support regex. The feature is used in less-loader.

Relate https://github.com/web-infra-dev/rspack-resolver/pull/60

## Checklist


crates/node_binding/src/options/raw_resolve.rs
```diff
@@ -254,6 +265,16 @@ pub fn normalize_raw_resolve_options_with_dependency_type(
         })
         .transpose()?;
 
+      let restrictions = raw.restrictions.map(|estriction| {
```
stormslowly, line 268: typo

stormslowly (approved):

PR #9895, opened 2025-04-03
fix(wasm): shutdown async runtime and thread workers after finishing

CPunisher:
## Summary

The reason I make this pr separately is that I'm not sure if my solution is best. So discussions are welcome.

Rspack wasi will get stuck after finishing, because:

1. We should manually shutdown napi async runtime.
2. Rayon thread pool doesn't exit. Normally rayon threads run and steal works until the main thread exits. But wasi simulates threads with nodejs workers. I find the main thread in nodejs doesn't exit if nodejs worker get stuck. So we should also shutdown all the workers manually.

Temporarily I stop napi generating js glue code and export the function to shutdown workers. And I'm not sure the replace where I put shutdown code is correct.

## Checklist

### DIFF
--- a/crates/node_binding/package.json
+++ b/crates/node_binding/package.json
@@ -47,8 +47,15 @@
       "aarch64-unknown-linux-gnu",
       "aarch64-apple-darwin",
       "aarch64-unknown-linux-musl",
-      "aarch64-pc-windows-msvc"
-    ]
+      "aarch64-pc-windows-msvc",
+      "wasm32-wasip1-threads"
+    ],
+    "wasm": {
+      "initialMemory": 16384,
+      "browser": {
+        "fs": true
+      }
+    }
   },
   "optionalDependencies": {
     "@rspack/binding-darwin-arm64": "workspace:*",
--- a/crates/node_binding/package.json
+++ b/crates/node_binding/package.json
@@ -47,15 +47,8 @@
       "aarch64-unknown-linux-gnu",
       "aarch64-apple-darwin",
       "aarch64-unknown-linux-musl",
-      "aarch64-pc-windows-msvc",
-      "wasm32-wasip1-threads"
-    ],
-    "wasm": {
-      "initialMemory": 16384,
-      "browser": {
-        "fs": true
-      }
-    }
+      "aarch64-pc-windows-msvc"
+    ]
   },
   "optionalDependencies": {
     "@rspack/binding-darwin-arm64": "workspace:*",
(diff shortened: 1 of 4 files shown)


h-a-n-a:
Would you please elaborate on "I find the main thread in nodejs doesn't exit if nodejs worker get stuck."? Is this problem exist when nodejs workers encounter some error or exist on every executions?

CPunisher:
> Would you please elaborate on "I find the main thread in nodejs doesn't exit if nodejs worker get stuck."? Is this problem exist when nodejs workers encounter some error or exist on every executions?

Here is an example. When you run `node main.js`, you will get stuck.
```js
// main.js
const { Worker } = require('worker_threads');
const worker = new Worker('./worker.js')
console.log("main end")

// worker.js
console.log("worker start")
while (true) {}
console.log("worker end")
```

Now let me clarify the problems with `rayon`:
1. Rspack wasi (powered by emnapi) create threads by creating node workers
2. Although `rayon` falls back to sequential iteration in wasm,  I find it still create its worker threads which lives in the whole lifetime of program. In normal native program, it's ok because the main thread doesn't wait for child theads without `join`.
3. Rolldown avoids using `rayon` by wasm shims: https://github.com/rolldown/rolldown/blob/main/crates/rolldown_utils/src/rayon.rs. It's very hard to apply it in rspack because some rayon iter api is not consistent with std iter api such as `fold`. We may need to do more to improve the wasm shims or write lots of `#[cfg]`.

h-a-n-a:
> Although rayon falls back to sequential iteration in wasm, I find it still create its worker threads which lives in the whole lifetime of program. 

That's weird. I haven't really dug into the implementation in WASI node. But I thought rayon tasks would be sent to `worker_threads` for execution instead of being executed in sequential.

> In normal native program, it's ok because the main thread doesn't wait for child threads without join.

Is it possible to `unref()` all `Worker`s? This should make the worker threads exit if main thread finishes its execution?

CPunisher:
> That's weird. I haven't really dug into the implementation in WASI node. But I thought rayon tasks would be sent to `worker_threads` for execution instead of being executed in sequence.

You could be right! Actually I know this from https://github.com/rayon-rs/rayon?tab=readme-ov-file#usage-with-webassembly, but I think it's very valuable to optimize in the future.

> Is it possible to unref() all Workers? This should make the worker threads exit if main thread finishes its execution?

Thanks! I didn't know this api before. I quickly tried but didn't solve the problem. I will do more research about it.

CPunisher:
It seems that `worker.on('message', ...)` will call `worker.ref()` again, see: https://github.com/nodejs/node/issues/53036. A hacky way is adding `worker.ref = () => {}` to prevent workers being refed.

h-a-n-a (commented):
Thanks

h-a-n-a (approved):

PR #9925, opened 2025-04-07
fix(deps): `tinypool` not being able to run on older node versions

h-a-n-a:
## Summary

Use pre-compiled `tinypool` to avoid breaking changes to node versions that are lower than 18.

`tinypool` is copied from `node_modules` and removes `engines` field in `package.json`. Tinypool does not have dependencies, so we don't need to prebundle it.

## Checklist

### DIFF
--- a/packages/rspack/move-tinypool.js
+++ b/packages/rspack/move-tinypool.js
@@ -9,11 +9,9 @@ const tinypool = require
 	.replace("/package.json", "");
 const dest = path.resolve(__dirname, "./compiled/tinypool");
 
-fs.cpSync(tinypool, path.resolve(__dirname, "./compiled/tinypool"), {
-	recursive: true
-});
+fs.cpSync(tinypool, dest, { recursive: true });
 
 const pkg = JSON.parse(fs.readFileSync(path.join(dest, "package.json")));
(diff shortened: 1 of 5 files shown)


packages/rspack/move-tinypool.js
```diff
@@ -0,0 +1,17 @@
+// This script copies the tinypool package from node_modules to the compiled directory.
+// It removes the engines field from package.json and ensures that the package is working under `engine-strict=true` on lower node versions.
+
+const fs = require("node:fs");
+const path = require("node:path");
+
+const tinypool = require
+	.resolve("tinypool/package.json")
+	.replace("/package.json", "");
```
chenjiahan, line 9: This may break on Windows

packages/rspack/move-tinypool.js
```diff
@@ -0,0 +1,17 @@
...
+const dest = path.resolve(__dirname, "./compiled/tinypool");
+
+fs.cpSync(tinypool, dest, { recursive: true });
```
chenjiahan, line 12: Should we add `force: true` to override files?

chenjiahan (approved):

chenjiahan:
LGTM 👍

PR #9939, opened 2025-04-08
test: remove new code splitting tests as now its the default

JSerFeng:
## Summary

We make newCodeSplitting test which enables experiments.parallelCodeSplitting and runs all config/stats tests, now its true by default, all tests are already using experiments.parallelCodeSplitting, we don't need these tests anymore

## Checklist


JSerFeng:
> so are tests still running for old splitting?

No

LingyuCoder (approved):

PR #9968, opened 2025-04-09
fix: stablelize order of concated modules

JSerFeng:
## Summary

Stablelize the order of concatenated modules.

Iterate `FxHashSet` is not stable, `ModuleConcatenatePlugin` uses `dep.sourceOrder` to ensure the order, however the `sourceOrder` can be the same.

For example:

```
export { a, b } from 'lib';
```

Code above contains 2 [`ExportImportedDependency`](https://webpack-dependencies-visualize.vercel.app/#code=ZXhwb3J0IHsgYSwgYiB9IGZyb20gJ2xpYic7), and they have the same `sourceOrder`. So the sorting algorithm is not safe

## Checklist

### DIFF
--- a/crates/node_binding/src/module_graph.rs
+++ b/crates/node_binding/src/module_graph.rs
@@ -167,7 +167,7 @@ impl JsModuleGraph {
     let (compilation, module_graph) = self.as_ref()?;
     Ok(
       module_graph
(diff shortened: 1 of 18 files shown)


Kingwl:
It's very helpful to me！

LingyuCoder (approved):

PR #9978, opened 2025-04-10
chore: use RefCell instead of Mutex in thread_local!

quininer:
## Summary

It doesn't make sense to use a `Mutex` in this case, `RefCell` is a lower-cost option. but this probably won't give any performance improvement, since there's no access to `GLOBAL_TRACE_STATE` outside of `register_global_trace` and `cleanup_global_trace`.

## Checklist


SyMind:
@quininer maybe Cell is better?

SyMind (approved):

Issue #9985, opened 2025-04-11
[Bug]: sometimes content hash of runtime chunk changes but code of chunk loading do not change


LingyuCoder:
### System Info

x

### Details

x

### Reproduce link

https://github.com/hikerpig/module-runtimechunk-repro

### Reproduce Steps

npm ci
npm install
npm run dev
edit index.js

PR #9986, opened 2025-04-11
fix: sometimes entry chunk hash not changes with full hash runtime modules

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/9985

In this repro https://github.com/hikerpig/module-runtimechunk-repro, the combination of `chunkFormat: module` and `runtimeChunk: true` causes the entry chunk to import the runtime chunk via `import __webpack_require__ from "./runtime~main.857eaa3b.js";`.

This leads to a situation, when relying on full hash (such as the entry injected by the dev server), during incremental builds in the dev mode, changes of the full hash will cause changes in the content of the runtime chunk. If the filename of the runtime chunk contains `[contenthash]`, then its final filename will change. At this time, due to the above mentioned `import` statement, the content of the entry chunk will also change, and consequently, its `[contenthash]` will change.

However, when rspack calculates the `contenthash` of a chunk, it doesn't take this scenario into account. Because the entry chunk itself has no real relation to these changes, and the import statement is startup code injected during the code generation of the chunk. So when these contents change, the contenthash and chunkhash of the entry chunk in rspack won't change. As a result, the chunk code generation cache takes effect, and the runtime chunk from the previous compilation is referenced.

In Webpack, when adding a runtime module, the runtime module is associated with this chunk. When handling the chunk hash, for normal chunks, it also checks whether they contain full hash modules just like runtime chunks, and then includes the chunk in the hash update process of full hash chunks. However, in rspack, this associative relationship doesn't exist, and the chunk hash only deals with the full hash update of runtime chunks. 

So this PR only applies to the scenarios of `commonjs` and `module` chunkFormats. If they introduce the loading statement of the runtime chunk, then this entry chunk will be added to the full - hash update, causing the cache of this entry chunk to become invalid, thus fixing this issue.

However, in the `new incremental`, `Mutation::ModuleSetHashes` only affects the chunks associated with the module. This causes the chunk hashing process to be skipped, and the problem still persists. I haven't found a good solution to this.

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/NewIncremental-watch.test.js
+++ b/packages/rspack-test-tools/tests/NewIncremental-watch.test.js
@@ -20,6 +20,8 @@ describeByWalk(
 	},
 	{
 		source: path.resolve(__dirname, "./watchCases"),
-		dist: path.resolve(__dirname, `./js/new-incremental/watch`)
+		dist: path.resolve(__dirname, `./js/new-incremental/watch`),
+		// TODO: fix this test case
+		exclude: [/runtimeChunkFullHash/]
 	}
 );
--- a/crates/rspack_core/src/compiler/compilation.rs
+++ b/crates/rspack_core/src/compiler/compilation.rs
@@ -2019,18 +2019,18 @@ impl Compilation {
 
     fn try_process_chunk_hash_results(
       compilation: &mut Compilation,
-      chunk_hash_results: Vec<Result<(ChunkUkey, (RspackHashDigest, ChunkContentHash, bool))>>,
+      chunk_hash_results: Vec<Result<(ChunkUkey, ChunkHashResult)>>,
       full_hash_chunks: &mut UkeySet<ChunkUkey>,
     ) -> Result<()> {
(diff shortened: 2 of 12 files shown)


packages/rspack-test-tools/tests/NewIncremental-watch.test.js
```diff
@@ -20,6 +20,8 @@ describeByWalk(
 	},
 	{
 		source: path.resolve(__dirname, "./watchCases"),
-		dist: path.resolve(__dirname, `./js/new-incremental/watch`)
+		dist: path.resolve(__dirname, `./js/new-incremental/watch`),
+		// TODO: fix this test case
+		exclude: [/runtimeChunkFullHash/]
```
LingyuCoder, line 25: This case can not work with `new incremental`, perhaps we need to handle this in https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/incremental/mutations.rs#L146-L148 @ahabhgk

ahabhgk (approved):

Issue #9877, opened 2025-04-01
[Bug]: `compilation.errors` undefined in CircularDependenciesPlugin


bostondv:
### System Info

System:
    OS: macOS 15.3.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 81.34 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.nvm/versions/node/v22.11.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.11.0/bin/yarn
    npm: 10.9.0 - ~/.nvm/versions/node/v22.11.0/bin/npm
    pnpm: 10.7.0 - /opt/homebrew/bin/pnpm
    Watchman: 2024.11.18.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 134.0.6998.166
    Safari: 18.3
    Safari Technology Preview: 18.4
  npmPackages:
    @rspack/cli: latest => 1.3.1 
    @rspack/core: latest => 1.3.1 

### Details

Circular Dependencies plugin docs recommend using `compilation.errors.push` to add compilation errors when cycle is detected but this does not work.

Doc link: https://rspack.dev/plugins/rspack/circular-dependency-rspack-plugin#examples

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/871532c0-eaf2-4c7f-8def-ebe95ed10492" />

Are the docs wrong or is there a bug?

Additionally, clarification how to push errors to the build would be appreciated.

Thank you

### Reproduce link

https://github.com/bostondv/rspack-repro/tree/cycle

### Reproduce Steps

Run example provided.

fireairforce:
sorry i will update the doc for this, i use to rust api instead of js api here.

fireairforce:
fixed~

PR #10006, opened 2025-04-13
refactor: adjust circularDependencyRspackPlugin compilation api

fireairforce:
## Summary

In the method https://rspack.dev/zh/plugins/rspack/circular-dependency-rspack-plugin#ondetected , the `compilation` param is `JsCompilation` not `Compilation` in rust side(see docs: https://rspack.dev/api/javascript-api/compilation), which means:

image

I adjust the `compilation` inject time stage, just inject the `compilation` in the JS side, you can see my code at `packages/src/builtin-plugin/CircularDependencyRspackPlugin.ts`.

So the user can get the `compilation` api at js side:

image

closes issue: #9877

I add the case at `packages/rspack-test-tools`

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -1513,10 +1513,10 @@ export interface RawCircularDependencyRspackPluginOptions {
   allowAsyncCycles?: boolean
   exclude?: RegExp
   ignoredConnections?: Array<[string | RegExp, string | RegExp]>
-  onDetected?: (entrypoint: Module, modules: string[], compilation: JsCompilation) => void
-  onIgnored?: (entrypoint: Module, modules: string[], compilation: JsCompilation) => void
-  onStart?: (compilation: JsCompilation) => void
-  onEnd?: (compilation: JsCompilation) => void
+  onDetected?: (entrypoint: Module, modules: string[]) => void
+  onIgnored?: (entrypoint: Module, modules: string[]) => void
+  onStart?: () => void
+  onEnd?: () => void
 }
 
 export interface RawConsumeOptions {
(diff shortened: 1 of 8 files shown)


chenjiahan (commented):
@SyMind should we expose `JsCompilation` to users?

fireairforce:
i am also curious about this, rspack don't export `Compilation` in `crates/node_binding/binding.d.ts`(), so i have to use `JsCompilation` here(https://github.com/web-infra-dev/rspack/blob/main/crates/node_binding/binding.d.ts#L194).

SyMind:
@fireairforce In rspack code now. I think we support push to compilation.errors

SyMind:
I recommend implementing `compilation.errors.push`

fireairforce:
ok, i will add this

chenjiahan:
@h-a-n-a I'm also curious if we should expose `compilation.pushDiagnostic' to users. It seems to be an internal API.

h-a-n-a:
> @h-a-n-a I'm also curious if we should expose `compilation.pushDiagnostic' to users. It seems to be an internal API.

We've actually done this in `loaderContext.experiments.emitDiagnostic` and it's been used by an internal project in ByteDance. Exposing this will just need you to copy and paste some code hre. It totally a LGTM.

chenjiahan:
@h-a-n-a Cool, can you add documentation for `loaderContext.experiments.emitDiagnostic` first? 😄

fireairforce:
I just update the pr and don't use `JsCompilation` here anymore.

SyMind (approved):

chenjiahan (approved):
LGTM 👍

PR #10008, opened 2025-04-13
refactor: simplify fs code

nilptr:
## Summary

small refactors:

- refactor: change `JsResolver::resolve_sync` to `JsResolver::resolve`, making it async
   because `JsResolver::resolve_sync` is a napi function, and on the js side, it's only exposed in a callback-style API
- refactor: remove some unnecessary `spawn_blocking`
    Since` ReadableFileSystem` already provides asynchronous methods, there's no need to use `spawn_blocking` anymore.
- refactor: remove `ReadableFileSystem::read_async` as it's duplicated

changes that may cause regression:

- refactor: use non-blocking fs in `impl ReadableFileSystem for NativeFileSystem`

## Checklist

I have run `cargo test --lib --bins --examples --tests --benches --all-targets` and `./x test ci` in my local, It passed.

image

### DIFF
--- a/packages/rspack/src/Resolver.ts
+++ b/packages/rspack/src/Resolver.ts
@@ -27,12 +27,6 @@ export class Resolver {
 		this.binding = binding;
 	}
 
-	resolveSync(context: object, path: string, request: string): string | false {
-		const data = this.binding.resolveSync(path, request);
-		if (data === false) return data;
-		return data.resource;
-	}
-
 	resolve(
 		context: object,
 		path: string,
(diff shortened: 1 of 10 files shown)


packages/rspack/src/Resolver.ts
```diff
@@ -27,12 +27,6 @@ export class Resolver {
 		this.binding = binding;
 	}
 
-	resolveSync(context: object, path: string, request: string): string | false {
-		const data = this.binding.resolveSync(path, request);
```
hardfist, line 31: this should be keeped

PR #10009, opened 2025-04-14
chore: merge rust & js loader

hardfist:
## Summary
merge rust & js side tracing

## Checklist


packages/rspack/src/loader-runner/index.ts
```diff
@@ -1020,11 +1048,15 @@ export async function runLoaders(
 					}
 					if (!fn) continue;
 
-					const args = await isomorphoicRun(fn, [
-						loaderContext.remainingRequest,
-						loaderContext.previousRequest,
-						currentLoaderObject.loaderItem.data
-					]);
+					const args = await isomorphoicRun(
```
hardfist, line 1051: @h-a-n-a I don't follow this function, it seems the args passed here is very different with the following(remainingRequest vs content

Cargo.toml
```diff
@@ -78,7 +78,7 @@ sugar_path       = { version = "1.2.0", features = ["cached_current_dir"] }
 syn              = { version = "2.0.95" }
 tokio            = { version = "1.42.0", features = ["rt", "rt-multi-thread"] }
 # don't include debug & trace in release binary to save release binary size
-tracing            = { version = "0.1.41", features = ["max_level_trace", "release_max_level_info"] }
+tracing            = { version = "0.1.41", features = ["max_level_trace", "release_max_level_trace"] }
```
hardfist, line 81: support trace level trace in release binary cause it cause 0.4M increase before gzip which seems acceptable

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1460,29 +1463,36 @@ impl Compilation {
...
+      .instrument(info_span!("hook:optimize_chunk_modules"))
       .await?;
     logger.time_end(start);
 
     // ChunkGraph is frozen for now on, we have a chunk graph that won't change
     // so now we can start to generate assets based on the chunk graph
 
     let start = logger.time("module ids");
-    tracing::info_span!("Compilation:module_ids")
-      .in_scope(|| plugin_driver.compilation_hooks.module_ids.call(self))
```
hardfist, line 1479: in_scope not support async fn

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1437,19 +1437,22 @@ impl Compilation {
         .compilation_hooks
         .optimize_modules
         .call(self)
-        .instrument(info_span!("Compilation:optimize_modules"))
+        .instrument(info_span!("hook:optimize_modules"))
```
ahabhgk, line 1440: We had tracing for every plugin hook written in macro https://github.com/web-infra-dev/rspack/blob/5a49a0169c7ae795ffa06cddf648ea48ce23c1c7/crates/rspack_macros/src/plugin.rs#L186-L193
you can add tracing at the macro so you don't need to add it at every callsite

crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs
```diff
@@ -109,7 +109,9 @@ impl ParserAndGenerator for JavaScriptParserAndGenerator {
     module.source().map_or(0, |source| source.size()) as f64
   }
 
-  #[tracing::instrument("JavaScriptParser:parse", skip_all)]
+  #[tracing::instrument("JavaScriptParser:parse", skip_all,fields(
+    resource_path =parse_context.resource_data.resource_path.as_ref().map(|p| p.as_str()),
```
ahabhgk, line 113:
```suggestion
    resource = parse_context.resource_data.resource.as_str(),
```

crates/rspack_core/src/compiler/compilation.rs
hardfist: this comment out macro should be added back

hardfist: will refactor all instrument using macro in following pr

ahabhgk (approved):

Issue #8998, opened 2025-01-13
[Feature]: Support run SWC transform via a standalone API


9aoy:
### What problem does this feature solve?

Currently, to utilize the `builtin:swc-loader` transformation capability in rspack, a complete rspack bundling compilation is required. 

If rspack provides a standalone transform API, it can be used directly to invoke rspack's `builtin:swc-loader`  for some file transformations.

Potential scenarios:
- Loading of some configuration files. When rsbuild loads the configuration file, it needs to transform the configuration file from typescript to javascript.
- Rslib's bundleless scenario, currently requires a complete rspack compilation. However, it seems that only rspack's resolve and transform capabilities are needed.

### What does the proposed API of configuration look like?

It may be an API on the rspack instance that reads rspack's configuration. Or a util method that calls rspack `builtin:swc-loader` alone.

```js
import { rspack } from '@rspack/core';
import fs from 'node:fs'

const compiler = rspack({
  // ...
});
const code = await compiler.transform('./index.ts');

fs.writeFileSync('./dist/index.js', code);
```

or 

```js
import { swcTransform } from '@rspack/core';
import fs from 'node:fs'

const code = await swcTransform('./index.ts', swcLoaderOptions);

fs.writeFileSync('./dist/index.js', code);
```

chenjiahan:
We can align the API interface with [@swc/core](https://swc.rs/docs/usage/core#transform).

For example:

```js
import { rspack } from '@rspack/core';

rspack.util.swc.transform("source code", {
  // some options
});

rspack.util.swc.parse("source code", {
  // some options
});

rspack.util.swc.minify("source code", {
  // some options
});
```

GiveMe-A-Name:
bump

GiveMe-A-Name:
Should we provide api `parse`? Actually, sometimes AST is too large. The communication between Rust and JavaScript can be quite costly. 
Moreover, the side Js is rather limited in its capabilities, as it only has access to the AST

chenjiahan:
I prefer to implement `transform` and `minify` first as the `parse` API needs further discussion and design.

GiveMe-A-Name:
It finished by https://github.com/web-infra-dev/rspack/pull/10201

chenjiahan:
Awesome 👍 Can we add documentation for these new APIs?

GiveMe-A-Name:
> Awesome 👍 Can we add documentation for these new APIs?

I‘m pleasure

[...]

PR #10017, opened 2025-04-14
feat: provide javascript compiler api to transform, parse, minify js 

GiveMe-A-Name:
## Summary

Consider upper layer (rslib, rsbuild or rspress) and js plugin need parse [t|j]s file or minify js. We should provide a compiler api to transform js & minify it.
So we reuse swc transform and minify feature. 

## Related issue (if exists):
https://github.com/web-infra-dev/rspack/issues/8998
## Checklist


GiveMe-A-Name:
This PR would split two parts: 
1. refactor: unit swc api 
2. export swc api to node_bind
This two parts will split two PR.

Issue #7364, opened 2024-07-30
Optional Environment Variables


Blankeos:
### System Info

System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 88.97 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.14.0 - ~/.nvm/versions/node/v20.14.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.14.0/bin/yarn
    npm: 10.7.0 - ~/.nvm/versions/node/v20.14.0/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v20.14.0/bin/pnpm
    bun: 1.1.21 - ~/.bun/bin/bun
  Browsers:
    Chrome: 127.0.6533.73
    Safari: 17.5

### Details

When trying to access env vars via:
process.env.PUBLIC_NICE
import.meta.env.PUBLIC_NICE

And I don't have PUBLIC_NICE in my .env. It throws an error instead of just being undefined.

Afaik in Vite, I'm allowed to have optional env variables. I usually setup my config in a typescript file with default values for local like:
```
export const config = {
   NICE: process.env.PUBLIC_NICE ?? "my default";
};
```

### Reproduce Steps

1. Make an rspack project
2. Go to App.tsx and write process.env.PUBLIC_NICE somewhere
3. Run the app (expect: error)

4. Add PUBLIC_NICE="Hello" to .env
5. Run the app (expect: no errors)

CPunisher:
Maybe you can define an empty process.env with DefinePlugin.

https://rspack.dev/zh/plugins/webpack/define-plugin
https://rsbuild.dev/config/source/define

```js
new rspack.DefinePlugin({
  "process.env": {},
}),
```

abenhamdine:
it's indeed a non documented breaking change in rspack beta 6, see https://github.com/web-infra-dev/rsbuild/issues/3029

LingyuCoder:
Currently Rspack/Webpack does not support `import.meta.env`. You can use DefinePlugin and `process.env` instead. The previous version of Rspack would replace `import.meta.env.xxxx` with `undefined`, this may result in entering an unexpected branch without knowing.

abenhamdine:
[...]

PR #10030, opened 2025-04-15
fix: covert optional import meta property to `undefined`

LingyuCoder:
## Summary

fix comment https://github.com/web-infra-dev/rspack/issues/7364#issuecomment-2469633093

according to https://bundlers.dev/javascript/modules/import.meta/, the `import.meta.env?.X` should be convert to `undefined`, just align with webpack.

## Checklist


crates/rspack_plugin_javascript/src/visitors/dependency/parser/mod.rs
```diff
@@ -713,6 +713,7 @@ impl<'parser> JavascriptParser<'parser> {
           }
           members_optionals.push(in_optional_chain);
           object = *expr.obj.take();
+          in_optional_chain = false;
```
LingyuCoder, line 716: This is a bugfix. Before this change, `import.meta.kkk.env?.X` will generate `members: [kkk, env, X]` and `member_optionals: [true, true, true]`, after this change, it will be `[false, false, true]`

packages/rspack-test-tools/tests/normalCases/esm/import-meta/index.js
LingyuCoder, line 1: All cases in this test file has been included in https://github.com/web-infra-dev/rspack/blob/main/tests/webpack-test/cases/esm/import-meta/index.js

ahabhgk (approved):

PR #10033, opened 2025-04-15
chore: bump swc version to 22.4.0

GiveMe-A-Name:
## Summary

- Bump swc version to latest
- Only emit help_msg when swc error include `load plugin errors`.

## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/compiler.rs
+++ b/crates/rspack_loader_swc/src/compiler.rs
@@ -209,6 +209,8 @@ fn read_config(opts: &Options, name: &FileName) -> Result<Option<Config>, Error>
   res.with_context(|| format!("failed to read .swcrc file for input file at `{}`", name))
 }
 
+const SWC_MIETTE_DIAGNOSTIC_CODE: &str = "Builtin swc-loader error";
+
 pub(crate) struct SwcCompiler {
   cm: Arc<SourceMap>,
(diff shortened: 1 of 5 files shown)


crates/rspack_loader_swc/src/compiler.rs
```diff
@@ -359,21 +362,37 @@ impl SwcCompiler {
...
+                swc_core::common::errors::DiagnosticId::Lint(_) => false,
+              }
+            },
+            None => false,
+        }
+      ) {
+          // swc errors includes plugin error;
+          let error_msg = err.to_pretty_string();
+          let swc_core_version = env!("RSPACK_SWC_CORE_VERSION");
+          // FIXME: with_help has bugs, use with_help when diagnostic print is fixed
```
chenjiahan, line 383: Can you provide any context for these bugs?

h-a-n-a (approved):

crates/rspack_loader_swc/src/compiler.rs
GiveMe-A-Name: Acorrding this PR https://github.com/swc-project/swc/pull/10241, swc api try_with_handler return TWithDiagnosticArray struct, we can get the diagnostic id to judge swc error type.
Only emit help message when swc load plugin error.

Issue #9077, opened 2025-01-22
[Bug]: webpack-dev-server should be a dependency not a devDependency


rrosenshain-sc:
### System Info

N/A

### Details

webpack-dev-server is being used directly to export types:
```export interface DevServer extends webpackDevServer.Configuration {```

But webpack-dev-server isn't a direct dependency so TS fallbacks to Any for `webpackDevServer.Configuration`.
I'm using pnpm v8.

### Reproduce link

will provide if needed

### Reproduce Steps

install rspack without directly install webpack dev server. Consume `RspackDevServerOptions` and notice webpack config are Any.

LingyuCoder:
bump

PR #10038, opened 2025-04-15
fix: dev server types

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/9077

Use the built-in types on `DevServer` instead of inheriting from `webpack-dev-server`.

## Checklist


chenjiahan (dismissed):
LGTM, it would be better if we could add JSDoc for these devServer options in the future.

chenjiahan (approved):

Issue #9064, opened 2025-01-21
[Bug]: Cache timestamp changes lead to invalidation?


cskunkun:
### System Info

macos，linux也是如此

### Details

使用缓存构建
```
tools: {
  rspack: {
    cache: true,
    experiments: {
      cache: { type: 'persistent' },
    },
  },
}
```
把node_modules下的.cache/rspack压缩一遍，把原文件夹删掉，再解压获得一个文件夹，会导致缓存验证失败，理论上缓存没有变化。
⚠ validate scope `occasion_make_dependencies` failed due to some packs are modified。

是否是因为timestamp变化导致缓存验证失败，如何解决？

### Reproduce Steps

1. 使用缓存构建
2. 把缓存文件压缩一遍，删掉原缓存文件
3. 解压还原缓存文件
4. 缓存验证失败

jerrykingxyz:
cache 的失效检测目前会校验文件的 修改时间，所以解压之后这些文件的 修改时间 都更新了，导致缓存校验失败。这个问题预计会在下个版本修复。
---
The cache invalidation detection currently verifies the modification time of the files, so after decompression, the modification time of these files is updated, causing the cache verification to fail. This problem is expected to be fixed in the next version.

jerrykingxyz:
bump

Narretz:
Hi @jerrykingxyz do you know approximately when this will be implemented? Cache relying on mtime makes it unusable in many CI situations.

hardfist:
> Hi [@jerrykingxyz](https://github.com/jerrykingxyz?rgh-link-date=2025-04-02T16%3A34%3A57.000Z) do you know approximately when this will be implemented? Cache relying on mtime makes it unusable in many CI situations.

we're working on new solution for this now, will fix it in near future

PR #10050, opened 2025-04-16
feat: persistent cache snapshot save file hash by default

jerrykingxyz:
## Summary

The persistent cache snapshot will store the compile times and use them to calculate which files were modified when hot start. However, this strategy is invalid in scenarios such as CI because the update time of all files will be refreshed.

This PR will make snapshot store both compile time and file hash, and when calculate modified files the file hash will be compared when the file modification time has been updated.

This PR will introduce some performance regression due to the addition of hash calculations, and I manually benchmarked the performance of snapshot generation in cold start.

| Name | File Count | Hash Count | Before | After |
| -- | -- | -- | -- | -- |
| Docusaurus | 4059 | 2143 | 22.32ms | 73.274ms |
| Large real project | 44905 | 8891 | 106.189ms | 365.41ms |

As can be seen from the table above, this takes three times longer than before, and the main time consumption occurs in file IO. This is currently acceptable in large real projects.

PS. For hot start, it will take 3 times longer if you update a lot of files like in CI, otherwise it will be the same as before. HMR will create snapshots incrementally so the performance impact is smaller.

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/mod.rs
+++ b/crates/rspack_core/src/cache/persistent/mod.rs
@@ -23,7 +23,7 @@ use crate::{
   Compilation, CompilerOptions,
 };
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Hash)]
 pub struct PersistentCacheOptions {
   pub build_dependencies: Vec<PathBuf>,
   pub version: String,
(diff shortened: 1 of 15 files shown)


LingyuCoder (approved):

LingyuCoder:
can this issue https://github.com/web-infra-dev/rspack/issues/9064 be closed?

jerrykingxyz:
> can this issue #9064 be closed?

No, this is another bug caused by storage invalidation, not snapshot invalidation. I will fix it in another PR.

PR #10057, opened 2025-04-16
feat(diagnostic): support minification error

h-a-n-a:
## Summary

This PR supports `StatsError.code` and `StatsWarning.code`:

- `ChunkMinificationError`: Generated when chunk minification encounters an error
- `ChunkMinificationWarning`: Generated when chunk minification encounters an warning
- `ModuleParseWarning`: Generated when module parsing encounters an warning
- `ModuleParseError`: Generated when module parsing encounters an error

The following example shows when minifcation encounters an error:
```
ERROR in main.js
  × Chunk minification failed:
  ╰─▶   × JavaScript parsing error: 'import', and 'export' cannot be used outside of module code
          ╭─[8:0]
        6 │   \******************/
        7 │ (function (module, exports) {
        8 │ import "./abc";
          · ──────
        9 │
       10 │
          ╰────
```

It's error code is `ChunkMinificationError`.

## Checklist


crates/rspack_core/src/diagnostics.rs
```diff
@@ -258,3 +243,36 @@ pub fn map_box_diagnostics_to_module_parse_diagnostics(
...
+    write!(f, "Chunk minification failed:")
+  }
+}
+
+impl miette::Diagnostic for MinifyError {
+  fn code<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
+    match self.severity().unwrap_or(miette::Severity::Error) {
+      miette::Severity::Advice => unreachable!("miette::Severity::Advice should not be used"),
+      miette::Severity::Warning => Some(Box::new("ChunkMinificationWarning")),
+      miette::Severity::Error => Some(Box::new("ChunkMinificationError")),
```
hardfist, line 269: I think all erorr_code should be maintained in a separate file like error_code.json, so it can be easily reviewed otherwise developers may not be aware of this is a public api

crates/node_binding/binding.d.ts
```diff
@@ -1198,6 +1198,7 @@ export interface JsStatsCompilation {
 export interface JsStatsError {
   moduleDescriptor?: JsModuleDescriptor
   message: string
+  code?: string
```
hardfist, line 1201: I'm not sure whether code should be string or number,and there're some drawback of using string as code
* if we make some typo in code, we can't just change it because it's a public api and may break user code like following

```js
// notice we made a typo here, but we can't just change it
if(error.code== 'ChunkMinificationErro') {
  
}
```
if we want to use string as code, maybe we should at least export these message
```js
if(error.code == RSPACK_ERROR_CODE.CHUNK_MINIFICATION_ERROR){
}
```

hardfist (approved):

crates/node_binding/binding.d.ts
h-a-n-a: > if we make some typo in code, we can't just change it because it's a public api and may break user code like following

IMO, this happens with numbers anyway.

hardfist: not sure how to typo number?

h-a-n-a: For example, typing 10001 as 10010. This is a similar problem. This is not a flaw unique to strings—any hardcoded value can suffer from this issue.

crates/rspack_core/src/diagnostics.rs
h-a-n-a: I would expose some constants. But for generating diagnostics, it's a whole new complex problem that couldn't be done without some hassle. https://github.com/microsoft/TypeScript/blob/069de743dbd17b47cc2fc58e1d16da5410911284/Herebyfile.mjs#L86

hardfist: maybe we don't need generating diagnostics for now, the errorType is limited now so manuallly maintain is acceptable

packages/rspack/src/stats/statsFactoryUtils.ts
```diff
@@ -159,8 +159,33 @@ export type KnownStatsModuleIssuer = {
 
 export type StatsModuleIssuer = KnownStatsModuleIssuer & Record<string, any>;
 
+export enum StatsErrorCode {
+	/**
+	 * Warning generated when either builtin `SwcJsMinimizer` or `LightningcssMinimizer` fails to minify the code.
+	 */
+	ChunkMinificationError = "ChunkMinificationError",
```
chenjiahan, line 166: Should we use `SwcMinificationError` and `LightningcssMinificationError`? So that users can know the error is throwed by SWC or lightmingcss, not Rspack

h-a-n-a: We can, but I don't know if we should. For example, `ModuleBuildError` or `ModuleParseError` code does not give the exact module type. The module type is located in `moduleIdentifier`. IMO, this also applies to `ChunkMinificationError`. It might not necessary to give the exact type.

chenjiahan: If we can add the minimizer type to the error message, it should be good enough

image

h-a-n-a: This makes me thinking about the current implementation of `StatsError`. 
If need to give user the ability to create there own stats printer. Then we should add a field `pluginName` and let users define their own `StatsPrinterPlugin`.
If we only want users to rely on the current `DefaultStatsPrinterPlugin`, we just need to make it clear by adding the plugin name into the error.

chenjiahan: Adding the plugin name to the error seems simpler..

h-a-n-a: An internal project in ByteDance is using `StatsError.code` and string matcher to match errors into structural object. We might just let them decide this part as most of the users don't really care about this type usage, including knowing what error belongs to what type and plugin.

chenjiahan: > most of the users don't really care about the type for each error, including knowing which error belongs to what type and plugin

I disagree with this. Letting normal users know that the minification bug is from SWC or lightningcss can help them debug or report issues.

h-a-n-a: This is not what I meant. When I was talking about "most of the users don't really care about the type for each error", what I'm actually talking about is, for **ADVANCED** users, adding the plugin name to the error is not an idea choice for them. This is in addition to basic usages. For basic usages, we just need to add the plugin name.

chenjiahan: get 👌

chenjiahan (approved):

h-a-n-a:
Rebased cc @chenjiahan

chenjiahan:
👍

PR #10063, opened 2025-04-17
perf: watcher incrementally adds or remove listening files

jerrykingxyz:
## Checklist

### DIFF
--- a/packages/rspack/src/node/NodeWatchFileSystem.ts
+++ b/packages/rspack/src/node/NodeWatchFileSystem.ts
@@ -46,29 +46,12 @@ export default class NodeWatchFileSystem implements WatchFileSystem {
 		) => void,
 		callbackUndelayed: (fileName: string, changeTime: number) => void
 	): Watcher {
-		if (!files || typeof files[Symbol.iterator] !== "function") {
-			throw new Error("Invalid arguments: 'files'");
+		if (this.watcherOptions !== options) {
+			// close old watcher, and create.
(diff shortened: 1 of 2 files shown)


packages/rspack/src/node/NodeWatchFileSystem.ts
```diff
@@ -46,29 +46,12 @@ export default class NodeWatchFileSystem implements WatchFileSystem {
 		) => void,
 		callbackUndelayed: (fileName: string, changeTime: number) => void
 	): Watcher {
-		if (!files || typeof files[Symbol.iterator] !== "function") {
-			throw new Error("Invalid arguments: 'files'");
+		if (this.watcherOptions !== options) {
```
chenjiahan, line 49: @LingyuCoder Can you help confirm if this might be causing some issues? I remember you mentioned that the previous implementation was for a reason

LingyuCoder: I'm not sure either, watchpack is too complex. Perhaps we should add an experimental option to control this. @jerrykingxyz

chenjiahan: Can we release a canary version and test it in some real world applications?

An experimental option seems to complex for our users.

chenjiahan:
Closed due to https://github.com/webpack/watchpack/pull/249 and https://github.com/web-infra-dev/rspack/pull/10658

PR #10073, opened 2025-04-17
Add minSizeReduction to SplitChunks CacheGroup

kbruccoleri:
## Summary

Fixes the following error:

Invalid configuration object. Rspack has been initialized using a configuration object that does not match the API schema.
- Unrecognized key(s) in object: 'minSizeReduction' at "optimization.splitChunks.cacheGroups.{group}"

that is caused by valid config:

```
const config = {
  ...
  optimization: {
    splitChunks: {
      ...
      vendor: {
        priority: 5,
        test: /[\\/]node_modules[\\/]/,
        minSize: 20000,
        minSizeReduction: 50000,
        reuseExistingChunk: true,
      },
    }
  }
};
```

## Checklist

Both of the following checks didn't seem pertinent:


JSerFeng (approved):

JSerFeng:
Thank u !

Issue #10056, opened 2025-04-16
[Bug]: RsBuild version 1.3.7. Crashes 


Dushik71:
### System Info

Crashes when trying to build if it is imported from an external css package that has @media in the middle of the text.

× Module build failed:
 ╰─▶ × Invalid media query at ...\node_modules\...\theme.css:10808:29

theme.css:
@media screen and (max-width: 64em) {
...
}

### Details

Version RSBuild 1.3.7
Crashes build 

### Reproduce Steps

yarn build

chenjiahan:
Related PR: https://github.com/web-infra-dev/rspack/pull/9979

chenjiahan:
Should be fixed by https://github.com/web-infra-dev/rspack/pull/10076

Also you can set ` errorRecovery: true` to fix this issue in Rspack v1.3.7, see https://rspack.dev/guide/features/builtin-lightningcss-loader#errorrecovery

```js
export default {
  module: {
    rules: [
      {
        test: /\.css$/,
        use: [
          {
            loader: 'builtin:lightningcss-loader',
            /** @type {import('@rspack/core').LightningcssLoaderOptions} */
            options: {
              errorRecovery: true,
            },
          },
        ],
      },
    ],
  },
};
```

Dushik71:
Thanks! It helped!

PR #10076, opened 2025-04-18
revert: lightningcss loader error recovery default value

SyMind:
## Summary

Revert #9979

Related issue: https://github.com/web-infra-dev/rspack/issues/10056

The syntax validation of lightning CSS is too strict and not user-friendly for Rspack users, so the default value for error recovery revert to true.

But when error recovery is true, the warning will give to the user now.

## Checklist


chenjiahan (commented):
Can you add some documentation to tell (this can be done in another PR):

- The default error recovery behavior of Rspack's lightningcss-loader
- An example for enabling error recovery in lightningcss-loader
- An example for ignore lightningcss warnings using https://rspack.dev/config/other-options#ignorewarnings

chenjiahan (approved):

PR #10080, opened 2025-04-18
refactor: improve render external module for module output

JSerFeng:
## Summary

Improve the rendering of ExternalModule.

Before:

```js
import * as __WEBPACK_EXTERNALS_foo__ from 'foo'

__WEBPACK_EXTERNALS_foo__.value
```

This PR:

```js
import { value } from 'foo'

value
```

## Design detail

Add a new concatenation hook named `concatenation_info`, this hook is used to let module prepare some data in the correct **topological** order, for example, moduleA, moduleB can use this hook to add their globally used names.

ExternalModule has global names because they will insert some import statements through `InitFragment`.

```
// a.js
import { a } from 'foo';
// b.js
import { a } from 'bar';
```

Tap this hook, and insert the `a` to the globally used names, so we can deconflict the conflict symbol name.

## Checklist


packages/rspack-test-tools/tests/configCases/library/modern-module-named-import-externals/rspack.config.js
```diff
@@ -0,0 +1,85 @@
...
+	},
+	plugins: [
+		function () {
+			const handler = compilation => {
+				compilation.hooks.processAssets.tap("testcase", assets => {
+					const source = assets["test.js"].source();
+					expect(source).toMatchInlineSnapshot(`
+				import { a, b } from "externals0";
+				import { a as a_0 } from "externals1";
+				import default_0 from "externals2";
```
fi3ework, line 38: The source code here before transform is

```js
import defaultValue from 'externals2'
```

Could we reserve the `defaultValue` and append index suffix to de-conflict, it will be much better for readability when default import is heavily used (just like how esbuild behaves now [esbuild - Try in the browser](https://esbuild.github.io/try/#YgAwLjI1LjIALS1idW5kbGUgLS1leHRlcm5hbDpsb2Rhc2ggLS1mb3JtYXQ9ZXNtAGUAZW50cnkuanMAaW1wb3J0IGxvZCBmcm9tICdsb2Rhc2gnCmltcG9ydCB7IHN1YiB9IGZyb20gJy4vZmlsZS5qcycKCmxvZC5hZGQoKQpzdWIoKQAAZmlsZS5qcwBpbXBvcnQgbG9kIGZyb20gJ2xvZGFzaCcKCmV4cG9ydCBjb25zdCBzdWIgPSBsb2Quc3Vi))

packages/rspack-test-tools/tests/configCases/library/modern-module-named-import-externals/rspack.config.js
```diff
@@ -0,0 +1,85 @@
...
+				import * as __WEBPACK_EXTERNAL_MODULE_externals3__ from "externals3";
```
fi3ework, line 39: I think we could remove `__WEBPACK_EXTERNAL_MODULE_` prefix, it's way too long.

packages/rspack-test-tools/tests/configCases/library/modern-module-named-import-externals/rspack.config.js
JSerFeng: Yes that'd be better and I've thought about it but encountered some issues to implement.
exportsInfo doesn't care about the imported name, a default value maybe imported as different names in other file, I thought the `getUsedName()` can get the used export name but it's not what I think. Maybe I can achieve this another PR

packages/rspack-test-tools/tests/configCases/library/modern-module-named-import-externals/rspack.config.js
JSerFeng: I think It's good, we can make it in another PR, as it will change a lots of snapshot, I want to keep this PR simpler

fi3ework: Got it.

fi3ework:
I manually checked test case of Rslib in eco-ci, changes are expected, the output code is much neater. 👍

fi3ework (approved):

PR #10097, opened 2025-04-21
fix(wasm): add wasm fallback in `checkVersion`

CPunisher:
## Summary

This eliminate errors if users only install `@rspack/binding-wasm32-wasi`. Usually users are aware of they are choosing wasm binding so I think we can avoid introducing very detailed messages.

## Checklist


h-a-n-a (approved):

h-a-n-a:
Thank you!

PR #10102, opened 2025-04-21
chore: fork tracing-chrome for better support for perfetto

hardfist:
## Summary
tracing-chrome crate has some limitation which cause it's hard to add better perfetto support for rspack, even maybe we can make pr to tracing-chrome but there may more custom behavior in the future which is coupled with rspack, so I think fork the implementation is better and we'll add perfetto support in following ways:

* support merge rust & js tracing file( the current tracing merge is too hacky, use regex to merge two json file to avoid parse large json file https://github.com/web-infra-dev/rspack/pull/10072/files#diff-b8118343bf83f27b63382f2a28551be6da81ef2bcbde3d5f88ece45a3f20b4f2R85

* support separate span in same track using `id2`(current overlap span will cause perfetto UI render wrong result, related to https://github.com/thoren-d/tracing-chrome/issues/24)

I don't separate this crate in separate repo because the custom behavior is highly coupled to rspack itself, and may needs lots of change in the future

## Checklist

### DIFF
--- a/.github/workflows/reusable-build-bench.yml
+++ b/.github/workflows/reusable-build-bench.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Install cargo-codspeed binary
         uses: taiki-e/install-action@be7c31b6745feec79dec5eb79178466c0670bb2d # v2
         with:
-          tool: cargo-codspeed@2.7.2
+          tool: cargo-codspeed@2.10.1
 
       - name: Build Benchmark
         env:
--- a/crates/rspack_tracing_chrome/README.md
+++ b/crates/rspack_tracing_chrome/README.md
@@ -1,0 +1,5 @@
+# rspack_tracing_chrome
+this is a fork of https://github.com/thoren-d/tracing-chrome to add better perfetto support for rspack like
+* support merge Rust & JS tracing
+* support use `id2` to different span in same track for perfetto, related to https://github.com/thoren-d/tracing-chrome/issues/24
+* ...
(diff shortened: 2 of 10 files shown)


crates/rspack_tracing_chrome/README.md
```diff
@@ -0,0 +1,5 @@
+# rspack_tracing_chrome
+this is a fork of https://github.com/thoren-d/tracing-chrome to add better perfetto support for rspack like
```
chenjiahan, line 2:
```suggestion

This is a fork of [tracing-chrome](https://github.com/thoren-d/tracing-chrome) to add better support for perfetto to Rspack like:

```

crates/rspack_tracing_chrome/README.md
```diff
@@ -0,0 +1,5 @@
...
+* support merge Rust & JS tracing
+* support use `id2` to different span in same track for perfetto, related to https://github.com/thoren-d/tracing-chrome/issues/24
+* ...
```
chenjiahan, line 5: Can we add some credits?

hardfist: in root README.md ?

chenjiahan: Add to this file is ok

```md

## Credits

Original [tracing-chrome](https://github.com/thoren-d/tracing-chrome) created by [thoren-d](https://github.com/thoren-d), licensed under the MIT License.
```

stormslowly (approved):

.github/workflows/reusable-build-bench.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       - name: Install cargo-codspeed binary
         uses: taiki-e/install-action@be7c31b6745feec79dec5eb79178466c0670bb2d # v2
         with:
-          tool: cargo-codspeed@2.7.2
+          tool: cargo-codspeed@2.10.1
```
hardfist, line 47: this fix cargo-codspeed error

Issue #9971, opened 2025-04-10
[Bug]: Vue builds errors in version 1.3.4


lzxb:
### System Info

  System:
    OS: macOS 15.3
    CPU: (8) arm64 Apple M1
    Memory: 104.00 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.nvm/versions/node/v22.11.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.11.0/bin/yarn
    npm: 10.9.0 - ~/.nvm/versions/node/v22.11.0/bin/npm
    pnpm: 10.6.3 - ~/.nvm/versions/node/v22.11.0/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  Browsers:
    Safari: 18.3

### Details

```log
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  Formatting argument out of range
Location: crates/rspack_error/src/graphical.rs:700

Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 1: _blake3_compress_in_place_portable
    at <unknown source file>
 2: _napi_register_module_v1
    at <unknown source file>
 3: _napi_register_module_v1
    at <unknown source file>
 4: _napi_register_module_v1
    at <unknown source file>
 5: _napi_register_module_v1
    at <unknown source file>
 6: _wasmer_vm_f64_nearest
    at <unknown source file>
 7: _napi_register_module_v1
    at <unknown source file>
 8: _napi_register_module_v1
    at <unknown source file>
 9: _napi_register_module_v1
    at <unknown source file>
10: _napi_register_module_v1
    at <unknown source file>
11: _napi_register_module_v1
    at <unknown source file>

```

### Reproduce link

https://github.com/esmnext/esmx/tree/767240b7b8a263d6900cfe619e621c235b244e87

### Reproduce Steps

Run
```bash
./build.sh
cd examples/ssr-demo-vue2 
RUST_BACKTRACE=full pnpm build
```

PR #10109, opened 2025-04-21
fix(diagnostic): fix excessively large width causing `format_args` panics

h-a-n-a:
## Summary

This PR fixes the panic when calling `format_args!` with excessively large width (> u16::MAX), which is an constraint introduced in rust PR https://github.com/rust-lang/rust/issues/99012.

This is not an ideal way for handling a minimized source code. The way of printing diagnostics of minimized source remains to be discussed.

closes https://github.com/web-infra-dev/rspack/issues/9971

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/__snapshots__/NewCodeSplitting-stats-output.test.js.snap
+++ b/packages/rspack-test-tools/tests/__snapshots__/NewCodeSplitting-stats-output.test.js.snap
@@ -64,8 +64,7 @@ Rspack compiled successfully (27eddb11e03f28d2)
 exports[`new code splitting stats output new code splitting stats output/builtin-swc-loader-parse-error should print correct stats for: NewCodeSplittingStatsOutput 1`] = `
 ERROR in ./index.ts
   × Module build failed:
-  ├─▶   ×
-  │     │   x Expected '{', got 'error'
+  ├─▶   ×   x Expected '{', got 'error'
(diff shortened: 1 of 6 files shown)


packages/rspack-test-tools/tests/__snapshots__/NewCodeSplitting-stats-output.test.js.snap
```diff
@@ -64,8 +64,7 @@ Rspack compiled successfully (27eddb11e03f28d2)
 exports[`new code splitting stats output new code splitting stats output/builtin-swc-loader-parse-error should print correct stats for: NewCodeSplittingStatsOutput 1`] = `
 ERROR in ./index.ts
   × Module build failed:
-  ├─▶   ×
-  │     │   x Expected '{', got 'error'
+  ├─▶   ×   x Expected '{', got 'error'
```
chenjiahan, line 67: There are two `× x` in one line, can we dedupe them?

h-a-n-a: Two errors, nested one is formatted by swc, the outer one is formatted by us.

chenjiahan: ok, seems hard to handle

jtsorlinis:
This PR has broken our builds, we now get
image

PR #10113, opened 2025-04-22
chore: trivial static variable improve

quininer:
## Summary

This is mostly some trivial optimizations I found while investigating `static` usage, and I open a separate PR to submit it.

## Checklist


crates/rspack_plugin_javascript/src/parser_plugin/check_var_decl.rs
```diff
@@ -9,7 +9,7 @@ use swc_core::{
 use super::JavascriptParserPlugin;
 use crate::visitors::{create_traceable_error, JavascriptParser};
 
-static STRICT_MODE_RESERVED_WORDS: LazyLock<FxHashSet<String>> = LazyLock::new(|| {
+static STRICT_MODE_RESERVED_WORDS: LazyLock<FxHashSet<&'static str>> = LazyLock::new(|| {
```
quininer, line 12: Because there is so little data, I think using `.binary_search` is also fast enough, which can save the overhead of `LazyLock` and `HashSet`. But I haven't really tested it, so keep original logic.

SyMind (approved):

PR #10119, opened 2025-04-22
fix(core): fix async event tracing showing in perfetto

hardfist:
### `id2`
It's harder for trace viewer(like perfetto) to explain overlap async event with same `name` and `id` and `cat`
```json
[
  {
    "name": "parse",
    "ph": "b",
    "ts": 0,
    "id": "1",
    "pid": 1,
    "tid": 1
  },
  {
    "name": "parse",
    "ph": "e",
    "ts": 4000,
    "id": "1",
    "pid": 1,
    "tid": 1
  },
  {
    "name": "parse",
    "ph": "b",
    "ts": 2000,
    "id": "1",
    "pid": 1,
    "tid": 1
  },
  {
    "name": "parse",
    "ph": "e",
    "ts": 5000,
    "id": "1",
    "pid": 1,
    "tid": 1
  }
]
```
it can explain the json in two different ways
* first
image
* second
image
chrome trace event supports use ["id2"](https://docs.google.com/document/d/1CvAClvFfyA5R-PhYUmn5OOQtYMH4h6I0nSsKchNAySU/preview?tab=t.0#heading=h.l7g30knytzzc) to different span, so we use "id2" instead to control how perfetto show overlap, it's not only useful to specify range but also can bind span with different name into same flamechart(we choose to bind all build phase tracing to resource for better view)
### Separate Build Overview and Build Detail Track
since the tracing contains too much information(especially for module.build phase which contains thousands of async events for large projects), it looks better that we split track to overview track(which covers important phase of compilation) and detail track(which covers detailed info about all module build process, we may add more detailed track like for code splitting and memory in the future)

image

### Know Limitations
* Perfetto UI: It seems perfetto's UI is not perfect for showing large async events which may be harder to read for too many build events for large projects compared with turbopack's trace viewer, but we can still search in it and we can build our own trace viewer in the future if https://github.com/google/perfetto/issues/1333 is not supported
* trace.json limitation: the current usage of trace.json is kind of hacky, since trace.json itself is second class support(legacy) in perfetto, we may switch to perfetto proto if trace.json doesn't meet more needs in the future.

## Checklist

### DIFF
--- a/crates/rspack_core/src/normal_module.rs
+++ b/crates/rspack_core/src/normal_module.rs
@@ -278,7 +278,11 @@ impl NormalModule {
     &mut self.presentational_dependencies
   }
 
-  #[tracing::instrument("NormalModule:build_hash", skip_all)]
+  #[tracing::instrument(
+    "NormalModule:build_hash", skip_all,fields(
+    id2 = self.resource_data.resource.as_str()
+  )
+)]
   fn init_build_hash(
     &self,
     output_options: &OutputOptions,
@@ -364,6 +368,7 @@ impl Module for NormalModule {
   #[tracing::instrument("NormalModule:build", skip_all, fields(
     module.resource = self.resource_resolved_data().resource.as_str(),
     module.identifier = self.identifier().as_str(),
+    id2 = self.resource_data.resource.as_str(),
     module.loaders = ?self.loaders.iter().map(|l| l.identifier().as_str()).collect::<Vec<_>>())
   )]
   async fn build(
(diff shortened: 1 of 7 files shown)


crates/rspack_core/src/normal_module.rs
```diff
@@ -364,6 +368,7 @@ impl Module for NormalModule {
   #[tracing::instrument("NormalModule:build", skip_all, fields(
     module.resource = self.resource_resolved_data().resource.as_str(),
     module.identifier = self.identifier().as_str(),
+    id2 = self.resource_data.resource.as_str(),
```
h-a-n-a, line 371: Is it possible to make it working agnostic to tracing frameworks? This does not look quite reasonable here.

hardfist: anyway we need two id, one id to bound to same compilation and another id to bound to module resource for module-releated events, so no matter what we tracing framework we choose we do need manually attach special id(it's just called id2 in our case now), I'm open to more general naming

stormslowly (approved):

PR #10121, opened 2025-04-22
perf: define Rspack version to reduce require calls

chenjiahan:
## Summary

The `rspackVersion` and `webpackVersion` are static values and can be defined as env vars in `rslib.config.ts`, this helps to reduce `require()` calls and make the startup faster.

The `release-canary.yml` has been updated to build JS packages after updating the package.json#verison. Tested in https://github.com/web-infra-dev/rspack/actions/runs/14592747054.

## Checklist


chenjiahan:
@h-a-n-a There are two CI issues:

1. Lint CI is not triggered correctly when modifying `*.cjs`
2. Lint CI will not prevent merge even if it fails (does not set as `required`)

@stormslowly is working on resolving them 😄

h-a-n-a:
> @h-a-n-a There are two CI issues:
> 
> 1. Lint CI is not triggered correctly when modifying `*.cjs`
> 2. Lint CI will not prevent merge even if it fails (does not set as `required`)
> 
> @stormslowly is working on resolving them 😄

Thanks for the heads-up. IMO, part of the issue is located at lint-staged. Maybe we didn't trigger it locally.

SoonIter (dismissed):

Timeless0911 (approved):

PR #10122, opened 2025-04-22
perf(core): remove json-parse-even-better-errors dependency

chenjiahan:
## Summary

This PR removes the `json-parse-even-better-errors` dependency to remove ~100 loc from the `dist/index.js`.

I think this package can be removed due to two reasons:

1. The "loader string option parsing error" is a rare case, and I think few users will encounter this error.
2. `json-parse-even-better-errors` does not seem to improve the readability of the error, but makes the log more complicated:

```js
import jsonParse from "json-parse-even-better-errors";

const invalidJson = "{foo:1,bar:2,baz:`}";

try {
  console.log("=== JSON.parse ===");
  JSON.parse(invalidJson);
} catch (e) {
  console.log(e);
  console.log('\n');
}

try {
  console.log("=== json-parse-even-better-errors ===");
  jsonParse(invalidJson);
} catch (e) {
  console.log(e);
}
```

image
 
## Checklist

### DIFF
--- a/packages/rspack/package.json
+++ b/packages/rspack/package.json
@@ -56,7 +56,6 @@
     "cross-env": "^7.0.3",
     "enhanced-resolve": "5.18.1",
     "graceful-fs": "^4.2.11",
-    "json-parse-even-better-errors": "^3.0.2",
     "prebundle": "^1.3.3",
     "tsx": "^4.19.3",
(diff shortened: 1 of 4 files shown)


h-a-n-a (approved):
Looks removable

Issue #9931, opened 2025-04-07
[Bug]: loaderContext.import module with inline loader causes rust panic


dtothefp:
### System Info

System: Apple M3 Max Sequoia 15.3.2 
Binaries: binding-darwin-arm64
NPM: @rspack/(core|cli) 1.3.2

### Details

This looks related to https://github.com/web-infra-dev/rspack/issues/8536 and seems to be fixed in https://github.com/web-infra-dev/rspack/pull/8590/files but maybe this issue is related to using the[ inline loader syntax ](https://rspack.dev/api/loader-api/inline).

When using `loaderContext.importModule` with an inline loader a rust panic occurs that is not caught by the callback / returned promise.

```
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
Message:  should have module
Location: crates/rspack_core/src/compiler/module_executor/execute.rs:80

Backtrace omitted.

Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

### Reproduce link

https://github.com/web-infra-dev/rspack-repro/compare/main...dtothefp:rspack-repro:dfp-loader-context-import-module

### Reproduce Steps

In the repo above run `pnpm install && pnpm run build`

PR #10136, opened 2025-04-23
fix: panic when factorize failed in module executor

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/9931

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/module_executor/execute.rs
+++ b/crates/rspack_core/src/compiler/module_executor/execute.rs
@@ -72,6 +72,45 @@ impl Task<MakeTaskContext> for ExecuteTask {
     compilation.plugin_driver = compilation.buildtime_plugin_driver.clone();
 
     let id = EXECUTE_MODULE_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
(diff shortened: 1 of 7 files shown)


crates/rspack_core/src/compiler/module_executor/execute.rs
```diff
@@ -72,6 +73,50 @@ impl Task<MakeTaskContext> for ExecuteTask {
     compilation.plugin_driver = compilation.buildtime_plugin_driver.clone();
 
     let id = EXECUTE_MODULE_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+    let errors = compilation
+      .make_artifact
+      .diagnostics()
+      .into_iter()
+      .filter(|d| matches!(d.severity(), RspackSeverity::Error))
+      .collect_vec();
```
quininer, line 81: There doesn't seem to be any consumption of `errors` vec, maybe we can use `.any()` instead of `.filter()` + `.collect_vec()`

quininer (changes requested):

quininer (approved):

Issue #9699, opened 2025-03-17
[Bug]: Build hangs on error when setting api: "modern-compiler" in sass-loader


Thiry1:
### System Info

  System:
    OS: macOS 14.7
    CPU: (8) arm64 Apple M2
    Memory: 90.69 MB / 24.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 22.14.0 - ~/.nodebrew/current/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.9.2 - ~/.nodebrew/current/bin/npm
    pnpm: 8.15.3 - ~/.nodebrew/current/bin/pnpm
  Browsers:
    Chrome: 134.0.6998.89
    Chrome Canary: 136.0.7072.0
    Edge: 134.0.3124.68
    Safari: 18.0
  npmPackages:
    @rspack/cli: ^1.2.8 => 1.2.8 
    @rspack/core: ^1.2.8 => 1.2.8 

### Details

Build hangs on error when setting api: "modern-compiler" in sass-loader.
In webpack, the process exits when an error occurs.
Removing the following setting allows the process to exit in rspack as well:
```diff
{
  loader: "sass-loader",
  options: {
-    api: "modern-compiler",
     implementation: require("sass-embedded"),
  }
}
```

### Reproduce link

https://github.com/Thiry1/rspack-bug-repro/compare/hang-sass-loader

### Reproduce Steps

1. clone repro repo
2. git switch hang-sass-loader
3. npm i
4. npm run build:rspack

kei-sakamoto-dwango:
I suspect that the issue is caused by the shutdown hook not being called on build errors, preventing sass-embedded from being disposed of.

I have confirmed that when Sass builds successfully, this hook is called and sass-embedded is properly disposed of.
https://github.com/webpack-contrib/sass-loader/blob/300ac87f3cd04183fd0d6ec9b931d52764d8ca64/src/utils.js#L775-L777

Thiry1:
I suspect that the issue is caused by the shutdown hook not being called on build errors, preventing sass-embedded from being disposed of.

I have confirmed that when Sass builds successfully, this hook is called and sass-embedded is properly disposed of.
https://github.com/webpack-contrib/sass-loader/blob/300ac87f3cd04183fd0d6ec9b931d52764d8ca64/src/utils.js#L775-L777

kei-sakamoto-dwango:
[...]

PR #10154, opened 2025-04-24
fix: should call `compiler.close` when having errors

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/9699

Persistent cache of rspack is implemented in Rust and should always call `compiler.close` after building.

## Checklist


chenjiahan (approved):
LGTM 👍

Issue #9635, opened 2025-03-12
[Bug]: val-loader and val-loader-modernizr fails


kamranayub:
### System Info

```
  System:
    OS: Windows 10 10.0.26100
    CPU: (12) x64 AMD Ryzen 5 5600X 6-Core Processor
    Memory: 24.87 GB / 63.91 GB
  Binaries:
    Node: 20.15.1 - ~\AppData\Local\fnm_multishells\78476_1741742697525\node.EXE
    npm: 10.7.0 - ~\AppData\Local\fnm_multishells\78476_1741742697525\npm.CMD
    pnpm: 9.9.0 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Spartan (38.35), Chromium (38.3), ChromiumDev (38.21)
    Internet Explorer: 11.0.26100.1882
  npmPackages:
    @rspack/cli: ^1.2.8 => 1.2.8
    @rspack/core: ^1.2.8 => 1.2.8
```

### Details

I was testing out rspack moving from Webpack and I use `val-loader` together with https://www.npmjs.com/package/val-loader-modernizr

When I execute the build, I get this error:

```sh
> rspack -c rspack.config.js

ERROR in ./scripts/.modernizrrc
  × Error:  on JsLoaderContext.additionalData
  │     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)


Rspack compiled with 1 error in 58.20 s
```

The time taken probably reflects my `ts-loader` as it takes about 60 seconds to typecheck.

The rule looks like:

```js
{
                test: /\.modernizrrc$/,
                use: [{
                    loader: 'val-loader',
                    options: {
                        executableFile: require.resolve('val-loader-modernizr')
                    }
                }]
}
```

Note: In ESM module in repro, I use `fileURLToPath(import.meta.resolve("val-loader-modernizr"))` but in my project I'm still using CommonJS config.

### Reproduce link

https://github.com/kamranayub/rspack-bug-val-loader

### Reproduce Steps

1. `pnpm install` (or `bun install`)
2. `npm run build` or (`pnpm run build`)
3. Observe error

kamranayub:
I also tried the minimal Modernizr example here: https://www.npmjs.com/package/val-loader#modernizr

Same deal, just FYI.

kamranayub:
It seems like this one works!

https://www.npmjs.com/package/@sect/modernizr-loader

[...]

PR #10155, opened 2025-04-24
fix: pass `null` to additional data of loader context

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/9635

same as https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/loader-runner/index.ts#L1023

## Checklist


h-a-n-a:
I don't quite get this issue. Would you please point me out the exact case we're on?

LingyuCoder:
> I don't quite get this issue. Would you please point me out the exact case we're after? I can't grab much information from the test case in this PR.

The val-loader will pass `null` to `loaderContext.additionalData` and `null` can not be transform from node binding (only accept `undefined` as empty value).

PR #10164, opened 2025-04-24
fix(parallel-loader): should not pass options for which loaders are not set to `parallel`

h-a-n-a:
## Summary

Worker threads would execute loaders until it's not set to `parallel` or it's a builtin loader that should be yielded to rust.
Previously, loader runner would pass loaders with `options` into worker thread without checking if the loader enabled `parallel`. In this PR, only `options` of which loader is set to `parallel` are passed. 

## Checklist


chenjiahan (approved):
LGTM

PR #10168, opened 2025-04-24
chore: test runner

deanjingshui:
see title

## Checklist


.github/workflows/pr-labeler.yaml
```diff
@@ -24,3 +24,12 @@ jobs:
           enable-versioned-regex: 0
           include-title: 1
           sync-labels: 1
+
+  check-linux-runner:
```
jerrykingxyz, line 28: should add this to .github/workflows/ci.yml.

PR #10204, opened 2025-04-27
test: clean full snapshoted stats api test cases

LingyuCoder:
## Summary

Using stats output instead of full snapshot to test stats

## Checklist


chenjiahan (approved):
👍

PR #10220, opened 2025-04-28
feat: support loaderContext.utils access swc api

GiveMe-A-Name:
## Summary

- Add loaderContext.utils.swc to helper loader editor can access swc api from this.utils.swc
## Checklist

### DIFF
--- a/packages/rspack/etc/core.api.md
+++ b/packages/rspack/etc/core.api.md
@@ -293,7 +293,7 @@ interface AssignmentPattern extends PatternBase {
 }
 
 // @public
-interface AssignmentPatternProperty extends Node_4, HasSpan {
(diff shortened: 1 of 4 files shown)


chenjiahan (commented):
I just checked webpack documentation and found that webpack exposes `loaderContext.webpack` to allow loader developers to access the webpack instance. See https://webpack.js.org/api/loaders/#thiswebpack

Maybe we can expose `loaderContext.rspack` so that we do not need to add `loaderContext.util.swc`, users can use `loaderContext.rspack.experiments.swc` instead.

@hardfist cc

GiveMe-A-Name:
I think using `this._compiler.rspack.swc` is better then `this.util.swc`

PR #10222, opened 2025-04-28
fix: copy zod types instead of bundling them

Timeless0911:
## Summary

In https://github.com/web-infra-dev/rspack/pull/9899, I use default import to import `zod` to make types clear like below.

image

But if user use named import in daily development, the types will be emitted like:

image

And there exists an ast-grep issue when we try to redirect the import path, see https://github.com/ast-grep/ast-grep/issues/1959

Since the named import is the recommended one, so this PR, we use named import to import zod uniformly, and modify the prebundle script to copy the dts file directly instead of using `rollup-plugin-dts` for bundling to avoid the ast-grep issue and make the types simple and clear.

> BTW, zod use namespace and `rollup-plugin-dts` has some issues for namespace support, so we just use copy when prebundle.

## Checklist

### DIFF
--- a/packages/rspack/prebundle.config.mjs
+++ b/packages/rspack/prebundle.config.mjs
@@ -9,7 +9,10 @@ const __dirname = dirname(__filename);
 /** @type {import('prebundle').Config} */
 export default {
 	dependencies: [
-		"zod",
+		{
+			name: "zod",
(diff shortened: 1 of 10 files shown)


chenjiahan (approved):
LGTM

PR #10224, opened 2025-04-28
perf: try add cache for hotpath function

hardfist:
## Summary
try add cache for hotpath function

## Checklist


h-a-n-a (dismissed):
`Stats.toJson()` test got about `7%` performance improvement. LGTM

hardfist:
> `Stats.toJson()` test got about `7%` performance improvement. LGTM

but it will have memory leak issue, since it never evict and I'm not sure whether it's a issue

Issue #9724, opened 2025-03-18
[Feature]: Display diagnostic title in error logs


chenjiahan:
### What problem does this feature solve?

Rspack can display diagnostic title in error logs to help users quickly locate the source of the problem.

- Example:

https://github.com/web-infra-dev/rspack/blob/7045e4827ccb255dc39afe0f1cd93189668300ad/crates/rspack_plugin_sri/src/lib.rs#L114-L118

- Current:

![Image](https://github.com/user-attachments/assets/19ae3ebe-fc4c-4fc2-9150-5ebe8f456d93)

### What does the proposed API of configuration look like?

- Expected:

```bash
ERROR in × SubresourceIntegrity: rspack option output.crossOriginLoading not set, code splitting will not work!
```

Issue #10081, opened 2025-04-18
[Feature]: Support printing namespaces for plugins and loaders in diagnostic


h-a-n-a:
### What problem does this feature solve?

Support printing namespaces for plugins and loaders in diagnostic.

### What does the proposed API of configuration look like?

Does not expose public APIs.

SyMind:
Closed by https://github.com/web-infra-dev/rspack/pull/12213

PR #10227, opened 2025-04-28
refactor: print diagnostic code in front of message

LingyuCoder:
## Summary

close https://github.com/web-infra-dev/rspack/issues/9724
perhaps relate to https://github.com/web-infra-dev/rspack/issues/10081

Print diagnostic code in front of message. This enables us to quickly identify the scope where the error lies.

## Checklist

### DIFF
--- a/crates/rspack_core/src/diagnostics.rs
+++ b/crates/rspack_core/src/diagnostics.rs
@@ -46,14 +46,14 @@ impl std::error::Error for ModuleBuildError {
 }
 
 impl std::fmt::Display for ModuleBuildError {
-  fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-    write!(f, "Module build failed:")
+  fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    Ok(())
   }
 }
 
 impl miette::Diagnostic for ModuleBuildError {
   fn code<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
-    Some(Box::new("ModuleBuildError"))
+    Some(Box::new("Module build failed"))
   }
   fn severity(&self) -> Option<miette::Severity> {
     self.0.severity()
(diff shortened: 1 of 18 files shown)


crates/rspack_core/src/diagnostics.rs
```diff
@@ -46,14 +46,14 @@ impl std::error::Error for ModuleBuildError {
...
-    write!(f, "Module build failed:")
+  fn fmt(&self, _f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    Ok(())
   }
 }
 
 impl miette::Diagnostic for ModuleBuildError {
   fn code<'a>(&'a self) -> Option<Box<dyn std::fmt::Display + 'a>> {
-    Some(Box::new("ModuleBuildError"))
+    Some(Box::new("Module build failed"))
```
h-a-n-a, line 56: Diagnostic code should be `PascalCase`.

crates/rspack_plugin_copy/src/lib.rs
```diff
@@ -526,7 +523,7 @@ impl CopyRspackPlugin {
             .lock()
             .expect("failed to obtain lock of `diagnostics`")
             .push(Diagnostic::error(
-              "CopyRspackPlugin Error".into(),
+              "CopyPlugin".into(),
```
h-a-n-a, line 526: Both `Diagnostic::warn` and `Diagnostic::error` are legacy APIs. So the `title` is not the same as which in the `TraceableError`. I prefer to align this part with `TraceableError`.

h-a-n-a:
IMO, If we need to print the `Error.code` to user, which is not currently available in rspack, we might need to add something new to `DefaultStatsXxxxPlugin` API instead of changing the current error code and stuff.

LingyuCoder:
> IMO, If we need to print the `StatsError.code` to user, which is not currently available in rspack, we might need to add something new to `DefaultStatsXxxxPlugin` API instead of changing the current error code and stuff. This has partly been done in this PR. #10057

Ok, I will close this PR and perhaps do some work after #10057 is merged.

h-a-n-a:
> > IMO, If we need to print the `StatsError.code` to user, which is not currently available in rspack, we might need to add something new to `DefaultStatsXxxxPlugin` API instead of changing the current error code and stuff. This has partly been done in this PR. #10057
> 
> Ok, I will close this PR and perhaps do some work after #10057 is merged.

LGTM

Issue #10215, opened 2025-04-27
[Bug]: MF shared dependencies are resolved differently


jbroma:
### System Info

```
System:
  OS: macOS 15.3.2
  CPU: (10) arm64 Apple M1 Pro
  Memory: 186.83 MB / 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 20.13.1 - ~/Library/Caches/fnm_multishells/36922_1745753953673/bin/node
  npm: 10.5.2 - ~/Library/Caches/fnm_multishells/36922_1745753953673/bin/npm
  pnpm: 9.15.2 - ~/Library/Caches/fnm_multishells/36922_1745753953673/bin/pnpm
  bun: 1.1.34 - /opt/homebrew/bin/bun
  Watchman: 2024.11.04.00 - /opt/homebrew/bin/watchman
Browsers:
  Chrome: 135.0.7049.115
  Safari: 18.3.1
npmPackages:
  @rspack/cli: ^1.3.7 => 1.3.7 
  @rspack/core: ^1.3.7 => 1.3.7 
```

### Details

Adding dependency as shared causes a resolution error which was not present before and results in a build error. Removing the MF Plugin (or simply removing the dependency from shared) results in correct resolution and inclusion of dependency into the bundle.

This bug is not present on `@rspack/core` <= `1.3.4` (version `1.3.5` is the first with this regression)

### Reproduce link

https://github.com/jbroma/rspack-mf-resolve-repro

### Reproduce Steps

1. `pnpm i`
2. `pnpm build:rspack`

jbroma:
Thank you @ahabhgk for taking care of this so promptly 🙏

PR #10231, opened 2025-04-28
fix: clever merge resolve options

ahabhgk:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/10215

which is introduced by https://github.com/web-infra-dev/rspack/pull/10021, a wrong fix about merge resolve options by dependency, the real cause is the bug on `clever_merge`

## Checklist

### DIFF
--- a/crates/rspack_core/src/normal_module_factory.rs
+++ b/crates/rspack_core/src/normal_module_factory.rs
@@ -331,8 +331,6 @@ impl NormalModuleFactory {
           dependency_type: &dependency_type,
           dependency_category: &dependency_category,
           span: dependency_source_span,
-          // take the options is safe here, because it
-          // is not used in after_resolve hooks
           resolve_options: data.resolve_options.clone(),
           resolve_to_context: false,
           optional: dependency_optional,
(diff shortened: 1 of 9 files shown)


crates/rspack_core/src/normal_module_factory.rs
```diff
@@ -666,9 +659,7 @@ impl NormalModuleFactory {
         }
       }
     }
-    resolved
-      .map(|r| r.merge_by_dependency(dependency_type))
-      .map(Arc::new)
+    resolved.map(Arc::new)
```
ahabhgk, line 662: the merge is done at here https://github.com/web-infra-dev/rspack/blob/ff40041f2d2a864db21c1c1bfbe1d4c924007f81/crates/rspack_core/src/resolver/resolver_impl.rs#L201, and for the `worker-alias` test case it's actually have nothing related with `byDependency`, there isn't something like this in the config:

```js
module: {
  rules: [{
    test: /\.js/,
    resolve: {
      byDependency: {
        worker: ...
      }
    }
  }]
}
```

LingyuCoder (approved):

PR #10243, opened 2025-04-29
ci: remove stale workflow

LingyuCoder:
## Summary

Remove the workflow which is used to add `stale` label to issue and PR.

## Checklist


chenjiahan (approved):
LGTM, this workflow causes more noise than benefit

PR #10246, opened 2025-04-29
feat(wasm): support `tracing-chrome` in wasm

CPunisher:
## Summary

Debugging wasm is very painful. So I hope this pr could help us find where the wasm crushes or get stuck. There are two things to do:

- [x] `tracing-chrome` send the writer to a isolated thread to write the traces. However, threads in wasm are actually node workers. They can't share file descriptors with each other https://github.com/nodejs/node/issues/30507. The solution is passing a writer creator instead of a writer when building chrome tracer.
- [x] `cleanup` function is called by `exit-hook` and sends an flush message to the tracing thread. This means `cleanup` function is called when there's no active handlers (workers) running, which means the chrome tracing thread is not running. So we may choose another way to call `cleanup`. @hardfist @h-a-n-a Do you have any suggestion? 

This pr can be holding on until there is a good solution.

## Checklist

### DIFF
--- a/crates/rspack_tracing/src/chrome.rs
+++ b/crates/rspack_tracing/src/chrome.rs
@@ -18,7 +18,7 @@ impl Tracer for ChromeTracer {
       .trace_style(TraceStyle::Async)
       .include_args(true)
       .category_fn(Box::new(|_| "rspack".to_string()))
-      .writer(move || trace_writer.writer())
+      .writer(trace_writer.writer())
       .build();
     self.guard = Some(guard);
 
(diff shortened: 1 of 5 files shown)


h-a-n-a:
> cleanup function is called by exit-hook and sends an flush message to the tracing thread. This means cleanup function is called when there's no active handlers (workers) running, which means the chrome tracing thread is not running. So we may choose another way to call cleanup.

I'm confused of what kind of problem you're resolving. What is wasm tracing cleanup related to? Does it mean that cleanup function should be invoked before workers are destroyed?

CPunisher:
> If that is the case, is it possible for you to leverage `rspack.experiments.globalTrace.cleanup` to eagerly flush the tracer?

I think so. Just now I realize that `cleanupGlobalTrace` just flushes the writer rather than shutdowns everything. So we can call it anywhere.

h-a-n-a (approved):

h-a-n-a:
Thanks!

PR #10257, opened 2025-04-29
chore: change RA targetdir

hardfist:
## Summary
RA will call cargo check which share same lock with cargo build, so change RA's targetDir to avoid conflict 

## Checklist


quininer (approved):
LGTM

PR #10268, opened 2025-05-04
feat: compilation.addEntry support

henryqdineen:
## Summary

In our webpack-based internal framework we use `compilation.addEntry()` to conditionally add entry points based on configuration and some file system conventions. I am aware that dynamic entrypoints exist but we have found that `compilation.addEntry()` to be more powerful and allow better encapsulation using plugins. I saw that `compilation.addInclude()` was implemented and it seemed like `addEntry()` could also be. 

This PR adds support for the `compilation.addEntry()` entry and is mostly based off of the existing `addInclude()` implementation. They _could_ probably share more code but I'm not sure if that is wanted. 

Thanks!

## Checklist

- [x] Tests updated (or not required). - added `configCases/compilation/add-entry` test
- [x] Documentation updated (or not required). - ran API extractor


SyMind (approved):

watsonhaw5566:
@henryqdineen 你好，我想问一下 其实这里的 compilation.addEntry 算不算 ompilation.addInclude 的别名？

PR #10289, opened 2025-05-07
fix(types): export PitchLoaderDefinitionFunction type

henryqdineen:
## Summary

When building a loader I noticed that `PitchLoaderDefinitionFunction` is not exported from `@rspack/core` so I updated the code to fix that. I see some JSDoc comments reference that type so this is technically a fix for those. Example:

https://github.com/web-infra-dev/rspack/blob/87ff2360a85d17d754868c112f7c77d7ec85e2a7/tests/webpack-test/configCases/css/css-auto/loader.js#L1-L2

Thanks!

## Checklist


chenjiahan (approved):
Thank you ❤️

PR #10290, opened 2025-05-07
feat(create-rspack): better default browserslist target

vegerot:
Summary: The current browserslist default has two problems
1. It only covers [31% of users](https://browsersl.ist/#q=chrome+%3E%3D+87%2C+edge+%3E%3D+88%2C+firefox+%3E%3D+78%2C+safari+%3E%3D+14)
2. The versions are hardcoded, so needs to be manually updated

This commit changes the default browserslist to dynamically update as browser usage changes, and moves coverage from 31% to [91%](https://browsersl.ist/#q=last+2+versions%2C+Firefox+ESR%2C+%3E+0.2%25%2C+not+dead)

### DIFF
--- a/packages/create-rspack/template-react-js/rspack.config.mjs
+++ b/packages/create-rspack/template-react-js/rspack.config.mjs
@@ -8,7 +8,7 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const isDev = process.env.NODE_ENV === "development";
 
 // Target browsers, see: https://github.com/browserslist/browserslist
-const targets = ["chrome >= 87", "edge >= 88", "firefox >= 78", "safari >= 14"];
+const targets = ["last 2 versions", "> 0.2%",  "not dead",  "Firefox ESR"];
 
 export default defineConfig({
 	context: __dirname,
--- a/packages/create-rspack/template-react-ts/rspack.config.ts
+++ b/packages/create-rspack/template-react-ts/rspack.config.ts
@@ -5,7 +5,7 @@ import * as RefreshPlugin from "@rspack/plugin-react-refresh";
 const isDev = process.env.NODE_ENV === "development";
 
 // Target browsers, see: https://github.com/browserslist/browserslist
-const targets = ["chrome >= 87", "edge >= 88", "firefox >= 78", "safari >= 14"];
+const targets = ["last 2 versions", "> 0.2%",  "not dead",  "Firefox ESR"];
 
 export default defineConfig({
 	context: __dirname,
--- a/packages/create-rspack/template-vanilla-js/rspack.config.mjs
+++ b/packages/create-rspack/template-vanilla-js/rspack.config.mjs
@@ -2,7 +2,7 @@ import { defineConfig } from "@rspack/cli";
 import { rspack } from "@rspack/core";
 
 // Target browsers, see: https://github.com/browserslist/browserslist
-const targets = ["chrome >= 87", "edge >= 88", "firefox >= 78", "safari >= 14"];
+const targets = ["last 2 versions", "> 0.2%",  "not dead",  "Firefox ESR"];
 
 export default defineConfig({
 	entry: {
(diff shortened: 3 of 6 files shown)


chenjiahan:
Sorry, this was not an expected change, here are the reasons:

1. "It only covers 31% of users" is not true, browsersl.ist calculates incomplete coverage. For example, chrome >= 87 actually covers ChromeAndroid >= 87 (44.7%), which browsersl.ist ignores.

2. Rspack's current default browserslist is the same as [Rsbuild](https://rsbuild.dev/guide/advanced/browserslist#web-target) and [Vite](https://vite.dev/config/build-options.html#build-target), which is in line with common practice in the community.

vegerot:
1. Fair enough.  You're right that the current has more coverage.
2. I disagree. Just because Vite does it doesn't mean it's good.
   * All of the browsers Vite targets were released in 2020. Vite picked the 2020 versions back in 2021 because [that's what browserslist told them](https://github.com/vitejs/vite/commit/05fd1e26be1c64c665b5c01161f2e789c1deb6aa#diff-aa53520bfd53e6c24220c44494457cc66370fd2bee513c15f9be7eb537a363e7R233) were the most modern and widely supported browsers at the time
       - Vite claims they picked these because they support ES Modules, however that's not the whole story.  ES Modules were supported for _years_  (Chrome 2016, Safari 2017, Firefox 2018).  Vite picked their target by requiring it to be the **most modern** and **support modules**.
       - <details> <summary> slightly related </summary> Before the patch I sent, it seems that Vite's _only_ requirement was supporting ESM, which we can see by comparing their <a href="https://github.com/vitejs/vite/commit/05fd1e26be1c64c665b5c01161f2e789c1deb6aa#diff-aa53520bfd53e6c24220c44494457cc66370fd2bee513c15f9be7eb537a363e7L235"> previous browserslist</a> to <a href="https://caniuse.com/es6-module"> ESM support </a>. (off-topic: it's funny that commit is `refactor!:`. How can refactors be breaking changes 😁? </details>
       - Using the same query Vite did back in 2021 [today reveals we'd want to support back to Chrome 109, Safari 15.6, and Firefox 128](https://browsersl.ist/#q=defaults+and+supports+es6-module+and+supports+es6-module-dynamic-import)
   * There have been many new web features released that are now [widely available](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility).  Ignoring these new features means larger bundles and slower code execution.

tl;dr: @chenjiahan I understand wanting to be conservative with our default browser target, but do not think that Vite did anything more than just use the most modern, widely-supported browsers at the time.  Sticking with five-year old browsers leaves performance on the table for the vast majority of our users.  This PR proposes a way that we can continuously stay with the most widely supported browsers while also introducing new features as they become widely available.

vegerot:
btw, while doing research for my last post I discovered Vite will probably change it too 😁 https://github.com/vitejs/vite/pull/20007

chenjiahan:
Thanks for the detailed introduction, I agree that it makes sense to preset a dynamic range in the Rspack templates.

I noticed that the [default value](https://github.com/browserslist/browserslist?tab=readme-ov-file#best-practices) of browserslist is "> 0.5%, last 2 versions, Firefox ESR, not dead", while this PR introduces a different order of values ​​(e.g. Firefox ESR is at the end). Is this intentional?

vegerot:
> Is this intentional?

No 🙂

chenjiahan (approved):
LGTM 👍 Thank you

vegerot:
@chenjiahan should we make similar patches for Rsbuild and Modern.js?

chenjiahan:
It's a breaking change for Rsbuild so I prefer to update the value in major release.

vegerot:
@chenjiahan makes sense.  Are we already working on a major release in a `next` branch or something that I could add the patch to?

chenjiahan:
Yeah keep Rsbuild's templates clean is quite import as it's an zero config build tool.

We haven't have a clear plan for Rsbuild's next major version. Maybe later in the year we'll start thinking about it.

vegerot:
What about Modern.js?  We already provide a `.browerslistrc` in the template so we can change that.

I see Modern.js will drop support for Node 16 next month.  When will Rsbuild drop support?

chenjiahan:
For Morern.js, would you like to create an issue in the Modern js repo?

Rsbuild will drop Node 16 support in Rsbuild 1.5, which should be released in Q3 2025.

vegerot:
@chenjiahan let's discuss the Rsbuild change in https://github.com/web-infra-dev/rsbuild/issues/5235 and the Modern.js patch in https://github.com/web-infra-dev/modern.js/pull/7147

PR #10304, opened 2025-05-08
fix(template-react-ts): work with newer Node

vegerot:
Summary:
On my machine running Node v23.11.0, running this example fails with two errors:
1. `__dirname` is undefined
2. `RefreshPlugin` is an object instead of a class

This commit fixes both of these errors.

Test Plan:
1. use Node v23.11.0
2. run `pnpm create rspack@latest`
3. select react-typescript config
4. run `pnpm run dev`

Before this commit: it doesn't work
After this commit: it works

### DIFF
--- a/packages/create-rspack/template-react-js/package.json
+++ b/packages/create-rspack/template-react-js/package.json
@@ -14,7 +14,7 @@
   "devDependencies": {
     "@rspack/cli": "workspace:*",
     "@rspack/core": "workspace:*",
-    "@rspack/plugin-react-refresh": "^1.4.1",
+    "@rspack/plugin-react-refresh": "^1.4.3",
     "@types/react": "^19.1.3",
     "@types/react-dom": "^19.1.3",
     "react-refresh": "^0.17.0"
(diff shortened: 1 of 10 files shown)


chenjiahan (changes requested):
This change breaks Node < 23, we need to find another way to support Node >= 23

- Node 18

image

- Node 22

image

chenjiahan:
@hardfist It seems that ts-node will have this problem. Rsbuild uses jiti and it works fine.

chenjiahan:
Here is another issue with `ts-node`: https://github.com/rspack-contrib/rspack-plugin-react-refresh/issues/33

vegerot:
In the meantime I suggest we make two templates that the `create` script will automatically pick between.  Right now people using the latest version of Node will run into problems starting new projects.

hardfist:
@vegerot I fix the template by
* remove __dirname: which is unnecessary
* change `import RefreshPlugin from '@rspack/plugin-react-refresh'` to `import { ReactRefreshRspackPlugin } from '@rspack/plugin-react-refresh'` which can be compatible with esm&cjs

chenjiahan (approved):
Thank you!

PR #10309, opened 2025-05-09
fix: environment.dynamicImport should affect chunkFormat and chunkLoading

LingyuCoder:
…ding

## Summary

Align with webpack, `chunkFormat` and `chunkLoading` should be set default value according to `environment.dynamicImport`

## Checklist


tests/webpack-test/package.json
```diff
@@ -28,6 +28,7 @@
     "is-ci": "4.1.0",
     "less": "4.3.0",
     "less-loader": "^12.2.0",
+    "lodash-es": "^4.17.15",
```
fi3ework, line 31: Why `lodash-es` is added?

LingyuCoder: This is used in test case `issue-16040`

fi3ework (approved):

PR #10326, opened 2025-05-12
feat: JS API can access buildInfo.assets

SyMind:
## Summary

support use `buildInfo.assets` to get data from rust.

Used in Next.js.

```js
Object.keys(module.buildInfo.assets)
```

## Checklist


crates/node_binding/src/modules/context_module.rs
```diff
@@ -1,6 +1,7 @@
 use crate::{impl_module_methods, Module};
 
 #[napi]
+#[repr(C)]
```
LingyuCoder, line 4: what is this

LingyuCoder (approved):

Issue #10310, opened 2025-05-09
[Bug]: Critical dependency: require function is used in a way in which dependencies cannot be statically extracted


xxvv:
### System Info

System:
  OS: Windows 11 10.0.26100
  CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-12400
  Memory: 10.02 GB / 31.73 GB
Binaries:
  Node: 18.20.5 - ~\AppData\Local\Volta\tools\image\node\18.20.5\node.EXE
  Yarn: 4.1.1 - ~\AppData\Local\Volta\tools\image\yarn\4.1.1\bin\yarn.CMD
  npm: 10.8.2 - ~\AppData\Local\Volta\tools\image\node\18.20.5\npm.CMD
  pnpm: 9.15.9 - C:\Program Files\Volta\pnpm.EXE
  bun: 1.2.10 - ~\.bun\bin\bun.EXE
Browsers:
  Edge: Chromium (132.0.2957.140)
  Internet Explorer: 11.0.26100.1882
npmPackages:
@rspack/cli: ^1.3.9 => 1.3.9
@rspack/core: ^1.3.9 => 1.3.9

### Details

![Image](https://github.com/user-attachments/assets/f58acb08-e3b2-4c9e-87af-12f49e2b83c4)
我是使用require.context语法后出现的这个警告，这是老项目并不想修改这块功能。所有根据文档设置[exprContextCritical=false](https://rspack.dev/zh/config/module#moduleparserjavascriptexprcontextcritical)，希望关闭这个警告，发现这个配置没有效果

### Reproduce Steps

我是使用require.context

markmssd:
@xxvv were you able to fix it? We also see this warning on the `monaco-editor` package.

xxvv:
I solved it by configuring this ignoreWarnings property.
https://rspack.dev/config/other-options#ignorewarnings

markmssd:
Ah that's also what I had done, couldn't find a proper fix :P

PeterCardenas:
@markmssd looks like for monaco you can use the ContextReplacementPlugin like so
```typescript
new rspack.ContextReplacementPlugin(/^vs\/editor\/editor\.main$/, path.resolve(__dirname, "node_modules/monaco-editor/esm/vs/editor/editor.main.js"),
```

markmssd:
[...]

PR #10329, opened 2025-05-12
fix: respect expr_context_critical option for require.context dependency

PeterCardenas:
## Summary

In [this issue](https://github.com/web-infra-dev/rspack/issues/10310), it was expected that setting `module.parser.javascript.exprContextCritical` to `false` would silence the warning with the message: `Critical dependency: require function is used in a way in which dependencies cannot be statically extracted`. However the option is not respected in this case. I believe there is a deeper issue here, but at least we should respect the option here.

Have a repro here with monaco-editor: https://github.com/PeterCardenas/rspack-critical-dependency-repro

## Checklist

  - Do let me know if a test here is apropriate


ahabhgk:
LGTM, could you add a test for this? just for avoiding regression, you can take `packages/rspack-test-tools/tests/configCases/context-module/critical-warning` as an exmaple

PeterCardenas:
@ahabhgk added a couple tests let me know if it's acceptable 🙏

PR #10333, opened 2025-05-12
chore: 🔧 use less docker in self hosted runner

stormslowly:
## Summary
to avoid network error while running jobs in docker

1. build wasm in runner directly, no need to run in docker
2. run e2e test in self hosted runner, it takes similar time while in docker

## Checklist


hardfist (approved):

stormslowly:
https://github.com/web-infra-dev/rspack/pull/10344 is better. close.

PR #10334, opened 2025-05-12
feat(compiler-builder): support `browserslist` as the target

h-a-n-a:
## Summary

Supports using `browserslist` as the target.

### Ceveat

`browserslist_rs` did not export `load_config`, this causes a difference between webpack and rspack. In webpack, if the bundler is not able to find a config, then it will fallback to target `web`. In rspack, it fallbacks to the default browserslist query instead.

Tests are being reviewed by my eyes to align with webpack's `target-browserslist` unit tests. Nevertheless, there could be some cases that I'm not aware of. Please file an issue if you find one.

Snapshot of `crates/rspack/tests/snapshots/defaults__default_options.snap` has been changed due to the previous reason.

## Checklist

### DIFF
--- a/crates/rspack/src/builder/browserslist_target.rs
+++ b/crates/rspack/src/builder/browserslist_target.rs
@@ -1,0 +1,508 @@
+use std::collections::HashMap;
+
+use browserslist::Opts;
+
+use super::target::TargetProperties;
+
+// Macro for defining HashMaps with less boilerplate
+macro_rules! hashmap {
+    ($( $key: expr => $val: expr ),* $(,)?) => {{
+         let mut map = ::std::collections::HashMap::new();
+         $( map.insert($key, $val); )*
+         map
+    }}
+}
+
+/// Configuration parsed from input string and context directory
+#[derive(Debug, Default)]
+pub struct BrowserslistHandlerConfig<'a> {
+  /// Optional absolute path to config file
+  pub config_path: Option<String>,
+  /// Optional environment name
+  pub env: Option<String>,
+  /// Optional query string
+  pub query: Option<String>,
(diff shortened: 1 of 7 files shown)


h-a-n-a:
> Can we check the performance regression: [#10334 (comment)](https://github.com/web-infra-dev/rspack/pull/10334#issuecomment-2871985708)

It's within the expectation. Browserslist needs to be invoked to discover possible configs and convert it to `environment` in rspack option.

chenjiahan:
Get👌 How about renaming to case name from "basic" to "rust-ai-basic" to make it clearer

h-a-n-a:
> Get👌 How about renaming to case name from "basic" to "rust-api-basic" to make it clearer

Sounds good, what about making rust api starts with `rust@basic`, for js , it starts with `js@xxx`?
`

chenjiahan:
LGTM 👍

h-a-n-a:
> LGTM 👍

@chenjiahan Would you please approve?

crates/rspack/src/builder/browserslist_target.rs
chenjiahan, line 1: It seems that the target logic of builder/browserslist_target.rs is similar to that of Node.js side. Do we have plans to merge them into one implementation to reduce maintenance costs?

h-a-n-a: We cannot do that. These two implementations are not the same, one from js and one from rust.

h-a-n-a: However, there's no easy way to move part of the default logic to rust. That is being said some of the default applies rely on js functions which cannot be integrated without napi. But we don't want napi to interfere rust only code.

Issue #10118, opened 2025-04-22
[Bug]: Panic on arm64 with `performance.buildCache = true`


regisb:
### System Info

```System:
  OS: Linux 5.15 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
  CPU: (4) arm64 Neoverse-N1
  Memory: 13.17 GB / 15.15 GB
  Container: Yes
  Shell: 5.1.4 - /bin/bash
Binaries:
  Node: 20.18.0 - /usr/local/bin/node
  Yarn: 1.22.22 - /usr/local/bin/yarn
  npm: 10.8.2 - /usr/local/bin/npm
```

### Details

I am attempting to build an application in Docker, both on arm64 and amd64. The build is successful on amd64, but fails on arm64. Build is successful on arm64 when I disable caching.

Here's the error message (or rather, the lack thereof):

```
1.047 info    Rspack persistent cache enabled (experimental)
4.211 Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues
4.211 Message:  created a new `Panic` from: out of range integral type conversion attempted
4.212 Location: index.crates.io-1949cf8c6b5b557f/rancor-0.1.0/src/lib.rs:648
4.214
4.214 Run with COLORBT_SHOW_HIDDEN=1 environment variable to disable frame filtering.
4.215 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ BACKTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
4.228  1: start_thread
4.228     at <unknown source file>
4.362 Aborted (core dumped)
```

This issue is consistently reproduced on an Apple M1 or M2. I do not own an Apple computer, but I was able to reproduce the issue by building on a remote arm64 server.

This issue was first reported here: https://github.com/overhangio/tutor-mfe/pull/252#issuecomment-2796362845

### Reproduce link

https://github.com/regisb/rsbuild-arm64-failure

### Reproduce Steps

```
git clone https://github.com/regisb/rsbuild-arm64-failure .
make build-arm64-with-cache
```

Note that the following commands complete successfuly:

```
make build-arm64-no-cache
make build-amd64-with-cache
```

1714080902120:
[...]

PR #10346, opened 2025-05-13
refactor: use OwnedOrRef instead of #[cacheable(as=...)]

jerrykingxyz:
## Summary

The implementation of persistent cache relies on passing references to avoid clone structures. However, it is currently necessary to define a ref struct, and some unsafe methods will be used to make the data structure and the ref struct reuse the same serialized data. This can lead to some bug, such as https://github.com/web-infra-dev/rspack/issues/10118
``` rust
#[cacheable]
struct Node {
  mgm: ModuleGraphModule
  ...
}
#[cacheable(as=Node)]
struct NodeRef<'a> {
  mgm: &'a ModuleGraphModule
  ...
}

fn main() {
   let node_ref = NodeRef { ... };
   let bytes = to_bytes(node_ref, ...)?;
   let node: Node = from_bytes(bytes, ...)?;
}
```

This PR will use the `OwnedOrRef` struct instead of `#[cacheable(as=)]` to avoid using unsafe methods.

``` rust
#[cacheable]
struct Node<'a> {
  mgm: OwnedOrRef<'a, ModuleGraphModule>
  ...
}

fn main() {
   let node = Node { mgm: OwnedOrRef::Borrowed(...) };
   let bytes = to_bytes(node, ...)?;
   let node: Node = from_bytes(bytes, ...)?;
   assert!(matches!(node.mgm, OwnedOrRef::Owned(_)));
}
```

## Checklist


LingyuCoder (approved):

regisb:
Thanks for the fix @jerrykingxyz @LingyuCoder!

PR #10348, opened 2025-05-13
perf: skip UTF-8 validation if JS loader returned a string

h-a-n-a:
## Summary

Skip `UTF-8` validation on the Rust side in normal module build if the last loader returned a string. Also removes source map `UTF-8` validation as it's always `UTF-8` encoded.

## Checklist


h-a-n-a:
!bench

ahabhgk (approved):

PR #10351, opened 2025-05-13
fix: bailout when optional module dependency

LingyuCoder:
## Summary

If a dependency is optional and it is not found, then the make artifact should bailout and swap the make artifact. Currently if bailout, the make artifact will not be swapped and then the modules will be empty.

## Checklist


LingyuCoder:
hold on this PR. perhaps we can find a better way to handle this.

chenjiahan:
Closed as stale. Feel free to reopen it if needed.

PR #10365, opened 2025-05-14
fix: loader in module user request should be resolved path

SyMind:
## Summary

Loader in module user request should be resolved path.

```js
import "loader!resource";
```

the module.userRequest should be `/path/to/loader.js!/path/to/resource`.

## Checklist

### DIFF
--- a/crates/rspack_core/src/utils/identifier.rs
+++ b/crates/rspack_core/src/utils/identifier.rs
@@ -45,14 +45,12 @@ pub async fn stringify_loaders_and_resource<'a>(
       .collect::<Vec<_>>();
     let resolve_results = futures::future::join_all(features).await;
     let mut s = String::new();
-    for resolve_result in resolve_results {
-      if let Ok(result) = resolve_result {
-        if let ResolveResult::Resource(resource) = result {
-          s.push_str(&resource.full_path());
-        }
+    for resolve_result in resolve_results.into_iter().flatten() {
+      if let ResolveResult::Resource(resource) = resolve_result {
+        s.push_str(&resource.full_path());
       }
     }
-    s.push_str("!");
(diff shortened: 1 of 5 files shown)


crates/rspack_core/src/utils/identifier.rs
```diff
@@ -32,17 +32,42 @@ pub fn to_identifier(v: &str) -> Cow<str> {
...
-      .map(|i| &*i.loader)
-      .collect::<Vec<_>>()
-      .join("!");
-    Cow::Owned(format!("{s}!{resource}"))
+      .map(|loader| {
+        let loader_request = &loader.loader;
+        async move {
+          (
+            loader_resolve
+              .resolve(context.as_path().as_std_path(), loader_request)
```
ahabhgk, line 49: `resolve` here?

SyMind: yes, should get the absolute path.

ahabhgk: This is bad for performance, there should have already resolved loader absolute path in somewhere

crates/rspack_core/src/utils/identifier.rs
```diff
@@ -32,17 +32,23 @@ pub fn to_identifier(v: &str) -> Cow<str> {
   }
 }
 
-pub fn stringify_loaders_and_resource<'a>(
-  loaders: &'a [ModuleRuleUseLoader],
+pub async fn stringify_loaders_and_resource<'a>(
```
ahabhgk, line 35:
```suggestion
pub fn stringify_loaders_and_resource<'a>(
```

ahabhgk (approved):

PR #10371, opened 2025-05-15
fix: inner graph block scopes

LingyuCoder:
## Summary

Fix inner graph block scope variables analysis
- Use TopLevelSymbol as inner graph map keys which align with [webpack](https://github.com/webpack/webpack/blob/main/lib/optimize/InnerGraph.js#L342-L349)

## Checklist


JSerFeng:
Thank you !

JSerFeng (approved):

PR #10382, opened 2025-05-15
Enrich client modules types with supported api (closes #10335)

### LINKED ISSUE #10335: [Bug]: @rspack/core/module types mismatch
### System Info

System:
OS: macOS 15.3.2
CPU: (12) arm64 Apple M2 Pro
Memory: 2.25 GB / 32.00 GB
Shell: 5.9 - /bin/zsh
Binaries:
Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
Yarn: 4.1.0 - /opt/homebrew/bin/yarn
npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
bun: 1.2.9 - ~/.bun/bin/bun
Browsers:
Chrome: 136.0.7103.93
Safari: 18.3.1
npmPackages:
@rspack/cli: 1.3.9 => 1.3.9
@rspack/core: 1.3.9 => 1.3.9
@rspack/dev-server: 1.1.1 => 1.1.1
@rspack/plugin-react-refresh: 1.4.2 => 1.4.2

### Details

I see the mismatch in types exported by the `@rsapck/core/module` 

First of all, I am questioning myself, what should be the source of truth in this case?
we have the types based on https://github.com/webpack/webpack/blob/main/module.d.ts
but we also have https://www.npmjs.com/package/@types/webpack-env

And it seems that the `@types/webpack-env` is a better one. Why?

I am trying to get rid of webpack stuff in our web application, and I want to sue the built-in rspack types.

- we do not use `@types/node` at all. and the Rspack types do not define the globals at all. In our project this mean I have to copy parts of `@types/webpack-env` in a separate d.ts file. Specificaly:

```typescript
declare namespace Rspack {
	type ModuleId = string | number

	interface RequireFunction {
		(path: string): any
		<T>(path: string): T
	}

	interface Module {
		id: ModuleId
		filename: string
	}

	interface NodeProcess {
		env?: any
	}
}

declare namespace NodeJS {
	interface Module extends Rspack.Module {}
	interface Require extends Rspack.RequireFunction {}
}

interface NodeModule extends Rspack.Module {}

declare var module: NodeJS.Module
declare var process: NodeJS.Process
declare var require: NodeJS.Require
```

- parts of the interfaces are missing. you can see the `Module` is missing the `id` field and so one.

- all these interfaces are proved to be exposed with RSPack and work fine. it is just a type mismatch

Suggestion - to use `@types/webpack-env` as a source of types. I am open for the advice here and if I can find the proper interfaces in the code of RSPack directly.

I can sync the type definitions if we agree on the approach.

Thank you!

### Reproduce link

-

### Reproduce Steps

I am switching the type definitions from `@types/webpack-env` to `@rspack/core/moodule`

n0th1ng-else:
## Summary

RSPack client types are missing the key variable that web applications usually rely on:
- better module api (specifically module.hot and module.id)
- require and the underlying api

This becomes actual when the application does not fetch @types/node and hence the Typescript does not know such constructs. We are developing purely web application, so adding @types/node does not make any sense for us. Thus the change.

In this commit, we are adding the module variable and the require() function, based on the api supported by rspack. The type definitions are inspired by [@types/webpack-env](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack-env/index.d.ts#L71)

## Checklist


packages/rspack/module.d.ts
```diff
@@ -220,25 +252,10 @@ declare var __non_webpack_require__: (id: any) => unknown;
 declare const __system_context__: object;
 
 declare namespace NodeJS {
-	interface Module {
-		hot: Rspack.Hot;
-	}
-
-	interface Require {
-		ensure(
```
n0th1ng-else, line 228: Does not exists (see PR description)

packages/rspack/module.d.ts
```diff
@@ -220,25 +252,10 @@ declare var __non_webpack_require__: (id: any) => unknown;
...
-		): void;
-		context(
-			request: string,
-			includeSubdirectories?: boolean,
-			filter?: RegExp,
-			mode?: "sync" | "eager" | "weak" | "lazy" | "lazy-once"
-		): Rspack.Context;
-		include(dependency: string): void;
-		resolveWeak(dependency: string): void;
-		onError?: (error: Error) => void;
```
n0th1ng-else, line 242: no supported (see the warning in PR description)

packages/rspack/module.d.ts
```diff
@@ -220,25 +252,10 @@ declare var __non_webpack_require__: (id: any) => unknown;
...
-			errorCallback?: (error: Error) => void,
-			chunkName?: string
-		): void;
-		context(
-			request: string,
-			includeSubdirectories?: boolean,
-			filter?: RegExp,
-			mode?: "sync" | "eager" | "weak" | "lazy" | "lazy-once"
-		): Rspack.Context;
-		include(dependency: string): void;
```
n0th1ng-else, line 240: no supported (see the warning in PR description)

packages/rspack/module.d.ts
```diff
@@ -169,6 +171,36 @@ declare namespace Rspack {
 		id: string | number;
 		(dependency: string): unknown;
 	}
+
+	interface Module {
```
n0th1ng-else, line 175: There is also a `filename` property for the webpack Modules (see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack-env/index.d.ts#L74)

but I do not see this property with rspack

n0th1ng-else:
There is also a [definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/webpack-env/index.d.ts#L382) for the `process` variable. But it is not a real variable, the DefinePlugin injects the values provided, so I am not sure if I need to add it. 

My idea is that it is not part of the core, and if you use the DefinePlugin, then you will describe the used variables manually anyway.

but I am fine to add it as well

packages/rspack/module.d.ts
```diff
@@ -9,6 +9,8 @@
  */
 
 declare namespace Rspack {
+	type ModuleId = string | number;
```
n0th1ng-else, line 12: to double check. in Webpack the moduleId is `string | number`. I assume the number comes from the concatenated modules. do we have the same behavior in rspack?

chenjiahan: Yes, Rspack's moduleId should be the same as webpack

chenjiahan (approved):
LGTM. Thank you! ❤️

PR #10391, opened 2025-05-16
fix: assets imported by css should not generate js modules

LingyuCoder:
## Summary

If an asset is imported by uri in css files, it should not generate a JavaScript module through code generation. This makes sense in the scenario when importing an image with asset/inline, which transforms that image to base64 and replaces the asset URI with it. If the JavaScript module of the image is generated, the long base64 string will be generated again, which significantly increases the size of the final assets.

## Checklist

### DIFF
--- a/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
+++ b/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
@@ -411,7 +411,7 @@ impl ChunkGraph {
         {
           module_source_types.contains(&source_type)
         } else {
-          module.source_types(&module_graph).contains(&source_type)
+          module.source_types(module_graph).contains(&source_type)
         }
       })
       .collect::<Vec<_>>();
@@ -429,7 +429,7 @@ impl ChunkGraph {
       .modules
       .iter()
       .filter_map(|uri| module_graph.module_by_identifier(uri))
-      .filter(move |module| module.source_types(&module_graph).contains(&source_type))
+      .filter(move |module| module.source_types(module_graph).contains(&source_type))
       .map(|m| m.as_ref())
   }
 
--- a/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
+++ b/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
@@ -102,8 +102,9 @@ impl ChunkGraphChunk {
 
 fn get_modules_size(modules: &[&BoxModule], compilation: &Compilation) -> f64 {
   let mut size = 0f64;
+  let module_graph = compilation.get_module_graph();
   for module in modules {
-    for source_type in module.source_types() {
+    for source_type in module.source_types(&module_graph) {
       size += module.size(Some(source_type), Some(compilation));
     }
   }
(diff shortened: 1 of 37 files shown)


JSerFeng (approved):

dkempner:
@LingyuCoder this may have introduced a bug when trying to import css from [mapbox-gl](https://github.com/mapbox/mapbox-gl-js).

```
import 'mapbox-gl/dist/mapbox-gl.css'
```

I noticed we were getting incorrect chunk ids being generated in our webpack stats while the file was not actually created:

`Assets not found for chunk: data_image_svg_xml_charset_utf-8_3Csvg_width_24_height_24_viewBox_0_0_20_20_xmlns_http_www_w3-e2eefd`

`Assets not found` is from our internal system but the rest is the bad chunk id

It seems like it's because of `data:image` images like this:

```css
.mapboxgl-ctrl button.mapboxgl-ctrl-zoom-out .mapboxgl-ctrl-icon {
    background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg width='29' height='29' viewBox='0 0 29 29' xmlns='http://www.w3.org/2000/svg' fill='%23333'%3E %3Cpath d='M10 13c-.75 0-1.5.75-1.5 1.5S9.25 16 10 16h9c.75 0 1.5-.75 1.5-1.5S19.75 13 19 13h-9z'/%3E %3C/svg%3E")
}
```

LingyuCoder:
> @LingyuCoder this may have introduced a bug when trying to import css from [mapbox-gl](https://github.com/mapbox/mapbox-gl-js?rgh-link-date=2025-06-20T19%3A05%3A17Z).
> 
> ```
> import 'mapbox-gl/dist/mapbox-gl.css'
> ```
> 
> I noticed we were getting incorrect chunk ids being generated in our webpack stats while the file was not actually created:
> 
> `Assets not found for chunk: data_image_svg_xml_charset_utf-8_3Csvg_width_24_height_24_viewBox_0_0_20_20_xmlns_http_www_w3-e2eefd`
> 
> `Assets not found` is from our internal system but the rest is the bad chunk id
> 
> It seems like it's because of `data:image` images like this:
> 
> ```css
> .mapboxgl-ctrl button.mapboxgl-ctrl-zoom-out .mapboxgl-ctrl-icon {
>     background-image: url("data:image/svg+xml;charset=utf-8,%3Csvg width='29' height='29' viewBox='0 0 29 29' xmlns='http://www.w3.org/2000/svg' fill='%23333'%3E %3Cpath d='M10 13c-.75 0-1.5.75-1.5 1.5S9.25 16 10 16h9c.75 0 1.5-.75 1.5-1.5S19.75 13 19 13h-9z'/%3E %3C/svg%3E")
> }
> ```

Can you provide a repro？

nathanmarks:
@LingyuCoder `rspack` is now outputting a chunk that has no files associated with it. Wouldn't it be correct to just not have a chunk altogether?

LingyuCoder:
> @LingyuCoder `rspack` is now outputting a chunk that has no files associated with it. Wouldn't it be correct to just not have a chunk altogether?

Sounds like a bug. I need a repro to fix it

PR #10416, opened 2025-05-20
fix: CopyRspackPlugin's transformer function should handle Buffer input

CPunisher:
## Summary

https://webpack.js.org/plugins/copy-webpack-plugin/#transform is actually out of date.
From https://github.com/webpack-contrib/copy-webpack-plugin/blob/0cae504f3044eef77a62b338cdf0c092675748fe/types/index.d.ts#L123 we can see that `input` parameter should be `Buffer`.

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -1650,7 +1650,7 @@ export interface RawCopyPattern {
    * @default false
    */
   copyPermissions?: boolean
-  transform?: { transformer: (input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  } | ((input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>)
+  transform?: { transformer: (input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  } | ((input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>)
 }
 
 export interface RawCopyRspackPluginOptions {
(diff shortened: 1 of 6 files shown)


crates/node_binding/binding.d.ts
```diff
@@ -1650,7 +1650,7 @@ export interface RawCopyPattern {
    * @default false
    */
   copyPermissions?: boolean
-  transform?: { transformer: (input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  } | ((input: string, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>)
+  transform?: { transformer: (input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>  } | ((input: Buffer, absoluteFilename: string) => string | Buffer | Promise<string> | Promise<Buffer>)
```
LingyuCoder, line 1653: This should be a breaking change?

CPunisher: I'm not sure. But in the lastest version we have assume that the input is a Buffer in the rust side: https://github.com/web-infra-dev/rspack/blob/eec2f5e3740d344c013f9f44982622e51d363d45/crates/node_binding/src/raw_options/raw_builtins/raw_copy.rs#L15

CPunisher: This change is introduced by https://github.com/web-infra-dev/rspack/pull/10398/files#diff-fd42ec611cd5acb13724f10d47d47067d31c35ae5a06062825c02fb81a1d8450R1653

We can see `input` is a `Buffer` in the latest version

LingyuCoder (approved):

PR #10426, opened 2025-05-20
fix: allow multiple runtime instances to receive HMR updates

GuilleX7:
## Summary

The current rspack runtime assigns a global HMR JSONP loader that replaces any existing loader with the same name. While usually a single instance of the same runtime is running on the same context, we found some cases in Module Federation where running the same remoteEntry more than once (and therefore the webpack runtime) in the same page allows to have coexisting instances of the same runtime. However, since the HMR JSONP loader is overwritten by the latest instance, only the latest instance gets the updates, causing errors in the earlier instances.

See the following scenario.

Before the fix:

https://github.com/user-attachments/assets/e1b7141c-f026-4921-94e3-e373bfb925ce

After the fix:

https://github.com/user-attachments/assets/d558f880-6241-4b0e-ac75-1bf4206a08bd

The proposed fix consists of installing a JSONP loader that calls previous existing JSONP loaders in the scope. Since all instances trigger HMR updates sequentially, we also rely on `currentUpdate` to determine which instance needs the HMR update (if an old instance is waiting for the HMR update and the JSONP loader is executed before newer instances have requested the update, the newer instances might not even have set `currentUpdate` up at that time)

Our real use case is:

> We have a microfrontend that we load using Module Federation and we want to have multiple instances of it running at the same time. The problem is that it uses singleton libraries, so mounting it twice is not enough, as the runtime would be the same and both would share the same instance.
> 
> What we are doing right now is to run remoteEntry multiple times, creating different runtimes so multiple instances of the library are instantiated. However, with this approach we only get HMR in the latest instance that has been created.

For completeness, here's the discussion I've had on webpack. Same changes have been proposed, since the runtime seems to be the same.

https://github.com/webpack/webpack/pull/19543

## Checklist

- [X] Tests updated (or not required): E2E test added
- [X] Documentation updated (or not required): likely not needed


ahabhgk:
Same response with webpack: https://github.com/webpack/webpack/pull/19543#issuecomment-2898431577, for Module Federation you can use `output.uniqueName` to avoid the hotUpdateGlobal override to let the multiple hotUpdateGlobals co-exist at the same page, and each runtime (mf container) update their own part on the page, but it's impossible to let one runtime to update multiple part on the page

PR #10440, opened 2025-05-21
fix: css prefetch and preload

LingyuCoder:
## Summary

fix missing css prefetch and preload runtime.

> not fix css extract plugin yet, should process it in another PR

## Checklist


crates/rspack_plugin_css/src/runtime/css_loading_with_preload.js
```diff
@@ -0,0 +1,17 @@
+__webpack_require__.H.s = (chunkId) => {
```
GiveMe-A-Name, line 1: Maybe we need using arrow function and normal function refer output.environment 
The output code maybe need run the code is incompatible es5.

LingyuCoder: Changed it to normal function. All css runtime modules are not rendered by dojang so they are not generated according to `output.environment` yet. Perhaps we need to refactor them in another PR

GiveMe-A-Name (approved):

PR #10445, opened 2025-05-22
refactor: avoid read rust side json in js side

hardfist:
## Summary
sometimes the rust side json file is too big too read in js side, so avoid reading it in js side
```js
Error: Cannot create a string longer than 0x1fffffe8 characters
    at Object.readFileSync (node:fs:448:20)
    at JavaScriptTracer.dumpJavaScriptTrace
```

## Checklist

### DIFF
--- a/packages/rspack/src/trace/index.ts
+++ b/packages/rspack/src/trace/index.ts
@@ -49,6 +49,11 @@ export class JavaScriptTracer {
 		}
 	}
 
+	/**
+	 *
+	 * @param isEnd true means we are at the end of tracing,and can append ']' to close the json
+	 * @returns
+	 */
 	static async cleanupJavaScriptTrace() {
 		if (!this.layer.includes("chrome")) {
 			return;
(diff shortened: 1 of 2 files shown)


packages/rspack/src/trace/index.ts
```diff
@@ -92,20 +92,17 @@ export class JavaScriptTracer {
...
 			// this is hack, [] is empty and [{}] is not empty
-			const originTraceIsEmpty = !originTrace.includes("{");
-			const eventMsg =
-				(this.events.length > 0 && !originTraceIsEmpty ? "," : "") +
-				this.events
-					.map(x => {
-						return JSON.stringify(x);
-					})
-					.join(",\n");
+			const eventMsg = this.events
```
stormslowly, line 96: if eventMsg's length hit the limit, it's better to write the content by fs stream api .

hardfist: stream is async I change it to manually stream write to avoid intermediate large string

stormslowly (approved):

PR #10451, opened 2025-05-22
feat: inline const for leaf modules

ahabhgk:
## Summary

Pre-PR for enum inlining

ref: https://github.com/evanw/esbuild/commit/7bf3046f723f4c76b3c61ce9d2724468150c1722

This PR only handle the cross-module part of const inlining, the idea is very similar to mangle exports:
1. Add `inlinable` on `export_info`, which is similar to `can_mangle`, and add `UsedName::Inlined` for inlined value and `UsedName::Normal` for maybe mangled names
2. Add `inline_const` parser plugin to find all the inlinable const variable, and "tag" it to tell the `ESMExportSpecifierDependency` that this exported value is inlinable

How to use:

```js
{
  module: {
    rules: [{
      test: ...,
      parser: {
        inlineConst: false // use parser options to opt-out
      }
    }]
  },
  experiments: {
    inlineConst: true // enable the experimental switch
  }
}
```

## Checklist

- [x] Documentation updated (~or not required~ update in future PR).


ahabhgk:
CodSpeed regression is caused by extra `get_used_name`, will optimize the perf of `get_used_name` in future PR

PR #10482, opened 2025-05-26
docs: Add Miro into the list of consumers

n0th1ng-else:
## Summary

I am happy to say that in Miro we have migrated our biggest application to RSPack and it goes pretty well so far. We have migrated all operational parts including the dev server, production asset generation, Storybooks as well.

This being said, I suggest adding the Miro in the list of companies that actively use RSPack now.

## Checklist

Let me know if I need to adjust the width
image


website/theme/components/Landingpage/WhoIsUsing/index.tsx
```diff
@@ -143,6 +144,12 @@ const companyList: Company[] = [
     url: 'https://verkada.com',
     width: 180,
   },
+  {
+    name: 'Miro',
+    logo: miroLogo,
+    url: 'https://miro.com',
+    width: 140,
```
chenjiahan, line 151: Adjust the width to 120 to keep it similar to the other logos in terms of visual size:

```suggestion
    width: 120,
```

chenjiahan (approved):
🎉 We are so happy that Miro is using Rspack! Thank you for your contribution ❤️

Issue #10504, opened 2025-05-28
[Bug]: `moduleIds` set to `"natural"` results in unsorted modules


dinofx:
### System Info

macOS

### Details

In webpack, when "natural" is used for `optimization.moduleIds`, the output looks like:

```js
var __webpack_modules__ = ([
/* 0 */,
/* 1 */
/***/ (function(module, exports) {

module.exports = "first module"

/***/ }),
/* 2 */
/***/ (function(module, exports) {

module.exports = "second module"
...
```

Note that the structure is an array, and therefore the position must determine the "id" of each module.

In Rspack, the generated output is instead an object with unpredictable ordering. Sometimes it looks like:

```js
var __webpack_modules__ = ({
2: (function (module) {
module.exports = "second module"

}),
1: (function (module) {
module.exports = "first module"
...
```

We have test fixtures that we run both during local development (on macOS) or in our build system (Linux).  On our local laptops, the ordering of this trivial example is consistently 1,2, but in our build system 2,1.  This makes it impossible for us to generate the expected output on our development machines. And ,for more complex outputs the ordering seems fragile and susceptible to randomly changing.

How can we prevent this order from changing in different environments?  I suspect that this code is related:

https://github.com/web-infra-dev/rspack/blob/20dce0d068a9ed1494bcf77431b7ba476a711ce3/crates/rspack_plugin_javascript/src/runtime.rs#L59

Note that in the above example the two modules that were imported (by a trivial entry point module) were actually loaded by different rules that matched different sequences of loaders.  The requests to the two modules would have been something like:

```
/our-build-tool-path/my-loader!/our-build-tool-path/another-loader!/some-temp-folder-path/xyz123/fixture-project/static/styles.css
/our-build-tool-path/my-loader!/some-temp-folder-path/xyz123/fixture-project/static/file.txt
```

Where `our-build-tool-path` is the absolute path to our build tool, and `some-temp-folder-path` is the absolute path to the stuff it is building.  We don't want any of the following to affect the output of our builds:

 - where we install our build tool (which includes rspack)
 - where a user decides to clone the source repo being built
 - where the dependencies of the source repo are installed

With webpack we have been able to use "natural" module IDs to prevent absolute paths from affecting build output. This no longer seems to work in Rspack.

### Reproduce Steps

Build some code with imports that result in different loader chains like described, such that the absolute paths to the loaders alters the sorting of module identifiers.  For example:

main.js
```js
import foo from 'noop-loader!./foo';
[...]

PR #10516, opened 2025-05-29
fix(core): use module_id to sort module_code

hardfist:
## Summary
fix https://github.com/web-infra-dev/rspack/issues/10504, and ensure sorted_modules in generated artifacts

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/configCases/source-map/verify-es6/index.js
+++ b/packages/rspack-test-tools/tests/configCases/source-map/verify-es6/index.js
@@ -11,10 +11,10 @@ it("verify es6 (esmodule) bundle source map", async () => {
 	const source = fs.readFileSync(__filename + ".map", "utf-8");
 	const map = JSON.parse(source);
 	expect(map.sources).toEqual([
-		"webpack:///../../../../dist/helper/util/checkSourceMap.js",
 		"webpack:///./b-dir/c-dir/c.js",
 		"webpack:///./b-dir/b.js",
 		"webpack:///./a.js",
+		"webpack:///../../../../dist/helper/util/checkSourceMap.js",
 		"webpack:///./index.js",
 	]);
 	expect(map.file).toEqual("bundle0.js");


hai-x:
Maybe we should find a more ideal solution. Sorting by module identifier can improve compression and it brings ~2KB gzipped size improvement in the repro of following reported issue.

https://github.com/webpack/webpack/pull/19184#issuecomment-2849272596

chenjiahan:
This change caused the bundle size of our benchmark case to degrade by 6%, so we decided to revert it first and then explore better solutions.

> Rspack CLI 1.3.12 -> 1.3.13: 1381.9kB -> 1466.1kB (gzip size degraded by 84.2kB, non-gzip size remained unchanged)

> Steps to reproduce:
> 1. https://github.com/rspack-contrib/build-tools-performance
> 2. CASE=large TOOLS=rspack RUN_TIMES=1 WARMUP_TIMES=0 node ./benchmark.mjs

hai-x:
> This change caused the bundle size of our benchmark case to degrade by 6%, so we decided to revert it first and then explore better solutions.
> 
> > Rspack CLI 1.3.12 -> 1.3.13: 1381.9kB -> 1466.1kB (gzip size degraded by 84.2kB, non-gzip size remained unchanged)
> 
> > Steps to reproduce:
> > 
> > 1. [rspack-contrib/build-tools-performance](https://github.com/rspack-contrib/build-tools-performance)
> > 2. CASE=large TOOLS=rspack RUN_TIMES=1 WARMUP_TIMES=0 node ./benchmark.mjs

@chenjiahan 

I prefer to add one **moduleSort** option: enabling to sort modules based on such as `module identifier`, `module id`, or `module order`... 

Feel free to share additional ideas and have discussion.

chenjiahan:
Yes, adding an option is a feasible solution. But I am concerned that it will make things more complicated for users.

If we can find a stable and gzip-friendly way to sort modules, it would be perfect.

hai-x:
Hi, I think the fixing in https://github.com/webpack/webpack/pull/20287 looks good. Feel free to port it.

Issue #9513, opened 2025-02-28
[Feature]: Warn if using `[contenthash]` in filenames in development mode


h-a-n-a:
### What problem does this feature solve?

Using `[contenthash]` in filenames in development mode would bloat memory. This should be a part of the compilation diagnostics.

### What does the proposed API of configuration look like?

No extra APIs are exposed

blueavee:
Is this issue still available? I'd like to work on it if possible.

cseas:
Hi @h-a-n-a, is anyone actively working on this or is it open to contributions?

h-a-n-a:
@cseas Hi! It's open for contribution. Feel free to open a PR!

jaredallard:
Is there a good recommendation that should go along with this?

ahabhgk:
Why using `[contenthash]` bloat memory, from the implementation rspack will always calculate all the hashes, including modulehash, chunkhash, contenthash, and fullhash, just replace different hash with different placeholder: `[xxxhash]`

PR #10521, opened 2025-05-29
feat: warn if using [contenthash] in filenames in development mode

cseas:
## Summary

Using [contenthash] in filenames in development mode would bloat memory. Added a warning to the console if `[contenthash]` is added to the filename in development mode.

Fixes https://github.com/web-infra-dev/rspack/issues/9513

Gave a shot at implementing the feature. Please let me know if the warning should be added somewhere else in the codebase or if this needs an accompanying documentation update as well. I'd imagine the warning should go somewhere in the output docs as well: https://rspack.rs/config/output

**Update:** I noticed that the warning for this is already documented but it's too far down the page in another section and I just happened to stumble across it as I was reading the docs.

Screenshot 2025-05-31 at 02 26 02

IMO it'd be easier to come across this warning if it came earlier in the output.filename docs, right where the usage of contenthash is first suggested:

Screenshot 2025-05-31 at 02 28 02

Let me know, I'll move the warning up in the page if it makes sense.

## Checklist

### DIFF
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -544,6 +544,17 @@ const applyOutputDefaults = (
 		futureDefaults: boolean;
 	}
 ) => {
+	// Warn if [contenthash] is used in output.filename in development mode
+	if (
+		development &&
+		typeof output.filename === "string" &&
+		output.filename.includes("[contenthash]")
+	) {
+		console.warn(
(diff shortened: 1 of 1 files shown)


cseas:
@h-a-n-a Please review

h-a-n-a:
We need to discuss about this feature internally. Please bear with us and thanks for your patience.

PR #10522, opened 2025-05-29
fix: Narrow the Dev Server Middleware type

n0th1ng-else:
## Summary

It will be convenient to have the Dev Server Middleware types exported from the rspack. All our middleware is written in separate modules and annotated with JSDoc types.
By exporting the type, we can ge rid of webpack-dev-server mentions in the annotation and use `import('@rspack/core').DevServerMiddleware` to type the middleware we have.

I have also made an effort to narrow the Middleware type so it is aligned with the one in the webpack-dev-middleware.

(see https://github.com/webpack/webpack-dev-middleware/blob/master/types/index.d.ts#L370)

## Checklist

### DIFF
--- a/packages/rspack/etc/core.api.md
+++ b/packages/rspack/etc/core.api.md
@@ -32,7 +32,7 @@ import type { ExternalObject } from '@rspack/binding';
 import { fs } from 'fs';
 import { default as fs_2 } from 'graceful-fs';
 import { HookMap } from '@rspack/lite-tapable';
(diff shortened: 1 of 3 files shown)


chenjiahan:
@LingyuCoder cc~

chenjiahan:
@LingyuCoder Pin again..

LingyuCoder (approved):
thanks

n0th1ng-else:
Thank you both! Appreciate your time

Issue #7417, opened 2024-08-02
[Bug]: Getting "should have __webpack_require__.f.consumes" error with module federation enabled


hverlin:
### System Info

Using `1.0.1-beta.9` (but visible on earlier versions as well).

### Details

With module federation 1.5 (using https://rsbuild.dev/config/module-federation/options), I am getting the following error:

```
Uncaught Error: should have __webpack_require__.f.consumes
    at __webpack_require__.f.consumes (_define_property.js:8:1)
    at _define_property.js:8:1
    at Array.reduce (<anonymous>)
    at __webpack_require__.e (_define_property.js:8:1)
    at Array.map (<anonymous>)
    at __webpack_require__.x (_define_property.js:8:1)
    at _define_property.js:8:1
    at _define_property.js:8:1
```

(seems similar to https://github.com/web-infra-dev/rspack/issues/6773)

### Reproduce Steps

I think this is related to [web-workers](https://rspack.dev/guide/features/web-workers). In my code, I think that it's failing because of this line. 
```js
import { wrap } from 'comlink';
...
const w = wrap(new Worker(new URL('workers/worker.js', import.meta.url)));
```

If I inline the web-worker in my code, the error goes away. This not a good workaround, however, as the performance of the application will suffer.

dungnguyenBKA:
I also faced this bug. My application have 1 worker and it bug occur. When I comment out worker then app start working. Is that any update on how to fix that temporary @hverlin ?

hverlin:
I did not find a way around this issue so far.

hverlin:
bump

chebotaevroman:
bump

workaround: use - ModuleFederationPluginV1

hverlin:
I can still reproduce in my project with the latest rsbuild version

technikhil314:
Same here I am trying to use webworker and i get this error in webworker chunk

technikhil314:
Any update help needed on this? If anyone can give me pointers I can start debugging? @ScriptedAlchemy Can you plz help here?

CKanishka:
Another workaround that worked for me was to use the [worker-rspack-loader](https://github.com/rspack-contrib/worker-rspack-loader). Not the recommended approach, but it works for the time being until the module federation fix goes live.

ScriptedAlchemy:
That PR referenced here may address the problem as side effect of solving other issues. Similar issue webpack had was also solved with hoisted references when I introduced some of these capabilities back to webpack core

PR #10524, opened 2025-05-30
feat(plugin_mf): Module Hoisting, Entry initialization via __webpack_require__.x, Custom Hooks

ScriptedAlchemy:
## Summary

This pull request introduces significant enhancements to the Module Federation system by implementing new runtime modules, plugins, and dependency management features. The changes focus on enabling runtime dependency tracking, improving federation runtime initialization, and adding new hooks for better plugin communication.

### Module Federation Enhancements:

* **Embed Federation Runtime Module**: Added a new runtime module, `EmbedFederationRuntimeModule`, which ensures federation runtime dependencies execute before other modules by wrapping the startup function. This module generates a startup wrapper pattern to manage execution order. (`crates/rspack_plugin_mf/src/container/embed_federation_runtime_module.rs`)

* **Embed Federation Runtime Plugin**: Introduced `EmbedFederationRuntimePlugin` to manage federation runtime initialization. It adds runtime requirements to federation-enabled chunks, injects the `EmbedFederationRuntimeModule` into runtime chunks, and handles explicit startup calls for entry chunks. (`crates/rspack_plugin_mf/src/container/embed_federation_runtime_plugin.rs`)

* **Federation Runtime Dependency**: Implemented a new dependency type, `FederationRuntimeDependency`, to track runtime dependencies for federation modules. This dependency is cacheable and integrates with the existing dependency system. (`crates/rspack_plugin_mf/src/container/federation_runtime_dependency.rs`)

### Plugin Communication and Hooks:

* **Federation Modules Plugin**: Added `FederationModulesPlugin`, a central hook management system for Module Federation. It provides hooks for plugins to share dependency information and communicate through a publish-subscribe pattern. (`crates/rspack_plugin_mf/src/container/federation_modules_plugin.rs`)

* **Runtime Module for Federation**: Added `FederationRuntimeModule`, a runtime module that initializes the federation runtime environment. It includes a chunk matcher and root output directory logic to support federation runtime initialization. (`crates/rspack_plugin_mf/src/container/federation_runtime_module.rs`)

### Additional Changes:

* **New Options for Module Federation Runtime Plugin**: Added `RawModuleFederationRuntimePluginOptions` to support custom configurations like `entry_runtime` and `runtime_chunk`. These options are used to initialize the `ModuleFederationRuntimePlugin`. (`crates/node_binding/src/raw_options/raw_builtins/raw_mf.rs`)

* **Dependency Updates**: Updated `Cargo.toml` to include new dependencies such as `rspack_plugin_javascript` and `rspack_sources`, which are required for the new runtime modules and plugins. (`crates/rspack_plugin_mf/Cargo.toml`)
## Other notes

Should reduce bundle size by 70kb per entrypoint when there are multiple entrypoints with `runtimeChunk: single`


crates/rspack_plugin_mf/src/container/embed_federation_runtime_module.rs
```diff
@@ -0,0 +1,111 @@
...
+    let mut runtime_requirements = RuntimeGlobals::default();
+    let mut module_executions = Vec::new();
+
+    for dep_id in federation_runtime_modules {
+      let module_str = module_raw(compilation, &mut runtime_requirements, &dep_id, "", false);
+      module_executions.push(format!("\t\t{}", module_str));
+    }
+
+    // Generate oldStartup wrapper pattern
+    let result = format!(
```
ScriptedAlchemy, line 87: patch __webpack_require__.x which is entry startup function. 

first initialize federation, then call previous startup if exists - if it exists, it will be webpack require of entrypoint module. 

If startup doesnt exist in entrypoint bootstrap, like in single runtime chunk, another plugin will manually add require.x() to the entrypoint chunks at the begenning of startup

crates/rspack_plugin_mf/src/container/embed_federation_runtime_plugin.rs
```diff
@@ -0,0 +1,244 @@
+//! # EmbedFederationRuntimePlugin
```
ScriptedAlchemy, line 1: Embeds the module federation runtime module into runtime chunks, handles management of startup calls and rendering the startup of entrypoints.

crates/rspack_plugin_mf/src/container/federation_modules_plugin.rs
```diff
@@ -0,0 +1,118 @@
...
+  federation_runtime_dependency::FederationRuntimeDependency,
+};
+
+// Hook definitions
+define_hook!(AddContainerEntryDependencyHook: Series(dependency: &ContainerEntryDependency));
+define_hook!(AddFederationRuntimeDependencyHook: Series(dependency: &FederationRuntimeDependency));
+define_hook!(AddRemoteDependencyHook: Series(dependency: &dyn Dependency));
+
+// Struct to hold instances of the hooks for a specific compilation
+pub struct FederationModulesPluginCompilationHooks {
```
ScriptedAlchemy, line 31: adds compilation hooks module federation so that communication between all plugins and modules can be simplified.

For instance we need to know when container plugin is added, we need to know when federation runtime dependency is injected, and have to track various modules in the graph throughout the compilation.

crates/rspack_plugin_mf/src/container/module_federation_runtime_plugin.rs
```diff
@@ -1,122 +1,78 @@
...
   _runtime_requirements: &mut RuntimeGlobals,
 ) -> Result<()> {
+  // Add base FederationRuntimeModule
   compilation.add_runtime_module(chunk_ukey, Box::<FederationRuntimeModule>::default())?;
+
+  Ok(())
+}
+
+#[plugin_hook(CompilerFinishMake for ModuleFederationRuntimePlugin)]
+async fn finish_make(&self, compilation: &mut Compilation) -> Result<()> {
```
ScriptedAlchemy, line 56: instead of injecting a entrypoint or patching entry dependencies into each entrypoint. We will use a dependency to addInclude the module. 

This will inject it with no require calls, so that our runtime module can handle startup of federation ahead of the entrypoint.

crates/rspack_plugin_mf/src/container/embed_federation_runtime_plugin.rs
```diff
@@ -0,0 +1,244 @@
...
+  // Skip build time chunks
+  if chunk.name() == Some("build time chunk") {
+    return Ok(None);
+  }
+
+  // Only inject EmbedFederationRuntimeModule into runtime chunks
+  let has_runtime = chunk.has_runtime(&compilation.chunk_group_by_ukey);
+  if has_runtime {
+    // Collect federation dependencies snapshot
+    let collected_ids_snapshot = self
```
ScriptedAlchemy, line 110: hooks work a little different, since i cannot nest hooks easily, i am passing the tap results from the parent

ahabhgk:
This seems fix https://github.com/web-infra-dev/rspack/issues/7417 since the federation runtime is injected by runtime module instead of entry

ahabhgk (dismissed):
LGTM

ScriptedAlchemy:
Yeah this fixed workers in v1 webpack 
V2 may need some extra adjustment but it'll likely fix the bundler side and if theres any remaining issue it'll be that I need to add importScript to my runtime

ahabhgk (approved):

kambing86:
just want to check if this commit breaks `@module-federation/retry-plugin` with `build plugin registration` in React

found error `Uncaught ReferenceError: $RefreshReg$ is not defined` when upgrading `@rsbuild/core` from `1.4.8` to `1.4.9`

ScriptedAlchemy:
Hmm do you have a sample repo? @kambing86

kambing86:
@ScriptedAlchemy I just created one https://github.com/kambing86/react-module-federation-test

it seems like using `@module-federation/enhanced/rspack` is fine

but using `@module-federation/rsbuild-plugin` has issue

the last commit in `master` branch is the one that will cause the issue

screenshot here
Screenshot 2025-07-31 at 2 54 56 PM

kambing86:
and you could checkout branch `downgrade` in that sample repo that should be working by downgrading `@rsbuild/core` to `1.4.8`

ScriptedAlchemy:
gotcha, it seems react refresh is being applied too aggressively to all modules.

2heal1:
@kambing86 @ScriptedAlchemy I find the reason , `@module-federation/rsbuild-plugin` will add `@module-federation/` to source.include which will cause react-refresh-loader consume it as well . 

I change it to detail packages to avoid this case:   https://github.com/module-federation/core/pull/3971

kambing86:
> @kambing86 @ScriptedAlchemy I find the reason , `@module-federation/rsbuild-plugin` will add `@module-federation/` to source.include which will cause react-refresh-loader consume it as well .
> 
> I change it to detail packages to avoid this case: [module-federation/core#3971](https://github.com/module-federation/core/pull/3971)

@ScriptedAlchemy @2heal1 thanks, I can confirm that the latest version fixed the issue by using the latest version in my [repo](https://github.com/kambing86/react-module-federation-test)

PR #10531, opened 2025-05-30
feat: support dependences in build info

SyMind:
## Summary

Support follow js api:
- buildInfo.fileDependencies
- buildInfo.contextDependencies
- buildInfo.missingDependencies
- buildInfo.buildDependencies

And fix the module.buildInfo typescript type.

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -24,8 +24,8 @@ export interface Module {
 	get useSourceMap(): boolean;
 	get useSimpleSourceMap(): boolean;
 	get _readableIdentifier(): string;
-	buildInfo: Record<string, any>;
-	buildMeta: Record<string, any>;
+	buildInfo: BuildInfo & Record<string, any>;
+	buildMeta: BuildInfo & Record<string, any>;
 }
 
 interface NormalModuleConstructor {
@@ -76,6 +76,10 @@ export declare class AsyncDependenciesBlock {
 
 export declare class BuildInfo {
   get _assets(): Assets
+  get _fileDependencies(): Array<string>
+  get _contextDependencies(): Array<string>
+  get _missingDependencies(): Array<string>
+  get _buildDependencies(): Array<string>
 }
 
 export declare class Chunks {
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -24,8 +24,8 @@ export interface Module {
 	get useSourceMap(): boolean;
 	get useSimpleSourceMap(): boolean;
 	get _readableIdentifier(): string;
-	buildInfo: BuildInfo & Record<string, any>;
-	buildMeta: BuildInfo & Record<string, any>;
+	buildInfo: KnownBuildInfo & Record<string, any>;
+	buildMeta: KnownBuildInfo & Record<string, any>;
 }
 
 interface NormalModuleConstructor {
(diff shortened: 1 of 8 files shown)


jerrykingxyz (approved):

SyMind:
## Community usage of module.buildInfo dependency fields

Verified on 2026-09-09. These are downstream Rspack reads; their functional necessity needs to be assessed separately.

- **OpenSearch Dashboards** creates its compilers with `@rspack/core` and directly iterates `module.buildInfo.fileDependencies` for SCSS modules in the compiler's `done` callback. It records the paths in the bundle's referenced-file metadata. Removing the getter without adapting this caller would make the iteration throw. See the [consumer](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a20f5cc0502d76fc6cec6755bbc37c50ee2c9740/packages/osd-optimizer/src/worker/run_compilers.ts#L127) and [worker entry point](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a20f5cc0502d76fc6cec6755bbc37c50ee2c9740/packages/osd-optimizer/src/worker/run_worker.ts#L129).

  **Current purpose is limited:** the old application-level cache was removed in [OpenSearch Dashboards #11866](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/11866), and its external watcher was removed in [#12070](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/12070). In the current repository search, calls to `getReferencedFiles()` were found only in tests, alongside the getter's definition; see the [integration assertions](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/a20f5cc0502d76fc6cec6755bbc37c50ee2c9740/packages/osd-optimizer/src/integration_tests/basic_optimization.test.ts). This file-list collection may be leftover bookkeeping. It establishes a compatibility dependency, but does not demonstrate that the current build needs this data for cache invalidation or watching.

- **`@octanejs/rspack-plugin`** reads all four dependency collections from CSS provider modules during its optional `cssModuleConstants` optimization. It folds proven CSS class strings into importer output, then forwards the dependencies of actually consumed providers through the importer's loader context using `addDependency()`, `addContextDependency()`, `addMissingDependency()`, and `addBuildDependency()`. See the [reads](https://github.com/octanejs/octane/blob/f86fb48c47a70de868c14ebcbbb11266376d172f/packages/rspack-plugin-octane/src/css-module-constants.js#L126) and [propagation](https://github.com/octanejs/octane/blob/f86fb48c47a70de868c14ebcbbb11266376d172f/packages/rspack-plugin-octane/src/css-module-data.js#L70).

  **Scope matters:** the optimization is disabled by default, runs only in one-shot production builds, and makes eligible importers uncacheable; ordinary stylesheet references are preserved. See the [documented scope](https://github.com/octanejs/octane/blob/f86fb48c47a70de868c14ebcbbb11266376d172f/packages/rspack-plugin-octane/README.md#L86) and [loader implementation](https://github.com/octanejs/octane/blob/f86fb48c47a70de868c14ebcbbb11266376d172f/packages/rspack-plugin-octane/src/css-module-data.js#L18). Removing the getters would omit this dependency propagation because the reads tolerate missing values. These sources alone do not demonstrate a resulting output, watch, or persistent-cache regression.

The existing snapshot getters support both read paths. This is evidence to account for when changing the API, rather than proof that the API is indispensable. A removal proposal should distinguish adapting existing callers from demonstrating whether module-level dependency propagation is required for their supported behavior.

PR #10535, opened 2025-05-30
fix: Add $schema field for the swc config validation

n0th1ng-else:
## Summary

According to swc documentation, we are able to narrow the json type for the configuration object.

https://swc.rs/docs/configuration/swcrc

```json
{
  "$schema": "https://swc.rs/schema.json",
}
```

In fact, rspack now throws a validation error

```
- Invalid options of 'builtin:swc-loader': Invalid configuration object. Rspack has been initialized using a configuration object that does not match the API schema.
- Unrecognized key(s) in object: '$schema' at "module.rules[0].oneOf[6]"
```

I suggest fixing the validation schema

## Checklist


chenjiahan (approved):
LGTM, thank you!

PR #10537, opened 2025-05-31
perf: regex from js and js resolver api

SyMind:
## Summary

There are tow perf for RespckRegex:

1. Eliminated redundant initialization
Removed duplicate RspackRegex::new calls in crates/node_binding/src/raw_options/raw_module/mod.rs by caching instances.

2. Reduced N-API calls
Optimized FromNapiValue implementation for RspackRegex to minimize underlying N-API function invocations.

## Checklist


quininer (approved):
LGTM

PR #10538, opened 2025-05-31
fix: Use correct tag for canary release

irinaklimova:
## Summary

I have noticed that in the documentation its suggested to use `latest` tag for the rspack-canary packages. But latest released version for the package actually has a `canary` tag. The version by `latest` tag seems to be outdated and not updated anymore.

Because of the alleged inconsistency of documentation and versions tags, I suggest changing the tag in the canary release workflow from `canary` to `latest`.

Alternative would be to update the documentation.

Links:
https://rspack.rs/guide/start/quick-start#install-canary-version

Screenshot 2025-05-31 at 11 35 50

https://www.npmjs.com/package/@rspack-canary/core?activeTab=versions

Screenshot 2025-05-31 at 11 37 40

## Checklist

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -104,6 +104,6 @@ jobs:
         run: |
           ./x version snapshot
           pnpm run build:js:canary
-          ./x publish snapshot --tag canary
+          ./x publish snapshot --tag latest
         env:
           NPM_TOKEN: ${{ secrets.RSPACK_CANARY_RELEASE_TOKEN }}


chenjiahan (commented):
It seems more reasonable to use "latest" dist tag, but we should also update the https://github.com/rspack-contrib/install-rspack package because it uses the canary dist-tag of `@rspack-canary/core` by default.

irinaklimova:
@chenjiahan thank you for the feedback. I had a look into intall script and opened another PR https://github.com/rspack-contrib/install-rspack/pull/6
Please, have a look.

SoonIter (approved):
LGTM❤️️

PR #10540, opened 2025-05-31
skip calling js loaders that don't have a pitch function

### LINKED ISSUE #10140: [Feature]: skip call js side pitch loader-runner if loader doesn't implement pitch
It seems rspack will call js side pitch loaderRunner even js loader doesn't implement pitch method which will cause extra tsfn call cost, so it would be better if rspack can skip call loaderRunner if no pitch is implemented

nilptr:
## Summary

close #10140

## Checklist


h-a-n-a (approved):
Thanks!

PR #10551, opened 2025-06-03
chore: Enable no-embed-metadata

quininer:
## Summary

see https://kobzol.github.io/rust/rustc/2025/06/02/reduce-cargo-target-dir-size-with-z-no-embed-metadata.html

This reduced 700M in a fresh profiling build.

before:
```
4.3G ┌─┴ target
```

after:
```
3.6G ┌─┴ target
``` 

## Checklist

### DIFF
--- a/.github/workflows/reusable-build-build.yml
+++ b/.github/workflows/reusable-build-build.yml
@@ -58,7 +58,6 @@ jobs:
       - name: Trim paths
         run: |
           echo $'\n' >> .cargo/config.toml
-          echo '[unstable]' >> .cargo/config.toml
           echo 'trim-paths = true' >> .cargo/config.toml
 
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-build-build.yml
```diff
@@ -58,7 +58,6 @@ jobs:
       - name: Trim paths
         run: |
           echo $'\n' >> .cargo/config.toml
-          echo '[unstable]' >> .cargo/config.toml
           echo 'trim-paths = true' >> .cargo/config.toml
```
hardfist, line 61: is this needed removed too?

quininer: Yes, because it is added to `.cargo/config.toml`

quininer: oh I misread the line. this does not need to removed, this helps keep ci cache stable.

PR #10553, opened 2025-06-03
feat: add support for SHA-256

pgoldberg:
## Summary

Rspack doesn't support SHA-256 hashes, which makes migrating from Webpack difficult for projects that rely on SHA-256 content hashes, since it only supports MD4 and xxHash64. This PR adds support for SHA-256.

## Checklist


hardfist:
@pgoldberg can you help explain more how the project relys on SHA-256 content hashes?

pgoldberg:
> @pgoldberg can you help explain more how the project relys on SHA-256 content hashes?

Sure! I use a content-addressable storage model for serving static files from many bundles, where files from these bundles are all accessible via the same route (e.g. `/assets/<SHA256>.js`). This comes with the benefit of sharing a browser cache between bundles, while removing a layer of separation that would exist if we also included a bundle name in the route (e.g. `/assets/<bundle-name>/<file-hash>.js`).

Using a cryptographic hash, like SHA256, has made the security model of this solution easier to reason about in the event these bundles may come from several sources, because it removes the possibility of a dev constructing an intentional hash collision that allowed a file from one bundle to be injected into a different bundle. The two options provided by Rspack today, `md4` and `xxhash64`, are both non-cryptographic hashes where it's easy to find/manufacture collisions.

pgoldberg:
@hardfist Just wanted to nudge this one – I'm hoping to get this in for a project I'm working on. I'm happy to clarify any questions you might have about it!

hardfist:
@pgoldberg thanks for your contribution, LGTM

hardfist (dismissed):

hardfist:
@pgoldberg you also need to update the api snapshot by running ./x ae update

quininer (dismissed):
LGTM.

Since we don't provide consistent output with webpack, we could use a more modern algorithm here (like blake3). but for now sha2 is good enough.

pgoldberg:
@hardfist Thank you – just pushed that up!

stormslowly:
@pgoldberg i tried to fix the CI and merged main branch. 
If this make you confused, sorry for that.

hardfist (approved):

PR #10578, opened 2025-06-05
feat: serialized build info custom fields from js

SyMind:
## Summary

This PR synchronizes user-defined fields on the JavaScript buildInfo object to Rust, enabling persistent caching of these custom fields on the Rust side.

Additionally, minor refactors are included leveraging napi-rs' new support for symbol-named properties in define_properties. Private properties are now consolidated using Symbols to prevent accidental exposure to users.

## Checklist

### DIFF
--- a/packages/rspack/src/loader-runner/index.ts
+++ b/packages/rspack/src/loader-runner/index.ts
@@ -61,6 +61,7 @@ import {
 	loadLoader,
 	runSyncOrAsync
 } from "./utils";
+import { syncCustomFields } from "../BuildInfo";
 
 function createLoaderObject(
 	loader: JsLoaderItem,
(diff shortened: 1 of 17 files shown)


packages/rspack/src/loader-runner/index.ts
```diff
@@ -1101,6 +1101,12 @@ export async function runLoaders(
 			id2: resource
 		}
 	});
+
+	if (compiler.options.experiments.cache && compiler.options?.cache) {
+		// Serialize buildInfo to the Rust side for later persistent caching by Rust.
+		context._module.buildInfo.serialize();
```
jerrykingxyz: Maybe we could rename it to `sync`? All data in buildinfo or other fields should be added on the rust side to ensure data consistency, not just for persistent cache.

SyMind: @jerrykingxyz KnownBuildInfo use getter/setter to sync data.

only custom fields need this function to sync data to Rust.

SyMind: rename to syncCustomFields

jerrykingxyz (approved):

Issue #10546, opened 2025-06-02
[Bug]: `webpackIgnore` magic comment is not working for `new URL` calls


leonardlib:
### System Info

System:

  OS: macOS 15.3.2
  CPU: (8) arm64 Apple M1
  Memory: 96.55 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh

Binaries:

  Node: 22.13.1 - ~/.nvm/versions/node/v22.13.1/bin/node
  Yarn: 1.22.22 - ~/.nvm/versions/node/v22.13.1/bin/yarn
  npm: 10.9.2 - ~/.nvm/versions/node/v22.13.1/bin/npm

Browsers:

  Chrome: 136.0.7103.114
  Edge: 137.0.3296.52
  Safari: 18.3.1

npmPackages:

  @rspack/cli: ^1.3.13 => 1.3.13 
  @rspack/core: ^1.3.13 => 1.3.13 
  @rspack/dev-server: ^1.1.2 => 1.1.2 
  @rspack/plugin-minify: ^0.7.5 => 0.7.5 
  @rspack/plugin-react-refresh: ^1.4.3 => 1.4.3 

### Details

You already made a fix for this behaviour: https://github.com/web-infra-dev/rspack/issues/10444

But now, for some reason, the build process is failing:

Here is my code:

```js
pdfjs.GlobalWorkerOptions.workerSrc = new URL(
  /* webpackIgnore: true */ 'pdfjs-dist/legacy/build/pdf.worker.min.js',
  import.meta.url,
).toString();
```

And the output is:

![Image](https://github.com/user-attachments/assets/e61da151-d0be-4d64-8f16-279d873d9b2d)

### Reproduce Steps

npx nx run my-app:build

PR #10591, opened 2025-06-06
fix: Webpack config test cases falsely pass when compilation error occur

nilptr:
## Summary

I noticed this issue when investigating why #10546 was not caught.

reference:

https://github.com/webpack/webpack/blob/19ca74127f7668aaf60d59f4af8fcaee7924541a/test/ConfigTestCases.template.js#L712

## Checklist


nilptr:
Sorry, this PR is a silly mistake. Please ignore. Sorry for the interruption.

Issue #10467, opened 2025-05-25
[Feature]: Upgrade to Zod V4


colinaaa:
### What problem does this feature solve?

[Zod](https://github.com/colinhacks/zod) has released a new major v4, which is faster, slimmer, more `tsc`-efficient, and implements some long-requested features. See [Introducing Zod 4](https://zod.dev/v4) for more details.

Key migration points include:

- [ ] Update deprecated APIs. E.g: `z.record()` with a single argument: https://zod.dev/v4/changelog#drops-single-argument-usage
- [ ] The result of `z.function()` is no longer a Zod schema: https://zod.dev/v4/changelog#zfunction
- [ ] The `ZodRspackCrossChecker` which introduced in #8369 and #10055 (for `externals` and options of `builtin:swc-loader`)
- [ ] The `zod-validation-error` does not support Zod v4 for now. See: https://github.com/causaly/zod-validation-error/issues/455

### What does the proposed API of configuration look like?

No changes to public APIs are required.

PR #10602, opened 2025-06-07
refactor(zod/v4): replace `z.function` with `z.custom`

colinaaa:
## Summary

The result of `z.function()` in Zod v4 is no longer a schema: https://zod.dev/v4/changelog#zfunction. We need to migrate when upgrading to v4.

Since we are not exporting the Zod-generated types now, I think it would be fine to replace all the `z.function` with `z.custom<(...args: unknown[]) => any>`.

```ts
export const anyFunction = z.custom<(...args: unknown[]) => any>(
	data => typeof data === "function",
	input => ({ message: `Expected function, received ${getParsedType(input)}` })
);
```

<details><summary>Alternatives</summary>
<p>

I've tried the workaround at https://github.com/colinhacks/zod/issues/4143#issuecomment-2845134912 (which is recommended in the migration guide). It works with type but it will make validation pass for all inputs.

I've also tried to use generics in `anyFunction`, which could keep all the types of Zod schema:

```ts
export function anyFunction<T>() {
	return z.custom<T>(
		data => typeof data === "function",
		input => ({ message: `Expected function, received ${getParsedType(input)}` })
	);
}
```

However, this approach would involve additional function calls (and some additional objects/functions are created), potentially reducing performance, and could also slow down TypeScript's type-checking process.

I decided to opt for a universal `anyFunction` instead.

</p>
</details> 

issue: #10467 

## Checklist

I add the tests in [ec1c292](https://github.com/web-infra-dev/rspack/pull/10602/commits/ec1c292b715fdfd431b2e6a183bcd33b98b24a1c) and replace the `z.function` in [1200be4](https://github.com/web-infra-dev/rspack/pull/10602/commits/1200be4d6f5a61a909c1d5eb896c4209a43cc772) without modifying the tests.

### DIFF
--- a/packages/rspack/src/config/utils.ts
+++ b/packages/rspack/src/config/utils.ts
@@ -18,6 +18,7 @@ import {
 	ZodUnion,
 	type ZodUnionOptions,
 	addIssueToContext,
+	getParsedType,
 	z
 } from "zod";
 import type { RspackOptions } from "./types";
@@ -222,3 +223,10 @@ export class ZodRspackCrossChecker<T> extends ZodType<T> {
 		return root.data;
 	}
 }
+
+export const anyFunction = z.custom<(...args: unknown[]) => any>(
+	data => typeof data === "function",
+	// Make the similar error message as zod v3
+	// https://github.com/colinhacks/zod/blob/64bfb7001cf6f2575bf38b5e6130bc73b4b0e371/packages/zod/src/v3/types.ts#L3821-L3828
+	input => ({ message: `Expected function, received ${getParsedType(input)}` })
+);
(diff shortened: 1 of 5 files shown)


packages/rspack/src/config/utils.ts
```diff
@@ -222,3 +223,10 @@ export class ZodRspackCrossChecker<T> extends ZodType<T> {
 		return root.data;
 	}
 }
+
+export const anyFunction = z.custom<(...args: unknown[]) => any>(
```
colinaaa, line 227: Switching to `z.custom().transform` might be a better approach. We could consider adopting this method when migrating to `zod/v4`.

See: https://github.com/colinhacks/zod/issues/4143#issuecomment-2931729793

chenjiahan (approved):
LGTM 👍 I agree that anyFunction is enough for us, there is no need to define the types of the function input and output.

PR #10623, opened 2025-06-10
feat(incremental): enable advance-slient preset by default

ahabhgk:
## Summary

enable `incremental: "advance-slient"` by default

tests remain the same, incremental is only tested in `rspack-test-tools/tests/Incremental-*.test.js`, other tests still test `incremental: "safe"` (before's behavior)

## Checklist


chenjiahan (approved):
🚀

PR #10624, opened 2025-06-10
feat(core): expose `StatsErrorCode`

h-a-n-a:
## Summary

Export `StatsErrorCode` to support users filtering stats errors based on its code.
Check the original PR for the supported code: https://github.com/web-infra-dev/rspack/pull/10057

```js
import { rspack, StatsErrorCode } from '@rspack/core';

rspack(<your-config-here>, (err, stats) => {
  if (err) {
    // handle error
    return;
  }

  // filter minification error
  const hasMinificationError = stats.toJson({ errors: true }).errors
    ?.some(e => e.code === StatsErrorCode.ChunkMinificationError);

  if (hasMinificationError) {
    // handle minification error
    return;
  }
});
```

## Checklist

### DIFF
--- a/packages/rspack/etc/core.api.md
+++ b/packages/rspack/etc/core.api.md
@@ -6296,6 +6296,7 @@ declare namespace rspackExports {
         StatsCompilation,
         StatsError,
         StatsModule,
+        StatsErrorCode,
         Stats,
         RuntimeModule,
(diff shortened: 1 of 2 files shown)


chenjiahan:
Our documentation need to be updated 😄

PR #10640, opened 2025-06-11
feat: add swc transformSync and minifySync api

GiveMe-A-Name:
## Summary

Support swc transformSync and minifySync api so that user can call it in sync function.

## Checklist

### DIFF
--- a/crates/node_binding/src/swc.rs
+++ b/crates/node_binding/src/swc.rs
@@ -23,8 +23,7 @@ impl From<CompilerTransformOutput> for TransformOutput {
   }
 }
 
-#[napi]
-pub async fn transform(source: String, options: String) -> napi::Result<TransformOutput> {
+fn _napi_transform(source: String, options: String) -> napi::Result<TransformOutput> {
   let options: SwcOptions = serde_json::from_str(&options)?;
   let compiler = JavaScriptCompiler::new();
   compiler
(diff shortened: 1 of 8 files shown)


crates/node_binding/src/swc.rs
```diff
@@ -40,7 +39,25 @@ pub async fn transform(source: String, options: String) -> napi::Result<Transfor
...
 
 #[napi]
-pub async fn minify(source: String, options: String) -> napi::Result<TransformOutput> {
+pub async fn transform(source: String, options: String) -> napi::Result<TransformOutput> {
+  _napi_transform(source, options)
+}
+
+#[napi]
+pub fn transform_sync(source: String, options: String) -> napi::Result<TransformOutput> {
+  let future = async move { _napi_transform(source, options) };
```
hardfist: why not just call `_napi_transform` in `within_runtime_if_available` directly, it seems we don't need to create a future

GiveMe-A-Name: Sorry, I don't known `within_runtime_if_available` means. Is means user provide a tokio runtime?
Swc transform api may import wasm plugin, if it has no tokio runtime it will panic in wasmer runtime;
image

So i warp it into a future so that tokio can run it.

hardfist: image
it's a napi  that can inject tokio_runtime for sync function

GiveMe-A-Name: Done~

hardfist: we also need add same test cases for transformSync especially about wasm plugin

GiveMe-A-Name: Done~

hardfist (approved):

PR #10643, opened 2025-06-11
perf: align number hash algorithm with webpack

### LINKED ISSUE #10614: [Bug]: `get_number_hash` add with overflow in 32bit arch
## Details

https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_util/src/number_hash.rs uses `usize` to compute some hash numbers. But it causes `add with overflow` in 32bit arch where `usize` is actually `u32`.

We may change it to `u64`. However, I find this is a port of https://github.com/webpack/webpack/blob/4b4ca3bb53f36a5b8fc6bc1bd976ed7af161bd80/lib/util/numberHash.js. But actually webpack has updated their algorithm https://github.com/webpack/webpack/blob/main/lib/util/numberHash.js

nilptr:
## Summary

close #10614 

## Checklist


CPunisher:
🤔 Is this a breaking change cc @hardfist

crates/rspack_util/src/number_hash.rs
```diff
@@ -5,34 +5,45 @@
...
-  if range <= SAFE_PART {
-    ((arr[0] + arr[1] + arr[2] + arr[3]) % range as u64) as usize
+  hash & MASK_31
+}
+
+fn fnv1a64(s: &str) -> u64 {
+  let mut hash = FNV_OFFSET_64;
+
+  for code_unit in s.encode_utf16() {
+    hash ^= code_unit as u64;
```
quininer, line 35: This is actually wrong, fnv requires calculations in bytes, not chars.

see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#:~:text=is%20an%208-bit%20unsigned%20integer.

crates/rspack_util/src/number_hash.rs
```diff
@@ -5,34 +5,45 @@
...
+    hash = hash.wrapping_mul(FNV_PRIME_64);
+  }
+
+  hash
+}
+
+pub fn get_number_hash(s: &str, range: usize) -> usize {
+  if range < FNV_64_THRESHOLD {
```
quininer, line 43: I think webpack's choice here doesn't make sense, the performance advantage between 32bit fnv and 64bit fnv doesn't correspond to input length, but cpu arch.

I suggestion just using 64bit fnv1a.

crates/rspack_util/src/number_hash.rs
nilptr: I don’t have a strong preference between using the original fnv or the webpack variant. I just assumed the issue was meant to follow webpack’s changes and stay consistent with it — which would also provide a useful reference point for validation.

If your team concludes that the original version is better, just let me know — I can update the code along with the snapshots and test cases.

quininer: I suggestion using original fnv, which avoids the overhead of encode utf16

CPunisher (approved):

CPunisher:
We can merge it first, and optimize it in the future. Also we can submit the optimal algorithm to webpack.

PR #10654, opened 2025-06-12
fix: rspack block after removing dynamic entry

SyMind:
## Checklist


crates/rspack_core/src/compiler/make/cutout/mod.rs
```diff
@@ -140,6 +140,12 @@ impl Cutout {
...
       }
 
+      // should not build remove_entry_dependencies
+      build_deps = build_deps
+        .drain()
+        .filter(|(dep_id, _)| !remove_entry_dependencies.contains(dep_id))
+        .collect();
+
       for dep_id in remove_entry_dependencies {
         artifact.revoke_dependency(&dep_id, true);
```
jerrykingxyz, line 150: the useless entry will be revoked force, and it dependency at module graph will be removed. So those dependency will be filter at  https://github.com/web-infra-dev/rspack/blob/3c8b4b88945fbce6fee36ad26ce4a1fd71a79b90/crates/rspack_core/src/compiler/make/cutout/mod.rs#L164

jerrykingxyz: `build_deps` can be understood as the maximum set of values ​​that need to be built. It should be a structure that can only be added but not deleted.

SyMind: The current issue is that Next.js utilizes dynamic entry. When deleting an entry such as `foo/page.js`, the rebuild process gets stuck at the repair stage.

Although the corresponding `EntryDependency` becomes invalid after removing `foo/page.js`, it persists in the module graph.

Consequently, it cannot be filtered out via `module_graph.dependency_by_id(dep_id)`.

PR #10584, opened 2025-06-05
perf: introduce `ModuleGraphCache` and cache the result of `get_mode` and `determine_export_assignments`

CPunisher:
## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #10657, opened 2025-06-13
perf: reduce memory allocation in `esm_export_imported_specifier_dependency`

CPunisher:
## Summary

Some performance improvements after https://github.com/web-infra-dev/rspack/pull/10584

1. Split `ExportsMode` to enum variants and box the `DynamicReexport` case since it's larger than other variants. Data size: 112 bytes -> 40 bytes. This change also eliminates unnecessary `Option::expect`.
2. Avoid unnecessary `Vec` to `HashSet` reallocation. The data structures I change from `Vec` to `HashSet` in this pr are actually `Set`s in webpack.
3. Use `map.clear` rather than reallocate a new map in the loop.

## Checklist

### DIFF
--- a/crates/rspack_core/src/dependency/mod.rs
+++ b/crates/rspack_core/src/dependency/mod.rs
@@ -37,7 +37,7 @@ pub use runtime_requirements_dependency::{
   RuntimeRequirementsDependency, RuntimeRequirementsDependencyTemplate,
 };
 pub use runtime_template::*;
-use rustc_hash::FxHashMap;
+use rustc_hash::{FxHashMap, FxHashSet};
 use serde::Serialize;
 pub use span::SpanExt;
 pub use static_exports_dependency::{StaticExportsDependency, StaticExportsSpec};
@@ -106,8 +106,8 @@ pub struct ExportsSpec {
   pub terminal_binding: Option<bool>,
   pub from: Option<ModuleGraphConnection>,
   pub dependencies: Option<Vec<ModuleIdentifier>>,
-  pub hide_export: Option<Vec<Atom>>,
-  pub exclude_exports: Option<Vec<Atom>>,
+  pub hide_export: Option<FxHashSet<Atom>>,
+  pub exclude_exports: Option<FxHashSet<Atom>>,
 }
 
 pub trait DependencyConditionFn: Sync + Send {
--- a/crates/rspack_core/src/exports/exports_info.rs
+++ b/crates/rspack_core/src/exports/exports_info.rs
@@ -129,7 +129,7 @@ impl ExportsInfo {
     &self,
     mg: &mut ModuleGraph,
     can_mangle: bool,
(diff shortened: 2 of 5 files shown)


crates/rspack_core/src/dependency/mod.rs
```diff
@@ -106,8 +106,8 @@ pub struct ExportsSpec {
   pub terminal_binding: Option<bool>,
   pub from: Option<ModuleGraphConnection>,
   pub dependencies: Option<Vec<ModuleIdentifier>>,
-  pub hide_export: Option<Vec<Atom>>,
-  pub exclude_exports: Option<Vec<Atom>>,
+  pub hide_export: Option<FxHashSet<Atom>>,
```
hardfist, line 109: should we need to keep insert order for these data structures? if so may need to use indexset instead?

CPunisher: I forget that js set is ordered...

CPunisher: 🤔 But the original code converts `HashSet` to `Vec`, which is also unordered

hardfist: if we know it's no need ordered then it's fine

crates/rspack_core/src/exports/exports_info.rs
```diff
@@ -129,7 +130,7 @@ impl ExportsInfo {
     &self,
     mg: &mut ModuleGraph,
     can_mangle: bool,
-    exclude_exports: Option<Vec<Atom>>,
+    exclude_exports: &Option<FxHashSet<Atom>>,
```
quininer: Generally always use `Option<&_>` instead of `&Option<_>`.

hardfist:
nice job! it's meaningful to remove unnecessary unwrap, we have too much unnecessary now

LingyuCoder (approved):

LingyuCoder (approved):

crates/rspack_core/src/exports/exports_info.rs
hardfist: what's the reason prefer Option<&> than &Option<_>? is it because Rust compiler can optimize Option<&>  better than &Option<_>?

quininer: It is always possible to get `Option<&_>` from `&Option<_>`, but it is difficult to get `&Option<_>` from `Option<&_>`. so as an API, `Option<&_>` is the more compatible choice.

In terms of performance, `Option<&_>` reduces one stack access in non-niche scenarios, but i don't think that constitutes an advantage. And if arguments passing is straightforward enough, the `&Option<_>` should be more advantage.

quininer: In this case specifically, I don't think it has an impact on performance. this is because the function is only called in one place. it should be inlined after lto and generate exactly the same code.

Issue #10642, opened 2025-06-11
[Bug]: Rspack + SWC cannot provide good enough tree shaking and DCE results for react-router-dom


chenjiahan:
### System Info

Rspack v1.3.x

### Details

Rspack + SWC cannot provide good enough tree shaking and DCE results for react-router-dom.

- Input:

```js
import { BrowserRouter, Routes, Route } from 'react-router-dom';

console.log(BrowserRouter, Routes, Route);
```

- `react-router-dom` has `sideEffects: false` in its package.json.

### Comparision

| Build Tool | Minimizer | Minified Size |
|------------|-----------|---------------|
| webpack | terser | 37 KB |
| Vite | esbuild | 42.67 kB |
| Rspack CLI 1.3 | terser | 43 KB |
| Rsbuild 1.3 | SWC | 83 kB |
| Rspack CLI 1.3 | SWC | 83 KB |
| webpack | SWC | 84 KB |

### Reproduce link

https://github.com/chenjiahan/rspack-repro-react-router-tree-shaking

### Reproduce Steps

1. pnpm i
2. npx rsbuild build
3. npx vite rsbuild
4. pnpm build:webpack
5. pnpm build:rspack

chenjiahan:
This issue can be broken down into two parts:

1. When using Terser, the bundle size of Rspack does not align with that of webpack in this case. @LingyuCoder cc

2. SWC minification is not as good as Terser in this case (according to our tests, SWC needs to set `passes` to 6~7 to get the similar size to Terser's `passes: 1`) @GiveMe-A-Name  cc

kdy1:
Before fix:

```
Total size: 83360
```

https://github.com/swc-project/swc/pull/10604 is the fix.

```
Total size: 40019
```

And, https://github.com/swc-project/swc/pull/10602 improves the situation further


```
Total size: 36503
```

kdy1:
I triggered the publish action for `v1.12.1` because I think this is an important bugfix.

- https://github.com/swc-project/swc/actions/runs/15620821137
-

chenjiahan:
Thank you Donny! It's great that SWC was able to resolve the issue so quickly ❤️

chenjiahan:
With the latest SWC:

| Build Tool | Minimizer | Minified Size |
|------------|-----------|---------------|
| Rspack CLI | SWC | 36 KB |
| webpack | SWC | 36 KB |
| Rsbuild | SWC | 36 kB |
[...]

PR #10660, opened 2025-06-13
chore: bump swc to 27.0.4

GiveMe-A-Name:
## Summary

- Update SWC dependencies to latest version 

SWC fixs dce bug, more detail see: [CHANGELOG-CORE.md](https://github.com/swc-project/swc/blob/main/CHANGELOG-CORE.md#swc_corev2704---2025-06-12)

## Relative PR
- https://github.com/web-infra-dev/rspack/issues/10642

## Checklist

### DIFF
--- a/crates/rspack_javascript_compiler/src/compiler/transform.rs
+++ b/crates/rspack_javascript_compiler/src/compiler/transform.rs
@@ -359,7 +359,6 @@ impl<'a> JavaScriptTransformer<'a> {
         parse_file_as_script(&fm, syntax, target, comments, &mut errors).map(Program::Script)
       }
       IsModule::Unknown => parse_file_as_program(&fm, syntax, target, comments, &mut errors),
-      // TODO:
       IsModule::CommonJS => {
         parse_file_as_commonjs(&fm, syntax, target, comments, &mut errors).map(Program::Script)
       }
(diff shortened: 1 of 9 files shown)


crates/rspack_javascript_compiler/src/compiler/transform.rs
```diff
@@ -477,6 +483,7 @@ impl<'a> JavaScriptTransformer<'a> {
         BoolOr::Bool(true) | BoolOr::Data(JsMinifyCommentOption::PreserveAllComments) => true,
         BoolOr::Data(JsMinifyCommentOption::PreserveSomeComments) => false,
         BoolOr::Bool(false) => false,
+        BoolOr::Data(JsMinifyCommentOption::PreserveRegexComments { .. }) => false,
```
GiveMe-A-Name, line 486: The value need confirm

hardfist:
should run bench

crates/rspack_javascript_compiler/src/compiler/transform.rs
CPunisher: If we stripe all comments with `JsMinifyCommentOption::PreserveSomeComments`, I think it's also ok to do so with `JsMinifyCommentOption::PreserveRegexComments { .. }`

CPunisher:
Accept if the known performance regression of minifier is acceptable.

CPunisher (approved):

chenjiahan:
yeah, smaller bundle size is more important than build performance and we can improve minifier performance through subsequent optimizations.

PR #10699, opened 2025-06-17
feat: introduce rspack_tasks 

hardfist:
## Summary
Introduce rspack_tasks and CURRENT_COMPILER_CONTEXT to set isolated compiler_context for multi compiler instance
## Motivation
currently all Id(DEPENDENCY,CODE_GEN_RESULT_ID,NEXT_CACHE_UKEY,EXECUTE_MODULE_ID ...) are all registed in global static, which makes mutli compiler will polluate each other, persistent cache will use these Id as storage key so it's important to make these ids isolated from other compilers.
To avoid passing id_generators every where to whole codebase, we introduce task_local isolated compiler_context to store these id_generator, which makes every compiler has its own implicit compiler_context. This is learning from [Rustc](https://github.com/rust-lang/rust/blob/e314b97ee54091b6bcf33db4770c93d82fded8bc/compiler/rustc_middle/src/ty/context/tls.rs#L53) and [Turbopack](https://github.com/vercel/next.js/blob/678ef8b5650871a730ca14c480c762ca53716575/turbopack/crates/turbo-tasks/src/manager.rs#L479) .

## Limitation
Need to make sure every id generation happens in compiler_context, normally this is not a big problem cause we already wrap top level api `build` and `rebuild` in compiler_context, so it's safe to get context in it, but needs be careful when call `spawn` and expose `napi` api, you need to explicit passing parent context to it

* use rspack_tasks::spawn_in_compiler_context to replace tokio::spawn
spaw_in_compiler_context will always passing parent context to child task, so it's always perferable to use spawn_in_compiler_context
```diff
- tokio::spawn(fut);
+ rspack_tasks::spawn_in_compiler_context(fut);
```

* explicit passing context for napi call
```rs
#[napi]
async fn xxx(){
  let compiler_context = compilation.compiler_context.clone();
  with_compiler_context(compiler_context, async {
     the_real_work().await;
  })
}
```
## Future Plan
* move all Id generator to CompilerContext
* introduce  DebuggingContext to simplify the debug work in debug environment like https://github.com/rust-lang/rust-analyzer/blob/a642aa8023be11d6bc027fc6a68c71c2f3fc7f72/crates/hir-ty/src/tls.rs#L17

## Checklist

### DIFF
--- a/tasks/benchmark/benches/groups/build_chunk_graph.rs
+++ b/tasks/benchmark/benches/groups/build_chunk_graph.rs
@@ -10,9 +10,7 @@ use rspack_core::{
   Compilation, Compiler, Experiments, Optimization,
 };
 use rspack_fs::{MemoryFileSystem, WritableFileSystem};
-use rspack_tasks::{
-  within_compiler_context_for_testing_sync, CompilerContext, CURRENT_COMPILER_CONTEXT,
-};
+use rspack_tasks::{within_compiler_context_for_testing_sync, CURRENT_COMPILER_CONTEXT};
 use tokio::runtime::Builder;
 
 static NUM_MODULES: usize = 10000;


crates/rspack_tasks/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+  f: F,
+) -> TaskLocalFuture<Arc<CompilerContext>, F>
+where
+  F: Future,
+{
+  CURRENT_COMPILER_CONTEXT.scope(compiler_context, f)
+}
+
+// this is only used for testing rust builder api, we need to find better api in the future
+pub fn within_compiler_context_for_testing_sync<F, R>(f: F) -> R
```
stormslowly, line 75:
```suggestion
#[cfg(test)]
pub fn within_compiler_context_for_testing_sync<F, R>(f: F) -> R
```

crates/rspack_tasks/src/lib.rs
```diff
@@ -0,0 +1,89 @@
...
+where
+  F: FnOnce() -> R,
+{
+  CURRENT_COMPILER_CONTEXT.sync_scope(Arc::new(CompilerContext::new()), f)
+}
+pub fn within_compiler_context_for_testing<F>(f: F) -> TaskLocalFuture<Arc<CompilerContext>, F>
```
stormslowly, line 82:
```suggestion
#[cfg(test)]
pub fn within_compiler_context_for_testing<F>(f: F) -> TaskLocalFuture<Arc<CompilerContext>, F>
```

crates/rspack/src/builder/mod.rs
```diff
@@ -3817,37 +3820,39 @@ mod test {
...
-    macro_rules! plugin_index {
-      ($ident:ident) => {
-        context
-          .plugins
-          .iter()
-          .position(|p| p.tag() == BuiltinPluginOptionsTag::$ident)
-          .expect("plugin should exist")
-      };
-    }
+    within_compiler_context_for_testing_sync(|| {
```
hardfist, line 3823: this is kind of workaround, I think we shouldn't call plugin out of compiler context, so it would be better that builder has some api like run_in_compiler to call these plugins like rustc's https://rustc-dev-guide.rust-lang.org/rustc-driver/getting-diagnostics.html#:~:text=rustc_interface%3A%3Arun_compiler(config%2C%20%7Ccompiler%7C%20%7B cc @h-a-n-a

crates/rspack_tasks/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+
+use tokio::{
+  task::{futures::TaskLocalFuture, JoinHandle},
+  task_local,
+};
+
+// don't overuse this and put everything here, it's mostly used for store isolated id generator
+#[derive(Debug)]
+pub struct CompilerContext {
+  dependenc_id_generator: AtomicU32,
```
jerrykingxyz, line 16: I don't like this encapsulation. rspack_tasks should be similar to a toolkit, encapsulating task local related functions. compiler_context should be what rspack_core needs to pay attention to.

hardfist: what's your suggest change?

quininer: If we do not consider `rspack_tasks` as an independent toolkit but a basic library for business services, it is reasonable to integrate some fields. this can effectively avoid unnecessary lookup overhead when dynamic access.

If we want rspack_tasks to be a independent toolkit crate, we can use `Vec<(TypeId, Box<dyn Any>)>` here instead.

jerrykingxyz: just expose within_context and get_context.
``` rust
// rspack_core
use rspack_tasks::{within_context, get_context, define_context};

// define_context will add a task_local macro here
#[define_context]
struct CompilerContext {
   dependency_id_generate: AtomicU32
}

fn build() {
   within_context(self.compiler_context, async { ... })
}

impl EntryDependency {
  pub fn new() {
    get_context::<CompilerContext>().fetch_new()
  }
}

```

quininer: I'm more concerned with the implement behind this `define_context` macro, which introduces some lookup overhead anyway. actually i don't think a macro is needed here.

hardfist: I personally don't think `rspack_tasks` is a generic toolkit, it's more like a specific core data structure which only works for compiler context, so I'm not sure `define_context` will bring benifits here

hardfist: And I'm open  with change the crate name to more specific name like `rspack_comiler_tasks`, It's just that I'm ok with current's naming

jerrykingxyz: Another small question, if debug_context is to be implemented later, the ModuleGraph type may need to be imported from rspack_core into compilerContext. How to solve this package circular reference? Use pointers directly?

hardfist: yeah just use dark magic, it's just my random thoughts, if it's causing problem, I'm totoally ok with remove the debugContext

jerrykingxyz (approved):

Issue #10406, opened 2025-05-19
[Feature] Config - Target - allow newer targets - es2023 etc.


natocTo:
Hi, is there reason why we can not use `ES2023` as target? Code - https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/config/types.ts#L1360

We can use in in `jsc` configuration, but it is not allowed in `target` now. It is only about updating the options or there is reason why it is not there? We would love to use same value for both.

chenjiahan:
I think we can add 'es2023' or 'es2024' as target

PR #10700, opened 2025-06-17
feat: extend target type to include es2023, es2024, and es2025

chenjiahan:
## Summary

Extend `target` type to include `es2023`, `es2024`, and `es2025`.

Since Rspack runtime currently only uses the `es2020` syntax at most, setting a higher es version will produce the same result as setting `es2020`.

resolve https://github.com/web-infra-dev/rspack/issues/10406

## Checklist

### DIFF
--- a/packages/rspack/etc/core.api.md
+++ b/packages/rspack/etc/core.api.md
@@ -108,7 +108,7 @@ interface AdditionalData {
 type AffectedHooks = keyof Compiler["hooks"];
 
 // @public (undocumented)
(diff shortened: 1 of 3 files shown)


GiveMe-A-Name:
It will more better, If we add those target into docs.
- https://github.com/web-infra-dev/rspack/blob/main/website/docs/en/config/target.mdx?plain=1#L23
- https://github.com/web-infra-dev/rspack/blob/main/website/docs/en/config/target.mdx?plain=1#L23

chenjiahan:
Sorry I missed that...Done 😄

GiveMe-A-Name (approved):

PR #10697, opened 2025-06-17
refactor: exports info getter part 4

LingyuCoder:
## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #10652, opened 2025-06-12
refactor: prefetch exports info data of getters part 3

LingyuCoder:
## Summary

- Introduce `PrefetchedExportsInfoUsed` to prevent fetching unnecessary named exports which caused the performance degression. Then modify the param of `get_used_name` to enum to support `PrefetchedExportsInfoUsed` for fast path when the names list is empty.
- Unify the way of accessing properties. Set all properties of `ExportsInfoData/ExportInfoData` to private and provide getter/setter to access them. 
- Move `get_target/find_target` related functions to `exports/target.rs`.
- Move referenced export related functions to `exports/referenced_export.rs`, and also rename `process_export_info` to `collect_referenced_export_items` to make the function name more informative

---
> Generate by AI

This pull request introduces multiple changes across several files to enhance the handling of export information and improve code efficiency. The updates include replacing direct field access with getter methods, introducing new enums for clarity, and optimizing hash computation. Below is a breakdown of the most important changes grouped by theme.

### Enhancements to Export Information Handling:
* Replaced direct field access in `ExportInfoData` with getter methods like `info.name()` and `info.global_used()` for better encapsulation and maintainability in `crates/rspack_core/src/exports/export_info_getter.rs`. [[1]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L8-R30) [[2]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L76-R60) [[3]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L124-R91) [[4]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L133-R106) [[5]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L156-R131)
* Updated `get_used_name` and `get_used` methods to use the new getter methods, ensuring consistent access patterns. [[1]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L76-R60) [[2]](diffhunk://#diff-5e159b75f36463e320bf142ca44e9f4ba0752110a9c02b63781ffe0d180829b0L156-R131)

### Code Simplification and Refactoring:
[...]

PR #10712, opened 2025-06-18
perf: improve flag exports plugin and flag usage plugin

LingyuCoder:
### Impove FlagDependencyExportsPlugin

Parallel collecting exports of dependencies by `dep.get_exports`.

Since there is a backtracking mechanism in the processing of dependency exports, when re - exporting, if there is a relationship A -> B, after A is computed, when computing B, if B's exports change, then A's exports need to be recomputed.

If the computation queue is `[B, A]`:
 - For serial computation, first compute B's exports and immediately update the B's exports information to the module graph. So when A computes its exports, the module graph contains B's exports information.
 - For parallel computation, since A and B will compute their exports simultaneously on an empty module graph, and the subsequent update of A and B's exports to the module graph is sequential. As a result, the state of the module graph when A computes its exports will not match the state when it is updated. Also, because B is updated first and then A, the backtracking and recomputation of A cannot be triggered, leading to incorrect computation results.

If the computation queue is `[A, B]`:
 - For serial computation, first compute A's exports and immediately update the A's exports information to the module graph, but this information is useless. Because later B's exports will be computed and updated to the module graph, then triggering the recomputation of A. At this time, the module graph has B's exports information when A is computing.
 - For parallel computation, after A and B compute their exports and update them to the module graph, since B is updated after A, it will cause A to be recomputed. At this time, the module graph has B's exports information when A is computing, so the correct result can be obtained.

Therefore, modules need to be sorted according to their dependency relationships to ensure that the backtracking computation will be triggered, and the results of the first parallel computation are discarded. Although this will result in some wasted computations, since the number of re - exported modules is not particularly large, parallel computation will still significantly improve performance. 

### Improve FlagDependencyUsagePlugin

parallel collecting referenced exports of dependencies by `dep.get_referenced_exports`

This plugin has the same issue with FlagDependencyExportsPlugin. But it has a queue internal and has handle the dependencies by recursive calling `collect_referenced_export_items`. So it is safe to just parallel getting their referenced dependencies and no need to sort them.

### Improve SideEffectsFlagPlugin

Just prevent CallExpr being cloned by sending the parent expr node to `is_pure_call_expr` function.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/flag_dependency_usage_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/flag_dependency_usage_plugin.rs
@@ -355,23 +355,21 @@ impl<'a> FlagDependencyUsagePluginProxy<'a> {
           let mut current_exports_info = mgm_exports_info;
           let len = used_exports.len();
           for (i, used_export) in used_exports.into_iter().enumerate() {
-            let export_info = current_exports_info.get_export_info(&mut module_graph, &used_export);
+            let export_info = current_exports_info
+              .get_export_info(&mut module_graph, &used_export)
+              .as_data_mut(&mut module_graph);
             if !can_mangle {
-              export_info
-                .as_data_mut(&mut module_graph)
-                .set_can_mangle_use(Some(false));
+              export_info.set_can_mangle_use(Some(false));
             }
             if !can_inline {
-              export_info
-                .as_data_mut(&mut module_graph)
-                .set_inlinable(Inlinable::NoByUse);
+              export_info.set_inlinable(Inlinable::NoByUse);
             }
             let last_one = i == len - 1;
             if !last_one {
-              let nested_info = export_info.as_data(&module_graph).exports_info();
+              let nested_info = export_info.exports_info();
               if let Some(nested_info) = nested_info {
                 let changed_flag = ExportInfoSetter::set_used_conditionally(
-                  export_info.as_data_mut(&mut module_graph),
+                  export_info,
                   Box::new(|used| used == &UsageState::Unused),
                   UsageState::OnlyPropertiesUsed,
                   runtime.as_ref(),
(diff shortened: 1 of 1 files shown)


LingyuCoder:
* **#10712** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/10712?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/10712?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#10652** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/10652?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#10697](https://github.com/web-infra-dev/rspack/pull/10697) <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/10697?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

crates/rspack_core/src/compiler/compilation.rs
```diff
@@ -1479,6 +1479,7 @@ impl Compilation {
   #[instrument("Compilation:seal", skip_all)]
   pub async fn seal(&mut self, plugin_driver: SharedPluginDriver) -> Result<()> {
     self.other_module_graph = Some(ModuleGraphPartial::default());
+    self.get_module_graph_mut().prepare_export_info_map();
```
LingyuCoder, line 1482: Cause all exports info will `get_mut` during flag exports and usage, so we can just clone the whole map to module graph partial to prevent loop partials of each export info. This should be removed after refactoring of export info.

JSerFeng:
LGTM cc @ahabhgk

ahabhgk (approved):
LGTM 👍

PR #10732, opened 2025-06-19
feat(rstest): more mock APIs

fi3ework:
## Summary

Support more mock APIs for Rstest, includes:

- `rs.mock`
- `rs.mockRequire`
- `rs.doMock`
- `rs.doMockRequire`
- `rs.importActual`
- `rs.requireActual`
- `rs.importMock`
- `rs.requireMock`
- `rs.unmock`
- `rs.doUnmock`
- `resetModules`

More test could be seen at https://github.com/web-infra-dev/rstest/pull/290 (using canary version @rspack/core of this PR).

Overall changes:

- Set Rstest adapted template for `ImportDependency` to 
   - provide support for `importActual`.
   - allow mock unresolved module, throw error on runtime instead of compiling, which makes mock un-exist module possible. 

## Checklist


crates/rspack_plugin_javascript/src/dependency/commonjs/common_js_require_dependency.rs
```diff
@@ -9,10 +9,10 @@ use rspack_core::{
 #[cacheable]
 #[derive(Debug, Clone)]
 pub struct CommonJsRequireDependency {
-  id: DependencyId,
-  request: String,
+  pub id: DependencyId,
```
LingyuCoder: use methods to get the properties, not recommend to set them to `pub`

fi3ework: Reverted `pub`, it was intended to use by set custom commonjs template dependency, which is not used at the end.

LingyuCoder (approved):

Issue #10607, opened 2025-06-08
[Bug]: Generated wrong esm asset


chenjiahan:
### Details

See https://github.com/bytedance/UI-TARS-desktop/issues/672 for details

JSerFeng:
I think it uses the old version of @rslib/core, I upgraded it and it works well

![Image](https://github.com/user-attachments/assets/46d8c67b-d529-4e09-a284-b8fbe4f72558)

![Image](https://github.com/user-attachments/assets/6491d0e0-3a46-4a03-958f-dcdb657fe8df)

JSerFeng:
This project uses `puppeteer-core` which imports `yargs`

![Image](https://github.com/user-attachments/assets/7c18b240-1977-42f2-b309-d38ac2b75158)

And `yargs` contains code uses `require.cache`, which makes your entry cannot be inlined at the top,

https://github.com/yargs/yargs/blob/main/lib/utils/which-module.ts#L3C3-L10C4

The normal case in the output is

```
// your module
const foo = 42

// export 
export { foo }
```

But now your module cannot be scope hoisted because the use of `require.cache`, rslib assume the entry is scope hoisted and render the exports at the top level, the output is wrong

```
const __webpack_modules__ = {
"foo"() { const foo = 42 }
}

export {foo}
```

JSerFeng:
1. Rslib cannot assume the entry will be inlined at the top, as it can be broken in certain cases like above
2. The auto shims of rslib seems have failed, if changing `require` to `createRequire` in the yargs output code, the output is correct

fi3ework:
Conclusion: we'll add a built-in Rslib plugin in Rspack, to bail out the parsing for `require.cache` and leave it as-is.

PR #10735, opened 2025-06-19
feat: initialize Rslib built-in plugin

fi3ework:
## Summary

Initialize Rslib built-in plugin, to fix https://github.com/web-infra-dev/rspack/issues/10607 as of now.

## Checklist


JSerFeng:
Should we implement this based on ModernModuleLibraryPlugin instead 🤔

fi3ework:
> Should we implement this based on ModernModuleLibraryPlugin instead 🤔

But modern-module will only works with ESM format, `require.cache` is a CJS method, IMO, it will mostly used in CJS format.

fi3ework:
Moved Rslib plugin to `experiments` field to prevent from used by others.

PR #10739, opened 2025-06-19
perf(zod/v4): disable Zod JIT

colinaaa:
## Summary

Zod V4 uses `new Function` to compile `z.object()` schema by default to make next parse faster.

https://github.com/colinhacks/zod/blob/d5e23683cab94dfd89be8a57193e33187a09ee2d/packages/zod/src/v4/core/schemas.ts#L1794

But since Rspack would only parse most of the schema only a single time, it would have better performance to use jitless mode.

For `examples/react`, this would reduce the time of `schema.safeParse` from 3.5 ms to 0.9 ms on my laptop.

## Related links

issue: #10467 

## Checklist


chenjiahan (approved):
👍

PR #10757, opened 2025-06-23
chore(wasm): enable wasm test ci

CPunisher:
## Checklist


.github/workflows/reusable-build-test.yml
```diff
@@ -68,7 +68,12 @@ jobs:
     strategy:
       fail-fast: true # for better utilize ci runners
       matrix:
-        node: ${{ fromJSON(contains(inputs.target, 'linux') && github.ref_name == 'main' && '[16, 18, 20]' || '[18]' )}}
+        node: ${{ fromJSON(
```
stormslowly, line 71: it's getting complicated now. 
Maybe, it can be refactoried by  include an exclude in matrix option.

stormslowly: `include` or `exclude` makes it hard to read.

.github/workflows/reusable-build-test.yml
```diff
@@ -128,6 +133,15 @@ jobs:
         if: ${{ inputs.target == 'x86_64-pc-windows-msvc' }}
         run: pnpm run test:ci
 
+      ### WASM
+      - name: Test WASM
+        timeout-minutes: 15 # Tests should finish within 15 mins, please fix your tests instead of changing this to a higher timeout.
+        if: ${{ inputs.target == 'wasm32-wasip1-threads' }}
+        run: |
+          export NODE_NO_WARNINGS=1
```
stormslowly: putting these two env variable to `env` fields is more actions  idiomatic.

PR #10758, opened 2025-06-23
fix(rstest): put Rstest plugin in `experiments` field

fi3ework:
## Summary

Place the Rstest plugin in the `experiments` field, similar to how Rsdoctor does it, to prevent misuse by users.

## Checklist


chenjiahan (approved):
👍

PR #10685, opened 2025-06-16
fix(binding): remove Node 10 compatibility code

chenjiahan:
## Summary

Rspack requires Node.js >= 16 and we can remove Node 10 compatibility code.

The same as https://github.com/web-infra-dev/rspack/pull/10464

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #10770, opened 2025-06-24
refactor: split `rspack_node` into separate `rspack_binding_api`

h-a-n-a:
## Summary

This PR moved C ABI to a separate Rust crate `rspack_binding_api`. Export the APIs from `rspack_node` which previously is the place defined these APIs. It enables us to create `rspack_binding_builder` that re-exports every symbols from `rspack_binding_api`.

## Checklist


jerrykingxyz (approved):

hardfist:
it would be better to do a canary release testing first for moving node_binding, in case it may break some release workflow

crates/rspack_binding_builder_testing/binding.js
```diff
@@ -0,0 +1,241 @@
...
+const { join } = require('path')
+
+const { platform, arch } = process
+
+let nativeBinding = null
+let localFileExisted = false
+let loadError = null
+
+function isMusl() {
+  // For Node 10
```
chenjiahan: Is this a mistake? See https://github.com/web-infra-dev/rspack/pull/10685/files

crates/rspack_binding_builder_testing/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@rspack/binding-builder-testing",
```
chenjiahan: What's the difference between this new private package and `@rspack/binding`

crates/rspack_binding_builder_testing/binding.js
h-a-n-a: Removed as this is just for initialization

crates/rspack_binding_builder_testing/package.json
h-a-n-a: Removed as this is just for initialization

h-a-n-a:
> it would be better to do a canary release testing first for moving node_binding, in case it may break some release workflow

https://github.com/web-infra-dev/rspack/actions/runs/15846389329

jerrykingxyz (approved):

pnpm-lock.yaml
```diff
@@ -129,6 +129,24 @@ importers:
         specifier: workspace:*
         version: link:../../npm/win32-x64-msvc
 
+  crates/rspack_binding_builder_testing:
```
chenjiahan, line 132: This appears to be an unintended lock change, as rspack_binding_builder_testing does not seem to exist anymore.

h-a-n-a: Thanks. Will fix in the next one.

PR #10777, opened 2025-06-24
chore: Use relr for relocs

quininer:
## Summary

This argument make linker to use relr instead of rela, which reduces the size of linux product by about 4M.

see https://maskray.me/blog/2021-10-31-relative-relocations-and-relr

## Checklist


quininer:
Since ubuntu 22.04 only ships with glibc 2.35 and not 2.36 (from debian 12 onwards), this means we won't be able to turn it on until 2027.

:/

PR #10778, opened 2025-06-24
perf(zod): prefer to use `partial` to make all properties optional

chenjiahan:
## Summary

Prefer to use `partial` to make all properties optional, this makes the code more readable and can reduce bundle size.

> https://zod.dev/api?id=partial

## Bundle size

- before:

```txt
File (cjs0)     Size       
dist/index.js   850.5 kB
```

- after:

```txt
File (cjs0)     Size       
dist/index.js   843.9 kB
```

## Runtime performance

`optional()` and `partial()` have similar performance.

```bash
hyperfine --warmup 5 --runs 20  'node ./optional.js' 'node ./partial.js'
Benchmark 1: node ./optional.js
  Time (mean ± σ):     201.0 ms ±  14.5 ms    [User: 332.4 ms, System: 25.3 ms]
  Range (min … max):   190.1 ms … 242.9 ms    20 runs
 
Benchmark 2: node ./partial.js
  Time (mean ± σ):     198.0 ms ±   4.1 ms    [User: 332.0 ms, System: 25.0 ms]
  Range (min … max):   192.7 ms … 211.5 ms    20 runs
 
Summary
  node ./partial.js ran
    1.02 ± 0.08 times faster than node ./optional.js
```

## Checklist


packages/rspack/src/builtin-loader/swc/collectTypeScriptInfo.ts
```diff
@@ -1,9 +1,3 @@
-import * as z from "zod/v4";
-
 export type CollectTypeScriptInfoOptions = {
 	typeExports?: boolean;
 };
-
-export const ZodSwcCollectTypeScriptInfo = z.strictObject({
```
chenjiahan, line 7: moved to `swc/types.ts`

packages/rspack/src/config/zod.ts
```diff
@@ -1176,94 +1208,98 @@ export const getRspackOptionsSchema = memoize(() => {
...
-				force: z
-					.boolean()
-					.or(z.array(z.enum(["version", "uniqueId"])))
-					.optional()
+				version: z.string(),
+				bundler: z.string(),
+				force: z.boolean().or(z.array(z.enum(["version", "uniqueId"])))
 			})
+			.partial()
 			.optional()
```
colinaaa: nitpick: a missing `.optional()` :)

Although changing to `.partial()` does not reduce the calls to create schema, it would be great to keep the same style and future properties adding to `rspackFuture` could avoid the `.optional()`.

colinaaa: ditto for:

- `assetParserDataUrlOptions`
- `assetParserOptions`
- `cssParserOptions`
- `cssModuleParserOptions`

chenjiahan: Nice catch 👍 I've changed the related objects to use `partial()` (although it may only has one property currently)

PR #10782, opened 2025-06-24
chore(ci): including codspeed perfermance comparison into required checks

stormslowly:
Codspeed performance comparison is a neutral *Check*. And it's written by codspeed's GitHub App.
So we can't get the comparison result in workflow directly.

Here we use a workaround to get the comparison result by taking advance of Codspeed's pull request comment.

After comparison, Codspeed will write/update [comment](https://codspeed.io/docs/integrations/ci/buildkite#3-check-the-results) to the pull request about the performance status.

So we can poll pull request's codspeed bot's comment, check if it contains the commit id of current CI targeted. If so we get the latest comparison, then we do a simple grep to see if the pull request degrade the performance or not, then fail or pass the CI checking.

1. The polling is written in the `check-comment.sh`
2. The polling action should be put in `CI.yaml` other than `reusable-build-bench.yml`; it seems that codspeed will write/update comment after job finishes, so if put polling in `reusable-build-bench.yml`, it will trigger a dead-lock

Here is a failure log: https://github.com/web-infra-dev/rspack/actions/runs/15863460473/job/44726161423?pr=10782#step:3:9
Here is a success log: https://github.com/web-infra-dev/rspack/actions/runs/15865447682/job/44731781135#step:3:9

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,6 +114,9 @@ jobs:
         run: bash .github/actions/codspeed/check-comment.sh
 
   test_required_check:
+    # this job will be used for GitHub actions to determine required job success or not;
+    # When code changed, it will check if any of the test jobs failed.
+    # When *only* doc changed, it will run as success directly
     name: Test Required Check
     needs: [check-codspeed, test-linux, test-windows, test-mac, check-changed]
     if: ${{ always() && !cancelled() }}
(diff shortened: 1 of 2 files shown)


jerrykingxyz (approved):

chenjiahan (changes requested):
> The type `feat` MUST be used when a commit adds a new feature to your application or library.

https://www.conventionalcommits.org/en/v1.0.0/#specification

It would be more accurate to change `feat(ci)` to `chore(ci)`.

PR #10813, opened 2025-06-26
chore: add volta's node and pnpm configuration

RJiazhen:
## Summary

Add Volta’s Node and pnpm version configuration to package.json to enable automatic switching of Node and pnpm versions during development.

Also, add the word 'volta' to the cSpell plugin configuration to prevent the “unknown word” warning.

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -41,6 +41,7 @@
     "threejs",
     "tokio",
     "ukey",
-    "Ukey"
+    "Ukey",
+    "volta"
   ]
 }


chenjiahan (commented):
Hi, I was just wondering if Volta can read the .node-version file and the package.json#packageManager field.

RJiazhen:
> Hi, I was just wondering if Volta can read the .node-version file and the package.json#packageManager field.

@chenjiahan I don’t think Volta recognizes these configurations—at least that’s what I found in my testing.

That’s also why the [TypeScript’s package.json](https://github.com/microsoft/TypeScript/blob/02672d281c26e561708127da1d8d1a6cae45fee2/package.json#L114) includes a separate volta field, even though it already has a packageManager field.

chenjiahan:
Sorry, I prefer not to merge this PR because:

- Volta does not recognize these standard fields or files, and we do not want to maintain the same information in different places.
- Our team does not use Volta, so we cannot guarantee that the volta configuration can be updated in time.

RJiazhen:
Okay, I understand your decision. Thanks for the explanation.

PR #10823, opened 2025-06-27
docs: introduce @swc/react-compiler to make react-compiler fast

GiveMe-A-Name:
## Summary

We are introducing a recommendation to use `@swc/react-compiler` for pre-compilation to reduce the number of files processed by `babel-loader`, which is known to add significant compilation time.

## Checklist

### DIFF
--- a/website/docs/en/guide/tech/react.mdx
+++ b/website/docs/en/guide/tech/react.mdx
@@ -207,8 +207,10 @@ export default {
       {
         // Must be run before 'builtin:swc-loader' to ensure that the file is processed by the React Compiler
         // So we add it to the beginning of the rule
-        test: (resource) => /\.jsx$/.test(resource) && isReactCompilerRequiredSync(fs.readFileSync(resource)),
-        loader: 'babel-loader'
+        test: resource =>
+          /\.jsx$/.test(resource) &&
+          isReactCompilerRequiredSync(fs.readFileSync(resource)),
+        loader: 'babel-loader',
       },
     ],
   },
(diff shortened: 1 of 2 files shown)


website/docs/en/guide/tech/react.mdx
```diff
@@ -171,6 +171,50 @@ export default {
...
+            options: {
+              // SWC options for JSX
+            },
+          },
+        ],
+      },
+      {
+        // Must be run before 'builtin:swc-loader' to ensure that the file is processed by the React Compiler
+        // So we add it to the beginning of the rule
+        test: (resouce) => /\.jsx$/.test(resouce) && isReactCompilerRequiredSync(fs.readFileSync(resouce)),
```
chenjiahan: I'm curious about the performance impact of readFileSync , can we do a performance comparison test to verify the performance difference between `@swc/react-compiler` and the previous babel-loader?

GiveMe-A-Name: Got it, i will check it later

suguanYang:
Just out of curiosity, are we already use @swc/react-compiler to use the react-compiler compling on jsx files? or only use the `isReactCompilerRequired` to skip some files as we still use babel to ReactCompiler?  

I was looking for use react-compiler with swc, but the links leads me here. 😀

chenjiahan:
Currently @swc/react-compiler only provides the `isReactCompilerRequired` method to skip some files, and React Compiler still depends on running Babel ([ref](https://github.com/facebook/react/pull/32219)).

Unfortunately, in our early tests, `isReactCompilerRequired` did not bring significant performance improvements.

PR #10839, opened 2025-06-30
docs: remove the guideline of tailwindcss v4 in postcss-loader

CPunisher:
## Summary

Tailwindcss v4 still replies on postcss so we need to remove the wrong guideline.

## Checklist


chenjiahan:
Let's wait for https://github.com/rspack-contrib/rsbuild-plugin-tailwindcss to provide better support for Tailwind CSS v4.

PR #10841, opened 2025-06-30
feat(rspack): add create fn exporter

ityuany:
## Summary

I need to customize a Rust-based plugin and would prefer to directly use built-in functions when bridging at the JS layer.

## Checklist


ityuany:
> This is an internal API and not ready for public use yet. @hardfist cc

Are you considering the explicit release of `unStableDefineRspackPlugin`?

hardfist:
@ityuany this plugin register api is too internal to become a public api(even for experimental usage), we have designed some official way to register builtin plugin for rust api integration, @h-a-n-a will share it

ityuany:
> @ityuany this plugin register api is too internal to become a public api(even for experimental usage), we have designed some official way to register builtin plugin for rust api integration, @h-a-n-a will share it

Okay, then I will first implement it in a more wizard-like way, looking forward to @h-a-n-a's sharing.

h-a-n-a:
@ityuany Moved to https://github.com/web-infra-dev/rspack/pull/10857

PR #10855, opened 2025-07-01
refactor: use length as id instead of module id for lazy compilation

### LINKED ISSUE #9829: [Bug]: lazy compilation url request too long
### System Info

no need

### Details

Lazy compilation use EventSource to make connection with server, which passes data(triggered module identifier) through url, but if there are a large amount of modules that are triggered, the url can be too long and throw error by browser

### Reproduce link

_No response_

### Reproduce Steps

Clone rspack repo

```bash
pnpm i
cd website
npm run dev
```

Then open the page and navigate very fast, check the console, you can see a error complains about 'url too long'

JSerFeng:
## Summary

fix #9829

Use a map to record a mapping between module id and number, when passing what modules should be compiled, use the number id to avoid long url

## Checklist


chenjiahan:
Lazy compilation URL after this PR:

PR #10857, opened 2025-07-01
feat(binding-builder): support using custom binding

h-a-n-a:
## Summary

This PR supports using custom binding with environment variable `RSPACK_BINDING`.
_Path-to-binding_ is a path to the root of the binding package.

```
RSPACK_BINDING=path-to-binding
```
This is achieved by our already used `ast-grep` package to scan and replace the original content of module `@rspack/binding` and `@rspack/binding/package.json`.

In order to either use or test the new binding package (which contains support for custom plugin), this PR also adds support for exposing `experiments.createNativePlugin` which is an alias to the original plugin `create` and would throw error if there's any naming collision with the internal builtin plugins. `createNativePlugin` is used to create wrappers around native plugins:

```js
const MyPlugin = createNativePlugin("MyPlugin", (options) => {
 return options;
});

new MyPlugin({
 foo: "bar"
}).apply(compiler);
```

Tests are now containing two parts, the original binding and the builder testing binding. The latter part is to test if `rspck-binding-builder` successfully extended all the exposed symbols and functionalities provided by `crates/node_binding`. In order to cover that part, tests are written in a mixed style with the original rspack tests. With `test.filter.js`, these tests are only enabled when  running `test:base-builder` script in `rspack-test-tools`. I talked to @stormslowly about these changes to the workflows. Due to the complexity, I would separate the workflow change to a separate PR.

## Checklist

- [ ] Tests updated (or not required). Required, but to heavy to implement in the same PR.


crates/rspack_binding_builder_testing/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@rspack/binding-testing",
```
hardfist, line 2: I'm not sure whether we could just maintain one binding crate, another binding crate may cause extra maintaince burden, especially it contains lots of duplicated code with node_binding

h-a-n-a: Unfortunately, no, we need to create a new binding crate to make sure `rspack_binding_builder` exposed the right symbols. This testing crate will later be running the same rspack testing to make sure that point. I was thinking about making it as simple as possible, but running the e2e test seems to be the absolute correct way to achieve that.

hardfist: can we maintain this using features in node_binding crate?
```
napi build --features custom-binding-test 
```

chenjiahan: Could you measure the increase in CI time?

h-a-n-a: > can we maintain this using features in node_binding crate?
> 
> ```
> napi build --features custom-binding-test 
> ```

`node_binding` does not utilize `rspack_binding_builder`, so we might not be able to do that.

h-a-n-a: > Could you measure the increase in CI time?

I will give it a try in the next PR. Additionally, we don't need to run on every platform just to make sure it handles linking correctly. Only a single linux and maybe wasm in the future looks good to me.

hardfist: > node_binding does not utilize rspack_binding_builder, so we might not be able to do that.

why it can't, it seems it can be controlled by features and make rspack_binding_builder optional dependency
TBH I'm not sure which way is better, but I think we should reduce unnecessary workflow as much as possible, which is already super complex and hard to maintain now.

h-a-n-a: Well, technically, it can. Just want to be honest with you, I prefer not adding another layer of complexity here in the codebase as well. It requires more build flags to control the build script. However, adding a new workflow script does the job right on and will not introduce that much hassle?
One thing i would like to mention is the current publishing CI workflow supports both testing and releasing. If we were to enable testing in releasing, then the binding with that testing feature enabled will eventually be published as a part of it, causing a size bloat.

hardfist: > However, adding a new workflow script does the job right on and will not introduce that much hassle?

That's something I'm not sure but we can give it a try, if it doesn't complicate the workflow and cause much CI burden I'm ok with it.

There's also another alternative way to do this, maintain the custom_binding crate in a separate repo like `custom_rspack` and integrate using git url or published crate

> If we were to enable testing in releasing, then the binding with that testing feature enabled will eventually be published as a part of it, causing a size bloat.

I don't mean include testing code in release code, I mean separate build script for testing like
```sh
"build:dev": "napi build",
"build:dev:custom": "napi build --features custom_binding"
```
and only run build:dev:cutom in testing environment, if so, it seems it will not bloat the release code size?

hardfist: > This testing crate will later be running the same rspack testing to make sure that point.

I also have question for this, do we need run all tests for custom_binding? for me custom_binding more like another js api or options, maybe we should treat it as a test cases?

hardfist:
@ityuany can you help explain your current workflow of integration and see whether this solution meets your needs

crates/rspack_binding_builder_testing/package.json
h-a-n-a: > I also have question for this, do we need run all tests for custom_binding?

That's what I was also thinking about. The difference between a normal feature change and this custom binding thing is the latter one is technically an extended part of the current features, so it might be worthy to test them all instead of tesing just a small portion of the custom plugin?

I will think twice about the proposal you made earlier. But thanks for the recommendations.

h-a-n-a: I would suggest we merging this first and deal with these questions in a separate PR since this does not block anything from merging this one. cc @chenjiahan @hardfist

hardfist (approved):

ityuany:
> @ityuany can you help explain your current workflow of integration and see whether this solution meets your needs

https://github.com/sheinsight/spack

We are currently forking rspack to attempt customized feature development, and then releasing our own binding. In our internal framework, we forward requests to @rspack/binding to @shined/spack-binding by hijacking require.resolve.

nanianlisao:
https://github.com/web-infra-dev/rspack/blob/330f80db58217d32bc6c22136d556ee415368428/packages/rspack/src/util/bindingVersionCheck.ts#L43

Did you miss it here @h-a-n-a

PR #10870, opened 2025-07-02
feat(wasm): support `@rspack/browser`

CPunisher:
## Summary

In Rspack 1.4, we successfully compiled @rspack/binding to wasm32-wasip1-threads and enabled support for WebContainers.

The goal of this PR is to remove the reliance on WebContainers by providing equivalent polyfills in a standalone package, allowing users to use the Rspack JS API directly in the browser.

Todo list:
- [x] Send pr to support injecting `Buffer` in wasi template
- [x] ESM Loader/CJS loader

There are some known issues/features which will be supported in the future: 
1. `Atomics.wait cannot be called in this context` when calling some rust api from js
2. `import` and `require` in loader code
3. Resolving node modules from unpkg
4. Rspack builtin plugins

I will submit the scripts and ci for releasing in another pr

## Checklist


crates/rspack_browser/src/oneshot.rs
```diff
@@ -0,0 +1,82 @@
...
+}
+
+unsafe impl<T: Send> Send for Inner<T> {}
+unsafe impl<T: Send> Sync for Inner<T> {}
+
+pub struct Sender<T> {
+  inner: Arc<Inner<T>>,
+}
+
+impl<T> Sender<T> {
```
quininer, line 46: `Drop` should be implement for `Sender`, otherwise the `Receiver` will never end if `Sender` is released.

crates/rspack_napi/src/threadsafe_function.rs
```diff
@@ -100,7 +103,12 @@ impl<T: 'static + JsValuesTupleIntoVec, R> ThreadsafeFunction<T, R> {
 
   async fn call_async<D: 'static + FromNapiValue>(&self, value: T) -> Result<D> {
     let rx = self.call_with_return(value);
-    rx.await.expect("failed to receive tsfn value")
+    #[cfg(feature = "browser")]
+    let ret =
+      tokio::task::unconstrained(async { rx.await.expect("failed to receive tsfn value") }).await;
```
quininer:
```suggestion
      tokio::task::unconstrained(rx).await.expect("failed to receive tsfn value");
```

crates/rspack_browser/src/panic.rs
```diff
@@ -0,0 +1,19 @@
+//! See [install_panic_handler]
+
+/// Install panic handler for browser environment
+///
+/// Rust builtin panic handler or other community implementation always lock backtrace and stderr before printing error messages.
+/// However, requesting locks finally calls `atomics.wait`, and it's forbidden to do it in the main thread of browser.
+/// So this function provides a panic handler which directly writes error messages to stderr without locks.
+pub fn install_panic_handler() {
```
hardfist, line 8: what's the behavior if we don't install panic_handler, not sure whether panic_handler is necessary

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
+{
```
hardfist, line 1: maybe we should add some description of @rspack/browser and claims it doesn't follow semver now

crates/rspack_browser/src/panic.rs
CPunisher: 1. The default panic handler in rust std requests locks for backtrace and stderr, which is forbidden in the main thread of the browser.  https://doc.rust-lang.org/src/std/panicking.rs.html#268
2. If the main thread get panic when calling tsfn, the panic handler is executed in the main thread.

Not sure if this is the best panic handler in browser wasm, but at least it reduces the confusion of the error message. 
`wasm-bindgen` gets the error backtrace by calling `Error::new()` in the js side and logs it to `console.err`. But it's not easy to casually inject the js api(e.g. `Error`) to rust with napi-rs

packages/rspack-browser/package.json
chenjiahan: We should also add a MIT license file to this folder

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "name": "@rspack/browser",
+  "version": "1.4.1",
+  "webpackVersion": "5.75.0",
+  "license": "MIT",
```
chenjiahan, line 5: We should also add a MIT license file to this folder

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@rspack/browser",
+  "version": "1.4.1",
+  "webpackVersion": "5.75.0",
```
chenjiahan, line 4: Do we need to declare `webpackVersion` for this package?

CPunisher: This is used in something related to stats like https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/MultiStats.ts#L92

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "license": "MIT",
+  "description": "The fast Rust-based web bundler with webpack-compatible API for running in the browser. This is still in early stage and may not follow the semver.",
```
chenjiahan:
```suggestion
  "description": "Rspack for running in the browser. This is still in early stage and may not follow the semver.",
```

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "devDependencies": {
+    "@rspack/binding": "workspace:*",
+    "memfs": "4.20.0"
+  },
+  "homepage": "https://rspack.rs",
+  "bugs": "https://github.com/web-infra-dev/rspack/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/web-infra-dev/rspack",
+    "directory": "packages/rspack"
```
chenjiahan:
```suggestion
    "directory": "packages/rspack-browser"
```

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "packageManager": "pnpm@10.10.0"
```
chenjiahan:
```suggestion
```

packages/rspack/src/util/require.ts
```diff
@@ -0,0 +1,73 @@
...
+type RequireFn = (id: string) => Promise<any>;
+
+export function nonWebpackRequire(compiler: Compiler): RequireFn {
+	return async function (id: string) {
+		if (IS_BROWSER) {
+			// Why is IS_BROWSER used here:
+			// Loading modules in @rspack/browser is difference from the @rspack/core.
+			// 1. It resolves the JavaScript in the memfs with Node.js resolution algorithm rather than in the host filesystem.
+			// 2. It customizes how to evaluate CJS/ESM because there's no `require` any more.
+			const resolver = compiler._lastCompilation!.resolverFactory.get("normal");
```
hardfist: this seems unnecessary and kind of weird, inputFilesystem should be used to load user's project other than load rpsack core itself, the rspack_core bundler logic maybe should put into separate build plugin and integrated into rspack's own build logic(rslib.build.js)

packages/rspack-browser/package.json
chenjiahan: get

packages/rspack-browser/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@rspack/browser",
+  "version": "1.4.1",
```
chenjiahan, line 3: Should we update the version to keep consistent with other packages

CPunisher: I think we'd better to do that. Because the *.wasm binding is always consistent with native bindings. They are always compiled from the same rust code.

packages/rspack/src/util/require.ts
CPunisher: We have shimmed `node:fs` to `memfs` so the `const inputFileSystem = compiler.inputFileSystem!` below is really unnecessary. 

Here the main problem is that how we can use the resolver with `node:fs` (will be shimmed by `memfs`) to get the path of an import specifier. Do you think it's appropriate to copy some of the code in https://github.com/web-infra-dev/rspack-resolver/blob/main/napi/src/lib.rs to rspack binding crate?

hardfist: we already have planned to expose resolver to binding and after it we can just use binding resolver api

CPunisher: Done with the help of https://github.com/web-infra-dev/rspack/pull/11211

PR #10876, opened 2025-07-02
feat(binding-builder): add binding-builder cli

h-a-n-a:
## Summary

Introduce a new package `@rspack/binding-builder-cli` to build binding (bin: `rspack-builder`).

The new `@rspack/binding-builder-cli` will be utilized for both `crates/node_binding` and projects using `rspack-binding-template` or similar.

The original build script contains special environment variables and features that are specialized for `crates/node_binding`. With that being said, I kept the original `scripts/build.js` to call `rspack-builder` to achieve the same effect.

### Rationale

This package is designed to provide a command-line interface for building custom bindings for rspack.

Rspack internally performs post-processing modifications on the generated TypeScript definition (.d.ts) files from NAPI-RS bindings. The @rspack/binding-builder-cli synchronizes these modifications, ensuring that the generated .d.ts files are complete and provide access to all type information available in @rspack/binding plus the newly added types in custom binding.

### Debug

To debug `rspack-builder`, set `DEBUG=rspack-builder:*` as a prefix to the command.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4697,7 +4697,6 @@ name = "rspack_node"
 version = "0.2.0"
 dependencies = [
  "napi-derive",
- "parking_lot",
  "rspack_binding_api",
  "rspack_binding_build",
 ]
--- a/crates/node_binding/Cargo.toml
+++ b/crates/node_binding/Cargo.toml
@@ -18,11 +18,9 @@ sftrace-setup = ["rspack_binding_api/sftrace-setup"]
 ignored = [
   # Add napi-derive as a dependency to workaround an issue where `dts` will no longer work without it.
   "napi-derive",
-  "parking_lot",
 ]
 
 [dependencies]
-parking_lot        = { version = "0.12.3", features = ["nightly"] }
 rspack_binding_api = { workspace = true }
 
 napi-derive = { workspace = true, features = ["compat-mode"] }
--- a/packages/rspack-test-tools/package.json
+++ b/packages/rspack-test-tools/package.json
@@ -19,7 +19,7 @@
     "dev": "tsc -b -w",
     "test": "cross-env RUST_BACKTRACE=full pnpm run --stream /^test:.*/",
     "testu": "pnpm run --stream /^test:.*/ -u",
-    "test:base": "cross-env node --no-warnings --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --logHeapUsage --colors --config ./jest.config.js --passWithNoTests",
+    "test:base": "cross-env node --no-warnings --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --logHeapUsage --colors --config ./jest.config.js --passWithNoTests --runInBand",
     "test:hot": "cross-env RSPACK_HOT_TEST=true node --no-warnings --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --logHeapUsage --colors --config ./jest.config.hot.js --passWithNoTests",
     "test:diff": "cross-env RSPACK_DIFF=true node --no-warnings --expose-gc --max-old-space-size=8192 --experimental-vm-modules ../../node_modules/jest-cli/bin/jest --logHeapUsage --colors --config ./jest.config.diff.js --passWithNoTests",
     "api-extractor": "api-extractor run --verbose",


packages/binding-builder-cli/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "version": "1.4.2",
+  "description": "CLI for rspack binding builder",
+  "homepage": "https://rspack.rs",
+  "bugs": "https://github.com/web-infra-dev/rspack/issues",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/web-infra-dev/rspack",
+    "directory": "packages/binding-builder-cli"
+  },
+  "license": "MIT",
```
chenjiahan, line 12: The new package should contain a license file: https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/LICENSE

.github/actions/pnpm/install-binding-dependencies/action.yml
```diff
@@ -31,7 +31,7 @@ runs:
       shell: bash
       run: |
         cd ./crates/node_binding
-        pnpm install --ignore-workspace --no-lockfile
```
h-a-n-a, line 34: The bad part is we have to link `binding-builder-cli` to `crates/node_binding`. This requires the binding to fallback to workspace install instead of plain package install. cc @stormslowly

stormslowly, line 34: remove `--ignore-workspace` will install all workspace depenceies and slow down installation.

pnpm-lock.yaml
```diff
@@ -8957,6 +8967,41 @@ snapshots:
       - supports-color
       - ts-node
 
+  '@jest/core@29.7.0(ts-node@10.9.2(@swc/core@1.12.0(@swc/helpers@0.5.17))(@types/node@22.16.0)(typescript@5.8.3))':
```
chenjiahan, line 8830: maybe we can run `pnpm dedupe` to reduce duplicated dependencies

crates/node_binding/scripts/build.js
```diff
@@ -30,77 +26,46 @@ build().then((value) => {
...
 	return new Promise((resolve, reject) => {
 		let args = [
-			"build",
 			"--platform",
 			"--dts",
 			"binding.d.ts",
 			"--no-js",
-			// "--no-const-enum",
 			"--no-dts-header",
-			"--pipe",
```
stormslowly, line 40: it seems that `--pipe` removed and does not add back.
Maybe the reason why d.ts modified

.github/actions/pnpm/install-binding-dependencies/action.yml
h-a-n-a: We could probably install the binding builder cli as a dependency (not from the current workspace) after the first release, then we can revert the workflow back.

crates/rspack_binding_builder_testing/package.json
```diff
@@ -11,17 +11,17 @@
     "binding.d.ts"
   ],
   "scripts": {
-    "build:dev": "node scripts/build.js",
-    "build:debug": "node scripts/build.js --profile release-debug",
-    "build:ci": "node scripts/build.js --profile ci",
-    "build:profiling": "node scripts/build.js --profile profiling",
+    "build:dev": "rspack-builder",
```
hardfist, line 14: if all the build logic moved to binding-builder-cli why the builder_testing's script/build.js is not moved?

crates/node_binding/scripts/build.js
```diff
@@ -30,77 +26,46 @@ build().then((value) => {
...
 			envs.RUSTFLAGS = "-Zinstrument-xray=always";
 		}
+
 		if (features.length) {
-			args.push("--features " + features.join(","));
+			args.push("--features", features.join(","));
 		}
 
 		if (positionals.length > 0) {
-			// napi need `--` to separate options and positional arguments.
```
stormslowly, line 74: This is used to pass through cargo extra option to npai build.
e.g.: We expect to `pnpm run build:release -- --timings` to enable cargo options.

crates/node_binding/scripts/build.js
h-a-n-a: Not really, d.ts is modified because I didn't add `banner.d.ts` to biome's ignore list. We didn't happen to format it before.

packages/binding-builder-cli/scripts/build.js
```diff
@@ -0,0 +1,414 @@
...
+	strict: true,
+	allowPositionals: true
+});
+
+if (values.help) {
+	showHelp();
+	process.exit(0);
+}
+
+const { NapiCli } = require("@napi-rs/cli");
```
hardfist, line 272: I'm not sure it's a stable public api or just an internal api, I'm afraid whether we'll miss some logic in the napi-cli/bin

crates/node_binding/scripts/build.js
h-a-n-a: `positionals` have already included `--` and `rspack-builder` will parse that part.

packages/binding-builder-cli/scripts/build.js
h-a-n-a: https://github.com/napi-rs/napi-rs/blob/7e34e30b66b2c35de31a3f4386d166c2154e4e8c/cli/package.json#L26 Looks like public APIs to me

packages/binding-builder-cli/scripts/build.js
```diff
@@ -0,0 +1,414 @@
+const path = require("node:path");
+const { readFileSync, writeFileSync } = require("node:fs");
+
+const debug = require("debug")("rspack-builder:build");
+
+const OPTIONS_CONFIG = {
```
hardfist, line 6: is this duplicate of https://github.com/napi-rs/napi-rs/blob/main/cli/src/def/build.ts#L5 or we add custom logic here?

stormslowly:
If `rspack-builder` is mainly for rspack template user,  in my opinion, put how we use `napi` in `package.json`'s script is better.

1. easy for napi veteran to tweak
2. less abstraction; In my opinion, the template repo is more of a rust napi project other than rspack plugin project.

packages/binding-builder-cli/scripts/build.js
h-a-n-a: Yea, it's a bunch of code supposed to achieve the same functionalities with napi-rs cli. I changed it to programmatic API to make our CLI output nice and clean.

packages/binding-builder-cli/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@rspack/binding-builder-cli",
+  "version": "1.4.2",
```
hardfist, line 3: I don't fully understand why we need make our own cli instead of using napi-cli? can you help explain in README.md so outside contributors can understand it

h-a-n-a:
> If `rspack-builder` is mainly for rspack template user, in my opinion, put how we use `napi` in `package.json`'s script is better.
> 
> 1. easy for napi veteran to tweak
> 2. less abstraction; In my opinion, the template repo is more of a rust napi project other than rspack plugin project.

I had thought about this in the first place. But rspack's binding heavily requires some tweaking to the `binding.d.ts` and we need to sync these changes to the users and this is the most painful part.

packages/binding-builder-cli/package.json
h-a-n-a: Being explained in https://github.com/web-infra-dev/rspack/pull/10876#issuecomment-3027471751

h-a-n-a: +README.md https://github.com/web-infra-dev/rspack/blob/55995fcc35e6518fca990d507cb11dfbed9d8aa6/packages/binding-builder-cli/README.md

crates/node_binding/scripts/build.js
h-a-n-a: FYI: 
image

pnpm-lock.yaml
h-a-n-a: Done

packages/binding-builder-cli/package.json
h-a-n-a: Done

crates/rspack_binding_builder_testing/package.json
h-a-n-a: Done

Issue #7608, opened 2024-08-19
[Feature]: builtin-lightningcss-loader 如何 支持 visitor 选项


ckken:
### What problem does this feature solve?

https://rspack.dev/zh/guide/features/builtin-lightningcss-loader#options
https://lightningcss.dev/transforms.html

### What does the proposed API of configuration look like?

LightningcssLoaderOptions 增加 visitor

JSerFeng:
It's cool but not that much high priority for us at present, and if using JavaScript `visitor`, there will be interop between Rust and JavaScript, so the performance gain of builtin-loader is gone, in this case you can try calling lightningcss-node binding yourself, or you can try [this loader](https://github.com/fz6m/lightningcss-loader) from community

We will support visitors in the future, but still, not high priority for us at present, of course we are happy to help if you are willing to contribute for this feature

ckken:
Without visitor, from a business perspective, built-in functions are useless, such as px to vw, etc.

ckken:
https://empjs.dev/plugin/tool/plugin-lightningcss.html

JSerFeng:
bump

ckken:
请问 这个有进度吗

GiveMe-A-Name:
No progress temporarily, we will do this in two week recently.

sowtame:
> No progress temporarily, we will do this in two week recently.

Really looking forward

sowtame:
guys, if there's any news on this feature, we have 80 percent of the build taken up by postcss plugins on our project, we want to switch to lightningcss.
Sincerely, a dedicated rspack user

sowtame:
bump

hardfist:
consider visitor will cause lots of rust <-> js commutation which will make builtin-lightningcss-loader slow, we suggest use the js version of lightningcss-loader instead if you want to use lightningcss with visitor

ckken:
与其放个半成品进去，真心建议把内置的删了！

PR #10884, opened 2025-07-03
perf: remove visitor feature of lightningcss

hardfist:
## Summary

it seems visitor feature is used for js visitor api which we don't use 
## Checklist


ckken:
https://github.com/web-infra-dev/rspack/issues/7608 感觉看到进度又破灭了～

PR #10885, opened 2025-07-03
perf: override wasmer-wasix to remove all unused wasi extension feature

hardfist:
## Summary
override wasmer-wasix to remove all unused wasi extension to reduce binary size
## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,7 +285,7 @@ dependencies = [
  "bitflags 2.9.1",
  "cexpr",
  "clang-sys",
- "itertools 0.13.0",
+ "itertools 0.10.5",
  "log",
  "prettyplease",
  "proc-macro2",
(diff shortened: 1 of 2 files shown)


syrusakbary:
Hey! If you open a PR to Wasmer upstream, we would be happy to have it there, so you don't need to fork.
Note: we would need this to be feature gated.

Thanks!

hardfist:
> Hey! If you open a PR to Wasmer upstream, we would be happy to have it there, so you don't need to fork. Note: we would need this to be feature gated.
> 
> Thanks!

@syrusakbary  Thanks, this pr is actually used to verify how much binary reduction we can get if remove unused wasix feature, no intention to bifurcate, I will make a PR soon.

hardfist:
@syrusakbary I'm not sure whether I should make sys feature empty or add another feature like `wasi-common` which contains only minimal wasi feature

PR #10888, opened 2025-07-03
perf: improve parallelism of real content hash plugin

LingyuCoder:
## Summary

Improve parallelism when calculating the real content hashes.

By recording the processed hashes and the hashes on which the current asset depends, if all have been processed, it is added to the current processing batch. Then process the batch together to prevent the number of tasks processed in parallel at one time from being too small, which would make the process approach sequential processing.

Before:
image

After:
image

## Checklist

### DIFF
--- a/packages/rspack-test-tools/tests/hashCases/real-content-hash-consistent/test.filter.js
+++ b/packages/rspack-test-tools/tests/hashCases/real-content-hash-consistent/test.filter.js
@@ -1,0 +1,1 @@
+module.exports = () =>{return !process.env.WASM}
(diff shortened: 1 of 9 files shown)


packages/rspack-test-tools/tests/hashCases/real-content-hash-consistent/test.filter.js
```diff
@@ -0,0 +1 @@
+module.exports = () =>{return !process.env.WASM}
```
LingyuCoder, line 1: Not sure why these cases failed in wasm testing, so just ignore them until fixing by @CPunisher

ahabhgk (approved):

PR #10892, opened 2025-07-03
fix(wasm): use `parking_lot` nightly to avoid panic in tests

h-a-n-a:
## Summary

Turn on `wasm_atomic` in `parking_lot`.
See: https://github.com/Amanieu/parking_lot/blob/master/core/src/thread_parker/wasm_atomic.rs

This fixes `Parking not supported on this platform` issue in wasm.

This also makes it possible to start multiple tokio workers.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4167,6 +4167,7 @@ dependencies = [
  "napi",
  "napi-derive",
  "once_cell",
+ "parking_lot",
  "ropey",
  "rspack_allocator",
  "rspack_browserslist",
--- a/crates/rspack_binding_api/Cargo.toml
+++ b/crates/rspack_binding_api/Cargo.toml
@@ -110,7 +110,7 @@ rustc-hash                             = { workspace = true }
 serde                                  = { workspace = true }
 serde_json                             = { workspace = true }
 swc_core                               = { workspace = true, default-features = false, features = ["ecma_transforms_react"] }
-tokio                                  = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "test-util", "tracing"] }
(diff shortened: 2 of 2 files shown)


CPunisher:
We can also remove `tokio          = { workspace = true, features = ["parking_lot"] }`  under `[target.'cfg(not(target_family = "wasm"))'.dependencies]`

h-a-n-a:
> We can also remove `tokio = { workspace = true, features = ["parking_lot"] }` under `[target.'cfg(not(target_family = "wasm"))'.dependencies]`

Okay, but isn't this supposed to make tokio running faster?

CPunisher:
> > We can also remove `tokio = { workspace = true, features = ["parking_lot"] }` under `[target.'cfg(not(target_family = "wasm"))'.dependencies]`
> 
> Okay, but isn't this supposed to make tokio running faster?

Not sure but from the `wasm_atomics.rs` in the `parking_lot` I guess the implementation is as same as it in `std::sync`.

h-a-n-a:
> > > We can also remove `tokio = { workspace = true, features = ["parking_lot"] }` under `[target.'cfg(not(target_family = "wasm"))'.dependencies]`
> > 
> > 
> > Okay, but isn't this supposed to make tokio running faster?
> 
> Not sure but from the `wasm_atomics.rs` in the `parking_lot` I guess the implementation is as same as it in `std::sync`.

`tokio` is under `not(target_family = "wasm")`, so I assume It's not supposed to interfere with `wasm`-ish targets, right?

CPunisher:
> `tokio` is under `not(target_family = "wasm")`, so I assume It's not supposed to interfere with `wasm`-ish targets, right?

Mostly yes. Just pay attention not to introduce unsupported features like `fs`. See https://docs.rs/tokio/latest/tokio/#wasm-support. The compiler also helps us guarantee.

PR #10907, opened 2025-07-04
fix: use AhoCorasick for real content hash

### LINKED ISSUE #10798: [Bug]: `rspack_plugin_real_content_hash` panics with `CompiledTooBig`
### System Info

  System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 101.30 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.11.0 - ~/.nvm/versions/node/v22.11.0/bin/node
    npm: 10.9.0 - ~/.nvm/versions/node/v22.11.0/bin/npm
    pnpm: 10.6.3 - ~/.nvm/versions/node/v18.18.0/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.120
    Edge: 137.0.3296.93
    Safari: 18.5

### Details

I'm in the process of migrating some large bundles to Rspack, and when bundling some of them I end up getting this error:

```
Message:  Invalid regex: CompiledTooBig(10485760)
Location: crates/rspack_plugin_real_content_hash/src/lib.rs:117
```

That points to this line, which looks like it's making a very large regex using every content hash for every asset: https://github.com/web-infra-dev/rspack/blob/6362a790f5487466eccf723c1dec9fc5a81d629c/crates/rspack_plugin_real_content_hash/src/lib.rs#L117

I haven't looked too deeply into the code here, but is it possible to do this without creating this very large regex?

### Reproduce link

_No response_

### Reproduce Steps

I don't have a good minimal repro, but I'm getting this issue when assembling large bundles with content hashed names

ahabhgk:
## Summary

fix #10798

~~The regex input source can be trust since those are content hash of the asset, which are generated by ourself, since this is only a check so should be fine to "skip" the check~~

AhoCorasick is perfect for the case

Can't found a good way to setup a test

## Checklist

### DIFF
--- a/crates/rspack_plugin_real_content_hash/src/lib.rs
+++ b/crates/rspack_plugin_real_content_hash/src/lib.rs
@@ -114,7 +114,7 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
   // use LeftmostLongest here:
   // e.g. 4afc|4afcbe match xxx.4afcbe-4afc.js -> xxx.[4afc]be-[4afc].js
   //      4afcbe|4afc match xxx.4afcbe-4afc.js -> xxx.[4afcbe]-[4afc].js
-  let hash_regexp = AhoCorasick::builder()
+  let hash_ac = AhoCorasick::builder()
     .match_kind(MatchKind::LeftmostLongest)
     .build(hash_list.iter().map(|s| s.as_bytes()))
     .expect("Invalid patterns");
@@ -128,7 +128,7 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
       asset.get_source().map(|source| {
         (
           name.as_str(),
-          AssetData::new(source.clone(), asset.get_info(), &hash_regexp),
+          AssetData::new(source.clone(), asset.get_info(), &hash_ac),
         )
       })
     })
@@ -187,11 +187,8 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
     let batch_sources = batch_source_tasks
       .into_par_iter()
       .map(|(hash, name, data)| {
-        let new_source = data.compute_new_source(
-          data.own_hashes.contains(hash),
-          &hash_to_new_hash,
-          &hash_regexp,
-        );
+        let new_source =
+          data.compute_new_source(data.own_hashes.contains(hash), &hash_to_new_hash, &hash_ac);
         ((hash, name), new_source)
       })
       .collect::<HashMap<_, _>>();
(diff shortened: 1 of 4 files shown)


crates/rspack_plugin_real_content_hash/src/lib.rs
```diff
@@ -114,7 +114,12 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
   // e.g. 4afc|4afcbe match xxx.4afcbe-4afc.js -> xxx.[4afc]be-[4afc].js
   //      4afcbe|4afc match xxx.4afcbe-4afc.js -> xxx.[4afcbe]-[4afc].js
   hash_list.par_sort_by(|a, b| b.len().cmp(&a.len()));
-  let hash_regexp = Regex::new(&hash_list.join("|")).expect("Invalid regex");
+  let hash_regexp = {
+    RegexBuilder::new(&hash_list.join("|"))
```
quininer: It seems unnecessary to use regex for a rule that only `or`. can we use [AhoCorasick](https://docs.rs/aho-corasick/1.1.3/aho_corasick/struct.AhoCorasick.html) directly?

ahabhgk: Yes we can, this is perfect for the case!

crates/rspack_plugin_real_content_hash/src/lib.rs
```diff
@@ -255,17 +258,14 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
...
-          let hash = c
-            .get(0)
-            .expect("RealContentHashPlugin: should have match")
-            .as_str();
-          hash_to_new_hash
-            .get(hash)
-            .expect("RealContentHashPlugin: should have new hash")
-        })
-        .into_owned();
+      let mut new_name = String::new();
```
quininer: Using `with_capacity` will avoid grow allocations ?

crates/rspack_plugin_real_content_hash/src/lib.rs
```diff
@@ -105,16 +106,18 @@ async fn inner_impl(compilation: &mut Compilation) -> Result<()> {
...
     // xx\xx{xx?xx.xx -> xx\\xx\{xx\?xx\.xx escape for Regex::new
     .map(|hash| QUOTE_META.replace_all(hash, "\\$0"))
     .collect::<Vec<Cow<str>>>();
-  // long hash should sort before short hash to make sure match long hash first in hash_regexp matching
+  // use LeftmostLongest here:
   // e.g. 4afc|4afcbe match xxx.4afcbe-4afc.js -> xxx.[4afc]be-[4afc].js
   //      4afcbe|4afc match xxx.4afcbe-4afc.js -> xxx.[4afcbe]-[4afc].js
-  hash_list.par_sort_by(|a, b| b.len().cmp(&a.len()));
-  let hash_regexp = Regex::new(&hash_list.join("|")).expect("Invalid regex");
+  let hash_regexp = AhoCorasick::builder()
```
quininer: It's no longer regex, maybe it could be called `hash_ac` :D

quininer (approved):
LGTM

PR #10914, opened 2025-07-04
fix: clear compiled modules from active modules

JSerFeng:
## Summary

When modules are compiled, they should be removed from the compiled map, otherwise it will be passed to server in next compiling, causing the url get longer by time

## Checklist

### DIFF
--- a/crates/rspack_plugin_lazy_compilation/src/module.rs
+++ b/crates/rspack_plugin_lazy_compilation/src/module.rs
@@ -209,7 +209,7 @@ impl Module for LazyCompilationProxyModule {
     );
 
     let keep_active = format!(
-      "var dispose = client.keepAlive({{ data: data, active: {}, module: module, onError: onError }})",
+      "var dispose = client.activate({{ data: data, active: {}, module: module, onError: onError }})",
       block.is_some()
     );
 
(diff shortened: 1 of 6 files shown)


chenjiahan:
E2E failed ⬆️

tests/e2e/cases/lazy-compilation/persistent-cache/index.test.ts
```diff
@@ -5,15 +5,8 @@ test("should load success", async ({ page, rspack }) => {
 	let component_count = await page.getByText("Component").count();
 	expect(component_count).toBe(1);
 
-	const responsePromise = page.waitForResponse(
-		response =>
-			response.url().includes("lazy-compilation-using") &&
```
chenjiahan, line 10: Why do we remove these lines

JSerFeng: It's not related to this test, this test contains no dynamic import at all, once page is loaded, no lazy-compilation request received, before this test passed because if module changed, the module will re-send a unnecessary request to server unexpectedly

chenjiahan: get 👌

chenjiahan (approved):

PR #10931, opened 2025-07-07
perf: prepare module outgoings for code splitting

LingyuCoder:
## Summary

Prepare module outgoings for code splitting while generating chunk graph.

Before: 2.8s
image

After: 1.7s
image
image

## Checklist


LingyuCoder:
find a better way to improve performance, so this pr will be closed

PR #10942, opened 2025-07-08
refactor: remove the encapsulation for `rspack_fs::Error` 

CPunisher:
## Summary

`rspack_fs::Error` contains only `std::io::Error` for two years. So here considering remove the encapsulation.

There could be some thing wrong with ci linux tools, which produce ~1MB larger artifact than the that in my pc with cross-compilation.

## Checklist

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -67,4 +67,6 @@ jobs:
           size-threshold: 51200
 
       - name: Debug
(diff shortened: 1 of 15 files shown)


CPunisher:
There could be some thing wrong with ci linux tools, which produce ~1MB larger artifact than the that in my pc with cross-compilation. But I have no time to investigate currently.

UPDATE: it seems the ci use GLIBC with newer version

PR #10947, opened 2025-07-08
refactor: enfored shared dependency

hardfist:
## Summary
enforce all workspace use workspace dependency

## Checklist


.vscode/settings.json
```diff
@@ -22,10 +22,7 @@
     "Cargo.toml"
   ],
   "rust-analyzer.cargo.targetDir": true, // Don't let RA blocking cargo build
-  "rust-analyzer.cargo.features": [
```
hardfist, line 25: found RA broken when develop deny-ext feature caused by clippy may pass non-supported features to crate

PR #10948, opened 2025-07-08
fix(binding-builder): pre-generate workspace info

h-a-n-a:
## Summary

This PR introduced a new `rspack_workspace` crate to support pre-generating workspace information:

- rspack_pkg_version
- rspack_swc_core_version

Also removed build scripts using `CARGO_WORKSPACE_DIR`. This is bound to the current workspace not for published crates.

`tasks/codegen` is used to do the real codegen job.

## Checklist

### DIFF
--- a/.github/workflows/size-limit.yml
+++ b/.github/workflows/size-limit.yml
@@ -35,6 +35,9 @@ jobs:
       - name: Setup Rust Target
         run: rustup target add x86_64-unknown-linux-gnu
 
+      - name: Run Cargo codegen
+        run: cargo codegen
+
       - name: Trim paths
         run: |
           echo $'\n' >> .cargo/config.toml


.github/workflows/ci-rust.yaml
```diff
@@ -37,6 +37,9 @@ jobs:
           save-if: ${{ github.ref_name == 'main' }}
           key: check
 
+      - name: Run Cargo codegen
+        run: cargo codegen
```
h-a-n-a, line 41: We can split this into an action if there're multiple steps to setup Rust crates. But currently there's only one. So I decided to explicitly use a simple `cargo codegen` here to avoid all these hassle.

tasks/codegen/src/main.rs
```diff
@@ -0,0 +1,77 @@
...
+  let package_json: serde_json::Value = serde_json::from_str(&package_json_content)?;
+  let rspack_version = package_json["version"]
+    .as_str()
+    .expect("version field in package.json is not a string");
+
+  // Generate combined content
+  let content = format!(
+    r#"//! This is a generated file. Don't modify it by hand! Run 'cargo codegen' to re-generate the file.
+/// The version of the `swc_core` package used in the current workspace.
+pub fn rspack_swc_core_version() -> &'static str {{
```
quininer: should it be const fn or const &'static str ?

h-a-n-a: Sure it can be! Thanks

quininer (approved):

quininer (approved):

PR #10950, opened 2025-07-08
fix: circular reference

SoonIter:
## Summary

fix: circular reference

## Checklist


SoonIter:
The js code in rspack core has a circular reference, if someone is willing to fix it, they can submit a pull request.

PR #10953, opened 2025-07-08
refactor: move chunk_artifacts occasion to artifact

hardfist:
## Summary
move old_cache parts to artifact to integrate with persistent cache in the future

## Checklist


crates/rspack_core/src/cache/persistent/storage/memory_gc.rs
```diff
@@ -0,0 +1,65 @@
+use std::sync::atomic::{AtomicU32, Ordering};
```
jerrykingxyz: It is not appropriate to place memory_gc in the persistent cache, storage in persistent cache is used to save buffer data to database/file. memory_gc is more like a util, maybe it would be better to move it to `src/utils`.

crates/rspack_core/src/artifacts/mod.rs
```diff
@@ -14,6 +15,7 @@ mod side_effects_do_optimize_artifact;
 pub use cgm_hash_artifact::*;
 pub use cgm_runtime_requirement_artifact::*;
 pub use chunk_hashes_artifact::*;
+pub use chunk_render_cache_artifact::*;
```
jerrykingxyz: do not export all mods, export what you use only.

hardfist: yes!

jerrykingxyz (approved):

PR #10960, opened 2025-07-09
feat: replace `std::fs` and `tokio::fs` with `inputFileSystem` and `outputFileSystem`

CPunisher:
## Summary

There are still some code with `std::fs` or `tokio::fs`, where it's hard to access the abstract fs on `Compilation`:
1. rspack::tracing
2. rspack_javascript_compiler::compiler::transform::load_swcrc

## Checklist

### DIFF
--- a/crates/node_binding/binding.d.ts
+++ b/crates/node_binding/binding.d.ts
@@ -2758,7 +2758,6 @@ export interface ThreadsafeNodeFS {
   readFile: (name: string) => Promise<Buffer | string | void>
   stat: (name: string) => Promise<NodeFsStats | void>
   lstat: (name: string) => Promise<NodeFsStats | void>
-  chmod: (name: string, mode: number) => Promise<void>
   realpath: (name: string) => Promise<string | void>
   open: (name: string, flags: string) => Promise<number | void>
   rename: (from: string, to: string) => Promise<void>
@@ -2768,6 +2767,7 @@ export interface ThreadsafeNodeFS {
   read: (fd: number, length: number, position: number) => Promise<Buffer | void>
   readUntil: (fd: number, code: number, position: number) => Promise<Buffer | void>
   readToEnd: (fd: number, position: number) => Promise<Buffer | void>
+  chmod?: (name: string, mode: number) => Promise<void>
 }
 
 export declare function transform(source: string, options: string): Promise<TransformOutput>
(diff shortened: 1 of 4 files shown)


quininer (dismissed):
LGTM

crates/rspack_binding_api/src/fs_node/node.rs
```diff
@@ -33,6 +33,8 @@ pub struct ThreadsafeNodeFS {
   pub stat: ThreadsafeFunction<String, Promise<Either<NodeFsStats, ()>>>,
   #[napi(ts_type = "(name: string) => Promise<NodeFsStats | void>")]
   pub lstat: ThreadsafeFunction<String, Promise<Either<NodeFsStats, ()>>>,
+  #[napi(ts_type = "(name: string, mode: number) => Promise<NodeFsStats | void>")]
```
stormslowly: is the type annotation right ?
```suggestion
  #[napi(ts_type = "(name: string, mode: number) => Promise<void>")]
```

crates/node_binding/binding.d.ts
```diff
@@ -2755,6 +2756,7 @@ export interface ThreadsafeNodeFS {
   readFile: (name: string) => Promise<Buffer | string | void>
   stat: (name: string) => Promise<NodeFsStats | void>
   lstat: (name: string) => Promise<NodeFsStats | void>
+  chmod: (name: string, mode: number) => Promise<NodeFsStats | void>
```
stormslowly:
```suggestion
  chmod: (name: string, mode: number) => Promise<void>
```

packages/rspack/etc/core.api.md
```diff
@@ -5309,6 +5309,8 @@ export type Output = {
 
 // @public (undocumented)
 export interface OutputFileSystem {
+    // (undocumented)
+    chmod: (arg0: string, arg1: number, arg2: (arg0?: NodeJS.ErrnoException | null) => void) => void;
```
hardfist, line 5313: is this compatible with webpack's OutputFileSystem type?

CPunisher: > is this compatible with webpack's OutputFileSystem type?

Actually not. There's no `chmod` in webpack's OutputFileSystem

hardfist: maybe we should make this optional to be compatible with webpack's type, I thinks it's not a required feature and we can fallback to noop for fs which not supporting permission

PR #10967, opened 2025-07-09
fix: resolveData.*_dependencies should work when factorize success

jerrykingxyz:
## Summary

Rspack will lose dependency factorize info include file_dependencies, missing_dependencies when it factorize success,
this behaivor will make the file_dependencies added by user at `normalModuleFactory.hooks.resolve` useless.

This PR will make dependency rebuilds behave more like modules, and the dependency will refactorize if modifed files in its factorize_info.file_dependencies.

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/make/mod.rs
+++ b/crates/rspack_core/src/compiler/make/mod.rs
@@ -84,24 +84,24 @@ impl MakeArtifact {
       .remove_batch_file(&build_info.build_dependencies);
     self.make_failed_module.remove(module_identifier);
 
-    // clean incoming & all_dependencies(outgoing) factorize info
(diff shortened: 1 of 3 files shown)


ahabhgk:
I'm ok with this if the `10000_production-mode_persistent-hot + exec` benchmark is expected

jerrykingxyz:
This PR will result in
1. During the make phase, all dependencies will be looped and checked if they depend on modified files, causing a lot of performance regressions.
2. All dependencies will save factorization info, increasing memory usage.
The probability that a dependency that has been factorized successfully needs to be rebuilt is low, so I will temporarily close this PR until there is a related request or bug.

PR #10974, opened 2025-07-10
refactor: move existing tasks to cargo-xtask

hardfist:
## Summary
move existing tasks to xtask which provides good summary of all existing task, benchmark task is special which is used for benchmark, so keep as it is now

image

## Checklist


xtask/src/release_check/mod.rs
```diff
@@ -43,32 +50,27 @@ fn main() -> Result<()> {
...
     Ok(())
   } else {
-    let error = ReleaseCheckError {
-      others: diagnostics,
-    }
-    .into();
-    Err(error)
+    let combined = diagnostics
+      .into_iter()
+      .map(|e| format!("{e:?}")) // 使用 `Display` 也可以
```
CPunisher:
```suggestion
      .map(|e| format!("{e:?}"))
```

CPunisher (approved):

PR #10982, opened 2025-07-10
refactor: avoid exposing `Ref` and `RefMut` of `DashMap`

CPunisher:
## Summary

Currently some functions  return `Ref` or `RefMut` of `DashMap`, which holds the inner sync rwlocks. When the `Ref`s are used across `.await`, deadlocks may happen.

So this pr wrapper all values of dashmap with `Arc<T>` or `Arc<tokio::sync::RwLock<T>>` depending on their usages. 

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -386,8 +386,10 @@ opt-level = "s"
 # This is for CI build based on release but with faster build time
 [profile.ci]
 codegen-units = 256
+debug         = "full"
 inherits      = "release"
 lto           = false
+strip         = false
 ## reduce little optimization to reduce build time
 opt-level = 2
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -386,10 +386,8 @@ opt-level = "s"
 # This is for CI build based on release but with faster build time
 [profile.ci]
 codegen-units = 256
-debug         = "full"
 inherits      = "release"
 lto           = false
-strip         = false
 ## reduce little optimization to reduce build time
 opt-level = 2
 
--- a/crates/rspack_plugin_html/src/plugin.rs
+++ b/crates/rspack_plugin_html/src/plugin.rs
@@ -1,7 +1,7 @@
 use std::{
   borrow::Cow,
   path::{Path, PathBuf},
-  sync::LazyLock,
+  sync::{Arc, LazyLock},
 };
 
 use cow_utils::CowUtils;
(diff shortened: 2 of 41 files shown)


quininer (approved):

Cargo.toml
```diff
@@ -386,8 +386,10 @@ opt-level = "s"
 # This is for CI build based on release but with faster build time
 [profile.ci]
 codegen-units = 256
+debug         = "full"
```
hardfist: why change this?

CPunisher: Sorry, I use this to debug

crates/rspack_plugin_html/src/plugin.rs
```diff
@@ -22,7 +23,7 @@ use crate::{
   BeforeAssetTagGenerationData, BeforeEmitData, HtmlPluginHooks,
 };
 
-static COMPILATION_HOOKS_MAP: LazyLock<FxDashMap<CompilationId, Box<HtmlPluginHooks>>> =
+static COMPILATION_HOOKS_MAP: LazyLock<FxDashMap<CompilationId, Arc<RwLock<HtmlPluginHooks>>>> =
```
hardfist, line 26: why use async lock other than sync lock here?

CPunisher: Sync locks could be better. I will replace the rwlocks that don't hold across `.await` to the sync ones.

crates/rspack_plugin_javascript/src/plugin/mod.rs
```diff
@@ -352,7 +354,12 @@ impl JsPlugin {
             allow_inline_startup = false;
           }
           let hooks = JsPlugin::get_compilation_hooks(compilation.id());
-          let bailout = hooks.inline_in_runtime_bailout.call(compilation).await?;
+          let bailout = hooks
+            .read()
+            .await
+            .inline_in_runtime_bailout
+            .call(compilation)
+            .await?;
```
hardfist, line 362: ok this relys on async lock

hardfist (approved):

crates/rspack_plugin_javascript/src/plugin/mod.rs
CPunisher: Yes, I also find it.

PR #10991, opened 2025-07-11
fix: do not test version in canary builds

h-a-n-a:
## Summary

In canary version, version bump is done after binding is built.
And to export `EXPECTED_RSPACK_CORE_VERSION` in binding, it relys on the bumped version of @rspack/core.
So we can't check the version of @rspack/core and @rspack/binding in canary version.
Here we ignore version check for canary version.

## Checklist

### DIFF
--- a/packages/rspack/src/util/bindingVersionCheck.ts
+++ b/packages/rspack/src/util/bindingVersionCheck.ts
@@ -12,7 +12,7 @@ export const checkVersion = () => {
 	}
 
 	// In canary version, version bump is done after binding is built.
-	// And to export `EXPECTED_RSPACK_CORE_VERSION` in binding, it relys on the bumped version of @rspack/core.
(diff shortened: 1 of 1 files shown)


packages/rspack/src/util/bindingVersionCheck.ts
```diff
@@ -11,6 +11,14 @@ export const checkVersion = () => {
 		return null;
 	}
 
+	// In canary version, version bump is done after binding is built.
+	// And to export `EXPECTED_RSPACK_CORE_VERSION` in binding, it relies on the bumped version of @rspack/core.
+	// So we can't check the version of @rspack/core and @rspack/binding in canary version.
+	// Here we ignore version check for canary version.
+	if (CORE_VERSION.includes("canary")) {
```
h-a-n-a, line 18: Looks like `CORE_VERSION` is bumped before building canary. Let me double test.

h-a-n-a: It works

PR #10994, opened 2025-07-11
chore: include crate version bump in `x version` command

h-a-n-a:
## Checklist

### DIFF
--- a/crates/rspack_macros/Cargo.toml
+++ b/crates/rspack_macros/Cargo.toml
@@ -11,5 +11,13 @@ proc-macro = true
 
 [dependencies]
 proc-macro2 = { workspace = true }
-quote       = { workspace = true }
-syn         = { workspace = true, features = ["full", "visit-mut"] }
+quote = { workspace = true }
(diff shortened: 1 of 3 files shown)


crates/rspack_macros/Cargo.toml
```diff
@@ -11,5 +11,13 @@ proc-macro = true
 
 [dependencies]
 proc-macro2 = { workspace = true }
-quote       = { workspace = true }
-syn         = { workspace = true, features = ["full", "visit-mut"] }
+quote = { workspace = true }
+syn = { workspace = true, features = [
+  "full",
```
stormslowly, line 16: ？？？ why appears in this script PR ?

h-a-n-a: Tried to run a publish test with this PR, but failed due to lacking features

quininer: This may be because we configured `default-features = false` for all dependencies.

stormslowly (approved):

PR #10999, opened 2025-07-11
fix: factorize hooks avoid update dependency request

jerrykingxyz:
## Summary

The `normalModuleFactory.hooks.resolve` will update dependency based on the request set by user, and it will make rebuild use rewrited request directly. This behavior is inconsistent with webpack.

## Checklist


crates/rspack_binding_api/src/plugins/interceptor.rs
```diff
@@ -1456,7 +1453,7 @@ impl NormalModuleFactoryBeforeResolve for NormalModuleFactoryBeforeResolveTap {
       .await
     {
       Ok((ret, resolve_data)) => {
-        dependency.set_request(resolve_data.request);
```
ahabhgk, line 1459: Seems this is the only usage about set_request, can we remove set_request on Dependency trait now?

ahabhgk (approved):

PR #11011, opened 2025-07-14
refactor: remove `oneshot`

CPunisher:
## Summary

We have refactored tsfn related functions to async ones, so here we can use `tokio::sync::oneshot` to replace `oneshot`. 

## Checklist


quininer (approved):
Good, AFAIK tokio has the only lock-free oneshot implement

PR #11020, opened 2025-07-14
perf: avoid duplicate calculations in executor task

### LINKED ISSUE #10987: [Bug]: `css-loader` causes build to hang (and eventually crash) with `modules: true` option set
### System Info

```
npx envinfo --system --npmPackages '@rspack/*' --binaries --browsers
sh: envinfo: command not found
```

```
pnpm list -r @rspack/cli @rspack/core @rspack/dev-server css-loader

devDependencies:
css-loader 7.1.2
@rspack/cli 1.4.6
@rspack/core 1.4.6
@rspack/dev-server 1.1.3
```

### Details

`rspack` pegs one of my machines CPU cores at ~100%, uses large amount of RSS memory (>20 GiB) and hangs for a very long time before eventually crashing with a `137` OOM error. By adding a thin wrapper around `css-loader` that logs the `.css` files it processes i see that it processes `139` files before hanging

```
Exit status 137
```

### Reproduce link

_No response_

### Reproduce Steps

I'll try to provide a more concrete repro later but for now here's our config snippet for the `.css` rule:

```javascript
        {
          test: /\.css$/i,
          use: [
            CssExtractRspackPlugin.loader,
            {
              loader: 'css-loader',
              ident: 'css',
              options: {
                modules: true,
                sourceMap: true,
              },
            },
          ],
          type: 'javascript/auto',
          exclude: [/node_modules/],
        },
```

hardfist:
## Summary
close #10987
The build hang is caused by redundant computation in ExecutorTask, resulting in approximately 90⁴ (i.e., 90×90×90×90) calculations, as seen in [this file](https://github.com/mjames-c/rspack-bug-10987/blob/main/src/dep-1.css)

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/module_executor/execute.rs
+++ b/crates/rspack_core/src/compiler/module_executor/execute.rs
@@ -130,11 +130,10 @@ impl Task<ExecutorTaskContext> for ExecuteTask {
     let mut has_error = false;
 
     while let Some(m) = queue.pop_front() {
-      // to avoid duplicate calculate in https://github.com/web-infra-dev/rspack/issues/10987
-      if modules.contains(&m) {
+      // to avoid duplicate calculations in https://github.com/web-infra-dev/rspack/issues/10987
+      if !modules.insert(m) {
         continue;
       }
-      modules.insert(m);
       let module = mg.module_by_identifier(&m).expect("should have module");
       let build_info = module.build_info();
       execute_result
--- a/crates/rspack_core/src/compiler/module_executor/execute.rs
+++ b/crates/rspack_core/src/compiler/module_executor/execute.rs
@@ -130,6 +130,10 @@ impl Task<ExecutorTaskContext> for ExecuteTask {
     let mut has_error = false;
 
     while let Some(m) = queue.pop_front() {
+      // to avoid duplicate calculate in https://github.com/web-infra-dev/rspack/issues/10987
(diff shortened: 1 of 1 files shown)


crates/rspack_core/src/compiler/module_executor/execute.rs
```diff
@@ -130,6 +130,10 @@ impl Task<ExecutorTaskContext> for ExecuteTask {
     let mut has_error = false;
 
     while let Some(m) = queue.pop_front() {
+      // to avoid duplicate calculate in https://github.com/web-infra-dev/rspack/issues/10987
+      if modules.contains(&m) {
```
ahabhgk: if !modules.insert(&m) {

crates/rspack_core/src/compiler/module_executor/execute.rs
```diff
@@ -130,6 +130,10 @@ impl Task<ExecutorTaskContext> for ExecuteTask {
     let mut has_error = false;
 
     while let Some(m) = queue.pop_front() {
+      // to avoid duplicate calculate in https://github.com/web-infra-dev/rspack/issues/10987
```
mjames-c:
```suggestion
      // to avoid duplicate calculations in https://github.com/web-infra-dev/rspack/issues/10987
```

jerrykingxyz (approved):

PR #11023, opened 2025-07-14
refactor: remove `redirect_to` in exports info

LingyuCoder:
## Summary

Remove `exports_info_data.redirect_to` which has made our architecture extremely complex. It is only used for handling json tree shaking.

In fact, since there are `exports_info` and `exports_info_owned` on `export_info`, when `exports_info_owned` is `false`, it means that `export_info` points to the `exports_info` of other modules, and when it is `true`, it represents a nested `exports_info`.

Therefore, we can handle JSON by directly bridging `export_info` to `nested_exports_info`, which is also consistent with the way of handling javascript.

Since the previous [PR](https://github.com/web-infra-dev/rspack/pull/10849) merged the `export_info_map` and `exports_info_map`, after removing redirect_to, it will greatly avoid recursive operations based on the module graph. Therefore, the performance compatible code previously used during recursive processing can be removed, which will be handled in the next PR.

## Checklist

### DIFF
--- a/crates/rspack_core/src/exports/export_info_setter.rs
+++ b/crates/rspack_core/src/exports/export_info_setter.rs
@@ -3,7 +3,7 @@ use std::collections::hash_map::Entry;
 use rspack_util::atom::Atom;
 
 use super::{ExportInfoData, ExportInfoTargetValue, Inlinable, UsageFilterFnTy, UsageState};
-use crate::{DependencyId, ExportsInfo, ExportsInfoData, ModuleGraph, Nullable, RuntimeSpec};
+use crate::{DependencyId, ExportsInfo, Nullable, RuntimeSpec};
 
 impl ExportInfoData {
   pub fn reset_provide_info(&mut self) {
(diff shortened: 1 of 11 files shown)


LingyuCoder:
* **#11051** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/11051?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#11023** <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/11023?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/web-infra-dev/rspack/11023?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

ahabhgk:
AFAIK this is also useful for esm-cjs interop + reexport, you can try this case:

```js
// index.js
import * as reexport from "./reexport.js"
console.log(reexport.lib.default)
// reexport.js
export * as lib from "./lib.js"
// lib.js
exports.a = 1
exports.b = 2
```

The correct log should be `{ a: 1, b: 2 }`, but after remove the `rediect_to` the log is `{}`

LingyuCoder:
seems it is not easy to resolve cjs interop issue, this PR needs more work so just convert it to draft

Issue #10997, opened 2025-07-11
[Bug]: Lazy compilation causes "should be added before using" panic


chenjiahan:
### System Info

- Rspack 1.4.6

### Details

Lazy compilation causes the "should be added before using" panic when running Rsbuild's E2E cases:

<img width="2422" height="1196" alt="Image" src="https://github.com/user-attachments/assets/4ae484e3-aeeb-4c3b-9ea2-a94aa7b8d114" />

This case will visit the `/rsbuild-dev-server` path of Rsbuild dev server (which is a served static page).

### Reproduce link

https://github.com/web-infra-dev/rsbuild/pull/5582

### Reproduce Steps

1. git clone git@github.com:web-infra-dev/rsbuild.git
2. git checkout enable_lazy_compilation_0711
3. pnpm i
4. pnpm e2e:rspack  viewing-served-files

GiveMe-A-Name:
After some debugging, we found that the panic occurs because we retrieve compilation.stats immediately after the first page load.

With lazy compilation enabled, the actual compilation process takes place only after the first page visit. This leads to a scenario where we are in the middle of building the moduleGraph and then the ChunkGraph. Meanwhile, when we access compilation.stats, it retrieves modules from the moduleGraph and then tries to fetch module information from the ChunkGraph. However, the ChunkGraph may not have finished building at this point, resulting in a panic.

Maybe we update stats generate logic is more suitable. I will fix it in the future.

PR #11030, opened 2025-07-15
fix: chunk_graph_module maybe undefined while compiling

GiveMe-A-Name:
## Summary

User maybe get stats while rspack is compiling.
When moduleGraph is built, but chunkGraph is not built finished, the module relative chunk_graph_module maybe undefined.
We should not panic in this case.

## Related links
fix: #10997

## Checklist


ahabhgk:
Need a test case to avoid regression

GiveMe-A-Name:
> Need a test case to avoid regression

Updated

PR #11068, opened 2025-07-16
ci: support publishing crates in release-otp workflow

h-a-n-a:
## Summary

Support publishing crates in release-otp workflow

## Checklist


.github/workflows/release-crates.yml
```diff
@@ -13,6 +13,22 @@ on:
...
+    inputs:
+      dry_run:
+        type: boolean
+        description: "DryRun release"
+        required: false
+        default: false
+      push_tags:
+        type: boolean
+        description: "Push tags to repository"
+        required: false
```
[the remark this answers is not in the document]
h-a-n-a: We don't require users passing options with `workflow_call`

.github/workflows/release-otp.yml
```diff
@@ -77,6 +77,17 @@ jobs:
       test: ${{ inputs.test }}
       profile: "release"
 
+  release_crates:
```
stormslowly, line 88: Release crate job has no needs configed, but i think it should be config as `needs : [build]`

In a case like this
1. User trigger the publishing workflow with testing enabled, 
2. A bug found in the testing. 

So the npm publishing will fail, but the crate will release successfully.

After the bug fixed, if user trigger the workflow again, crate will not be release with new code.

stormslowly: If config `needs: build` directly, the release workflow will be longger.
So, maybe we can split a the pre checking job run parrallly with build.
And release crate depends on both build and prechck job.

PR #11071, opened 2025-07-16
perf(wasm): set lto=fat and cgu=1

CPunisher:
## Summary

I'm not sure why this is not set before. At that time I just copy the config from rolldown, but they have updated the config so we can sync.

Wasm size: 36.75MB -> 28.88MB

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -402,12 +402,10 @@ split-debuginfo = "off"
 strip           = false
 
 [profile.release-wasi]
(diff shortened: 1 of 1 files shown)


hardfist:
why such big difference

CPunisher:
> why such big difference

@quininer says cgu and lto are just so powerful.

quininer (approved):

Issue #11123, opened 2025-07-22
[Bug]: FS Error with Yarn PnP and dynamic require


elbywan:
### System Info

  System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 4.99 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.15.0 - ~/.volta/tools/image/node/22.15.0/bin/node
    Yarn: 4.9.1-git.20250411.hash-1908ee79f - ~/.volta/tools/image/yarn/4.9.1/bin/yarn
    npm: 10.9.2 - ~/.volta/tools/image/node/22.15.0/bin/npm
    pnpm: 10.11.0 - ~/.volta/bin/pnpm
    Watchman: 2025.04.28.00 - /opt/homebrew/bin/watchman
  Browsers:
    Brave Browser: 137.1.79.126
    Chrome: 138.0.7204.158
    Safari: 18.5

### Details

With Yarn PnP, when a third party module uses some kind a dynamic require (!= static analysis) then rspack will crash with the following error:

```sh
[rspack-cli] [Error:   × Rspack FS Error: Not a directory (os error 20)
  ╰─▶ Not a directory (os error 20)
] {
  code: 'GenericFailure'
}
```

This is surely caused by these factors:

- The native filesystem does not support [read_dir](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_fs/src/native_fs.rs#L232) for pnp and uses a plain `fs::read_dir` call
- Th context module factory [calls read_dir](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/context_module_factory.rs#L413) to crawl folders:
```rust
if !fs
  .metadata_sync(dir) // <-- metadata_sync supports pnp and will return true for nested .zip paths
  .map(|m| m.is_directory)
  .unwrap_or(false)
{
  return Ok(());
}
let include = &options.context_options.include;
let exclude = &options.context_options.exclude;
for filename in fs.read_dir_sync(dir)? { // <-- read_dir_sync does not and will crash
```

### Reproduce Steps

With Yarn Berry installed…

```bash
yarn --version
4.9.1-git.20250411.hash-1908ee79f
```

…run these steps to create a small repro folder…

```bash
mkdir rspack-pnp-dynamic-require-repro
cd rspack-pnp-dynamic-require-repro
yarn init
[...]

PR #11124, opened 2025-07-22
fix(pnp): add yarn pnp support to read_dir

elbywan:
## Summary

Makes `read_dir` work when using yarn pnp, preventing a crash when a third party module contains a dynamic require and making it properly the zip crawl directories.

I do not think that the require context ever worked with yarn pnp before anyway, but at least it did not crash prior to https://github.com/web-infra-dev/rspack/pull/10960.

This is not efficient at all because it iterates over the whole index of the zip file every time, the proper fix would be to add a mapping between folders and children directly in the [pnp-rs](https://github.com/yarnpkg/pnp-rs?tab=readme-ov-file#filesystem-utilities) crate (cc @arcanis 👋) but at least it should stop rspack from crashing in the meantime:

```
[rspack-cli] [Error:   × Rspack FS Error: Not a directory (os error 20)
  ╰─▶ Not a directory (os error 20)
] {
  code: 'GenericFailure'
}
```

## Related links

Should solve https://github.com/web-infra-dev/rspack/issues/11123 

## Checklist

### DIFF
--- a/crates/rspack_fs/src/native_fs.rs
+++ b/crates/rspack_fs/src/native_fs.rs
@@ -1,6 +1,7 @@
 use std::{
   fs::{self, File},
   io::{BufRead, BufReader, BufWriter, Read, Write},
+  path::{Path, PathBuf},
 };
 
 use pnp::fs::{FileType, LruZipCache, VPath, VPathInfo, ZipCache};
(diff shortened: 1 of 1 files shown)


stormslowly:
@elbywan in which situation, we need to read dir under the `node_modules`？

PR #11140, opened 2025-07-23
perf: enhance native watcher with customizable ignore patterns

GiveMe-A-Name:
### What This PR Does
This PR significantly enhances Rspack's native watcher by introducing support for both glob pattern-based and function-based ignore patterns, while delivering substantial performance improvements over the previous implementation.
### Performance Optimizations
- Reduced NAPI Overhead: Replaced RegExp objects with string patterns to avoid expensive Thread-safe Function (tsfn) calls across the Rust-JavaScript boundary
- Native Glob Matching: Implemented fast, native glob pattern matching using the fast-glob crate instead of JavaScript regex operations
- Cross-Platform Path Handling: Added automatic path normalization for Windows compatibility (\ → /)

## Checklist

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -1511,7 +1511,7 @@ export interface NativeWatcherOptions {
   pollInterval?: number
   aggregateTimeout?: number
   /** Ignored paths or a function to determine if a path should be ignored. */
-  ignored?: string | string[] | ((path: string) => boolean)
+  ignored?: string | string[] | RegExp | ((path: string) => boolean)
 }
 
 export interface NodeFsStats {
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -1510,8 +1510,11 @@ export interface NativeWatcherOptions {
   followSymlinks?: boolean
   pollInterval?: number
   aggregateTimeout?: number
-  /** Ignored paths or a function to determine if a path should be ignored. */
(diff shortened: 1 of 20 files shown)


crates/node_binding/napi-binding.d.ts
```diff
@@ -1510,8 +1510,8 @@ export interface NativeWatcherOptions {
   followSymlinks?: boolean
   pollInterval?: number
   aggregateTimeout?: number
-  /** A function that will be called with the path of a file or directory that is ignored. */
-  ignored?: (path: string) => boolean
+  /** Ignored paths or a function to determine if a path should be ignored. */
+  ignored?: string | ((path: string) => boolean)
```
stormslowly: i think we should align this config to webpack watch options https://webpack.js.org/configuration/watch/#watchoptionsignored

GiveMe-A-Name:
I do a simple bench case in my local using rspack-ecosystem-benchmark.10000_development case 
```
  System:
    OS: Linux 5.15 veLinux GNU/Linux 2 (lyra) 2 (lyra)
    CPU: (64) x64 Intel(R) Xeon(R) Platinum 8457C
    Memory: 225.56 GB / 251.61 GB
    Container: Yes
    Shell: 5.2.15 - /bin/bash
 ```
- before this perf
image

- after this perf
image

crates/node_binding/napi-binding.d.ts
GiveMe-A-Name: Updated, it's already support receive `string[]`.
But limitation by napi, we can't pass RegExp from js to rust.

stormslowly (approved):

PR #11157, opened 2025-07-24
fix: update response headers in lazy compilation middleware

GiveMe-A-Name:
## Summary

Lazy compilation middleware should have `Access-Control-Allow-Origin`,  `Access-Control-Allow-Methods`, `Access-Control-Allow-Headers` headers like [webpack.lazyCompilationBackend](https://github.com/webpack/webpack/blob/main/lib/hmr/lazyCompilationBackend.js#L78-L83) to avoid tigger CORS. 
## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
+++ b/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
@@ -177,8 +177,12 @@ const lazyCompilationMiddlewareInternal = (
 		const indices = req.url.slice(lazyCompilationPrefix.length).split("@");
 		req.socket.setNoDelay(true);
 
-		res.setHeader("content-type", "text/event-stream");
-		res.writeHead(200);
+		res.writeHead(200, {
+			"content-type": "text/event-stream",
+			"Access-Control-Allow-Origin": "*",
+			"Access-Control-Allow-Methods": "*",
+			"Access-Control-Allow-Headers": "*"
+		});
 		res.write("\n");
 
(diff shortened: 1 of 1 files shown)


packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
```diff
@@ -177,8 +177,12 @@ const lazyCompilationMiddlewareInternal = (
 		const indices = req.url.slice(lazyCompilationPrefix.length).split("@");
 		req.socket.setNoDelay(true);
 
-		res.setHeader("content-type", "text/event-stream");
-		res.writeHead(200);
+		res.writeHead(200, {
+			"content-type": "text/event-stream",
+			"Access-Control-Allow-Origin": "*",
```
chenjiahan: This is unsafe, we should let the dev server to configure CORS headers

chenjiahan: See: https://github.com/web-infra-dev/rsbuild/pull/4399

JSerFeng:
It's not safe to use cors: '*', it's strange that dev server should already set cors headers as expected, is that lazy compile middleware run before that 🤔

packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
GiveMe-A-Name: Updated

JSerFeng (approved):

PR #11171, opened 2025-07-25
refactor: bench external getResolve & remove ResolveClosureContext

SyMind:
## Summary

add bench case for external getResolve in codspeed

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/raw_options/raw_external.rs
+++ b/crates/rspack_binding_api/src/raw_options/raw_external.rs
@@ -103,25 +103,6 @@ pub struct RawExternalItemFnCtxData<'a> {
   pub context_info: &'a ContextInfo,
 }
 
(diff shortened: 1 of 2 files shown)


crates/rspack_binding_api/src/raw_options/raw_external.rs
```diff
@@ -103,25 +103,6 @@ pub struct RawExternalItemFnCtxData<'a> {
   pub context_info: &'a ContextInfo,
 }
 
-#[derive(Debug, Clone)]
-struct ResolveClosureContextInner {
```
h-a-n-a, line 107: Maybe we should consider this PR as a `refactor` or `perf`?

SyMind: done

PR #11173, opened 2025-07-25
fix: should add incoming blocks for async entrypoint

JSerFeng:
## Summary

Should add incoming blocks for async_entrypoints like normal chunks, so that they can be invalidated during incremental rebuild

## Checklist


stormslowly:
Any test cases to make it easy to be caught？

JSerFeng:
> Any test cases to make it easy to be caught？

hard to repro this stable, as it runs rebuild in parallel, but e2e will most likely fail if no fix

PR #11177, opened 2025-07-25
perf: replace rwlock with `atomic_refcell`

CPunisher:
## Summary

We should make sure that there's no read-write and write-write conflicts for each hook instance.

## Checklist


quininer (approved):
This is good, but the usage pattern is uncomfortable. In future, `COMPILATION_HOOKS_MAP` should be replacedto avoid global variable.

PR #11182, opened 2025-07-25
feat: build dependencies recursively collect dependencies

jerrykingxyz:
## Summary

This PR will align build dependency behavior with webpack and make build dependencies recursively collect dependencies.

Here is a simple example that parses the config file and adds the plugin file to the build dependencies:
``` js
import plugin from "./config/plugin"

module.exports = {
  cache: true,
  experiments: {
    cache: {
      type: 'persistent',
      buildDependencies: [path.join(__dirname, "./rspack.config.js")]
    },
  },
}
```

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
+++ b/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
@@ -83,8 +83,8 @@ impl BuildDeps {
         // node package path skip recursive search.
         continue;
       }
-      if let Some(childs) = helper.resolve(current.assert_utf8()).await {
-        queue.extend(childs.iter().map(|item| item.as_path().into()));
+      if let Some(children) = helper.resolve(current.assert_utf8()).await {
+        queue.extend(children.iter().map(|item| item.as_path().into()));
       }
     }
 


crates/rspack_core/src/cache/persistent/build_dependencies/helper/mod.rs
```diff
@@ -0,0 +1,251 @@
...
+/// and continue processing them until all dependency files are calculated.
+pub struct Helper {
+  resolver: Resolver,
+  warnings: Vec<String>,
+}
+
+impl Helper {
+  pub fn new(fs: Arc<dyn ReadableFileSystem>) -> Self {
+    Helper {
+      resolver: Resolver::new(
```
LingyuCoder, line 32: can we use loader resolver?

jerrykingxyz: The build dependency resolver should be isolated from the loader. The webpack implementation is more complex, but this point is the same.

https://github.com/webpack/webpack/blob/main/lib/FileSystemInfo.js#L1544-L1570

Issue #11191, opened 2025-07-28
[Bug]: Missing query in `loaderContext.getResolve()`


colinaaa:
### System Info

  System:
    OS: macOS 26.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 635.97 MB / 96.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Binaries:
    Node: 22.15.1 - ~/.local/state/fnm_multishells/36819_1753434101409/bin/node
    npm: 10.9.2 - ~/.local/state/fnm_multishells/36819_1753434101409/bin/npm
    pnpm: 9.15.2 - ~/.local/state/fnm_multishells/36819_1753434101409/bin/pnpm
    bun: 1.2.8 - ~/.bun/bin/bun
  Browsers:
    Chrome: 138.0.7204.169
    Safari: 26.0
  npmPackages:
    @rspack/cli: 1.4.9 => 1.4.9 
    @rspack/core: 1.4.9 => 1.4.9

### Details

This is a regression from Rspack [v1.4.9](https://github.com/web-infra-dev/rspack/releases/tag/v1.4.9) (it works fine in v1.4.8).

When using `this.getResover()` with query in the request, the resolve result does not contain query.

```js
  const resolver = this.getResolve({})
  const result = await resolver(this.context, './bar.css?common') // Should have `?common` in result
```

In webpack / Rspack 1.4.8, it returns

```
<ROOT_DIR>/src/bar.css?common
```

In Rspack 1.4.9, it returns

```
<ROOT_DIR>/src/bar.css
```

Also see a real-world case in lynx-stack: https://github.com/lynx-family/lynx-stack/pull/1335#discussion_r2224682758

### Reproduce link

https://github.com/colinaaa-reproductions/rspack-repro-resolve-query

### Reproduce Steps

1. `pnpm install`
2. `pnpm run build`

colinaaa:
This is likely to be caused by https://github.com/web-infra-dev/rspack/pull/11102/files#diff-d5d662427a296d44968a916dc6b7abb907b4f29da035d1301d8f09c708555e1fR67 in https://github.com/web-infra-dev/rspack/pull/11102.

PR #11194, opened 2025-07-28
fix: should return `resource` instead of `path` in resolve

colinaaa:
## Summary

Return `resource` instead of `path` in `resolver.resolve`.

## Related links

close: #11191

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/resolver.rs
+++ b/crates/rspack_binding_api/src/resolver.rs
@@ -19,7 +19,6 @@ use crate::{
 #[derive(Debug, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ResolveRequest {
-  pub resource: String,
   pub path: String,
   pub query: String,
   pub fragment: String,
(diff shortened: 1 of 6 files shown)


SyMind:
Actually, webpack does not return the `resource`, so I am not sure if Rspack needs to return this field.

colinaaa:
> Actually, webpack does not return the `resource`, so I am not sure if Rspack needs to return this field.

What do you think of adding a new struct `ResolveRequestWithResource`:

```rust
pub struct ResolveRequestWithResource {
  pub resource: String,
  pub request: ResolveRequest,
}
```

And call `callback(error, result.resource, result.request)` in `Resolver.resolve`.

SyMind:
How about get the resource by concatenating the path, query, and fragment?

colinaaa:
> How about get the resource by concatenating the path, query, and fragment?

This approach could work as well. However, I initially avoided it because I felt it would be redundant, given that Rust already provides a `Resource::full_path` method to achieve the same functionality.

https://github.com/web-infra-dev/rspack/blob/6d120919fe4bb1c73b396021509f9e60d883ceee/crates/rspack_core/src/resolver/mod.rs#L87-L92

BTW, if we are concatenating in JS, should we also insert the zero width space for fragment?

SyMind (approved):

PR #11204, opened 2025-07-28
fix: fix_issuers not clean cycled modules children

jerrykingxyz:
## Summary

When fix_issuers cleans up cycled modules in step4, it will revoke many modules, and these modules may be issuer of other modules. lets look at a simple example for this problem:

image

This PR will remove ModuleC to make the module graph work properly.

## Checklist


ahabhgk (approved):

Narretz:
Thanks for the fix!

PR #11211, opened 2025-07-29
feat: add resolver JavaScript API

CPunisher:
## Summary

Generally copy everything from https://github.com/web-infra-dev/rspack-resolver/blob/main/napi/src/lib.rs except `tracing.rs`. Then do some simple fixes.

## Checklist


crates/node_binding/napi-binding.d.ts
```diff
@@ -1495,6 +1516,178 @@ export declare function minify(source: string, options: string): Promise<Transfo
...
+  /**
+   * Redirect module requests when normal resolving fails.
+   *
+   * Default `[]`
+   */
+  fallback?: Record<string, Array<string | undefined | null>>
+  /**
+   * Request passed to resolve is already fully specified and extensions or main files are not resolved for it (they are still resolved for internal requests).
+   *
+   * See also webpack configuration [resolve.fullySpecified](https://webpack.js.org/configuration/module/#resolvefullyspecified)
```
chenjiahan: Can we link to Rspack doc

website/docs/en/api/javascript-api/resolver.mdx
```diff
@@ -0,0 +1,20 @@
+# Resolver API
+
+Rspack uses the higher-performance [rspack-resolver](https://github.com/web-infra-dev/rspack-resolver) written in Rust to replace Webpack's [enhanced-resolve](https://github.com/webpack/enhanced-resolve), and experimentally re-exports these APIs through `rspack.experiments.resolver`. This allows you to directly call rspack-resolver's functions like `resolver.sync` or `resolver.async`.
```
chenjiahan: Use lower case for webpack

```suggestion
Rspack uses the higher-performance [rspack-resolver](https://github.com/web-infra-dev/rspack-resolver) written in Rust to replace webpack's [enhanced-resolve](https://github.com/webpack/enhanced-resolve), and experimentally re-exports these APIs through `rspack.experiments.resolver`. This allows you to directly call rspack-resolver's functions like `resolver.sync` or `resolver.async`.
```

crates/node_binding/napi-binding.d.ts
CPunisher: ok, do you think that we should also update some other the links to webpack in both rspack and rspack_resolver?

stormslowly (approved):

stormslowly (approved):

PR #11292, opened 2025-08-05
perf: try to create native regex instead of regress regex

LingyuCoder:
## Summary

> This pull request is inspired by [turbopack](https://github.com/vercel/next.js/blob/203adbd5d054609812d1f3666184875dcca13f3a/turbopack/crates/turbo-esregex/src/lib.rs#L71-L94)

- Try to create native regex instead of regress regex which has poor performance
- Remove `algo` and its fast path logic because native regex is fast enough

## Checklist


crates/rspack_regex/src/native.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut applied_flags = String::new();
+    // https://github.com/vercel/next.js/blob/203adbd5d054609812d1f3666184875dcca13f3a/turbopack/crates/turbo-esregex/src/lib.rs#L71-L94
+    for flag in &flags {
+      match flag {
+        // indices for substring matches: not relevant for the regex itself
+        'd' => {}
+        // global: default in rust, ignore
+        'g' => {}
+        // case-insensitive: letters match both upper and lower case
+        'i' => applied_flags.push('i'),
```
quininer, line 29: Should use https://docs.rs/regex/latest/regex/struct.RegexBuilder.html

PR #11300, opened 2025-08-05
feat: expose `experiments.swc` in parallel loader context

jbroma:
## Summary

This PR enhances the loader context in parallel loader by exposing `rspack.experiments.swc` API . This removes a need for importing `rspack` directly in the loader (or loader module) and instead makes it possible to access the `swc` API just like in non parallel loaders

## Related links

n/a

## Checklist

### DIFF
--- a/packages/rspack/src/loader-runner/worker.ts
+++ b/packages/rspack/src/loader-runner/worker.ts
@@ -246,6 +246,12 @@ async function loaderImpl(
 
 	loaderContext._compiler = {
 		...loaderContext._compiler,
+		rspack: {
+			// @ts-expect-error: some properties are missing.
+			experiments: {
+				swc: require("../swc")
+			}
+		},
 		// @ts-expect-error: some properties are missing.
 		webpack: {
 			util: {


packages/rspack/src/loader-runner/worker.ts
```diff
@@ -246,6 +246,12 @@ async function loaderImpl(
 
 	loaderContext._compiler = {
 		...loaderContext._compiler,
+		rspack: {
+			// @ts-expect-error: some properties are missing.
```
GiveMe-A-Name, line 250: Can we import `_compiler.rspack.experiments.swc` in parallel loader context?
Like follow doc example https://rspack.rs/api/javascript-api/swc#examples

h-a-n-a (approved):
Looks good! Thanks

packages/rspack/src/loader-runner/worker.ts
jbroma: yeah, that's exactly what this PR allows, can be imported just like in the example you've linked 👍

PR #11311, opened 2025-08-06
feat: propagate const for inline const

ahabhgk:
## Summary

Propagate const:

```js
const a = 1;
const b = 2;
export const value = a + b // before this won't inline, this PR will inline it
```

## Checklist

### DIFF
--- a/crates/rspack_core/src/exports/utils.rs
+++ b/crates/rspack_core/src/exports/utils.rs
@@ -5,7 +5,7 @@ use rspack_cacheable::{
   cacheable,
   with::{AsPreset, AsVec},
 };
-use rspack_util::{atom::Atom, json_stringify};
+use rspack_util::{atom::Atom, json_stringify, ryu_js};
 use rustc_hash::FxHashSet as HashSet;
 
 use crate::DependencyId;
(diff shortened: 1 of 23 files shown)


crates/rspack_core/src/exports/utils.rs
```diff
@@ -33,55 +33,76 @@ pub enum UsedExports {
...
-#[cacheable]
-#[derive(Debug, Clone, Hash, PartialEq, Eq)]
-pub struct EvaluatedInlinableValue(EvaluatedInlinableValueInner);
+impl Hash for EvaluatedInlinableValue {
+  fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+    match self {
+      EvaluatedInlinableValue::Null => 0.hash(state),
+      EvaluatedInlinableValue::Undefined => 1.hash(state),
+      EvaluatedInlinableValue::Boolean(v) => {
+        2.hash(state);
```
quininer: maybe use https://doc.rust-lang.org/stable/std/mem/fn.discriminant.html ?

SyMind (dismissed):

quininer (changes requested):

quininer (approved):

PR #11315, opened 2025-08-06
fix: export the missing `PathData` type

colinaaa:
## Summary

This type can be used when using function in `output.filename`. It would be great to have type exported.

## Related links

https://rsbuild.rs/config/output/filename#using-function

## Checklist

### DIFF
--- a/packages/rspack/src/exports.ts
+++ b/packages/rspack/src/exports.ts
@@ -8,10 +8,10 @@ export type {
 	Asset,
 	AssetInfo,
 	Assets,
+	ChunkPathData,
 	CompilationParams,
 	LogEntry,
-	PathData,
-	PathDataChunkLike
+	PathData
 } from "./Compilation";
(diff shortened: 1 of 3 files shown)


packages/rspack/src/exports.ts
```diff
@@ -9,7 +9,9 @@ export type {
 	AssetInfo,
 	Assets,
 	CompilationParams,
-	LogEntry
+	LogEntry,
+	PathData,
+	PathDataChunkLike
```
chenjiahan: To be consistent with webpack, how about renaming `PathDataChunkLike` to `ChunkPathData`?

https://github.com/webpack/webpack/blob/f2228eaa098b06e43d086f89f36a81c7888bad12/types.d.ts#L1628

colinaaa: Make sense! I've updated the exported type.

Issue #11319, opened 2025-08-06
Replace `LazyLock<Regex>` with compile-time regex


quininer:
Our current static regex are create using `LazyLock` on first access, which creates a parallelization bottleneck. and, create regex objects at runtime consumes unnecessary heap memory.

Switch to compile-time regex or parser would be beneficial. It's a simple experiment to replace the regex with [ere](https://crates.io/crates/ere) crate.

PR #11341, opened 2025-08-08
perf: Replace hot regex with parser

quininer:
## Summary

This eliminates most of the regex block in small benchmark.

sftrace image

Partially solved https://github.com/web-infra-dev/rspack/issues/11319, but there are still many regex not replaced

## Checklist

### DIFF
--- a/crates/rspack_util/src/identifier.rs
+++ b/crates/rspack_util/src/identifier.rs
@@ -111,15 +111,11 @@ pub fn make_paths_absolute(context: &str, identifier: &str) -> String {
   SEGMENTS_SPLIT_REGEXP
     .split(identifier)
     .map(|str| request_to_absolute(context, str))
-    .collect::<Vec<String>>()
-    .join("")
+    .collect()
 }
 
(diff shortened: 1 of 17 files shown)


crates/rspack_util/src/identifier.rs
```diff
@@ -111,15 +111,11 @@ pub fn make_paths_absolute(context: &str, identifier: &str) -> String {
...
-    .collect::<Vec<String>>()
-    .join("")
+    .collect()
 }
 
-static ZERO_WIDTH_SPACE: LazyLock<Regex> =
-  LazyLock::new(|| Regex::new("\u{200b}(.)").expect("invalid regex"));
-
 pub fn strip_zero_width_space_for_fragment(s: &str) -> Cow<'_, str> {
-  ZERO_WIDTH_SPACE.replace_all(s, "$1")
```
quininer, line 122: This is inconsistent with original regex, but symmetric with `insert_zero_width_space_for_fragment`. I'm not sure it's necessary to keep the logic of original regex.

ahabhgk: Should be just fine

Issue #11232, opened 2025-07-30
[Bug]: Panic occurred at runtime and LimitChunkCountPlugin bug


monjer:
### System Info

System:
+ OS: macOS 14.1.2
+ CPU: (8) arm64 Apple M1
+   Memory: 100.31 MB / 16.00 GB
+  Shell: 5.9 - /bin/zsh

Binaries:
+ Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
+ npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
+ pnpm: 9.11.0 - ~/.nvm/versions/node/v20.16.0/bin/pnpm

Browsers:
+ Chrome: 138.0.7204.169
+ Safari: 17.1.2

npmPackages:
+ @rspack/cli: ^1.4.10 => 1.4.10 
+ @rspack/core: ^1.4.10 => 1.4.10 
+ @rspack/plugin-react-refresh: ^1.4.3 => 1.4.3 

### Details

Use code splitting and set `new rspack.optimize.LimitChunkCountPlugin({ maxChunks: 1, })` in dev mode at the same time, when start the project , during the building process , and save any file , the rspack will exist with the error:
```
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues: panicked at crates/rspack_core/src/lib.rs:342:26:
Chunk(ChunkUkey(Ukey(2), PhantomData<rspack_core::chunk::Chunk>)) not found in ChunkByUkey
```

### Reproduce Steps

1. use rspack-cli create a new react project `npx create-rspack --dir my-project --template react`
2. install `@loadable/component` package , `npm i @loadable/component`
3. cretae `pages` folder and some test file and the folder structure like this 
```
src/pages/Page1.jsx
src/pages/Page2.jsx
src/pages/Page3.jsx
```
4. edit the code of file `src/App.jsx` like this 
```js
import { useState } from "react";
import loadable from '@loadable/component';
import "./App.css";

const fallback = {
  fallback: <div style={{ paddingTop: 16 }}></div>,
};

const Page1 = loadable(() => import(/* webpackChunkName:"Page1", webpackPrefetch:true */ 'src/pages/Page1'), fallback);
const Page2 = loadable(() => import(/* webpackChunkName:"Page2", webpackPrefetch:true */ 'src/pages/Page2'), fallback);
[...]

PR #11360, opened 2025-08-11
fix(incremental): LimitChunkCountPlugin panic with incremental turned on

pgoldberg:
## Summary

This fixes https://github.com/web-infra-dev/rspack/issues/11232.

This bug only happens with `experiments.incremental` set to `advance` (or `advance-slient`), and it happens because of the order in which we add mutations. Here's an example failure scenario with 3 chunks, `a`, `b`, and `c`, with the `LimitChunkCountPlugin` configured to 1 `maxChunks`:

1. `c` gets integrated into `b`, `integrated_chunks.insert(b)` and `removed_chunks.insert(c)`
1. `b` gets integrated into `a`, `integrated_chunks.insert(a)` and `removed_chunks.insert(b)`

https://github.com/web-infra-dev/rspack/blob/c40ac65cf85f423ea3eb7ddbf61ac82a2fe542f8/crates/rspack_plugin_limit_chunk_count/src/lib.rs#L210-L212

So now, we have:
`removed_chunks` contains `c` and `b`
`integrated_chunks` contains `b` and `a`

So we add these mutations (in this order):
`Mutation::ChunkRemove { chunk: c }`
`Mutation::ChunkRemove { chunk: b }`
`Mutation::ChunksIntegrate { to: b }`
`Mutation::ChunksIntegrate { to: a }`

https://github.com/web-infra-dev/rspack/blob/c40ac65cf85f423ea3eb7ddbf61ac82a2fe542f8/crates/rspack_plugin_limit_chunk_count/src/lib.rs#L302-L309

Later down the line when a file changes and we do an incremental rebuild, we get the affected chunks:
https://github.com/web-infra-dev/rspack/blob/973bd5cdd76730bcec047db9a04a8f488e9627fa/crates/rspack_core/src/compiler/compilation.rs#L1908

https://github.com/web-infra-dev/rspack/blob/81908623f3bba84ee1d76c2fca59e4906132b227/crates/rspack_core/src/incremental/mutations.rs#L146-L170

Because the `ChunkRemove` mutations come first, we end up with `b` and `a` as our affected chunks, even though we removed chunk `b`. That causes a panic at runtime when one of the taps of `additional_chunk_runtime_requirements` tries to get the non-existent chunk:
https://github.com/web-infra-dev/rspack/blob/973bd5cdd76730bcec047db9a04a8f488e9627fa/crates/rspack_core/src/compiler/compilation.rs#L2143-L2150

https://github.com/web-infra-dev/rspack/blob/8029c1dbc23a71a482fc5a869fd5feed598a5ff3/crates/rspack_plugin_runtime/src/array_push_callback_chunk_format.rs#L39-L46

This PR does the simpler thing of adding the `ChunkRemove` mutations after the `ChunksIntegrate` mutations, but we could instead keep track of mutations when integrating and removing chunks and add them in the order in which the mutations happen. I chose to do the most minimal change, but happy to do it the other way (or another way entirely) if that's preferred.

## Related links

https://github.com/web-infra-dev/rspack/issues/11232

## Checklist

### DIFF
--- a/crates/rspack_plugin_limit_chunk_count/src/lib.rs
+++ b/crates/rspack_plugin_limit_chunk_count/src/lib.rs
@@ -300,11 +300,13 @@ async fn optimize_chunks(&self, compilation: &mut Compilation) -> Result<Option<
   compilation.chunk_graph = new_chunk_graph;
 
   if let Some(mutations) = compilation.incremental.mutations_write() {
-    for chunk in removed_chunks {
-      mutations.add(Mutation::ChunkRemove { chunk });
-    }
+    // ChunkRemove mutations must be added last because a chunk can be removed after another chunk
+    // has been integrated into it
     for chunk in integrated_chunks {
       mutations.add(Mutation::ChunksIntegrate { to: chunk });
+    }
+    for chunk in removed_chunks {
+      mutations.add(Mutation::ChunkRemove { chunk });
     }
   }
 
--- a/packages/rspack-test-tools/tests/watchCases/chunks/limit-chunk-count-plugin/0/index.js
+++ b/packages/rspack-test-tools/tests/watchCases/chunks/limit-chunk-count-plugin/0/index.js
@@ -1,0 +1,12 @@
+async function main() {
+	const { utilA } = await import(/*webpackChunkName:"utilA"*/ "./utilA.js");
+	const { utilB } = await import(/*webpackChunkName:"utilB"*/ "./utilB.js");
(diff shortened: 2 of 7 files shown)


ahabhgk:
Great catch! This looks good to me, could you add a test for this, just for avoiding regression.

pgoldberg:
@ahabhgk Thanks! Just added a test case, and double-checked that it does fail as expected without this change:

image

ahabhgk (approved):
Thank you!

PR #11362, opened 2025-08-12
feat: port normal module replacement plugin to builtin plugin

ahabhgk:
## Summary

refactor and port normal module replacement plugin to builtin plugin

## Checklist


crates/rspack_binding_api/src/normal_module_factory.rs
```diff
@@ -51,20 +50,107 @@ pub struct JsCreateData {
...
+          .as_ref()
+          .map(|issuer| issuer.to_string())
+          .unwrap_or_default(),
+        issuer_layer: data.issuer_layer.clone(),
+      },
+      file_dependencies: data
+        .file_dependencies
+        .clone()
+        .into_iter()
+        .map(|item| item.to_string_lossy().to_string())
```
quininer:
```suggestion
        .map(|item| item.to_string_lossy().into_owned())
```

quininer: It would be better to handle this error if possible, because if lossy occurs, it is unlikely to be expected behavior.

crates/rspack_binding_api/src/normal_module_factory.rs
```diff
@@ -51,20 +50,107 @@ pub struct JsCreateData {
...
+        issuer: data
+          .issuer
+          .as_ref()
+          .map(|issuer| issuer.to_string())
+          .unwrap_or_default(),
+        issuer_layer: data.issuer_layer.clone(),
+      },
+      file_dependencies: data
+        .file_dependencies
+        .clone()
```
quininer: `.clone().into_iter()` seems unnecessary?

crates/rspack_plugin_module_replacement/src/normal_module_replacement.rs
```diff
@@ -0,0 +1,98 @@
...
+  data: &mut ModuleFactoryCreateData,
+  create_data: &mut NormalModuleCreateData,
+) -> Result<Option<bool>> {
+  if self
+    .resource_reg_exp
+    .test(&create_data.resource_resolve_data.resource)
+  {
+    match &self.new_resource {
+      NormalModuleReplacer::String(new_resource) => {
+        if new_resource.starts_with('/') || new_resource.chars().nth(1) == Some(':') {
```
quininer: maybe https://doc.rust-lang.org/stable/std/path/struct.Path.html#method.is_absolute ?

crates/rspack_binding_api/src/normal_module_factory.rs
ahabhgk: Just copied from origin code 🥲

crates/rspack_binding_api/src/normal_module_factory.rs
ahabhgk: We probably should store these file_dependencies as Utf8Path since rspack only support utf8 path, then no need to worry about the error here. But I'm not sure why file_dependencies is not using Utf8Path before

quininer (approved):

PR #11402, opened 2025-08-18
fix: correct export alias for json

### LINKED ISSUE #10722: [Bug]: Json transpiled to invalid js export alias
### System Info

```
System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 574.64 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.12.0 - ~/.nvm/versions/node/v22.12.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v22.12.0/bin/yarn
    npm: 10.9.0 - ~/.nvm/versions/node/v22.12.0/bin/npm
    bun: 1.2.10 - ~/.nvm/versions/node/v20.10.0/bin/bun
  Browsers:
    Chrome: 137.0.7151.105
    Edge: 137.0.3296.83
    Safari: 18.5
  npmPackages
    @rslib/core: 0.10.1
    @rsbuild/core: 1.4.0-beta.4
    @rspack/core: 1.4.0-beta.1
```

### Details

JSON first parent keys get transformed to invalid javascript. I found this issue while using rslib, but seems to be downstream in rspack

### Reproduce link

_No response_

### Reproduce Steps

1. Create a json file with a root key with a `-` in there;
```json
{
  "test-a": {}
}
```
2. Transpile, which results in;
```
var translation_namespaceObject = {"test-a":{}}
var __webpack_exports__test_a = translation_namespaceObject["test-a"];
export { __webpack_exports__test_a as test-a };
````

But is invalid as the `-` in the exports alias `as test-a` is not allowed

### PR DESCRIPTION
**Pull request by inottn**
## Summary
fix #10722 

## Checklist

### CODE REVIEW

File: packages/rspack-test-tools/tests/configCases/json/rspack-issue-10722/index.js [hunk 1]
```diff
@@ -0,0 +1,3 @@
+export * from "./foo.json";
+
+it("should run", () => {});
```
ahabhgk, line 3: This is not a build error so successfully build is not enough to ensure the case is covered. At least need to run the output and expect it has the correct exports

PR #11407, opened 2025-08-18
fix: NormalModuleReplacementPlugin doesn't update path, query and fragment

CPunisher:
## Summary

From https://github.com/web-infra-dev/rspack/blob/9a7358ac43501744c802e426b894ca9985de6fee/crates/rspack_binding_api/src/normal_module_factory.rs#L79, we can see `NormalModuleReplacementPlugin` updates not only the `resource` but also `path`, `query`, and `fragment`. These are missing in the https://github.com/web-infra-dev/rspack/pull/11362

## Checklist

### DIFF
--- a/crates/rspack_plugin_module_replacement/src/normal_module_replacement.rs
+++ b/crates/rspack_plugin_module_replacement/src/normal_module_replacement.rs
@@ -2,7 +2,7 @@ use derive_more::Debug;
 use futures::future::BoxFuture;
 use rspack_core::{
   ModuleFactoryCreateData, NormalModuleCreateData, NormalModuleFactoryAfterResolve,
-  NormalModuleFactoryBeforeResolve, Plugin, ResourceData, parse_resource,
+  NormalModuleFactoryBeforeResolve, Plugin,
 };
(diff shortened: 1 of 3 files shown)


ahabhgk:
Need a test case to avoid regression 😃

crates/rspack_plugin_module_replacement/src/normal_module_replacement.rs
```diff
@@ -65,12 +65,26 @@ async fn nmf_after_resolve(
   {
     match &self.new_resource {
       NormalModuleReplacer::String(new_resource) => {
+        // Same as [JsCreateData::update_nmf_data] in crates/rspack_binding_api/src/normal_module_factory.rs
+        fn update_resource_data(old_resource_data: &mut ResourceData, new_resource: String) {
```
ahabhgk: Let's extract this to rspack_util

CPunisher: I add this function to `ResourceData`. It could be a better way.

PR #11411, opened 2025-08-19
feat(cli): use SWC to do TS config transformamation

hardfist:
## Summary

Use  `experiments.swc` to transform TypeScript Rspack config files (CJS) at runtime and no need to install ts-node for most cases.

### Motivation

Provide out of box support of TypeScript config(CJS) support, TypeScript ESM is not covered and may be implemented in future PR.

### Breaking Change

@rspack/cli will not automatically try to load `ts-node/register` and `esbuild-resgister` to load TypeScript config, It will use builtin swc-register(based on rspack.experiments.transform) to load TypeScript config, It covers most cases of `ts-node` but if you still need more complex features(like declare const enum support), you can still use `ts-node/register` to  load typescript config by running following commands:

```
"build": "NODE_OPTIONS='--require ts-node/register' rspack build"
```

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/swc.rs
+++ b/crates/rspack_binding_api/src/swc.rs
@@ -57,7 +57,9 @@ fn _transform(source: String, options: String) -> napi::Result<TransformOutput> 
   compiler
     .transform(
       source,
-      Some(swc_core::common::FileName::Anon),
+      Some(swc_core::common::FileName::Real(
+        options.filename.clone().into(),
+      )),
       options,
       Some(module_source_map_kind),
       |_| {},
--- a/packages/rspack-cli/src/utils/loadConfig.ts
+++ b/packages/rspack-cli/src/utils/loadConfig.ts
@@ -54,7 +54,26 @@ const registerLoader = (configPath: string) => {
 				module: { type: "commonjs" },
 				sourceMaps: false,
 				isModule: true
-			});
+			let result;
+			try {
+				result = experiments.swc.transform(source, {
+					jsc: {
+						parser: {
+							syntax: "typescript",
+							tsx: false,
+							decorators: true,
+							dynamicImport: true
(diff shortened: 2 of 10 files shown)


crates/rspack_binding_api/src/swc.rs
```diff
@@ -57,7 +57,9 @@ fn _transform(source: String, options: String) -> napi::Result<TransformOutput>
   compiler
     .transform(
       source,
-      Some(swc_core::common::FileName::Anon),
```
hardfist, line 60: @GiveMe-A-Name transform missing filename cause sourcemap missing sources fields

packages/rspack-cli/src/utils/loadConfig.ts
```diff
@@ -1,50 +1,78 @@
...
+	return injectInlineSourceMap({ filename, code, map });
 }
-
 const DEFAULT_CONFIG_NAME = "rspack.config" as const;
-
-const registerLoader = async (configPath: string) => {
-	const ext = path.extname(configPath);
+// modified based on https://github.com/swc-project/swc-node/blob/master/packages/register/register.ts#L117
+const registerLoader = (configPath: string) => {
 	// TODO implement good `.mts` support after https://github.com/gulpjs/rechoir/issues/43
```
chenjiahan: Since we no longer depends on rechoir, should we remove the https://github.com/gulpjs/rechoir/issues/43 link?

hardfist: yes

crates/rspack_binding_api/src/swc.rs
GiveMe-A-Name: Got it

chenjiahan (approved):

PR #11420, opened 2025-08-19
fix: native watcher fails to work correctly with persistent cache

GiveMe-A-Name:
## Summary

This PR fixes an issue where the native file watcher doesn't work correctly with persistent cache. The main changes include:

1. Modified the getNativeWatcher method to return a tuple of [watcher, initialized], indicating whether the watcher is newly initialized or being reused.

2. Enhanced the watch method to handle file dependencies differently based on the watcher's initialization state:

- For newly initialized watchers: Pass all dependencies (files, directories, missing)
- For existing watchers: Only pass incremental changes (added/removed dependencies)

3. Improved code structure by extracting repeated logic into variables for better maintainability.

This fix ensures proper file watching behavior when persistent cache is enabled, preventing potential issues with file detection and updates.

## Checklist

### DIFF
--- a/packages/rspack/src/NativeWatchFileSystem.ts
+++ b/packages/rspack/src/NativeWatchFileSystem.ts
@@ -38,6 +38,7 @@ const toJsWatcherIgnored = (
 
 export default class NativeWatchFileSystem implements WatchFileSystem {
 	#inner: binding.NativeWatcher | undefined;
+	#isFirstWatch = true;
 	#inputFileSystem: InputFileSystem;
 
 	constructor(inputFileSystem: InputFileSystem) {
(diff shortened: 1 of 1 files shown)


packages/rspack/src/NativeWatchFileSystem.ts
```diff
@@ -163,9 +175,13 @@ export default class NativeWatchFileSystem implements WatchFileSystem {
 		};
 	}
 
-	getNativeWatcher(options: Watchpack.WatchOptions): binding.NativeWatcher {
+	getNativeWatcher(
+		options: Watchpack.WatchOptions
+	): [binding.NativeWatcher, boolean] {
```
stormslowly: getter returns a tuple (watcher, boolean)  not so conventional

GiveMe-A-Name: Updated

PR #11430, opened 2025-08-20
fix: bailout optimize if there are external modules in other concate scope

JSerFeng:
## Summary

Concate render in parallel, and each cannot have information of others. Thus when register global import, it must ensure that there is no other external modules in other concate scope.

eg.

```js
// foo.js in concate module 1
import {v} from 'externals1'

// bar.js in concate module 2
import {v} from 'externals2'
```

Result:

```js
import {v} from 'externals1'
import {v} from 'externals2' // <- duplicated imported specifier

__webpack_modules__ = {
  "concate1": function() { ... }
  "concate2": function() { ... }
}
```

## Checklist


Ryuurock:
到此一游

ahabhgk (approved):

PR #11451, opened 2025-08-21
fix: should update alias resolution when a higher-priority file is created in watch mode

SyMind:
## Summary

This PR fixes alias resolution in watch mode by ensuring that successful factorizations track missing dependencies, allowing the bundler to properly update when higher-priority alias files are created.

## Checklist


crates/rspack_core/src/compilation/make/graph_updater/cutout/mod.rs
```diff
@@ -66,14 +67,27 @@ impl Cutout {
...
+                      .module_identifier_by_dependency_id(dependency.id())
+                      .copied(),
+                  )
+                })
+            })
+            .collect::<Vec<_>>();
+          for (dependency_id, module_identifier) in affected_dependencies_and_modules {
+            force_build_deps.insert(dependency_id);
+            if let Some(module_identifier) = module_identifier {
+              force_build_modules.insert(module_identifier);
```
jerrykingxyz: why need rebuild target module？

SyMind: The main purpose is to disconnecting the module from the module graph. This is necessary because a dependency might resolve to a new module later. If the old module isn't disconnected, it will not be removed from the graph, even if it's no longer referenced.

jerrykingxyz: The force build deps will be revoked. And the fix issuer will check the target module whether isolation.

crates/rspack_core/src/dependency/context_dependency.rs
```diff
@@ -18,8 +20,8 @@ pub trait ContextDependency: Dependency {
   fn critical(&self) -> &Option<Diagnostic>;
   fn critical_mut(&mut self) -> &mut Option<Diagnostic>;
 
-  fn factorize_info(&self) -> &FactorizeInfo;
-  fn factorize_info_mut(&mut self) -> &mut FactorizeInfo;
+  fn factorize_info(&self) -> &Arc<FactorizeInfo>;
```
jerrykingxyz: Arc is useless, different dependencies will not share it, and even if they do, the persistent cache will be invalid after recovery.

crates/rspack_core/src/compilation/make/graph_updater/cutout/mod.rs
```diff
@@ -59,23 +60,32 @@ impl Cutout {
...
+                  Some(module.identifier())
+                } else {
+                  None
+                }
+              })
+              .collect::<Vec<_>>(),
+          );
+          force_build_deps.extend(
+            module_graph
+              .dependencies()
```
jerrykingxyz: The performance degradation may be caused by accessing all dependencies, which can far outnumber the number of modules. Maybe we can generate a new state of FactorizeInfo and save index at module artifact to make it faster.
``` rust
enum FactorizeInfo {
  Success,
  WithMissing { missing: ... }
  Failed { ... }
}

struct MakeArtifact {
   make_failed_dependencies: ...
   has_missing_info_dependencies: ...
}
```

jerrykingxyz: Another feasible solution is to use FactorizeInfo.Failed to store missing cases. Treat the case where missing_dependencies is not empty as the case where diagnostics is not empty.

``` rust
// factorize task
fn main() {
  if !factorize_info.diagnostics().is_empty() || !factorize_info.missing_dependencies.is_empty() {
      artifact
        .file_dependencies
        .add_batch_file(&factorize_info.file_dependencies());
      artifact
        .context_dependencies
        .add_batch_file(&factorize_info.context_dependencies());
      artifact
        .missing_dependencies
        .add_batch_file(&factorize_info.missing_dependencies());
      artifact
        .make_failed_dependencies
        .insert(*dependencies[0].id());
  }
}
```

crates/rspack_core/src/dependency/context_dependency.rs
SyMind: dependency in same module and with same `resource_identifier` will share it.

crates/rspack_core/src/compilation/make/graph_updater/cutout/mod.rs
SyMind: @jerrykingxyz Basically, every dependencies have missing dependencies

PR #11457, opened 2025-08-21
feat: tree shake statically analyse-able dynamic import

ahabhgk:
## Summary

Port: https://github.com/webpack/webpack/pull/19818

```js
let m = await import("./m");
m.a; // a is used, rest can be tree-shaken
m.f() // f is used, rest can be tree-shaken if strictThisContextOnImports = false
m // m namespace is used, can't be tree-shaken
```

## Checklist


ahabhgk:
Need to refactor JavaScriptParser.dependencies API before merge, I don't feel good for now we directly access the JavaScriptParser.dependencies

JSerFeng (approved):

PR #11463, opened 2025-08-22
refactor: run_loaders always return loader result

jerrykingxyz:
## Checklist


h-a-n-a (approved):
LGTM! This refactoring improves the separation of concerns between error information and loader context data. The new return type makes it clear that loader results are always available, which simplifies error handling and ensures dependency tracking is preserved regardless of execution success/failure.

PR #11475, opened 2025-08-25
perf: add fast path for `get_scheme`

CPunisher:
## Summary

Assume most modules don't contain scheme

## Checklist

### DIFF
--- a/crates/rspack_loader_runner/src/scheme.rs
+++ b/crates/rspack_loader_runner/src/scheme.rs
@@ -87,6 +87,12 @@ const HASH: char = '#';
 const QUERY: char = '?';
 
 pub fn get_scheme(specifier: &str) -> Scheme {
+  // Fast path: most modules don't have a scheme
+  let bytes = specifier.as_bytes();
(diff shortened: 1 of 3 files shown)


crates/rspack_loader_runner/src/scheme.rs
```diff
@@ -87,6 +87,12 @@ const HASH: char = '#';
 const QUERY: char = '?';
 
 pub fn get_scheme(specifier: &str) -> Scheme {
+  // Fast path: most modules don't have a scheme
+  let bytes = specifier.as_bytes();
+  if memchr::memchr(b':', bytes).is_none() {
```
quininer, line 92: just `specifier.contains(':')` ?

CPunisher: Yes, that's enough. I use `memchr` because I once think we can do more optimization based on the pos of `:`. But I think it's unnecessary because other code in this function is cold path.

quininer: Oh, I'm not sure if std's memchr is fast enough. use memchr should be fine.

Issue #11477, opened 2025-08-25
[Bug]: Module Federation fails to declare shared dependencies when using a custom resolver


elbywan:
### System Info

```sh
System:
  OS: macOS 15.6
  CPU: (10) arm64 Apple M1 Max
  Memory: 22.77 GB / 64.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 22.15.0 - ~/.volta/tools/image/node/22.15.0/bin/node
  Yarn: 4.9.1-git.20250411.hash-1908ee79f - ~/.volta/tools/image/yarn/4.9.1/bin/yarn
  npm: 10.9.2 - ~/.volta/tools/image/node/22.15.0/bin/npm
  pnpm: 10.14.0 - ~/.volta/bin/pnpm
Browsers:
  Brave Browser: 137.1.79.126
  Chrome: 139.0.7258.139
  Safari: 18.6
npmPackages:
  @rspack/cli: ^1.4.11 => 1.4.11 
  @rspack/core: ^1.4.11 => 1.4.11 
```

### Details

Hey 👋 

We are currently facing an issue when trying to upgrading rspack to `>= v1.4.7`.

When using the Module Federation plugin in conjunction with a custom resolver, the `__FEDERATION__.__SHARE__` global variable is empty during runtime - **whereas with older versions (`<= v1.4.6`) it was correctly populated.**

```js
// JSON.stringify(__FEDERATION__.__SHARE__)
'{"spa-federated:0.0.2":{"default":{}}}'
```

Our resolver plugin uses the [`unplugin`](https://unplugin.unjs.io/) library [which taps into the `normalModuleFactory.hooks.resolve` hook](https://github.com/unjs/unplugin/blob/main/src/rspack/index.ts#L64) - mutating the `resolveData.request` field.

I bisected the repo and the regression is likely coming from this refactoring of the hooks: 
- https://github.com/web-infra-dev/rspack/pull/10999

Before this change, `crates/rspack_plugin_mf/src/sharing/provide_shared_plugin.rs` was able to match the shared modules declared in the config using their name (`react`, `react-dom`…) but it is now trying to use the fully qualified path instead - which does not _feel right_ nor user friendly if it means that we have to resolve ahead of time and declare the shared dependencies using the absolute path.

[This simple change](https://github.com/web-infra-dev/rspack/compare/main...elbywan:rspack:fix-mf-shared-custom-resolution) seems to fix the problem by reverting to the old behaviour:

[...]

Issue #11437, opened 2025-08-20
[Bug]: NormalModuleReplacementPlugin isn't work together with ModuleFederationPlugin shared


SoonIter:
### System Info

Rspack v1.4.11

#### Before

rspack.NormalModuleReplacementPlugin work

```ts
  pluginModuleFederation({
      name: 'federation_consumer',
      remotes: {
        remote1: 'remote1@http://localhost:2001/mf-manifest.json',
      },
      shared: ['react', 'react-dom'],
    }),
```

<img width="1295" height="901" alt="Image" src="https://github.com/user-attachments/assets/7d98cfff-56e5-4de3-b80a-79f380a8b4f5" />

#### After

rspack.NormalModuleReplacementPlugin not work

```ts
  pluginModuleFederation({
      name: 'federation_consumer',
      remotes: {
        remote1: 'remote1@http://localhost:2001/mf-manifest.json',
      },
      shared: ['react', 'react-dom', '@arco-design/web-react'],
    }),
```

<img width="1445" height="655" alt="Image" src="https://github.com/user-attachments/assets/75524fb6-c36b-46b5-afdc-80d08cfd6f83" />

### Reproduce link

https://github.com/SoonIter/rsbuild-mf-normalModuleReplace-issue.git

### Reproduce Steps

1. pnpm install
2. pnpm run build

PR #11478, opened 2025-08-25
fix(mf): use unresolved request to match shared module

elbywan:
## Summary

Use `data.dependencies[0].request` instead of `create_data.raw_request` to check if a module is declared in the MF `shared` configuration.

Without it and when using a custom resolver hooking into `normalModuleFactory.hooks.resolve` the fully qualified path would be used and would end up not matching anything.

This PR should restore the same behaviour as before `v1.4.7` and specifically https://github.com/web-infra-dev/rspack/pull/10999

## Related links

- https://github.com/web-infra-dev/rspack/issues/11477

## Checklist


crates/rspack_plugin_mf/src/sharing/provide_shared_plugin.rs
```diff
@@ -242,7 +242,10 @@ async fn normal_module_factory_module(
   {
     return Ok(());
   }
-  let request = &create_data.raw_request;
+  let dependency = data.dependencies[0]
```
ahabhgk, line 245: I think a better way is use `data.request` here, should also solve the issue since `dependency.request` is assigned from `data.request`

elbywan: 👋 I tried to do that but it seems that `data.request` contains the mutated resolved path instead of the original request ([since this change?](https://github.com/web-infra-dev/rspack/pull/10999/files#diff-7898f75228fb06365ae84a67a02eb959e47b1e984c1e614a1e0631c8be4af473R1513)). I could not find another practical way of retrieving it except extracting it from `dependency[0].request` which does contain the pristine specifier.

ahabhgk: I see, have you tried webpack that they use the unresolved request or the changed request (changed by `NormalModuleFactory.hooks.afterResolve`)? If I remember correctly they will use the changed request. So at this point I'd like to keep the consistent with webpack

ahabhgk: Oh there is another issue about ConsumeSharedPlugin not using the changed request https://github.com/web-infra-dev/rspack/issues/11437 (your PR is about ProvideSharedPlugin not using original request)

elbywan: Ah I see this is more complicated than expected.

> I see, have you tried webpack that they use the unresolved request or the changed request (changed by NormalModuleFactory.hooks.afterResolve)? If I remember correctly they will use the changed request. So at this point I'd like to keep the consistent with webpack

Oh, I can double check 👍. 

What worries me is that if the logic is kept like this, it would make it very complicated to use a custom resolver - like with packages having subpath "exports" fields - because I'm not sure resolving all these paths ahead of time when generating the MF config + using the `import` field would be straightforward. 

Especially with packages that only export deep subpaths, without root exports. (`"exports": {"." { … } }`)

> Oh there is another issue about ConsumeSharedPlugin not using the changed request https://github.com/web-infra-dev/rspack/issues/11437 (your PR is about ProvideSharedPlugin not using original request)

I'm wondering, what would be the expected outcome in this case? The consumer would load the module from the resolved path and assume that it is not provided by a producer?

In addition to this PR changes, I tried modifying the request to use the resolved path `&data.request`  in the factorize hook [here](https://github.com/web-infra-dev/rspack/blob/c1ef1b58419360784f494769eaa8f2e6f1021784/crates/rspack_plugin_mf/src/sharing/consume_shared_plugin.rs#L411) and it seems to solve both issues but I'm not fully sure that I get all the implications yet. 

I'll try to dig into it and get a better understanding.

ahabhgk: I'm curious how you use the custom resolver that resolves paths ahead of time. If you resolve paths before compile, you can use the resolved paths to generate a MF plugin (configured with resolved paths instead of the original paths) and apply to the compiler

elbywan: I think that this is precisely the problem, I don't think that I can easily resolve all the possible subpaths for shared modules before compilation because I have no knowledge of the imports in the codebase. For simple cases it works, but as soon as I try to share a module exporting only subpaths it breaks (and the trailing slash syntax - `module/` - does not work when using an absolute path fallback).

For instance, for these kind of packages…

```json
{ "name": "package", "main": "index.js" }
```

…I cannot guess what the imports will look like in the provider code…

```ts
// any subpath is technically possible
import json from "package/package.json";
import stuff from "package/stuff";
import otherStuff from "package/other/stuff";
import someOtherStuff from "package/some/other/stuff";
```

…and it becomes very hard to generate the MF config:

```js
new ModuleFederationPlugin({
  name: "host",
  shared: {
    package: {
      version: "1.0.0",
      import: // I can add the resolved path here but what about nested paths??
    },
    "package/": {
      version: "1.0.0",
      import: // won't work
    },
    // Does every possible import paths have to be added as keys here?
  },
}),
```

Whereas this config used to work just fine with `v1.4.6`:

```js
new ModuleFederationPlugin({
  name: "host",
  shared: {
    package: {
      version: "1.0.0",
    },
    "package/": {
      version: "1.0.0",
    },
  },
}),
```

Nsttt:
Hey, got some feedback from the MF team, can you add the tests to https://github.com/web-infra-dev/rspack/tree/main/tests/rspack-test/configCases/sharing

elbywan:
Hey @Nsttt, that's great thanks for the update!

> Hey, got some feedback from the MF team, can you add the tests to https://github.com/web-infra-dev/rspack/tree/main/tests/rspack-test/configCases/sharing

Sure I'll try to dedicate some time ~this week~ next week to write them 👍 

**EDIT:** sorry I need a little more time 😬

elbywan:
👋 Just added a test case https://github.com/web-infra-dev/rspack/pull/11478/commits/8163e7972c238d030a01eaa54d8cfd1baf0014c3

```bash
pnpm run --filter="@rspack/tests" test:base -t configCases/sharing/provide-shared-with-custom-resolver
```

alessiopunturo:
Hello everyone! Any updates on this PR? :) We are trying to migrate our platform from webpack to rspack and we have module federation and yarn pnp. It seems there is an incompatibility between this components. Having this fix would be very very good.

PR #11482, opened 2025-08-26
refactor(browser): remove `esm.sh` default domain in `BrowserHttpImportEsmPlugin`

CPunisher:
## Summary

We cannot guarantee that the CDN is stable and secure

## Checklist


chenjiahan (approved):
LGTM 👍

PR #11484, opened 2025-08-26
perf(browser): disable `ProgressPlugin`, `RsdoctorPlugin` and `RstestPlugin`

CPunisher:
## Summary

I think `ProgressPlugin`, `RsdoctorPlugin` and `RstestPlugin` will not be used in browser.

Noted that I keep `RslibPlugin` because we may show rslib cases in browser?

Before: 27.01MB
After: 26.63MB

## Checklist


h-a-n-a (approved):

nicole-kustomer:
Doesn't this break progress for client?

https://rspack.rs/config/dev-server#progress
> Prints compilation progress in percentage in the browser.

`client: {
progress: true
}`

CPunisher:
> Doesn't this break progress for client?
> 
> https://rspack.rs/config/dev-server#progress
> 
> > Prints compilation progress in percentage in the browser.
> 
> `client: { progress: true }`

Thanks for your reminder. Yes, it breaks this progress, which I neglected. 

If it doesn't matter, we will add this broken option to [Browser API docs](https://rspack.rs/api/javascript-api/browser) to make it expected. Or if it has some useful cases, I can also add this plugin back.

What do you think?

PR #11494, opened 2025-08-27
ci: install `rustup` on Windows

colinaaa:
## Summary

Install `rustup` on Windows. This would avoid depending on having `rustup` on Windows runner (just like the Linux runner does).

## Related links

Fork from [`actions-rust-lang/setup-rust-toolchain`](https://github.com/actions-rust-lang/setup-rust-toolchain/blob/ab6845274e2ff01cd4462007e1a9d9df1ab49f42/action.yml#L144).

## Checklist

### DIFF
--- a/.github/workflows/reusable-rust-test.yml
+++ b/.github/workflows/reusable-rust-test.yml
@@ -104,7 +104,7 @@ jobs:
           cargo xtask deny-ext
 
       - uses: cargo-bins/cargo-binstall@79e4beb1e02f733a26129a6bf26c37dab4ab3307 # v1.14.4
-      - run: cargo binstall --no-confirm cargo-shear@1.1.12 --force
+      - run: cargo binstall --no-confirm cargo-shear@1.5.1 --force
       - run: cargo shear
 
   rust_test:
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-rust-test.yml
```diff
@@ -104,7 +104,7 @@ jobs:
           cargo xtask deny-ext
 
       - uses: cargo-bins/cargo-binstall@79e4beb1e02f733a26129a6bf26c37dab4ab3307 # v1.14.4
-      - run: cargo binstall --no-confirm cargo-shear@1.1.12 --force
+      - run: cargo binstall --no-confirm cargo-shear@1.5.1 --force
```
colinaaa, line 107: Due to the compile error:

```
Caused by:
  failed to select a version for the requirement `cargo-util-schemas = "^0.7.3"`
    version 0.7.3 is yanked
  location searched: crates.io index
  required by package `cargo-shear v1.1.12`
```

colinaaa:
Closed since we are not going to remove `rustup` on Windows.

PR #11548, opened 2025-09-01
feat: add css field support in Experiments configuration

ityuany:
## Summary

Enable `css` field to be properly passed from JavaScript/TypeScript interface layer through NAPI binding to Rust core layer.

## Changes

- Add `css` field to `Experiments` struct in `rspack_core`
- Add `css` field to `RawExperiments` struct in `rspack_binding_api`  
- Update type mapping from `RawExperiments` to `Experiments`
- Fix `ExperimentsBuilder` to properly handle css field conversion

## Test plan

- [x] All existing tests pass
- [x] Compilation succeeds for all relevant packages (`rspack_core`, `rspack_binding_api`, `rspack`)
- [x] Type system validates correct field transmission

This allows `experiments.css` configuration to be correctly transmitted from the JavaScript configuration to the Rust implementation, completing the configuration chain from NAPI layer to core.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/crates/rspack_binding_api/src/raw_options/raw_experiments/mod.rs
+++ b/crates/rspack_binding_api/src/raw_options/raw_experiments/mod.rs
@@ -26,6 +26,7 @@ pub struct RawExperiments {
   pub cache: RawExperimentCacheOptions,
   #[napi(ts_type = "false | Array<RegExp>")]
   pub use_input_file_system: Option<WithFalse<Vec<RspackRegex>>>,
+  pub css: bool,
   pub inline_const: bool,
   pub inline_enum: bool,
   pub type_reexports_presence: bool,
@@ -47,6 +48,7 @@ impl From<RawExperiments> for Experiments {
       top_level_await: value.top_level_await,
       rspack_future: value.rspack_future.unwrap_or_default().into(),
       cache: normalize_raw_experiment_cache_options(value.cache),
+      css: value.css,
       inline_const: value.inline_const,
       inline_enum: value.inline_enum,
       type_reexports_presence: value.type_reexports_presence,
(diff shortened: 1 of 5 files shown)


h-a-n-a:
May I ask for the use case?

ityuany:
> May I ask for the use case?

I need to customize some loaders to process CSS files, but I need to know if the user has enabled native CSS.

h-a-n-a:
CI failed. Please rerun `pnpm -w build:binding:dev` locally and see if `crates/node_binding/napi-binding.d.ts` is updated.

crates/rspack_binding_api/src/raw_options/raw_experiments/mod.rs
```diff
@@ -26,6 +26,7 @@ pub struct RawExperiments {
   pub cache: RawExperimentCacheOptions,
   #[napi(ts_type = "false | Array<RegExp>")]
   pub use_input_file_system: Option<WithFalse<Vec<RspackRegex>>>,
+  pub css: Option<bool>,
```
h-a-n-a, line 29: This should be `bool`. RawOptions need to be normalized from the JS side already

ityuany: https://github.com/web-infra-dev/rspack/blob/0fe3eeae0645589b50f816fc8e52c6d4a83a488d/packages/rspack/src/config/defaults.ts#L217-L223

here , the default value of experiments.css is undefined . 

i think the undefined and false should have different semantics .


https://github.com/web-infra-dev/rspack/blob/0fe3eeae0645589b50f816fc8e52c6d4a83a488d/packages/rspack/src/config/normalization.ts#L329-L372

here ,  normalized config options , but it's not have experiments.css .   i am not sure whether i should change the default value of experiments.css to false here

h-a-n-a: I see the case here. This is because `experiments.css` doesn't really need to be passed to the Rust side. It's converted into `CssModulesPlugin` on the fly. But I think your implementation is fine here though.

Issue #11551, opened 2025-08-29
[Bug]: @rsbuild/core v1.5 compiled code reports an error when running, the same code compiled with v1.4 does not have this problem


chenchengwork:
### Version

```sh
System:
    OS: Linux 6.14 Ubuntu 24.04.3 LTS 24.04.3 LTS (Noble Numbat)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1195G7 @ 2.90GHz
    Memory: 2.85 GB / 15.41 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Browsers:
    Chrome: 139.0.7258.154
```

### Details

代码运行报错信息:
index.ts:17 Uncaught TypeError: this.stack[(this.stack.length - 1)] is not a function

### Reproduce link

https://codesandbox.io/p/devbox/8gnf78

### Reproduce Steps

run `pnpm install && pnpm dev`
check console error

chenjiahan:
This might be a bug of Rspack or SWC.

- Input:

```js
function exitName(token) {
  const node = this.stack[this.stack.length - 1]
  assert(
    node.type === 'containerDirective' ||
      node.type === 'leafDirective' ||
      node.type === 'textDirective'
  )
  node.name = this.sliceSerialize(token)
}
```

- Output (Rspack 1.4.x):

```js
function exitName(token) {
  const node = this.stack[this.stack.length - 1]
  ;(0,devlop__WEBPACK_IMPORTED_MODULE_2__.ok)(
    node.type === 'containerDirective' ||
      node.type === 'leafDirective' ||
      node.type === 'textDirective'
  )
  node.name = this.sliceSerialize(token)
}
```

- Output (Rspack 1.5.x) (>=1.5.0-beta.0):

```js
function exitName(token) {
  const node = this.stack[this.stack.length - 1]
  (0,devlop__WEBPACK_IMPORTED_MODULE_2__.ok)(
    node.type === 'containerDirective' ||
      node.type === 'leafDirective' ||
      node.type === 'textDirective'
  )
  node.name = this.sliceSerialize(token)
}
```

PR #11555, opened 2025-09-02
fix: use stable swc lexer for correct asi

ahabhgk:
## Summary

The `swc_ecma_parser::Lexer` will output different tokens when using it in `parse` and `.collect_vec()`

For example: 

```js
import {assert} from './a'

const labelType = `${node}Label`

const node = []
assert(labelType)
```

In `parse` this will have correct tokens, but using with `.collect_vec()` it will have a lexing error: `Error { error: (60..97, UnterminatedTpl) }`

And the lexing error causes rspack collected the wrong semicolons (which is used for adding semicolons before replaced dependency to generate the correct code)

So this PR switch to the stable `swc_ecma_lexer::Lexer`.

## Related links

fix https://github.com/web-infra-dev/rspack/issues/11551

A better solution is using SWC's API https://github.com/swc-project/swc/issues/11053

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/Cargo.toml
+++ b/crates/rspack_plugin_javascript/Cargo.toml
@@ -50,6 +50,7 @@ swc_core = { workspace = true, features = [
   "base",
   "ecma_quote",
 ] }
+swc_ecma_lexer = { version = "=23.0.1" }
 swc_node_comments = { workspace = true }
 tokio = { workspace = true, features = ["sync"] }
 tracing = { workspace = true }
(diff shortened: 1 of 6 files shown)


crates/rspack_plugin_javascript/Cargo.toml
```diff
@@ -50,6 +50,7 @@ swc_core = { workspace = true, features = [
   "base",
   "ecma_quote",
 ] }
+swc_ecma_lexer = { version = "=23.0.1" }
```
CPunisher: write to workspace

PR #11520, opened 2025-08-28
chore: use pnpm in package.json scripts

ahabhgk:
## Summary

To avoid these npm warn:

Screenshot 2025-08-28 at 15 49 25

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #11572, opened 2025-09-03
chore: use local package for scripts

ahabhgk:
## Summary

ref https://github.com/web-infra-dev/rspack/pull/11520#discussion_r2306710529

## Checklist


chenjiahan:
I've intentionally used `pnpx cspell` before to reduce npm dependencies and speed up workflow (cspell is only used in specific CI). However, I can accept declaring these dependencies because it makes version management easier.

image

PR #11575, opened 2025-09-03
perf: improve deterministic chunk ids

LingyuCoder:
## Summary

We can make sure that chunk module exists when it is not a runtime module, then we can prevent finding all modules on module graph by `loop_partial` which needs to indexing multiple times.

Before:
image

After:
image

## Checklist

### DIFF
--- a/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
+++ b/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
@@ -12,8 +12,9 @@ use serde::{Serialize, Serializer};
 
 use crate::{
   BoxModule, Chunk, ChunkByUkey, ChunkGraph, ChunkGraphModule, ChunkGroupByUkey, ChunkGroupUkey,
-  ChunkIdsArtifact, ChunkUkey, Compilation, Module, ModuleGraph, ModuleGraphCacheArtifact,
-  ModuleIdentifier, RuntimeGlobals, RuntimeModule, SourceType, find_graph_roots, merge_runtime,
+  ChunkIdsArtifact, ChunkUkey, Compilation, DependencyType, Module, ModuleGraph,
+  ModuleGraphCacheArtifact, ModuleIdentifier, RuntimeGlobals, RuntimeModule, SourceType,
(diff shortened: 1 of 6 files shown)


crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
```diff
@@ -412,6 +413,27 @@ impl ChunkGraph {
     modules
   }
 
+  pub fn get_ordered_chunk_normal_modules_identifier(
```
ahabhgk, line 415: I think you don't need add a new method, you can just use `get_ordered_chunk_modules_identifier` here, at this phase runtime_modules is always empty.

crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
```diff
@@ -674,6 +697,14 @@ impl ChunkGraph {
       ) {
         for connection in module_graph.get_outgoing_connections(&module) {
           // https://github.com/webpack/webpack/blob/1f99ad6367f2b8a6ef17cce0e058f7a67fb7db18/lib/ChunkGraph.js#L290
+          if set.contains(connection.module_identifier()) {
+            let maybe_transitive = module_graph
+              .dependency_by_id(&connection.dependency_id)
+              .is_some_and(|dep| matches!(dep.dependency_type(), DependencyType::ExtractCSS));
```
ahabhgk: How's the benefit here, if improve the perf a lot I'm ok with this, but if not I don't recommend to depend on `DependencyType::ExtractCSS` here

LingyuCoder: This improve about 10~20% of calculating chunk ids. I can remove this because the total time of calculating chunk ids is small. Perhaps we can remove `ConnectionState::TransitiveOnly` in the future.

crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
LingyuCoder: I will refactor all getting chunk modules related methods in #11579 and remove in it.

PR #11576, opened 2025-09-03
fix: BuildDependencies resolve exports_fields incorrectly

jerrykingxyz:
## Checklist


h-a-n-a (approved):

jjoshi-space:
@jerrykingxyz Is the fix available within latest v1.5.2 ?
I'm unable to find the reference under the release [tag](https://github.com/web-infra-dev/rspack/releases/tag/v1.5.2)
If the fix is not published yet, can you share tentatively when it will be published?

Thanks

Issue #11583, opened 2025-09-04
[Bug]: rspack-split-chunk bug with webpackExports magic comments


zhangyuang:
### System Info

  System:
    OS: macOS 15.5
    CPU: (14) arm64 Apple M4 Pro
    Memory: 4.90 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.5.0 - ~/.nvm/versions/node/v24.5.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v24.5.0/bin/yarn
    npm: 11.5.1 - ~/.nvm/versions/node/v24.5.0/bin/npm
    pnpm: 10.15.1 - ~/.nvm/versions/node/v24.5.0/bin/pnpm
  Browsers:
    Chrome: 140.0.7339.80
    Safari: 18.5
  npmPackages:
    @rspack/cli: ^1.4.11 => 1.4.11
    @rspack/core: ^1.4.11 => 1.4.11
    @rspack/plugin-react-refresh: ^1.4.3 => 1.4.3

### Details

When I use magic comments with webpackChunkName and webpackExports together, it cannot generate the correct chunk name.

For the code below, the webpackChunkName can't take effect.

```js
const foo = () => import(
		/*webpackChunkName: "foo"*/
		/*webpackExports: "default"*/
		'./foo')
```

### Reproduce link

https://github.com/zhangyuang/rspack-bug-splitchunk/tree/bug/webpackExports

### Reproduce Steps

```bash
# yarn && npm run build && ls dist
787.js       787.js.map   index.html   main.css     main.css.map main.js      main.js.map
```

There will generate `787.js` instead of `foo.js`

zhangyuang:
I guess this bug may be caused by [webpack_comment.rs](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/webpack_comment.rs#L470).

If you'd welcome it, I would be happy to submit a pull request.

PR #11597, opened 2025-09-05
fix: webpackexports comment

zhangyuang:
## Summary

If magic comments with webpackExports, the iterator will be finished, which causes the comments to remain and can't be collected.

## Related links

#11583 

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,7 @@
   "[typescript]": {
     "editor.formatOnSave": true
   },
-  "editor.formatOnSave": false,
+  "editor.formatOnSave": true,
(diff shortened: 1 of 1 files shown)


LingyuCoder (commented):
LGTM, the test cases should be added in `tests/rspack-test`. The `test/webpack-test` is using to migrate test cases from webpack and should not be modified.

zhangyuang:
Okay, let me update it.

zhangyuang:
Has updated

LingyuCoder (approved):
thanks

i5ting:
棒棒哒

PR #11598, opened 2025-09-05
perf: improve create hash

LingyuCoder:
## Summary

- Use `RwLock::try_read` instead of `DashMap` in `JsPlugin::getCompilationHooks` to achieve better read performance when frequently accessing hooks, as this stage has no write operations so it's safe.
- When calculating content hash, write the chunk hash directly to prevent chunk hash from being calculated multiple times, as there are high hook dispatch costs involved.

Before:
image

After:
image

## Checklist


packages/rspack-test-tools/tests/hookCases/javascriptModulesPlugin#chunkHash/update-hash/hooks.snap.txt
ahabhgk, line 1: Is this expected?

LingyuCoder: Yes, the duplicated chunk hash will not be calculated, so the data that hasher received will be removed.

ahabhgk (approved):

PR #11619, opened 2025-09-09
refactor: replace once_cell crate with std

jerrykingxyz:
## Summary

1. replace `once_cell::sync::OnceCell` to `std::sync::OnceLock`
2. replace `once_cell::unsync::OnceCell` to `std::cell::OnceCell`

## Checklist


crates/rspack_util/src/once_cell.rs
```diff
@@ -0,0 +1,21 @@
+use std::sync::OnceLock;
+
+/// Alternative to OnceLock.get_or_try_init
+///
+/// TODO remove this method after OnceLock.get_or_try_init stable
+pub fn once_lock_get_or_try_init<T, E, F>(cell: &OnceLock<T>, init: F) -> Result<&T, E>
```
quininer, line 8: This is incorrect, as it would allow `init` to be called concurrently

quininer (changes requested):

crates/rspack_util/src/once_cell.rs
```diff
@@ -10,12 +12,13 @@ where
   if let Some(value) = cell.get() {
     return Ok(value);
   }
-  let new_value = init()?;
-  if cell.set(new_value).is_err() {
-    unreachable!()
+
+  let _guard = LOCK.lock().expect("should get lock success");
```
quininer, line 16: This cause a single global lock contention for all `once_lock_get_or_try_init`, which doesn't look good. :thinking:

jerrykingxyz: Yes, but considering that this function is currently used in few places, I think it is okay to sacrifice some performance.

jerrykingxyz: Or do you have any better idea?

PR #11629, opened 2025-09-10
fix: no mangle for enum member value

ahabhgk:
## Summary

Because we don't control the codegen of enum value (just use the loader generated), so we can't mangle it.

```js
var enum_E = /*#__PURE__*/ function(E) {
    E[E["Dynamic"] = Math.random()] = "Dynamic"; // we don't control the codegen here, it's generated by loader
    return E;
}({});

enum_E.Dynamic // should always use Dynamic here, we can't mangle it to something shorter
```

## Checklist


SyMind (approved):

chenjiahan:
LGTM 👍

PR #11634, opened 2025-09-10
fix(rslib): only treat "module" as normal variable when module have ESM syntax

fi3ework:
## Summary

This is a revised version of 15ba68e.

This PR mainly do:

when `parser.is_esm` is `true` (which means the CJS exports is useless due to the presence of ESM syntax) add parser hooks just like what `crates/rspack_plugin_javascript/src/parser_plugin/common_js_exports_parse_plugin.rs` do to `module` variable. and simply return `true` to bailout the afterwards hooks.

## Related links

revised version of https://github.com/web-infra-dev/rspack/pull/11588. only ignore`"module"` variable when ESM `import / export` are present.

## Checklist


fi3ework:
- rstest failed as expected
- Lynx failed since previous [PR](https://github.com/web-infra-dev/rspack/actions/runs/17823139372/job/50670434160) (at least)

going to merge

PR #11638, opened 2025-09-11
fix: mark import.meta.webpackHot type as optional

### LINKED ISSUE #11631: [Feature]: Mark import.meta.webpackHot type signature as optional

Currently, when referencing `@rsbuild/core/types`, ImportMeta type is 
``` ts
interface ImportMeta {
    url: string;
    // TODO: unsupported
    // webpack: number;
    webpackHot: Rspack.Hot;
   // ...
}
```

Cann we make webpackHot be optional, so that it matches the documentation that it may not necessarily exist? (If I got it right, during SSR it does not exist, if this is not expected I'll fire a bug report)
https://rspack.rs/api/runtime-api/hmr#example

### What does the proposed API look like?

-

yf-yang:
## Summary

closes #11631

## Checklist

### DIFF
--- a/packages/rspack/module.d.ts
+++ b/packages/rspack/module.d.ts
@@ -177,7 +177,7 @@ declare namespace Rspack {
 		loaded: boolean;
 		parents: NodeJS.Module["id"][] | null | undefined;
 		children: NodeJS.Module["id"][];
-		hot?: Hot | undefined;
+		hot?: Hot;
 	}
 
 	interface RequireResolve {


packages/rspack/module.d.ts
```diff
@@ -219,7 +219,7 @@ interface ImportMeta {
 	url: string;
 	// TODO: unsupported
 	// webpack: number;
-	webpackHot: Rspack.Hot;
+	webpackHot?: Rspack.Hot | undefined;
```
chenjiahan:
```suggestion
	webpackHot?: Rspack.Hot;
```

yf-yang:
Would you like to update this line, too?

https://github.com/web-infra-dev/rspack/blob/9dd9d3d07ca7a9eeaeae612c8b81d48ea3298054/packages/rspack/module.d.ts#L180

chenjiahan:
> Would you like to update this line, too?
> 
> https://github.com/web-infra-dev/rspack/blob/9dd9d3d07ca7a9eeaeae612c8b81d48ea3298054/packages/rspack/module.d.ts#L180

That would be great

PR #11643, opened 2025-09-11
fix: should update alias resolution when a higher-priority file is created in watch mode

SyMind:
## Summary

This PR enhances alias resolution in watch mode by ensuring that successful factorizations track missing dependencies. This allows the bundler to properly re-resolve and update module paths when higher-priority alias files or new resolution-relevant files (such as `package.json`) are created or modified.

### Problem

In watch mode, when a user adds or modifies a file that could affect module resolution (e.g., adding a higher-priority alias file or changing `package.json` exports), the resolver should re-resolve affected dependencies to ensure correctness. Previously, the bundler did not track these resolution dependencies, leading to outdated references.

### Example

Given the alias configuration:

```js
alias: ["a.js", "b.js"]
```

Where `a.js` has higher priority than `b.js`, if a user adds `a.js` after the initial build, the bundler should update the resolution from `b.js` to `a.js`. Without tracking missing dependencies during resolution, this update did not occur.

### Implementation Details

The fix tracks **missing dependencies** during alias resolution. When a module is resolved, the resolver records any missing paths that, if created, would change the resolution outcome. This allows the watcher to trigger re-resolution when those files are added.

### Additional Cases Handled

1. Extension priority: For extensions like `['js', 'ts', 'tsx']`, if `a.ts` is added after `a.js` was resolved, the resolver will now correctly update to use `a.ts` (if `ts` has higher priority).

2. Directory index resolution: When `a/index.js` exists and `a.js` is added, the resolution updates from the index file to the higher-priority `a.js`.

3. `package.json` tracking: Adding or modifying package.json (e.g., changing `"main"` or `"exports"`) now triggers re-resolution for modules that depend on it.

## Checklist

### DIFF
--- a/crates/rspack_core/src/compilation/make/artifact.rs
+++ b/crates/rspack_core/src/compilation/make/artifact.rs
@@ -142,21 +142,22 @@ impl MakeArtifact {
   pub fn revoke_dependency(&mut self, dep_id: &DependencyId, force: bool) -> Vec<BuildDependency> {
     let mut mg = ModuleGraph::new([None, None], Some(&mut self.module_graph_partial));
 
-    let revoke_dep_ids = if let Some(dependency) = mg.dependency_by_id(dep_id)
-      && let Some(info) = FactorizeInfo::get_from(dependency)
+    let revoke_dep_ids = if let Some(factorize_info) = mg
+      .dependency_by_id(dep_id)
+      .and_then(|dependency| FactorizeInfo::get_from(dependency))
     {
       let resource_id = ResourceId::from(dep_id);
       self
         .file_dependencies
-        .remove_files(&resource_id, info.file_dependencies());
+        .remove_files(&resource_id, factorize_info.file_dependencies());
       self
         .context_dependencies
-        .remove_files(&resource_id, info.context_dependencies());
+        .remove_files(&resource_id, factorize_info.context_dependencies());
       self
         .missing_dependencies
-        .remove_files(&resource_id, info.missing_dependencies());
+        .remove_files(&resource_id, factorize_info.missing_dependencies());
       // related_dep_ids will contain dep_id it self
-      info.related_dep_ids().to_vec()
+      factorize_info.related_dep_ids().to_vec()
     } else {
       vec![*dep_id]
     };
--- a/crates/rspack_core/src/compilation/make/artifact.rs
+++ b/crates/rspack_core/src/compilation/make/artifact.rs
@@ -52,8 +52,6 @@ pub struct MakeArtifact {
   // statistical data, which can be regenerated from module_graph_partial and used as index.
   /// Diagnostic non-empty modules in the module graph.
   pub make_failed_module: IdentifierSet,
-  /// Factorize failed dependencies in module graph
-  pub make_failed_dependencies: HashSet<DependencyId>,
   /// Entry dependencies in the module graph
   pub entry_dependencies: HashSet<DependencyId>,
   /// The files that current module graph depends on.
@@ -116,22 +114,19 @@ impl MakeArtifact {
       .iter()
       .chain(mgm.incoming_connections())
     {
-      if !self.make_failed_dependencies.remove(dep_id) {
-        continue;
-      }
-      // make failed dependencies clean it.
       let dep = mg.dependency_by_id(dep_id).expect("should have dependency");
-      let info = FactorizeInfo::get_from(dep).expect("should have factorize info");
-      let resource_id = ResourceId::from(dep_id);
-      self
-        .file_dependencies
-        .remove_files(&resource_id, &info.file_dependencies());
-      self
-        .context_dependencies
-        .remove_files(&resource_id, &info.context_dependencies());
-      self
-        .missing_dependencies
-        .remove_files(&resource_id, &info.missing_dependencies());
+      if let Some(info) = FactorizeInfo::get_from(dep) {
+        let resource_id = ResourceId::from(dep_id);
+        self
+          .file_dependencies
+          .remove_files(&resource_id, info.file_dependencies());
+        self
+          .context_dependencies
+          .remove_files(&resource_id, info.context_dependencies());
+        self
+          .missing_dependencies
+          .remove_files(&resource_id, info.missing_dependencies());
+      }
     }
 
     self.revoked_modules.insert(*module_identifier);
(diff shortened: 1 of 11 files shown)


crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -1218,3 +1261,80 @@ impl<'a> ModuleGraph<'a> {
...
+
+#[allow(clippy::type_complexity)]
+pub enum FactorizeInfos<'a> {
+  // An optimized iterator for when there's only a single, read-only layer of data.
+  // It directly iterates over the underlying `HashMap` without any merging overhead,
+  // filtering out deleted entries (`None` values).
+  Direct(
+    std::iter::FilterMap<
+      std::collections::hash_map::Iter<'a, DependencyId, Option<FactorizeInfo>>,
+      fn((&DependencyId, &'a Option<FactorizeInfo>)) -> Option<(DependencyId, &'a FactorizeInfo)>,
```
jerrykingxyz: This structure cannot be serialized. Is this incomprehensible structure due to performance considerations?

SyMind: Yes, and it does not need to be serialized since it is just an iterator, to avoid creating a HashSet.

crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -1218,3 +1261,80 @@ impl<'a> ModuleGraph<'a> {
...
+        .unwrap();
+      Self::Direct(
+        partial
+          .dependency_factorize_infos
+          .iter()
+          .filter_map(|item| item.1.as_ref().map(|info| (*item.0, info))),
+      )
+    } else {
+      let mut merged = HashMap::default();
+      for item in module_graph.partials.iter().flatten() {
```
jerrykingxyz: why do we need merge them? It seems that we can just search for it like connection.

crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -55,6 +55,8 @@ pub struct ModuleGraphPartial {
   /// Dependencies indexed by `DependencyId`.
   dependencies: HashMap<DependencyId, Option<BoxDependency>>,
 
+  dependency_factorize_infos: HashMap<DependencyId, Option<FactorizeInfo>>,
```
jerrykingxyz: This field need to be added to persistent cache MakeOccasion.

crates/rspack_core/src/module_graph/mod.rs
SyMind: just logic as same as `module_graph.dependencies()
`

crates/rspack_core/src/module_graph/mod.rs
SyMind: done

crates/rspack_core/src/module_graph/mod.rs
jerrykingxyz: This FactorizeInfos seems to be just for providing a fast path, and it is also very strange from a logical point of view. In fact, in the make stage, there are often only active and no partials.

I suggest you refer to the module_graph_module approach first and avoid over-optimization.

jerrykingxyz: If you must do it, consider this
``` rust
pub fn dependency_factorize_info_iter(&self) -> ... {
   // assert only one ModuleGraphPartial has dependency_factorize_infos
   let mut infos = self.active.and_then(|p| {                                                                                                     
      if p.dependency_factorize_infos.is_empty() {                                              
        None                                                                                                                                                                
      } else {                                                                                                                                                              
        Some(p.dependency_factorize_infos)                                                                                                                                                 
      }                                                                                                                                                                     
    });                                                                                                                                                                     
    for item in self.partials {                                                                                                                                             
      if infos.is_some() {                                                                                                                                                  
        panic!("xxx")                                                                                                                                                       
      }                                                                                                                                                                     
      infos = item.and_then(|p| {                                                                                                                                           
        if p.dependency_factorize_infos.is_empty() {                                                                                                                                       
          None                                                                                                                                                              
        } else {                                                                                                                                                            
          Some(p.dependency_factorize_infos)                                                                                                                                               
        }                                                                                                                                                                   
      })                                                                                                                                                                    
    }
    ...
}
```

SyMind: Here, the optimization is effective.

```rust
force_build_deps.extend(
  module_graph
    .dependency_factorize_info_iter()
    // The cost of `factorize_info.depends_on` can be highly non-uniform.
    // To enable Rayon's work-stealing for better load balancing, we first
    // materialize the iterator into a Vec. This allows `par_iter` to create
    // perfectly divisible slices, which is much more efficient than using
    // `par_bridge()` on a non-indexable iterator.
    .collect::<Vec<_>>()
    .par_iter()
    .filter_map(|(dependency_id, factorize_info)| {
      factorize_info.depends_on(&files).then_some(dependency_id)
    })
    .collect::<Vec<_>>(),
);
```

crates/rspack_core/src/compilation/make/graph_updater/cutout/mod.rs
```diff
@@ -59,23 +60,41 @@ impl Cutout {
...
-              .expect("should have dependency");
-            let info = FactorizeInfo::get_from(dep).expect("should have factorize info");
-            if info.depends_on(&files) {
-              force_build_deps.insert(*dep_id);
-            }
-          }
+          force_build_modules.extend(
+            module_graph
+              .modules()
+              .values()
```
jerrykingxyz: I'm not sure how much performance improvement will be achieved by switching to multithreading. The computational complexity of depends_on itself is not particularly large, and additional collect is required here. I suggest

``` rust
module_graph.modules().par_iter().filter_map(|(_, modules)| { ... }) 
```

SyMind: @jerrykingxyz I have reverted, because there are 2 times improvement in debug binding, but no improvement in release binding.

crates/rspack_core/src/compilation/make/artifact.rs
```diff
@@ -144,21 +146,18 @@ impl MakeArtifact {
...
       self
         .missing_dependencies
-        .remove_batch_file(&info.missing_dependencies());
+        .remove_batch_file(&factorize_info.missing_dependencies());
       // related_dep_ids will contain dep_id it self
-      info.related_dep_ids().into_owned()
+      factorize_info.related_dep_ids().to_vec()
     } else {
       vec![*dep_id]
     };
```
jerrykingxyz, line 171: mg.remove_dependency_factorize_info_by_id(&dep_id) ?

crates/rspack_core/src/dependency/factorize_info.rs
```diff
@@ -4,13 +4,15 @@ use rspack_cacheable::cacheable;
...
-use super::{BoxDependency, DependencyId};
+use super::DependencyId;
 
 #[cacheable]
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 pub enum FactorizeInfo {
-  #[default]
-  Success,
+  Success {
```
jerrykingxyz: merge them?

crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -1214,3 +1257,80 @@ impl<'a> ModuleGraph<'a> {
     }
   }
 }
+
+#[allow(clippy::type_complexity)]
+pub enum FactorizeInfos<'a> {
```
jerrykingxyz: remove it.

PR #11651, opened 2025-09-12
perf: introduce pre-computed hashing for ArcPath

SyMind:
## Summary

This PR introduces a significant performance optimization for ArcPath by pre-computing and caching its hash value. ArcPath is used extensively throughout the codebase as a key in HashMap and HashSet (including concurrent versions like DashMap). Previously, every hash operation (e.g., insert, get) required re-calculating the hash by iterating over the entire path string, leading to substantial and repetitive CPU overhead.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4461,7 +4461,6 @@ dependencies = [
  "rspack_plugin_javascript",
  "rspack_regex",
  "rspack_util",
- "rustc-hash",
  "serde_json",
  "tokio",
  "tracing",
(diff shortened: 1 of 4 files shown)


crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
```diff
@@ -24,11 +24,11 @@ pub struct BuildDeps {
   /// The build dependencies has been added to snapshot.
   ///
   /// This field is used to avoid adding duplicate build dependencies to the snapshot.
-  added: HashSet<ArcPath>,
+  added: ArcPathSet,
```
quininer, line 27: There is an optimization point here, the precomputed hash itself does not need to be stored in value, because it is already stored in hashmap

PR #11664, opened 2025-09-15
feat: parser support jsx

fi3ework:
## Summary

support JSX `preserve`.

## Checklist


crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs
```diff
@@ -179,8 +180,21 @@ impl ParserAndGenerator for JavaScriptParserAndGenerator {
     );
     let comments = SwcComments::default();
     let target = ast::EsVersion::EsNext;
+
+    let jsx = if let Some(js_parser_options) = module_parser_options
```
ahabhgk: use `ParserOptions::get_javascript()` here

crates/rspack_core/src/concatenated_module.rs
```diff
@@ -1989,10 +1989,27 @@ impl ConcatenatedModule {
       let comments = SwcComments::default();
       let mut module_info = concatenation_scope.current_module;
 
+      let jsx = compilation
+        .options
+        .module
+        .parser
+        .as_ref()
+        .and_then(|p| {
+          let options = p.get(ModuleType::JsEsm.as_str());
```
ahabhgk: This wouldn't respect parser options for specific rule

```js
module: {
  rule: [
    { test: /\.js/, parser: { jsx: true } }
  ]
}
```

fi3ework: yes, but i feel it's a little verbose to iter all js types, in Rslib side, Rslib will enable `jsx` field for all js types when `@rsbuild/plugin-react` is added to plugin, so not a issue for Rslib. any suggestions?

ahabhgk: Not about different js types, every normal module has its own parser options, different modules may use different options, even rslib only needs all modules to use the same options but user can still set different options by overriding the defaults. You need to get it from the normal module itself so you can get the correct options

fi3ework: made a modification, PTAL.

crates/rspack_core/src/concatenated_module.rs
```diff
@@ -1989,10 +1989,41 @@ impl ConcatenatedModule {
...
+            | ParserOptions::JavascriptDynamic(js_options) => js_options.jsx,
+            _ => None,
+          }
+        };
+
+        module
+          .as_ref()
+          .as_normal_module()
+          .and_then(|normal_module| normal_module.get_parser_options())
+          .and_then(get_jsx)
```
ahabhgk: Use `get_javascript().jsx` here, it's the simple way to handle get options from different js types

crates/rspack_core/src/concatenated_module.rs
```diff
@@ -1989,10 +1989,41 @@ impl ConcatenatedModule {
...
+          .or_else(|| {
+            compilation
+              .options
+              .module
+              .parser
```
ahabhgk: You don't need to read from compilation.options here if it's None, normal module will always has parser options once it's factorized, if it's None, it means the module is not normal module, you can just use `unwarp_or(false)` here, because the codegen of non-normal module won't contains jsx

fi3ework: PTAL, i think i made it right this time. 😇

ahabhgk: LGTM

fi3ework: thank u Kun!

tests/rspack-test/configCases/parsing/jsx-enabled/test.js
```diff
@@ -0,0 +1,115 @@
+const fs = require("fs");
+const path = require("path");
+
+it("should keep jsx in output when parser jsx is enabled", () => {
+  const bundle = fs.readFileSync(path.join(__dirname, "bundle0.jsx"), "utf-8");
+  expect(bundle).toBe(`import { App, App1A, App1B, App1C, app1cProps } from "./App1";
```
ahabhgk: This will not easy to update the snapshot, you can use `toMatchFileSnapshot(path.join(__SNAPSHOT__, 'bundle0.jsx.txt'))` here

PR #11668, opened 2025-09-16
chore(refactor): inline base64 utilities into rspack_util

hardfist:
## Summary
base64 is just a small util and it seems no need to create a separate crate for it and align with other rust project that use common util crate to hold these util functions

## Related links
* https://github.com/rust-lang/rust-analyzer/tree/master/crates/stdx
* https://github.com/rust-lang/rust/tree/master/compiler/rustc_data_structures

## Checklist


quininer (approved):
LGTM

Issue #11595, opened 2025-09-05
[Bug]: persistent cache should refresh when node module change


Austaras:
### System Info

  System:
    OS: macOS 15.6.1
    CPU: (16) arm64 Apple M4 Max
    Memory: 6.07 GB / 48.00 GB
    Shell: 4.0.2 - /opt/homebrew/bin/fish
  Binaries:
    Node: 24.7.0 - /opt/homebrew/bin/node
    npm: 11.5.1 - /opt/homebrew/bin/npm
    pnpm: 10.12.4 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 139.0.7258.155
    Safari: 18.6
  npmPackages:
    @rspack/cli: ^1.5.2 => 1.5.2
    @rspack/core: ^1.5.2 => 1.5.2

### Details

When node_modules change, persistent cache should not be reused.

### Reproduce link

https://github.com/Austaras/repro-rspack-hmr-version-change

### Reproduce Steps

1. run `pnpm run dev`
2. browser console logs 178
3. run `pnpm add three@0.180.0`
4. browser console still logs 178

Austaras:
Caused by symlink created by pnpm. Could be work aroud with `resolve.symlinks = true`.

Austaras:
Fixed by #11643

PR #11693, opened 2025-09-17
fix: add symlink to file dep

Austaras:
## Related links

#11595

## Checklist

### DIFF
--- a/tests/rspack-test/cacheCases/invalidation/symlink/driver.js
+++ b/tests/rspack-test/cacheCases/invalidation/symlink/driver.js
@@ -1,0 +1,3 @@
+export default 0;
+---
+export default 1;
--- a/tests/rspack-test/cacheCases/invalidation/symlink/index.js
+++ b/tests/rspack-test/cacheCases/invalidation/symlink/index.js
@@ -1,0 +1,12 @@
+import value from "./driver";
+import version from "./tool"
+
+it("should invalidate work when symlink changes", async () => {
+	if (COMPILER_INDEX == 0) {
+		expect(version).toBe(100);
+		await NEXT_START();
+	}
+	if (COMPILER_INDEX == 1) {
+		expect(version).toBe(200);
+	}
+});
--- a/tests/rspack-test/cacheCases/invalidation/symlink/rspack.config.js
+++ b/tests/rspack-test/cacheCases/invalidation/symlink/rspack.config.js
@@ -1,0 +1,32 @@
+const path = require("path");
+const fs = require("fs/promises");
+
+let index = 1
+const libPath = path.join(__dirname, "./tool")
+/** @type {import("@rspack/core").Configuration} */
+module.exports = {
+	context: __dirname,
+	experiments: {
+		cache: {
+			type: "persistent",
+			snapshot: {
+				managedPaths: [path.join(__dirname, "./tool")]
+			}
+		}
+	},
+	plugins: [
+		{
+			apply(compiler) {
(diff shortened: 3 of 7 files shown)


Austaras:
I have tested locally to verfiy that it fixed the issue. But I fail to see how to add a proper test.

jerrykingxyz:
You can try this test case. https://github.com/web-infra-dev/rspack/commit/0ec163a1a5fea98a90858ba1a72de046aded211a

jerrykingxyz:
You can use it by running `npm run test:base -- -t "lib-symlink"` under `test/rspack-test`.
I suggest you create a new project and copy the code in the test to debug and verify, because of some logs may be lost in rspack_test.

Austaras:
> I suggest you create a new project and copy the code in the test to debug and verify

I figure it out myself. It's also easier to compare it with webpack.

crates/rspack_core/src/normal_module_factory.rs
```diff
@@ -586,6 +586,14 @@ impl NormalModuleFactory {
       .call(data, &mut create_data, &mut module)
       .await?;
 
+    let build_info = module.build_info_mut();
+    build_info
+      .file_dependencies
+      .extend(file_dependencies.iter().map(|f| f.clone().into()));
```
jerrykingxyz, line 592: `file_dependencies` belong to the current dependency, not the module. Setting this here is unreasonable. If the module already exists, the value of this part may even be discarded. The correct approach is to make factorizeInfo effective for all dependencies.

jerrykingxyz:
I think https://github.com/web-infra-dev/rspack/pull/11643 may can fix the bug. We can wait for the above PR to be merged.

crates/rspack_core/src/normal_module_factory.rs
Austaras: But [compilation](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/compilation/make/graph_updater/cutout/mod.rs#L61) calcuates modules that need rebuild using [buildInfo](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/module.rs#L399).

> The correct approach is to make factorizeInfo effective for all dependencies.

Could you elaborate on this?

Austaras: Ah, if I understand #11643 correctly, it could be possible to leverage `FactorizeInfo` for this scenario.

Austaras:
> I think #11643 may can fix the bug. We can wait for the above PR to be merged.

I've tested on #11643 and it cannot fix this issue. But I'm sure this PR could be updated upon #11643.

crates/rspack_core/src/normal_module_factory.rs
jerrykingxyz: BuildInfo is used to check if a module needs to be rebuilt, and factorizeInfo is used to check if a dependency needs to be refactorize. The dependencies you add to buildInfo should actually be added to factorizeInfo, and rebuilding the dependencies will also fix this problem.

Austaras: Yes, but it seems that repack only rebuild failed dependencies at the moment.

jerrykingxyz: yes, https://github.com/web-infra-dev/rspack/pull/11643 will fix it and check all dependencies factorizeinfo when rebuilding.

Austaras:
Superseded by #11643.

Issue #11691, opened 2025-09-17
[Bug]: panicked at crates/rspack_plugin_lazy_compilation/src/module.rs:250:10


SoonIter:
### System Info

@rspack/core 1.5.4 1.5.5

### Details

```js
// index.js
import('./hello').then((module) => {
  const hello = module.default;
  console.log(hello)
})
```

remove the `src/hello.js` and panicked at crates/rspack_plugin_lazy_compilation/src/module.rs:250:10
```sh
  Rsbuild v1.5.7

  ➜  Local:    http://localhost:3001/
  ➜  Network:  http://100.84.247.90:3001/
  ➜  press h + enter to show shortcuts

info    port 3000 is in use, using port 3001.
start   build started...
ready   built in 0.04 s
start   build started...
ready   built in 0.04 s
start   building removed src/hello.js
Panic occurred at runtime. Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues: panicked at crates/rspack_plugin_lazy_compilation/src/module.rs:250:10:
```

### Reproduce link

https://github.com/SoonIter/rsbuild-lazy-compilation-panic-issue

### Reproduce Steps

1. pnpm install
2. pnpm dev
3. remove src/hello.js and panicked at crates/rspack_plugin_lazy_compilation/src/module.rs:250:10

stormslowly:
close by https://github.com/web-infra-dev/rspack/pull/11643

PR #11739, opened 2025-09-23
fix(lazycompilation): lazy compilation proxied module's removing should rebuild proxing module

stormslowly:
## Summary

close https://github.com/web-infra-dev/rspack/issues/11691

### Problem 
```txt
src/
├── hello.js
└── index.js
```
```js
import("./hello").then((module) => {
  const hello = module.default;
  console.log(hello);
});
```
After all lazy proxing modules are actived, remove `hello.js` file.
Rspack panic at lazy proxing module's codegen due to accessing `hello.js` in module graph.

### Solution
Removing `hello.js` should trigger `index.js` rebuild, but lazy proxying module stop the propogation from the module removing; because it depends on nothing.

Actuall proxing module depends on proxied module's existence other than its content.
So add a new API to tell a module whether a module should build when some files removed.

## Checklist

### DIFF
--- a/crates/rspack_plugin_lazy_compilation/src/module.rs
+++ b/crates/rspack_plugin_lazy_compilation/src/module.rs
@@ -164,10 +164,17 @@ impl Module for LazyCompilationProxyModule {
     }
   }
 
-  /// Lazy compilation module do not depends on any files.
-  /// The only way to make this module rebuild is self.need_build() return true.
+  /// Lazy compilation module depends on proxied module's existence only.
+  /// The only way to make this module rebuild to changing its content is self.need_build() return true.
   fn depends_on(&self, _modified_file: &ArcPathSet) -> bool {
     false
+  }
+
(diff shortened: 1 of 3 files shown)


crates/rspack_plugin_lazy_compilation/src/module.rs
```diff
@@ -164,12 +164,19 @@ impl Module for LazyCompilationProxyModule {
...
+  /// The only way to make this module rebuild to changing its content is self.need_build() return true.
   fn depends_on(&self, _modified_file: &ArcPathSet) -> bool {
     false
   }
 
+  // Lazy compilation module should be removed when proxied module is removed.
+  fn depends_on_removed(&self, removed_files: &ArcPathSet) -> bool {
+    removed_files
+      .iter()
+      .any(|removed_file| self.resource == removed_file.to_string_lossy())
```
quininer, line 177: Can it be just `removed_files.contains(&self.resource)`?

PR #11744, opened 2025-09-23
feat(parser): add `commonjs` options

fi3ework:
## Summary

The initial motivation was that when targeting ESM output, if the user wrote statements containing the `module` variable, such as `module.ok = 1`, `module` would be specially processed (`CommonJsExportsParserPlugin`).

However, the expectation is to retain `module` without processing it. I previously tried to use the same hook processing as `CommonJsExportsParserPlugin` in the Rslib parser plugin to shield it, but because the Rslib plugin is at the very front of all parser plugins, it would affect the logic of other necessary built-in plugins.

Since we currently do not support inserting parser plugins at arbitrary positions, after discussing with @JSerFeng, we decided to directly disable `CommonJsExportsParserPlugin` when `is_esm` is `true`, because when ESM syntax exists, there is no need to process CJS exports.

## Checklist


crates/rspack_plugin_javascript/src/parser_plugin/common_js_exports_parse_plugin.rs
```diff
@@ -265,6 +275,10 @@ impl JavascriptParserPlugin for CommonJsExportsParserPlugin {
     remaining: &[Atom],
     for_name: &str,
   ) -> Option<bool> {
+    if Self::should_suppress_commonjs_exports_in_esm(parser) {
```
ahabhgk: You don't need add this for every hook, actually you can add the condition at `JavascriptParser::new`, if it's false then don't add the plugin, for example: https://github.com/web-infra-dev/rspack/blob/2bf9b039da7f910191e2b0702aaffd4b05bf7d0f/crates/rspack_plugin_javascript/src/visitors/dependency/parser/mod.rs#L338-L344

crates/node_binding/napi-binding.d.ts
```diff
@@ -2317,6 +2317,11 @@ export interface RawJavascriptParserOptions {
    * @experimental
    */
   requireResolve?: boolean
+  /**
+   * This option is experimental in Rspack only and subject to change or be removed anytime.
+   * @experimental
+   */
+  suppressCommonjsExportsInEsm?: boolean
```
ahabhgk: Personally prefer just `commonjsExports`, BTW webpack has a `commonjs` parser options https://github.com/webpack/webpack/blob/cf17a4125972781bee4636e37d8ece29669523c7/lib/dependencies/CommonJsPlugin.js#L181-L237 maybe worth as reference for naming

crates/rspack_plugin_javascript/src/parser_plugin/common_js_exports_parse_plugin.rs
fi3ework: i'm afraid can't do it like this, as `parser.is_esm` is determined in https://github.com/web-infra-dev/rspack/blob/1682a94559cde7a78b4ffd2d927fc0c7af531f84/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk_pre.rs#L36-L50 `pre_walk_module_item` stage, which is after the `plugins` vec initialization.

crates/node_binding/napi-binding.d.ts
fi3ework: what about `commonjsExportsInEsm`? if naming to `commonjsExports` and being set to `false`, CJS exports will still work in modules that without ESM syntax. this option is dedicated to disable CJS exports when using ESM syntax (for Rslib).

ahabhgk: Maybe just `commonjs: boolean | { onlyInEsm: boolean, require: boolean, exports: boolean }`

packages/rspack/src/schema/config.ts
```diff
@@ -682,6 +683,7 @@ export const getRspackOptionsSchema = memoize(() => {
 			requireAsExpression: requireAsExpression,
 			requireDynamic: requireDynamic,
 			requireResolve: requireResolve,
+			suppressCommonjsExportsInEsm: suppressCommonjsExportsInEsm,
```
chenjiahan: Should we update documentation to include this option?

crates/node_binding/napi-binding.d.ts
fi3ework: sounds good to me, Rspack lacks `commonjs` field in parser options now which can be added alongside.

what are the uses of `require: boolean` and `exports: boolean`? Is there any logic that utilizes these?

packages/rspack/src/schema/config.ts
fi3ework: i'm inclined not to add it to the documentation as it's only for Rslib and rarely used in application, also it's not battle tested enough that may change in the future.

@ahabhgk mentioned the [`commonjs`](https://github.com/webpack/webpack/blob/c647cf193cb4033847971aa7c8a89da6cecb740c/declarations/WebpackOptions.d.ts#L3203) field, which does not exist in Rspack now. If we end up expanding webpack's original `boolean` type like the comment above suggests, I think it would be necessary for us to write out the complete type definition, because `commonjs` is also a parser option that webpack has.

crates/node_binding/napi-binding.d.ts
ahabhgk: `require: boolean` controls `CommonJsImportsParserPlugin` and `exports: boolean` controls `CommonJsExportsParserPlugin`, for rslib use case you can use `commonjs: { exportsInEsm: false, requireInEsm: true, exports: true, require: true }` (after a second thought `onlyInEsm` can't fulfill the needs)

fi3ework: to make the options more clear, what about: 

- `exports` (default to `true`): `CommonJsImportsParserPlugin`
- `require` (default to `true`): `CommonJsExportsParserPlugin`
- `enableInEsm` (default to `true`)`: 
   - `true`: the default behavior
   - `false`: bailout the hook if `is_esm` is true when `exports` / `require` are enabled.

ahabhgk: With `enableInEsm` how to write:
- require is enabled when `is_esm: true` and `is_esm: false`
- exports is enabled when `is_esm: false` but disable when `is_esm: false`

It's 2 * 2 combination scenarios, so we either have four `boolean` options (`commonjs: { exportsInEsm: false, requireInEsm: true, exports: true, require: true }`) or two `boolean | "suppressInEsm"` options

fi3ework: i'd prefer 

```js
commonjs: boolean | {
  exports: boolean | 'inNonEsm' // deafults to `true`
  require: boolean | 'inNonEsm' // deafults to `true`
}
```

to have a fine-grained control, and currently as we don't need `require` so it can be omitted as of now. the naming `inNonEsm` could be improved.

ahabhgk: This looks good to me

fi3ework: i think `skipInEsm` is better than `inNonEsm`, will implement it now.

PR #11757, opened 2025-09-24
refactor: file counter save the resource id which used the file

jerrykingxyz:
## Checklist


crates/rspack_core/src/utils/file_counter/resource_id.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{DependencyId, ModuleIdentifier};
+
+/// The resource using the path at file counter.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
```
SyMind, line 4: The hash calculation for ResourceId can be optimized. When it is a Module, the hash of ustr is directly used.

For Dependency, the fx hasher is used for calculation.

Then, HashSet and HashMap use the IdentityHasher.

jerrykingxyz: I am not sure whether this operation will cause hash conflicts. I will ignore it for now as there is no performance bottleneck. I can optimize it later through cpu perf.

PR #11779, opened 2025-09-28
feat(mf): inject remoteInfos to support lazy compilation

2heal1:
## Summary

Inject remotesInfos into `__webpack_require__.federation.bundlerRuntimeOptions.remotes` to help MF support lazy compilation . 

MF will update the remote chunkMapping and moduleMap with the updated module data . And it needs the `remotesInfos` which is generated at first build, so we need to inject it into bundler runtime to store it .

## Related links
https://github.com/module-federation/core/pull/4095

## Checklist

### DIFF
--- a/packages/rspack/src/runtime/moduleFederationDefaultRuntime.js
+++ b/packages/rspack/src/runtime/moduleFederationDefaultRuntime.js
@@ -161,6 +161,11 @@ module.exports = function () {
 		);
 		early(
 			__webpack_require__.federation.bundlerRuntimeOptions.remotes,
+			"remoteInfos",
+			() => __module_federation_remote_infos__
+		);
+		early(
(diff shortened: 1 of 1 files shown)


2heal1:
I will update the official MF version if the main logic is acceptable

packages/rspack/src/container/ModuleFederationPlugin.ts
```diff
@@ -40,6 +41,24 @@ export class ModuleFederationPlugin {
 			...this._options,
 			enhanced: true
 		}).apply(compiler);
+
+		// inject ensureChunkHandlers if enable lazy compilation
+		if (compiler.options.lazyCompilation && this._options.remotes) {
```
ahabhgk: Why inject runtime here? runtime should correctly injected automatically

PR #11796, opened 2025-09-30
fix: revert "fix: remove `serde` feature of `lightningcss` (#11706)"

colinaaa:
## Summary

This partially reverts commit db3cf1b7a3ab1455df52d1d32ecc4b01b273dcc8, with `lightningcss` crate bumping to 1.0.0-alpha.68.

Since `lightningcss` is now supporting `serde` >= 1.0.220, we can use the `serde` instead of serializing manually.

## Related links

- https://github.com/parcel-bundler/lightningcss/pull/1062

## Checklist


quininer (approved):
I think keep it is good for compile time. but if that is not important, revert it is also good.

PR #11819, opened 2025-10-09
refactor: remove experiments.layers

JSerFeng:
## Summary

Remove `experiments.layers` which is used to enable the `layers` feature, now we enable this by default. Now users only need to specify layer in `entry` or `moduleRule`, no need to enable it in `experiments`.

If you use this config before, just delete it will be fine

```diff
export default {
  experiments: {
-  layers: true
  }
}
```

## Checklist

### DIFF
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -218,8 +218,11 @@ const applyExperimentsDefaults = (
 	D(experiments, "lazyCompilation", false);
 	D(experiments, "asyncWebAssembly", experiments.futureDefaults);
 	D(experiments, "css", experiments.futureDefaults ? true : undefined);
-	D(experiments, "layers", false);
-
+	if ("layers" in experiments) {
+		console.warn(
+			"`experiments.layers` config has been deprecated and will be removed in Rspack v2.0. Feature layers will be always enabled. Please remove this option from your Rspack configuration."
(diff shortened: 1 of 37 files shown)


crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -93,7 +93,7 @@ pub struct ModuleGraphPartial {
 
 #[derive(Debug, Default)]
 pub struct ModuleGraph<'a> {
-  partials: [Option<&'a ModuleGraphPartial>; 2],
+  pub partials: [Option<&'a ModuleGraphPartial>; 2],
```
ahabhgk:
```suggestion
  partials: [Option<&'a ModuleGraphPartial>; 2],
```

keep it private

packages/rspack/src/config/defaults.ts
```diff
@@ -218,7 +218,6 @@ const applyExperimentsDefaults = (
 	D(experiments, "lazyCompilation", false);
 	D(experiments, "asyncWebAssembly", experiments.futureDefaults);
 	D(experiments, "css", experiments.futureDefaults ? true : undefined);
-	D(experiments, "layers", false);
```
ahabhgk, line 221: Just delete the config probably will break some users, do we need to delete the config more progressively like topLevelAwait?

PR #11840, opened 2025-10-11
feat: rslib supports add shims for js/esm

JSerFeng:
## Summary

Rslib supports add shims for javascript/esm, register the NodeStuff parser plugin if user uses shims

## Checklist


fi3ework:
cc @Timeless0911, a breaking change when bumping Rsbuild in Rslib side.

PR #11846, opened 2025-10-11
feat(mf): support manifest

2heal1:
## Summary

Ship MF manifest to rust, aims to speed up the build time which will be much slower in `stats.toJson()` in js side .

[MF manifest](https://module-federation.io/configure/manifest.html) is an important feature in MFV2 , it provides enough context to allow framework to extend features, like SSR/RSC/Server Preload.  

## Checklist

### DIFF
--- a/packages/rspack/src/container/ModuleFederationPlugin.ts
+++ b/packages/rspack/src/container/ModuleFederationPlugin.ts
@@ -54,7 +54,7 @@ export class ModuleFederationPlugin {
 			enhanced: true
 		}).apply(compiler);
 
-		if (this._options.manifest !== false) {
+		if (this._options.manifest) {
 			const manifestOptions: ModuleFederationManifestPluginOptions =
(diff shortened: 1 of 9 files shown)


packages/rspack/src/container/ModuleFederationPlugin.ts
```diff
@@ -38,6 +53,49 @@ export class ModuleFederationPlugin {
 			...this._options,
 			enhanced: true
 		}).apply(compiler);
+
+		if (this._options.manifest !== false) {
```
ahabhgk: Does this mean now user that uses rspack.container.ModuleFederationPlugin will also have a manifest by default? If yes then this should change to disabled by default, we should not give surprise to user, but this can be part of Rspack v2

2heal1: got, i will make it as false by default , and enable it in Rspack v2

PR #11899, opened 2025-10-16
feat(mf): add async startup promise gating for entrypoints

ScriptedAlchemy:
## Summary
- Add `experiments.asyncStartup` support for Module Federation by embedding federation runtime deps and exposing a guarded async startup hook.
- Gate **non-container entrypoints** on the async federation startup promise to avoid races with remote/consume initialization.
- Keep **container/remoteEntry** bundles synchronous (preserve `get`/`init` exports) while still running the embedded federation runtime setup.
- Fix hoisting of container references to the correct runtime chunk for `runtimeChunk: "single"`/multi-runtime.

## Notes
- No behavior change when `experiments.asyncStartup` is disabled.
- Container chunks (exposes) intentionally do **not** block startup on the async hook; they only run the base federation runtime setup to keep the container export sync.

## Tests
- Added MF async-startup config/serial test coverage (remoteEntry export shape, runtimeChunk single, multi-entry).
- CI: all required checks green.


ScriptedAlchemy:
@codex review pr

crates/node_binding/napi-binding.d.ts
```diff
@@ -2470,8 +2470,14 @@ export interface RawModuleFederationManifestPluginOptions {
...
 }
 
+export interface RawModuleFederationRuntimeExperimentsOptions {
+  asyncStartup?: boolean
+}
+
 export interface RawModuleFederationRuntimePluginOptions {
   entryRuntime?: string | undefined
+  experiments?: RawModuleFederationRuntimeExperimentsOptions
+  asyncStartup?: boolean
```
ahabhgk: duplicate with experiments.asyncStartup?

crates/rspack_plugin_javascript/src/plugin/mod.rs
```diff
@@ -619,14 +624,36 @@ var {} = {{}};
...
-          startup.push(
-            format!(
-              "var {} = {}();",
-              runtime_template.render_runtime_globals(&RuntimeGlobals::EXPORTS),
-              runtime_template.render_runtime_globals(&RuntimeGlobals::STARTUP)
-            )
-            .into(),
-          );
+          if has_expose_entry
+            && runtime_requirements.contains(RuntimeGlobals::ASYNC_FEDERATION_STARTUP)
```
ahabhgk: The mf specific logic shouldn't place at here

crates/rspack_plugin_library/src/module_library_plugin.rs
```diff
@@ -76,11 +76,15 @@ async fn render_startup(
...
+  let runtime_requirements = ChunkGraph::get_tree_runtime_requirements(compilation, chunk_ukey);
   let module_graph = compilation.get_module_graph();
   source.add(render_source.source.clone());
   // export { local as exported }
   let mut exports: Vec<(String, Option<String>)> = vec![];
-  if is_async {
+  // In async federation startup mode, STARTUP may return a Promise even when the entry module is
+  // sync (because it awaits federation init). Ensure we await the startup result before reading
+  // export fields so module outputs like MF containers still export live functions (get/init).
+  if is_async || runtime_requirements.contains(RuntimeGlobals::ASYNC_FEDERATION_STARTUP) {
```
ahabhgk: Can we just reuse the is_async? If user enable the asyncStartup, then we mark the mf entry as an async module

crates/rspack_plugin_runtime/src/array_push_callback_chunk_format.rs
```diff
@@ -53,7 +53,11 @@ async fn additional_chunk_runtime_requirements(
     .get_number_of_entry_modules(chunk_ukey)
     > 0
   {
-    runtime_requirements.insert(RuntimeGlobals::ON_CHUNKS_LOADED);
+    // When async federation startup is requested, STARTUP_ENTRYPOINT will be present and
+    // the MF wrapper handles startup; avoid adding ON_CHUNKS_LOADED in that case.
+    if !runtime_requirements.contains(RuntimeGlobals::ASYNC_FEDERATION_STARTUP) {
```
ahabhgk: Here is the same, I don't want the mf specific logic affects rspack_plugin_runtime and rspack_plugin_javascript

crates/rspack_plugin_runtime/src/startup_chunk_dependencies.rs
```diff
@@ -32,6 +32,10 @@ async fn runtime_requirements_in_tree(
   runtime_requirements_mut: &mut RuntimeGlobals,
 ) -> Result<Option<()>> {
   let is_enabled_for_chunk = is_enabled_for_chunk(chunk_ukey, &self.chunk_loading, compilation);
+  let mut async_chunk_loading = self.async_chunk_loading;
+  if runtime_requirements.contains(RuntimeGlobals::ASYNC_FEDERATION_STARTUP) {
```
ahabhgk: I think there is a more general design for this:

Add a new `RuntimeGlobals::ASYNC_STARTUP` to replace `RuntimeGlobals::ASYNC_FEDERATION_STARTUP`, and for each ChunkLoadingPlugin (JsonpChunkLoadingPlugin, ImportScriptsChunkLoadingPlugin...), if the chunk loading is async, add `RuntimeGlobals::ASYNC_STARTUP` at their `additional_tree_runtime_requirements` hook

crates/rspack_plugin_runtime/src/helpers.rs
```diff
@@ -260,9 +260,19 @@ pub fn generate_entry_startup(
...
   if chunks_ids.is_empty() {
     if !module_ids_code.is_empty() {
-      source.push_str(&format!("var {exports_name} = ("));
-      source.push_str(module_ids_code);
-      source.push_str(");\n");
+      if passive {
+        source.push_str(&format!("var {exports_name} = ("));
+        source.push_str(module_ids_code);
+        source.push_str(");\n");
+      } else {
```
ahabhgk, line 267: Not sure is this else branch is needed, since the `module_ids_code` is already empty, I'm ok with this change if this really have effect

crates/rspack_plugin_runtime/src/startup_chunk_dependencies.rs
ScriptedAlchemy: okay! i was trying to limit the amount of modifications across rspack, but thought of doing this initially, will adjust all plugins directly instead.

Issue #11903, opened 2025-10-17
[Bug]: swc env_name is empty in swc plugin metadata context


Zeahan:
### System Info

 System:
    OS: macOS 14.4
    CPU: (12) arm64 Apple M3 Pro
    Memory: 100.22 MB / 18.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.19.5
    pnpm: 10.15.1
  Browsers:
    Chrome: 127.0.6533.100
    Firefox: 140.0.2
    Safari: 17.4
  npmPackages:
    @rspack/cli: 1.3.1 => 1.3.1 
    @rspack/core: 1.3.1 => 1.3.1 
    @swc/plugin-emotion: 9.0.3

### Details

When using `@swc/plugin-emotion` in some rspack-based projects (including both Modern.js and Rsbuild projects), I found that the plugin’s configuration options—specifically the default values `autoLabel: 'dev-only' `and `sourceMap: true`—do not take effect as expected.

After reviewing the documentation and source code of `@swc/plugin-emotion`, I discovered that the plugin determines the current environment (development or production) by calling `TransformPluginProgramMetadata::get_context(&TransformPluginMetadataContextKind::Env)` to retrieve the environment name (`env_name`) from swc. This environment information is used to control different build behaviors.

To investigate further, I implemented a simple SWC plugin as a demo to print out the environment information from the metadata. I found that the environment field is an empty string. As a result, any SWC plugins that rely on this field are unable to correctly identify the current build environment, leading to incorrect or unexpected behavior.

### Reproduce link

https://github.com/Zeahan/rspack-repro-swc-env-name-error/tree/main

### Reproduce Steps

1. Run `pnpm install`
2. Check if the demo plugin is correctly built in the `packages/plugin-demo/target` directory
3. Run pnpm build in the project root directory and observe the env name information printed by the demo plugin in the console

PR #11906, opened 2025-10-17
fix(swcLoader): should correctly generate the default value of env_na…

Zeahan:
…me in swc options (#11903)

## Summary

In the previous implementation, the remaining fields of the `Options` struct (`swc_core::base::config::Options`) were initialized using `..Default::default()`. However, this approach can result in some fields having default values that do not match the expected behavior. The root cause is that the fields in Options are defined with serde macros, and custom logic is used to generate default values during deserialization.

```rust
pub(crate) fn default_env_name() -> String {
    if let Ok(v) = env::var("SWC_ENV") {
        return v;
    }

    match env::var("NODE_ENV") {
        Ok(v) => v,
        Err(_) => "development".into(),
    }
}

pub struct Options {
    #[serde(default = "default_env_name")]
    pub env_name: String,

    #[serde(default = "default_swcrc")]
    pub swcrc: bool,

    #[cfg(not(all(target_arch = "wasm32", not(target_os = "wasi"))))]
    #[serde(default = "default_cwd")]
    pub cwd: PathBuf,
    // ...
}
```

When initializing `Options` with `..Default::default()`, this custom logic is bypassed, leading to discrepancies between the generated fields and the original behavior in `swc-loader`. This inconsistency may cause certain swc plugins (such as `@swc/plugin-emotion`) to behave unexpectedly.

To address this issue and ensure consistency with the original swc implementation, the initialization of the remaining fields in Options has been changed to use deserialization from an empty JSON object via `serde_json` (which is the approach used in the swc source code). This ensures that the configuration is generated in a manner consistent with the original swc behavior.

## Related links
https://github.com/web-infra-dev/rspack/issues/11903

## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/options.rs
+++ b/crates/rspack_loader_swc/src/options.rs
@@ -187,9 +187,66 @@ impl TryFrom<&str> for SwcCompilerOptionsWithAdditional {
           schema,
           source_map_ignore_list,
         },
-        ..Default::default()
+        ..serde_json::from_value(serde_json::Value::Object(Default::default()))?
       },
       rspack_experiments: rspack_experiments.unwrap_or_default().into(),
     })
+  }
+}
+
+#[cfg(test)]
+mod tests {
+  use super::*;
+
+  #[test]
+  fn test_specially_default_values_in_swc_options() {
+    // Verifies that fields using `#[serde(default = "...")]`
+    // receive the correct default value from the initializations of
+    // both SwcCompilerOptionsWithAdditional and swc_core::base::config::Options.
+    let raw_options = "{}";
+    let swc_options_with_additional: SwcCompilerOptionsWithAdditional = raw_options
+      .try_into()
+      .expect("Parse SwcCompilerOptionsWithAdditional from empty JSON string failed");
+
+    let swc_options_from_rspack: Options = swc_options_with_additional.swc_options;
+
+    let swc_options_from_native_lib: Options =
+      serde_json::from_value(serde_json::from_str(raw_options).unwrap()).unwrap();
+
+    assert_eq!(
+      swc_options_from_rspack.env_name,
+      swc_options_from_native_lib.env_name
+    );
+    assert_eq!(swc_options_from_rspack.cwd, swc_options_from_native_lib.cwd);
+    assert_eq!(
+      swc_options_from_rspack.swcrc,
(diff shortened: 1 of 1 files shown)


CPunisher:
Could you please add a test case?

Zeahan:
> Could you please add a test case?

Thanks for your reply, a test case will be provided later.

Zeahan:
> Could you please add a test case?

@CPunisher rust test cases have been provided. 
Since the complete rspack build process requires an additional custom SWC plugin for testing, Node.js test cases are not implemented, sorry for that.

Zeahan:
@CPunisher Hi, I’ve tried running CI multiple times in my forked repo after syncing my PR branch with the main repo, but I noticed that the test Serial.test.js always fails.
截屏2025-10-21 17 58 40
Then, I tried running CI based on the latest version of the main branch ([a457128](https://github.com/web-infra-dev/rspack/commit/a457128f4329f224371c3396612e8e2109f610f4)) and encountered similar test errors.

I realized that these errors might be caused by another PR rather than mine, so I rolled back the base of my PR branch to [0f38cf19](https://github.com/web-infra-dev/rspack/commit/0f38cf19), and the CI finally passed in my forked repo.
截屏2025-10-21 18 15 29
I’ve now force-pushed the updated PR branch. Would you mind approving the CI workflows again?

CPunisher:
> ild process requires an additional custom SWC plugin for testing, Node.js test cases are not implemented, sorry for that.

😂 Sorry, recently our ci is not stable.

Zeahan:
@CPunisher Adding `default_env_name` is enough to solve this issue—that was my first thought as well. However, with this approach, some default field values(such as `cwd` and `swcrc`) would still be different from the original SWC behavior.

So in the end, I decided to copy this part of the code directly from the SWC repo 😂(https://github.com/swc-project/swc/blob/4848301edca71ed84d147e90baccae217b0a8938/bindings/binding_core_node/src/bundle.rs#L208), hoping to offer a more complete solution rather than just fixing the `env_name` problem. But of course, I’m happy to go with whichever approach your team prefers! 😁

CPunisher:
Yes, I have realized your consideration. I will merge this pr after we stablize CI. 

Thanks for your contribution!

CPunisher (approved):

PR #11914, opened 2025-10-18
fix(core): throw error when passing unsupported Node.js version browserslist query

hardfist:
## Summary
throw error when passing unsupported browserslist query 

## Related links
https://github.com/facebook/docusaurus/pull/11496

## Checklist


slorber:
Hey

I was wondering, has this issue been solved in a different way?

I'm under v1.7 and it doesn't seem to emit any error/warning atm:

CleanShot 2026-05-22 at 13 15 04@2x

PR #11920, opened 2025-10-20
perf: don't drop rust compiler instance in non-watch mode

hardfist:
## Summary
It's slow and unnecessary to drop the Compiler & Compilation in non-watch mode, so support to ignore drop of Compiler & Compilation instance in non-watch mode

img_v3_02r8_bfcdbf08-bf59-4d3b-bfd5-158e9fd89eeg

## Checklist


SyMind:
should we need move the option into experiments?

SyMind (approved):

PR #11927, opened 2025-10-20
feat: introduce allocative

quininer:
## Summary

This introduces `allocative` to debug memory leaks caused by global variables. when enabled, a global variable memory snapshot will be generated each rebuild.

This requires enable the feature at build time using env `ALLOCATIVE=1` and specify the output directory at runtime use env `RSPACK_ALLOCATIVE_DIR="$DIR"`.

The output can then be visualized as svg.

to svg: `cat allocative.0.flamegraph | inferno-flamegraph > 0.svg`
diff: `inferno-diff-folded allocative.0.flamegraph allocative.1.flamegraph | inferno-flamegraph > diff0v1.svg`

## Checklist

### DIFF
--- a/crates/rspack_cacheable/Cargo.toml
+++ b/crates/rspack_cacheable/Cargo.toml
@@ -10,6 +10,7 @@ version.workspace = true
 noop = []
 
 [dependencies]
+allocative      = { workspace = true, optional = true }
 camino          = { workspace = true }
 dashmap         = { workspace = true }
 hashlink        = { workspace = true }
@@ -27,3 +28,7 @@ smol_str        = { workspace = true }
 swc_core        = { workspace = true, features = ["ecma_ast"] }
 ustr            = { workspace = true }
 xxhash-rust     = { workspace = true, features = ["const_xxh64"] }
+
+[lints.rust.unexpected_cfgs]
+level = "warn"
+check-cfg = ['cfg(allocative)']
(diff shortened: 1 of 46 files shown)


crates/rspack_cacheable/Cargo.toml
```diff
@@ -10,6 +10,7 @@ version.workspace = true
 noop = []
 
 [dependencies]
+allocative      = { workspace = true }
```
hardfist: is allocative optional dependency?

quininer: I will modify it tomorrow and use both `feature` and `cfg` control.

crates/rspack_core/src/compiler/rebuild.rs
```diff
@@ -190,6 +190,38 @@ impl Compiler {
     self.compile_done().await?;
     self.cache.after_compile(&self.compilation).await;
 
+    #[cfg(allocative)]
```
hardfist, line 193: do we need add this to build_inner?

quininer: This focuses on memory usage during rebuild, and I think logging only during rebuild is sufficient. but there's nothing wrong with logging during build, so I'll add it.

hardfist: if we add both to build and rebuild maybe move it to init(like sftrace) is better?

hardfist (approved):

PR #11944, opened 2025-10-21
fix: perf regression caused by is_deferred

ahabhgk:
## Summary

We can record the `has_imported_by_defer` at make, and avoid iterate on incoming connections

## Checklist

### DIFF
--- a/packages/rspack-test-tools/src/helper/legacy/asModule.js
+++ b/packages/rspack-test-tools/src/helper/legacy/asModule.js
@@ -16,8 +16,7 @@ module.exports = async (something, context, unlinked) => {
 	const code = [...new Set(["default", ...Object.keys(something)])]
 		.map(
 			name =>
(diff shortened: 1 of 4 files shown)


crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -959,22 +959,7 @@ impl<'a> ModuleGraph<'a> {
     let module = self
       .module_by_identifier(module_id)
       .expect("should have module");
-    if module.build_meta().has_top_level_await {
-      return false;
-    }
-    for connection in self.get_incoming_connections(module_id) {
```
hardfist, line 965: if some connection is changed by following optimization whether could it affect is_defer result

ahabhgk: Yes, I'm changing to another solution

PR #11961, opened 2025-10-22
chore(ci): improve window cache speed with zstd

stormslowly:
## Summary

Now we install zstd, which is a faster compress/decompress program, 
We can benefit by caching and reducing the Windows build by about 1 minute.

## Checklist


chenjiahan (approved):
🚀

PR #11985, opened 2025-10-23
docs: add layer feature guide doc

JSerFeng:
## Summary

Add docs of how to use layer

## Checklist


chenjiahan (commented):
Could we add some links:

- layer guide -> layer configurations
- layer configurations -> layer guide

website/docs/en/guide/features/layer.mdx
```diff
@@ -0,0 +1,117 @@
...
+- Splitting modules from different layers into different output directories
+- Applying different optimization strategies for code in different environments (such as server-side and client-side)
+
+## Ways to specify layers
+
+You can specify layers for modules in the following ways:
+
+### 1. Specify via entry
+
+Directly specify layers for different entry points in the entry configuration:
```
chenjiahan: Could you make this part more detailed? Does the layer affect only the entry module, or all the modules imported by the entry module as well?

website/docs/en/guide/features/layer.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+```js
```
chenjiahan:
```suggestion
```js title="rspack.config.mjs"
```

website/docs/en/guide/features/layer.mdx
```diff
@@ -0,0 +1,117 @@
...
+      import: './src/client.js',
+      layer: 'client',
+    },
+  },
+};
+```
+
+### 2. Specify via loader rules
+
+You can use the `rule.layer` option to specify a layer for all modules that match the rule:
```
chenjiahan: If `entry.layer` and `rule.layer` are set at the same time, which one takes precedence?

JSerFeng: rule.layer, I'll point it out in the doc

chenjiahan (approved):
LGTM 👍

PR #11993, opened 2025-10-23
fix: use asset resource module as entry

LingyuCoder:
## Summary

Should not generate js file if the entry is a asset module with `asset/resource`.

## Checklist

### DIFF
--- a/tests/rspack-test/configCases/asset-modules/only-entry/test.filter.js
+++ b/tests/rspack-test/configCases/asset-modules/only-entry/test.filter.js
@@ -1,1 +1,0 @@
-module.exports = () => true;


tests/rspack-test/configCases/errors/multi-entry-missing-module/test.config.js
```diff
@@ -1,8 +1,9 @@
 module.exports = {
-	findBundle: function() {
+	findBundle: function () {
 		return [
-			"./a.js",
-			"./b.js",
+			// FIXME: the entry depenedency should generate module even when it is ignored
```
LingyuCoder, line 4: This is a bug of ignore plugin that the entry dependency should also generate a module which only have ignoring comments .

crates/rspack_plugin_runtime/src/helpers.rs
```diff
@@ -234,9 +240,11 @@ pub fn generate_entry_startup(
     .collect::<Vec<_>>()
     .join(", ");
   if chunks_ids.is_empty() {
-    source.push_str("var __webpack_exports__ = (");
-    source.push_str(module_ids_code);
-    source.push_str(");\n");
+    if !module_ids_code.is_empty() {
```
LingyuCoder, line 243: should not generate `var __webpack_exports__ = ()` if no module is need to load in startup

JSerFeng (approved):

PR #12012, opened 2025-10-28
feat: builtin react server component

SyMind:
## Summary

This PR adds built-in RSC plugins to Rspack. We recommend using them via the Rsbuild plugin wrapper available at https://github.com/rstackjs/rsbuild-plugin-rsc.

For usage examples, please refer to: https://github.com/rstackjs/rsbuild-plugin-rsc/tree/main/examples

The `builtin:swc-loader` has been enhanced to identify `"use client"` and `"use server"` directives. Within the RSC layer, it utilizes the `react-server-dom-rspack` package to transform these modules into the appropriate client and server references required by React.
- Note: This feature must be enabled via rspackExperiments.reactServerComponents.

RSC requires separate compilation for Server and Client bundles. This PR introduces `ServerPlugin` and `ClientPlugin` (created via `rsc.createPlugins()`), which orchestrate the two compilers to:
- Collect and compile modules with RSC directives.
- Generate the necessary manifests used by the RSC runtime.

The `ServerPlugin` includes an `onServerComponentChanges` callback. This is triggered whenever the Server Compiler detects changes in a Server Component.

## Checklist


Pablinho:
Hi @SyMind is there any ETA for this feature? Thanks in advance for the aswer.

SyMind:
@Pablinho I plan to expose this as an experimental feature in late December or early January.

This week or next, I will lay out the specific design in the Rspack discussion.

crates/rspack_plugin_rsc/Cargo.toml
```diff
@@ -0,0 +1,44 @@
...
+serde_json               = { workspace = true }
+simd-json                = { workspace = true }
+swc_core                 = { workspace = true }
+tokio                    = { workspace = true }
+tracing                  = { workspace = true }
+urlencoding              = { workspace = true }
+ustr                     = { workspace = true }
+
+[package.metadata.cargo-shear]
+ignored = ["tracing", "rspack_hash", "ustr-fxhash"]
```
chenjiahan: Is this config necessary?

crates/rspack_binding_api/src/plugins/rsc.rs
```diff
@@ -0,0 +1,114 @@
...
+use rspack_core::{Compiler, CompilerId};
+use rspack_error::ToStringResultToRspackResultExt;
+use rspack_plugin_rsc::{Coordinator, RscClientPluginOptions, RscServerPluginOptions};
+
+use crate::JsCompiler;
+
+type InvalidateTsFn = Arc<ThreadsafeFunction<(), (), (), Status, false, true, 0>>;
+
+#[napi]
+pub struct JsCoordinator {
```
hardfist, line 22: can you add comment about this struct?

crates/rspack_plugin_rsc/Cargo.toml
SyMind: Yes, if not add, cargo shear will throw error.

crates/rspack_core/src/compiler/mod.rs
```diff
@@ -200,11 +206,31 @@ impl Compiler {
     self.build().await?;
     Ok(())
   }
+
   pub async fn build(&mut self) -> Result<()> {
     let compiler_context = self.compiler_context.clone();
-    within_compiler_context(compiler_context, self.build_inner()).await?;
-    Ok(())
+    match within_compiler_context(compiler_context, self.build_inner()).await {
```
hardfist, line 212: why add this here? is it a previous bug that we don't call this hooks?

crates/rspack_binding_api/src/plugins/rsc.rs
SyMind: I have add comment in crates/rspack_plugin_rsc/src/coordinator.rs

crates/rspack_core/src/compiler/mod.rs
SyMind: In RSC Plugin, we need to track the complete lifecycle of both server compiler and client compiler on the Rust side, so we need to add done and failed hooks to confirm the end of the lifecycle.

SyMind: We are not have done and failed hook in Rust side.

crates/rspack_loader_swc/src/rsc_transforms/cjs_finder.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+        if &*i.id.sym == "exports" {
+          self.ignore_exports = true;
+        }
+      }
+    })
+  }
+}
+
+/// This visitor implementation supports typescript, because the api of `swc`
```
hardfist, line 44: what does this mean?  does it mean this visitor needs to handle typescript syntax directly?

crates/rspack_loader_swc/src/rsc_transforms/server_actions.rs
```diff
@@ -0,0 +1,2669 @@
...
+
+#[derive(Clone)]
+struct ServerReferenceExport {
+  ident: Ident,
+  export_name: ModuleExportName,
+  reference_id: Atom,
+}
+
+#[derive(Clone, Debug)]
+enum ServerActionsErrorKind {
```
hardfist, line 58: can you add comments about what related synatx does the following kind refers to

crates/rspack_loader_swc/src/rsc_transforms/cjs_finder.rs
SyMind: This comment means the visitor needs to handle both JavaScript and TypeScript files.

This code is ported from Next.js, and I'm preserved the original comments for reference.

crates/rspack_loader_swc/src/rsc_transforms/server_actions.rs
SyMind: you can get the details in `emit_error` function in this file.

crates/rspack_plugin_rsc/Cargo.toml
chenjiahan: I did not find `ustr-fxhash` related code

packages/rspack/src/Compiler.ts
```diff
@@ -131,6 +131,8 @@ export type CompilerHooks = {
   additionalPass: liteTapable.AsyncSeriesHook<[]>;
 };
 
+export const GET_COMPILER_ID = Symbol('getCompilerId');
```
hardfist, line 134: this is a public api ?

SyMind: private API, just used in RSC Plugin. It can not be got out @rspack/core

hardfist:
is the codspeed performance regression expected or just flaky?

SyMind:
> is the codspeed performance regression expected or just flaky?
@hardfist just flaky

crates/rspack_plugin_rsc/Cargo.toml
SyMind: it is in root Cargo.tom. maybe it is a cargo-shear bug? I do not know.
```
ustr = { package = "ustr-fxhash", version = "1.0.1", default-features = false }
```

SyMind: My mistake, the current RSC plugin does not using ustr.

crates/rspack_loader_swc/src/rsc_transforms/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+  loader_context: &mut LoaderContext<RunnerContext>,
+  filename: Arc<FileName>,
+  resource_path: &str,
+  comments: Rc<SingleThreadedComments>,
+  rsc_meta: &RefCell<Option<RscMeta>>,
+) -> impl Pass {
+  let module = &loader_context.context.module;
+  let is_react_server_layer = module
+    .get_layer()
+    .is_some_and(|layer| layer == "react-server-components");
```
fz6m, line 28: @copilot There might be a better approach to avoid hard coding the same string in multiple Rust crates.

PR #12017, opened 2025-10-28
fix: calculate side effects of call exprs in seq expr

LingyuCoder:
## Summary

- should calculate call expressions in seq expression.
- fix excluded normal cases in development and production mode.

## Checklist


tests/rspack-test/normalCases/parsing/harmony-deep-exports/index.js
```diff
@@ -45,29 +45,17 @@ it("should be able to call a deep function in commonjs", () => {
...
-		expect(x1.resetInfo).toBe(undefined);
-		expect(x1.unusedExport).toBe(true);
-		expect(x1.unusedExportInfo).toBe(undefined);
-		expect(x1.somethingElse).toBe(true);
-		expect(x1.somethingElseInfo).toBe(undefined);
-		expect(C.exportsInfo.nsInfo).toBe(undefined);
-		expect(C.exportsInfo.ns2).toBe(true);
-		expect(C.exportsInfo.ns2Info).toBe(undefined);
-	}
+	expect(x1.incrementInfo).toBe(true);
```
LingyuCoder, line 48: The `usedExports` and `providedExports` are also `true` in development mode. So the values should be the same with that in production mode. Webpack seems to have bug and the value of `process.env.NODE_ENV` is `test` then these assertion codes will never run.

JSerFeng (approved):

PR #12023, opened 2025-10-28
refactor: source return string or buffer

SyMind:
## Summary

This PR refactors the `Source` trait's `source()` method to return a new `SourceValue` enum instead of `Cow<str>`. This change introduces a unified representation that can handle both text and binary data more explicitly, providing better type safety and flexibility for handling various source file types in the build tool.

**Key Changes:**
- Introduced `SourceValue` enum with `String` and `Buffer` variants to represent source content
- Updated all `Source` implementations to return `SourceValue` instead of `Cow<str>`
- Updated all call sites to use `.into_string_lossy()` when string conversion is needed

## Checklist

### DIFF
--- a/crates/rspack/tests/basic.rs
+++ b/crates/rspack/tests/basic.rs
@@ -19,7 +19,10 @@ mod tests {
       assert!(errors.is_empty());
 
       let asset = &compiler.compilation.assets().get("main.js").unwrap();
-      assert_eq!(asset.source.as_ref().unwrap().source(), "console.log(123);");
+      assert_eq!(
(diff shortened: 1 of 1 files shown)


h-a-n-a:
Can we remove the downcasting code in `crates/rspack_binding_api/src/source.rs`? I believe the reason it exists is solely for differentiating `Buffer` from `Source` in `RawSource`.

SyMind:
@h-a-n-a I have already removed it.

h-a-n-a:
@SyMind Cool! Looks like I haven't updated rspack source code for a very long time 🥲

PR #12046, opened 2025-10-30
perf: EsmLibraryPlugin uses par_iter to deconflict symbols

JSerFeng:
## Summary

Use par_iter for deconflict symbols in different chunks

## Checklist

### DIFF
--- a/crates/rspack_plugin_esm_library/src/link.rs
+++ b/crates/rspack_plugin_esm_library/src/link.rs
@@ -1,6 +1,7 @@
 use std::{
   collections::{self},
   hash::BuildHasher,
+  hint::black_box,
   sync::Arc,
 };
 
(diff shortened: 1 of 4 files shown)


crates/rspack_plugin_esm_library/src/link.rs
```diff
@@ -69,6 +70,12 @@ struct ExportsContext {
   pub exported_symbols: FxHashMap<Atom, Atom>,
 }
 
+#[allow(clippy::mut_from_ref)]
+unsafe fn unsafe_get_mut<T>(r: &T) -> &mut T {
+  let a = black_box(r as *const T);
```
quininer, line 75: This is unsound. The `black_box` implies that operation has side effects, but it does not imply has alias.

crates/rspack_plugin_esm_library/src/link.rs
```diff
@@ -471,7 +480,12 @@ impl EsmLibraryPlugin {
       let exports_type = module.build_meta().exports_type;
       let default_object = module.build_meta().default_object;
 
-      let info = &mut concate_modules_map[id];
+      /*
+       * Safety:
+       * Scope hoisted module canonly exist in one chunk
+       * One chunk deconflict symbols won't access module in other chunk
```
quininer, line 486: If we want to use unsafe, the correct way is use `IdentifierIndexMap<UnsafeCell<ModuleInfo>>`.

A better way is to prove it to type system. We can lock the item (like `IdentifierIndexMap<AtomicRefCell<T>>`) or preemptively take `&mut T` to prove they don't intersect.

PR #12048, opened 2025-10-30
feat: add tracy profiler support

hardfist:
## Summary
add [tracy profiler](https://github.com/wolfpld/tracy) support for better memory profiling

```
cd crates/node_binding && TRACY=1 pnpm build:dev # build binding with tracy profiler support
```
## Checklist

### DIFF
--- a/crates/rspack_allocator/Cargo.toml
+++ b/crates/rspack_allocator/Cargo.toml
@@ -22,3 +22,6 @@ mimalloc = { workspace = true, features = ["v3"] }
 
 [target.'cfg(all(not(target_os = "linux"), not(target_os = "macos"), not(target_family = "wasm")))'.dependencies]
 mimalloc = { workspace = true }
+
+[package.metadata.cargo-shear]
+ignored = ["tracy-client-sys"]


Cargo.toml
```diff
@@ -146,10 +146,11 @@ swc_html_minifier   = { version = "35.0.0", default-features = false }
 swc_node_comments   = { version = "16.0.0", default-features = false }
 swc_plugin_runner   = { version = "21.0.0", default-features = false }
 
-wasi-common = { version = "35.0.0", default-features = false }
-wasmtime    = { version = "35.0.0", default-features = false }
-
-rspack_dojang = { version = "0.1.11", default-features = false }
+rspack_dojang    = { version = "0.1.11", default-features = false }
+tracy-client     = { version = "=0.18.2", features = ["enable", "sampling", "demangle", "system-tracing"] }
+tracy-client-sys = { version = "=0.26.0" }
```
stormslowly: Do we need to explicitly specify the client-sys version?

hardfist: yes, since it's deep coupled with tracy-server version and it may change IPC protocol during patch version

PR #12067, opened 2025-11-03
fix: use correct Cell for EsmLibraryPlugin

JSerFeng:
## Summary

Plugin needs to run without any mutable reference, however we do need to store some data in plugin.

EsmLibraryPlugin has some internal state that needs to be initialized in `make` phase, and consumed by `seal` phase, we use OnceCell before, which will panic if user rebuild again.

The case is a common write then read case. And because we need to support async, so use AtomicRefCell for such cases.

## Checklist

### DIFF
--- a/crates/rspack_plugin_esm_library/src/plugin.rs
+++ b/crates/rspack_plugin_esm_library/src/plugin.rs
@@ -251,6 +251,7 @@ async fn finish_modules(&self, compilation: &mut Compilation) -> Result<()> {
   // we mutably modify data in `self.concatenated_modules_map`
   let mut map = self.concatenated_modules_map_for_codegen.borrow_mut();
   *map = Arc::new(modules_map.clone());
+  drop(map);
 
   *self.concatenated_modules_map.write().await = Arc::new(modules_map);
   // mark all entry exports as used
--- a/crates/rspack_plugin_esm_library/src/render.rs
+++ b/crates/rspack_plugin_esm_library/src/render.rs
@@ -76,6 +76,9 @@ impl EsmLibraryPlugin {
     asset_info: &mut AssetInfo,
   ) -> Result<Option<RenderSource>> {
     let module_graph = compilation.get_module_graph();
+
+    // In this phase we only read from the lock, no write happen in this phase, the
+    // next write happen only happen for next compile start
     let chunk_link_guard = self.links.borrow();
     let chunk_link = &chunk_link_guard[chunk_ukey];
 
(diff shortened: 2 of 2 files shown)


crates/rspack_plugin_esm_library/src/plugin.rs
```diff
@@ -39,9 +40,10 @@ pub struct EsmLibraryPlugin {
   pub(crate) preserve_modules: Option<PathBuf>,
   // module instance will hold this map till compile done, we can't mutate it,
   // normal concatenateModule just read the info from it
-  pub(crate) concatenated_modules_map_for_codegen: OnceCell<Arc<IdentifierIndexMap<ModuleInfo>>>,
+  pub(crate) concatenated_modules_map_for_codegen:
+    AtomicRefCell<Arc<IdentifierIndexMap<ModuleInfo>>>,
```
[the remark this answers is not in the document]
JSerFeng: The access is guarrented to happen after the initialize

crates/rspack_plugin_esm_library/src/render.rs
```diff
@@ -76,9 +76,8 @@ impl EsmLibraryPlugin {
     asset_info: &mut AssetInfo,
   ) -> Result<Option<RenderSource>> {
     let module_graph = compilation.get_module_graph();
-    let chunk_link_guard = self.links.get().expect("should have chunk link");
-
-    let chunk_link = chunk_link_guard.get(chunk_ukey).expect("should have chunk");
+    let chunk_link_guard = self.links.borrow();
+    let chunk_link = &chunk_link_guard[chunk_ukey];
```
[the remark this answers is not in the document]
JSerFeng: The chunk is from chunk_graph, and the chunk_link built from chunk_graph.keys(), so there is certainly have the value

crates/rspack_plugin_esm_library/src/plugin.rs
```diff
@@ -247,10 +249,8 @@ async fn finish_modules(&self, compilation: &mut Compilation) -> Result<()> {
 
   // only used for scope
   // we mutably modify data in `self.concatenated_modules_map`
-  self
-    .concatenated_modules_map_for_codegen
-    .set(Arc::new(modules_map.clone()))
-    .expect("should set concatenated_modules_map_for_codegen");
+  let mut map = self.concatenated_modules_map_for_codegen.borrow_mut();
+  *map = Arc::new(modules_map.clone());
```
quininer, line 253: Do we need to drop this guard after assignment complete to reduce lock scope?

JSerFeng: No need, this is in `finish_modules` phase, this phase just write data not read at all

quininer: It may be safe in terms of behavior, but a smaller scope can produce better code.

JSerFeng: Thanks, done here and add some comments in later read stages

Issue #12072, opened 2025-11-03
[Bug]: should have __webpack_require__.f.consumes error


kwonoj:
### System Info

System:
    OS: macOS 26.0.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 2.82 GB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.15.0 - ~/.proto/tools/node/20.15.0/bin/node
    Yarn: 4.5.0 - ~/.proto/shims/yarn
    npm: 10.7.0 - ~/.proto/tools/node/20.15.0/bin/npm
  Browsers:
    Chrome: 142.0.7444.60
    Safari: 26.0.1
  npmPackages:
    @rspack/cli: 1.6.0-beta.0 => 1.6.0-beta.0 
    @rspack/core: 1.6.0-beta.0 => 1.6.0-beta.0 
    @rspack/plugin-react-refresh: ^1.5.2 => 1.5.2
    @module-federation/enhanced": "^0.17.1

### Details

Trying to upgrade rspack from 1.5.x to 1.6 and seeing the regressions

```
Uncaught Error: should have __webpack_require__.f.consumes
    at __webpack_require__.f.consumes (command-packages_ece-video-decoders_src_wrappedKeyHeaderDecoder_Argon2idWorker_KDFWorker_ts.chunk.js:650:87)
    at command-packages_ece-video-decoders_src_wrappedKeyHeaderDecoder_Argon2idWorker_KDFWorker_ts.chunk.js:153:30
    at Array.reduce (<anonymous>)
```

I recall previous fix was related with MF plugin or workers, and assume it's similar since error appears in the chunk around web workers or mf shared runtime related chunks.

Current mf runtime version is `"@module-federation/enhanced": "^0.17.1",`, but upgrading to the latest shows the same.

**Started occurring in 1.6.0-beta.0, 1.5.x works fine.**

### Reproduce link

[_No response_](https://github.com/web-infra-dev/rspack/pull/12074)

kwonoj:
Looking through changelogs https://github.com/web-infra-dev/rspack/pull/11779 seems somewhat related, but it's rough guess at the moment.

kwonoj:
I tried to make some repro https://github.com/web-infra-dev/rspack/pull/12074 and can see the error itself, while do not have confidence that this is the same root cause with my codebase.

Issue #12015, opened 2025-10-28
[Bug]: Federation module and web worker errors in the development server


baiwusanyu-c:
### System Info

System:
    OS: macOS 15.1.1
    CPU: (8) arm64 Apple M2
    Memory: 70.72 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.17.0 - /Users/baiwusanyu/.nvm/versions/node/v22.17.0/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.9.2 - /Users/baiwusanyu/.nvm/versions/node/v22.17.0/bin/npm
    pnpm: 9.15.0 - /usr/local/bin/pnpm
  Browsers:
    Chrome: 141.0.7390.123
    Safari: 18.1.1

### Details

-

### Reproduce link

https://github.com/baiwusanyu-c/test-rspack

### Reproduce Steps

1. Install dependencies correctly (the repository does not use a monorepo)
2.Run "build:child:mf" in the rspack-child folder
3.Run "dev" in the rspack-child folder
4.Run "dev" in the rspack-child folder

I got an error

<img width="677" height="540" alt="Image" src="https://github.com/user-attachments/assets/30c8f149-169b-49e6-9132-f419c101bad5" />

PR #12096, opened 2025-11-05
fix: inject global entry to async entrypoints

ahabhgk:
## Summary

If a entry name is undefined, then the entry is global entry, before we only add global entry's dependencies and includedDependencies into Entrypoints, after this PR global entry's dependencies and includedDependencies will also added into AsyncEntrypoints (created by `new Worker()` most of the time). So...
1. [Global entry that created by devServer](https://github.com/webpack/webpack-dev-server/blob/c53955de36d793345f203ae0f336bd885dd2efaa/lib/Server.js#L817-L819) now will also added into Worker runtime
2. [Global entry that created by ModuleFederationPlugin](https://github.com/web-infra-dev/rspack/blob/11814b5a116de80b07c6da3514ccb7576cfb4360/crates/rspack_plugin_mf/src/container/module_federation_runtime_plugin.rs#L67-L70) will also added into Worker runtime
3. Existing plugins that using global entry (`new EntryPlugin(..., { name: undefined }).apply(compiler)`, `addEntry(..., { name: undefined })`, or `addInclude(..., { name: undefined })`) will also be affected if the added code is not compatible with worker environment, but I don't think this kind of secret usage exists

## Related links

fix https://github.com/web-infra-dev/rspack/issues/7417
fix https://github.com/web-infra-dev/rspack/issues/12015
fix https://github.com/web-infra-dev/rspack/issues/12072

## Checklist

### DIFF
--- a/tests/e2e/cases/worker/mf-worker/rspack.config.js
+++ b/tests/e2e/cases/worker/mf-worker/rspack.config.js
@@ -1,4 +1,4 @@
-const path = require("path")
+const path = require("path");
 const rspack = require("@rspack/core");
 
 /** @type {rspack.Configuration} */
--- a/tests/e2e/cases/worker/worker-hmr/rspack.config.js
+++ b/tests/e2e/cases/worker/worker-hmr/rspack.config.js
@@ -1,4 +1,3 @@
-const path = require("path")
 const rspack = require("@rspack/core");
 
 /** @type {rspack.Configuration} */


JSerFeng (approved):

chenjiahan:
Can we document this behaviour?

chenjiahan:
@copilot submit a PR to document this behavior, adding to:

- website/docs/en/plugins/webpack/entry-plugin.mdx
- website/docs/zh/plugins/webpack/entry-plugin.mdx

PR #12133, opened 2025-11-10
fix: keep entry chunk signature

JSerFeng:
## Summary

Should always make sure entry chunk signature won't be ruined by other chunk imports.

When user use splitChunk to split module out from the entry chunk, there is a cycle and other chunks may import the symbol from the entry chunk. The entry chunk has to expose those symbols.

Check if other chunk has modules depends on modules inside entry chunk, if so, generate a chunk to avoid such things

## Checklist

### DIFF
--- a/tests/rspack-test/esmOutputCases/preserve-modules/conflict-symbol/__snapshots__/esm.snap.txt
+++ b/tests/rspack-test/esmOutputCases/preserve-modules/conflict-symbol/__snapshots__/esm.snap.txt
@@ -40,7 +40,6 @@ export { other_demo as demo };
 ```
 
 ```mjs title=index.mjs
-import "./demo/index.mjs";
 
 // ./src/index.js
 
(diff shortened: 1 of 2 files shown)


JSerFeng:
@copilot

crates/rspack_core/src/module_graph/mod.rs
```diff
@@ -222,6 +222,38 @@ impl<'a> ModuleGraph<'a> {
...
+
+    let mut map: HashMap<ModuleIdentifier, Vec<&ModuleGraphConnection>> = HashMap::default();
+    for dep_id in connections {
+      let con = self
+        .connection_by_dependency_id(dep_id)
+        .expect("should have connection");
+      if !con.is_active(module_graph, runtime, module_graph_cache) {
+        continue;
+      }
+      match map.entry(*con.module_identifier()) {
```
quininer: `.or_default().push()` ?

JSerFeng: Done

quininer (dismissed):

ahabhgk (approved):

Timeless0911:
Fix: https://github.com/web-infra-dev/rsbuild/pull/6524

PR #12169, opened 2025-11-12
chore: depreacte experiments.parallelCodeSplitting

JSerFeng:
## Summary

This options is depreacted, it has huge performance regression in some edge cases where the chunk graph has lots of cycles, we'll improve the performance of build_chunk_graph in the future instead.

## Checklist


website/docs/en/config/experiments.mdx
```diff
@@ -150,15 +150,15 @@ Starting from v1.4.0, Rspack enables incremental builds for all phases by defaul
 
 ## experiments.parallelCodeSplitting
 
-<ApiMeta addedVersion="1.2.0" />
+<ApiMeta deprecatedVersion="1.6.2" />
```
chenjiahan: We can move the deprecated options to the bottom of the page

chenjiahan: See: https://rspack.rs/config/experiments#experimentstoplevelawait

JSerFeng: Done

PR #12183, opened 2025-11-13
chore: bump rust toolchain

JSerFeng:
## Summary

Update rust toolchain.

In current version, if Cargo.lock has some specific syntax error, it may enter endless loop and run out of memory.

Known Issue: https://github.com/toml-rs/toml/pull/1003

## Checklist

### DIFF
--- a/crates/rspack/src/builder/builder_context.rs
+++ b/crates/rspack/src/builder/builder_context.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::enum_variant_names)]
 use enum_tag::EnumTag;
 use rspack_core::{
   BoxPlugin, ChunkLoadingType, CompilerOptions, EntryOptions, ExternalItem, ExternalType,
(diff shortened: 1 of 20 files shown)


crates/rspack/src/builder/builder_context.rs
```diff
@@ -7,8 +8,7 @@ use rspack_core::{
 /// Options of builtin plugins
 ///
 /// The order of this list is strictly ordered with respect to `rspackOptionsApply`.
-#[allow(clippy::enum_variant_names)]
-#[derive(Debug, EnumTag)]
+#[derive(EnumTag, Debug)]
```
JSerFeng, line 11: have to make clippy anotation at the root using `#![...]`, as the EnumTag macro will generate code like:
```rs
pub enum Value {
    AaaPlugin,
    BbbPlugin,
    CccPlugin,
}
const _: () = {
    pub enum ValueTag {
        AaaPlugin,
        BbbPlugin,
        CccPlugin,
    }
```

the #[allow(clippy::enum_variant_names)] can't work inside the second enum

PR #12186, opened 2025-11-13
refactor: split make phase

hardfist:
## Summary
split make phase into build_module_graph and collect_build_module_graph_effects phase which makes artifacts no overlap and has more accurate meaning
* build_module_graph: build module graph
* collect_build_module_graph_effects: collect diagnostics & mutations

Followup:
* change collect_build_module_graph_effects to access &Compilation

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/rebuild.rs
+++ b/crates/rspack_core/src/compiler/rebuild.rs
@@ -83,7 +83,7 @@ impl Compiler {
         .mutations_readable(IncrementalPasses::MAKE)
       {
         // copy field from old compilation
-        // make stage used
+        // build_module_graph stage used
         self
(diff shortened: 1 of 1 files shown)


crates/rspack_core/src/compiler/mod.rs
```diff
@@ -252,9 +252,7 @@ impl Compiler {
 
     Ok(())
   }
-
-  #[instrument("Compiler:compile", target=TRACING_BENCH_TARGET,skip_all)]
```
stormslowly, line 256: Should we add the instrument back?

hardfist: yes

PR #12197, opened 2025-11-14
ci(native-watcher): test multiple platforms

h-a-n-a:
## Summary

Add multi-platform test for rspack_watcher

## Checklist

### DIFF
--- a/.github/workflows/reusable-rust-test.yml
+++ b/.github/workflows/reusable-rust-test.yml
@@ -144,18 +144,15 @@ jobs:
             -- --nocapture
 
   rust_watcher_test:
-    name: Rust Watcher test ${{ matrix.array.os }}
+    name: Rust watcher test - ${{ matrix.array.runner }}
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-rust-test.yml
```diff
@@ -140,18 +140,58 @@ jobs:
...
+    needs: [check-changed]
+    if: ${{ needs.check-changed.outputs.code_changed == 'true' }}
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+
+      - name: Install Rust Toolchain
+        uses: ./.github/actions/rustup
+        with:
+          save-if: true
+          key: test
```
stormslowly: This key is too simple, should add os to avoid conflict.

h-a-n-a: done

PR #12226, opened 2025-11-18
feat: support shorthand function for module factory to reduce bundle size

JSerFeng:
## Summary

Introducing new `output.environment.methodShorthand`. If target environment support this, module factories will be rendered as object method shorthand form.

Before

```js
__webapck_modules__ = {
  "./src/app.ts": function() {}
}
```

This PR:

```js
__webpack_modules__ = {
"./src/app.ts"() {}
}
```

## Checklist


crates/rspack_plugin_javascript/src/runtime.rs
```diff
@@ -212,16 +220,29 @@ pub async fn render_module(
...
+        )));
+      }
       if module.build_info().strict && !all_strict {
         container_sources.add(RawStringSource::from_static("\"use strict\";\n"));
       }
       container_sources.add(render_source.source);
-      container_sources.add(RawStringSource::from_static("\n\n})"));
-      container_sources.add(RawStringSource::from_static(",\n"));
+
+      if compilation
```
stormslowly: use `use_method_shorthand`?

crates/rspack_plugin_javascript/src/runtime.rs
```diff
@@ -212,16 +220,29 @@ pub async fn render_module(
...
+        .options
+        .output
+        .environment
+        .supports_method_shorthand()
```
chenjiahan: Can we reuse the `use_method_shorthand` variable here

chenjiahan (approved):

PR #12232, opened 2025-11-18
fix(types): use namespace imports for Node.js built-in modules

SyMind:
## Summary

When users have `esModuleInterop: false` in their TypeScript configuration, importing Node.js built-in modules using default import syntax causes type errors:

```
Module '"node:http"' has no default export
```

This happens because Node.js built-in modules use CommonJS format and don't have default exports.

### Solution

Changed from default imports to namespace imports for Node.js built-in modules:

```typescript
// Before (fails with esModuleInterop: false)
import type http from "node:http";

// After (compatible with all TypeScript configurations)
import type * as http from "node:http";
```

## Checklist


chenjiahan (dismissed):
LGTM

PR #12235, opened 2025-11-18
refactor: processing more string concatenation in runtime template part 2

LingyuCoder:
## Summary

Should process all string concatenation in runtime templates of runtime modules and avoid processing them in rust dead code.

## Checklist

### DIFF
--- a/crates/rspack_plugin_sri/src/runtime.rs
+++ b/crates/rspack_plugin_sri/src/runtime.rs
@@ -18,9 +18,7 @@ use crate::{
 fn add_attribute(tag: &str, code: &str, cross_origin_loading: &CrossOriginLoading) -> String {
   format!(
     "{}\n{tag}.integrity = {}[chunkId];\n{tag}.crossOrigin = {};",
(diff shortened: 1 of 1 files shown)


crates/rspack_plugin_runtime/src/runtime_module/module_chunk_loading.rs
```diff
@@ -232,33 +240,16 @@ impl RuntimeModule for ModuleChunkLoadingRuntimeModule {
...
-            ""
-          },
-        )
-        .cow_replace("$CROSS_ORIGIN$", &cross_origin)
-        .to_string();
+        let link_prefetch_code = compilation.runtime_template.render(
+          &self.template(TemplateId::WithPrefetchLink),
+          Some(serde_json::json!({
+            "_charset": charset,
+            "_cross_origin": match cross_origin_loading {
```
stormslowly: This `match` expression is duplicated for 3 times, implement a `to_string` or  `to_corros_origin`

PR #12237, opened 2025-11-19
feat: support prefetch and preload in css extract plugin

LingyuCoder:
## Summary

- support prefetch and preload in css extract plugin.
- refactor css runtime module and css extract runtime module to process string concatenation in render templates.
- add `createLink` runtime plugin hooks to make sure the sri plugin can handle subresource integrity during creating link tags.

## Checklist

### DIFF
--- a/crates/rspack_plugin_extract_css/src/runtime.rs
+++ b/crates/rspack_plugin_extract_css/src/runtime.rs
@@ -233,12 +233,9 @@ impl RuntimeModule for CssLoadingRuntimeModule {
     }
 
     if with_hmr {
-      let hmr = compilation.runtime_template.render(
-        &self.template_id(TemplateId::WithHmr),
-        Some(serde_json::json!({
-          "_hmr_download_update_handlers": RuntimeGlobals::HMR_DOWNLOAD_UPDATE_HANDLERS.to_string(),
-        })),
-      )?;
+      let hmr = compilation
+        .runtime_template
+        .render(&self.template_id(TemplateId::WithHmr), None)?;
       res.push(hmr);
     } else {
       res.push("// no hmr".to_string());


crates/rspack_plugin_sri/src/runtime.rs
```diff
@@ -17,7 +17,7 @@ use crate::{
 
 fn add_attribute(tag: &str, code: &str, cross_origin_loading: &CrossOriginLoading) -> String {
   format!(
-    "{}\n{tag}.integrity = {}[chunkId];\n{tag}.crossOrigin = {};",
+    "{}\n{tag}.integrity = {}[chunkId];\n{tag}.crossOrigin = '{}';",
```
LingyuCoder, line 20: This is a bug fix, which is introduced in https://github.com/web-infra-dev/rspack/pull/12235

crates/rspack_plugin_css/src/runtime/mod.rs
```diff
@@ -224,63 +255,71 @@ installedChunks[chunkId] = 0;
...
+        let link_prefetch_raw = compilation.runtime_template.render(
+          &self.template_id(TemplateId::WithPrefetchLink),
+          Some(serde_json::json!({
+            "_charset": compilation.options.output.charset,
+            "_cross_origin": compilation.options.output.cross_origin_loading.to_string(),
+          })),
+        )?;
+
+        let link_prefetch = runtime_hooks
+          .borrow()
```
hardfist, line 274: need add safety comment about why .borrow here wouldn't panic here

hardfist: ok it seems it's only readable here so it seems safe here

PR #12250, opened 2025-11-20
feat: support MF share treeshake 

2heal1:
## Summary

Support shared tree shaking .

It support two modes: `runtime-infer` | `server-calc` to satisfy different demands. 

View this [discusstion](https://github.com/module-federation/core/discussions/4302) for more details . 

## Checklist

### DIFF
--- a/crates/rspack_plugin_mf/src/sharing/shared_container_plugin.rs
+++ b/crates/rspack_plugin_mf/src/sharing/shared_container_plugin.rs
@@ -88,7 +88,7 @@ async fn additional_tree_runtime_requirements(
   if let Some(name) = chunk.name()
     && name == self.options.name
   {
-    runtime_modules.push(ShareContainerRuntimeModule::new().boxed());
+    runtime_modules.push(ShareContainerRuntimeModule::new(&compilation.runtime_template).boxed());
   }
   Ok(())
 }
--- a/crates/rspack_plugin_mf/src/sharing/shared_container_runtime_module.rs
+++ b/crates/rspack_plugin_mf/src/sharing/shared_container_runtime_module.rs
@@ -1,37 +1,24 @@
-use rspack_collections::Identifier;
-use rspack_core::{ChunkUkey, Compilation, RuntimeModule, RuntimeModuleStage, impl_runtime_module};
+use rspack_core::{
+  Compilation, RuntimeModule, RuntimeModuleStage, RuntimeTemplate, impl_runtime_module,
+};
 
 #[impl_runtime_module]
 #[derive(Debug)]
-pub struct ShareContainerRuntimeModule {
-  id: Identifier,
-  chunk: Option<ChunkUkey>,
-}
+pub struct ShareContainerRuntimeModule {}
 
 impl ShareContainerRuntimeModule {
-  pub fn new() -> Self {
-    Self::with_default(
-      Identifier::from("webpack/runtime/share_container_federation"),
-      None,
-    )
+  pub fn new(runtime_template: &RuntimeTemplate) -> Self {
+    Self::with_name(runtime_template, "share_container_federation")
   }
 }
 
 #[async_trait::async_trait]
 impl RuntimeModule for ShareContainerRuntimeModule {
-  fn name(&self) -> Identifier {
-    self.id
-  }
-
   async fn generate(&self, _compilation: &Compilation) -> rspack_error::Result<String> {
     Ok(
       "__webpack_require__.federation = { instance: undefined,bundlerRuntime: undefined };"
         .to_string(),
     )
-  }
-
-  fn attach(&mut self, chunk: ChunkUkey) {
-    self.chunk = Some(chunk);
   }
 
   fn stage(&self) -> RuntimeModuleStage {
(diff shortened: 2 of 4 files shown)


crates/rspack_plugin_javascript/src/plugin/flag_dependency_usage_plugin.rs
```diff
@@ -276,17 +275,32 @@ impl<'a> FlagDependencyUsagePluginProxy<'a> {
...
         dep_id,
         self.build_module_graph_artifact.get_module_graph(),
         &self.compilation.module_graph_cache_artifact,
         runtime,
-      ) else {
-        continue;
-      };
+      );
+
+      // 直接使用 await 调用异步钩子
```
ahabhgk:
```suggestion

```

crates/rspack_binding_api/src/raw_options/raw_builtins/mod.rs
```diff
@@ -500,6 +514,18 @@ impl<'a> BuiltinPlugin<'a> {
         provides.sort_unstable_by_key(|(k, _)| k.to_string());
         plugins.push(ProvideSharedPlugin::new(provides).boxed())
       }
+      BuiltinPluginName::CollectSharedEntryPlugin => {
+        let options = downcast_into::<RawCollectShareEntryPluginOptions>(self.options)
+          .map_err(|report| napi::Error::from_reason(report.to_string()))?
+          .into();
+        plugins.push(CollectSharedEntryPlugin::new(options).boxed())
+      }
+      BuiltinPluginName::SharedContainerPlugin => {
```
ahabhgk, line 523: What is SharedContainerPlugin, can this merge into ContainerPlugin with an additional option?

crates/rspack_plugin_javascript/src/dependency/esm/esm_import_specifier_dependency.rs
```diff
@@ -106,6 +106,45 @@ impl ESMImportSpecifierDependency {
       .map_or_else(|| self.ids.as_slice(), |meta| meta.ids.as_slice())
   }
 
+  pub fn get_esm_import_specifier_referenced_exports(
```
ahabhgk: This is duplicate with `get_referenced_exports`, can you reuse `get_referenced_exports`?

crates/rspack_binding_api/src/raw_options/raw_builtins/mod.rs
2heal1: While both plugins share the underlying ContainerEntryDependency , I don't think they should be merged or reused at the plugin level due to their distinct responsibilities and configurations:

1. Different Responsibilities : ContainerPlugin is a general-purpose plugin for exposing multiple modules ( exposes ), whereas SharedContainerPlugin is specialized for wrapping a single shared dependency (like react ) into an isolated container, typically used with IndependentSharedPlugin .
2. Configuration Structure : ContainerPlugin handles a complex list of exposes map and enhanced mode settings. SharedContainerPlugin focuses on a single request and version without the concept of exposes . Reusing ContainerPlugin here would require awkward parameter conversion (e.g., faking request as an exposes entry), which increases complexity.
3. Runtime Behavior : They inject different RuntimeModule s ( ExposeRuntimeModule vs ShareContainerRuntimeModule ). Merging them would introduce unnecessary branching logic.
I believe reusing them at the Dependency level (as they currently do) is the correct level of abstraction.

crates/rspack_plugin_javascript/src/dependency/esm/esm_import_specifier_dependency.rs
2heal1: got ,reused

crates/rspack_plugin_javascript/src/plugin/flag_dependency_usage_plugin.rs
2heal1: removed

packages/rspack/package.json
```diff
@@ -70,7 +70,7 @@
     "@rspack/lite-tapable": "1.1.0"
   },
   "peerDependencies": {
-    "@module-federation/runtime-tools": ">=0.22.0",
+    "@module-federation/runtime-tools": "0.24.1",
```
chenjiahan, line 73: @2heal1 Why use a fixed version here? Should we include ^0.24.1 and ^2.0.0

packages/rspack/src/sharing/IndependentSharedPlugin.ts
```diff
@@ -0,0 +1,469 @@
...
+    return true;
+  }
+  return ![
+    'TreeShakingSharedPlugin',
+    'IndependentSharedPlugin',
+    'ModuleFederationPlugin',
+    'SharedUsedExportsOptimizerPlugin',
+    'HtmlWebpackPlugin',
+    'HtmlRspackPlugin',
+    'RsbuildHtmlPlugin',
```
chenjiahan, line 44: Hardcoding an Rsbuild internal plugin name to `@rspack/core` doesn't seem like good practice...

This plugin class name isn't a public API and is unstable.

PR #10573, opened 2025-06-05
feat: Support `import.meta.filename/dirname/resolve`

magic-akari:
## Summary

- Closes: #8008 

## Checklist

- [x] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #12255, opened 2025-11-20
chore: add `import_meta_property_in_destructuring` parser hook

LingyuCoder:
## Summary

As discussed in https://github.com/web-infra-dev/rspack/pull/10573, to support the parsing of `import.meta.filename/dirname` in the `NodeStuffPlugin`, interop with the `ImportMetaPlugin` is required.

Since webpack's parser plugins are implemented in the compiler plugins, compilation hooks can be added on the compiler plugin. However, rspack's parser plugins are implemented in javascript parser, and also the compilation id cannot be obtained inside. It needs to be passed in step by step from the outside. Moreover, the processing of the parser and the execution of hooks are synchronous, but on rspack's Rust side, only asynchronous hooks are retained, and synchronous hooks are only privately implemented inside the parser hook.

Therefore, a hook is added to the parser hook for interop between the `ImportMetaPlugin` and the `NodeStuffPlugin`, and the name is changed to `import_meta_property_in_desctructuring` to indicate that it is a plugin interop hook. 

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/node_stuff_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/node_stuff_plugin.rs
@@ -213,8 +213,8 @@ impl JavascriptParserPlugin for NodeStuffPlugin {
 
   fn import_meta_property_in_destructuring(
     &self,
-    _: &mut JavascriptParser,
-    __: &DestructuringAssignmentProperty,
+    _parser: &mut JavascriptParser,
+    _property: &DestructuringAssignmentProperty,
   ) -> Option<String> {
     // TODO: implement import.meta.filename/dirname in destructuring
     None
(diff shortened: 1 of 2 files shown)


LingyuCoder:
Can this PR meet your needs? @magic-akari

magic-akari:
Yes, this PR meets my needs! I've rebased onto this branch and implemented the feature for `import.meta.filename/dirname` destructuring support. The `import_meta_property_in_destructuring` hook works well. I've implemented the `NodeStuffPlugin` side to handle the following pattern:

```JavaScript
const { dirname, filename } = import.meta;
```

The implementation respects both `node.filename`/`node.dirname` options and the `importMeta` parser option. All the different modes (False, Mock, WarnMock, True, EvalOnly, NodeModule) are handled correctly. I also added `FxIndexSet` to preserve the property order in destructuring assignments.

PR #12262, opened 2025-11-21
fix: not try to add integrity to tags with remote url

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/12251

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #12265, opened 2025-11-21
fix: should match protocol-relative url of public path and tag src in SRI plugin

LingyuCoder:
## Summary

Support protocol-relative url when match public path and tag src. Both the public path and tag src can be protocol-relative url. If they are matched, the SRI integrity hash should be found and add to the tag.

_Originally posted by @nanianlisao in https://github.com/web-infra-dev/rspack/issues/12262#issuecomment-3562108748_
            

## Checklist


tests/rspack-test/configCases/sri/remote-src-protocol/rspack.config.js
LingyuCoder, line 1: Check this test case. Does this meet your requirements? @nanianlisao

nanianlisao: nice

hardfist (approved):

PR #12266, opened 2025-11-21
fix(javascript): prevent import.meta.url replacement in web target

MIreland:
## Description
This PR fixes an issue where `import.meta.url` was unconditionally replaced by the local file path during the build, causing `file:///` URLs to appear in web builds.

## Fix
Modified `ImportMetaPlugin` to check if the target environment supports `document`. If it does (web target), `import.meta.url` is preserved in the output.

## Verification
Added a new test case `tests/rspack-test/configCases/parsing/import-meta-url-web` which asserts that `import.meta.url` is not replaced by a `file://` URL when targeting web.

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
@@ -79,11 +79,15 @@ impl JavascriptParserPlugin for ImportMetaPlugin {
     if for_name == expr_name::IMPORT_META_VERSION {
       Some(eval::evaluate_to_number(5_f64, start, end))
     } else if for_name == expr_name::IMPORT_META_URL {
-      Some(eval::evaluate_to_string(
-        self.import_meta_url(parser),
-        start,
-        end,
-      ))
+      if parser.compiler_options.output.environment.supports_document() {
+        None
+      } else {
+        Some(eval::evaluate_to_string(
+          self.import_meta_url(parser),
+          start,
+          end,
+        ))
+      }
     } else {
       None
     }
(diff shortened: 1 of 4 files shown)


crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
```diff
@@ -79,11 +79,15 @@ impl JavascriptParserPlugin for ImportMetaPlugin {
     if for_name == expr_name::IMPORT_META_VERSION {
       Some(eval::evaluate_to_number(5_f64, start, end))
     } else if for_name == expr_name::IMPORT_META_URL {
-      Some(eval::evaluate_to_string(
-        self.import_meta_url(parser),
-        start,
-        end,
-      ))
+      if parser.compiler_options.output.environment.supports_document() {
```
chenjiahan: The script must be an ES module to use `import.meta.url`. We could not use `supports_document()` to determine this.

MIreland: Pushed an update

intellild:
Thanks. This would be a breaking change. Users may expect the original file path while preserve `import.meta.url` results in chunk serve url at runtime. A fine granted control is needed.
You can use `DefinePlugin` to workaround.

See the playground for detail 

https://playground.rspack.rs/#eyJyc3BhY2tWZXJzaW9uIjoiMi4xLjAtcmMuMCIsImlucHV0RmlsZXMiOlt7ImZpbGVuYW1lIjoicnNwYWNrLmNvbmZpZy5qcyIsInRleHQiOiJpbXBvcnQgKiBhcyByc3BhY2sgZnJvbSBcIkByc3BhY2svY29yZVwiXG5cbmV4cG9ydCBkZWZhdWx0IHtcbiAgbW9kZTogXCJkZXZlbG9wbWVudFwiLFxuICBkZXZ0b29sOiBmYWxzZSxcblx0ZW50cnk6IHtcblx0XHRtYWluOiBcIi4vaW5kZXguanNcIlxuXHR9LFxuICBvdXRwdXQ6IHtcbiAgICBtb2R1bGU6IHRydWUsXG4gIH0sXG4gIHBsdWdpbnM6IFtcbiAgICBuZXcgcnNwYWNrLkRlZmluZVBsdWdpbih7XG4gICAgICAnaW1wb3J0Lm1ldGEudXJsJzogSlNPTi5zdHJpbmdpZnkoJycpLFxuICAgICAgXCJfX2ltcG9ydF9tZXRhX19cIjogJ2ltcG9ydC5tZXRhJ1xuICAgIH0pXG4gIF0sXG59OyJ9LHsiZmlsZW5hbWUiOiJpbmRleC5qcyIsInRleHQiOiJjb25zb2xlLmxvZygnaW1wb3J0Lm1ldGEudXJsJywgaW1wb3J0Lm1ldGEudXJsKTtcblxuY29uc3QgeyB1cmwgfSA9IF9faW1wb3J0X21ldGFfXztcblxuY29uc29sZS5sb2coJ21ldGEnLCB1cmwpO1xuIn0seyJmaWxlbmFtZSI6ImxpYi5qcyIsInRleHQiOiJleHBvcnQgZGVmYXVsdCBcImxpYlwiOyJ9XX0=

PR #12277, opened 2025-11-24
feat(loader-runner): Allow limiting worker pool size for parallel loaders

Pablinho:
## Summary

This is related to **feat: support experimental loader parallelism** ( https://github.com/web-infra-dev/rspack/pull/9807 ). Currently worker threads are spawned on demand and, by default, use all available CPU threads (effectively `cpus() - 1`, fallback to `1`)

### Problem
- For very large projects that build multiple targets (for example, browser and node in dev mode), enabling parallel loaders (for project still using babel-loader for example) can cause OOMs or build failures with little diagnostic output. I’ve seen high CPU usage just before the OOM. The root cause appears to be the total number of worker threads tinypool is using.

### Solution
- Tools like  [thread-loader](https://github.com/webpack/thread-loader/blob/v4.0.4/src/workerPools.js#L19) already let you control worker count via a worker option: `the number of spawned workers, defaults to (number of cpus - 1) or fallback to 1 when require('os').cpus() is undefined.` 
- Similarly, the [terser-webpack-plugin](https://github.com/webpack/terser-webpack-plugin/tree/main?tab=readme-ov-file#number) provides a numberic `parallel` option in order to enable the multi-process parallel and lets you control how many concurrent workers are used. 

As recomended I extended the **parallel** setting to accept ` booleann | {  maxWorkers?: number }`.
 - If **parallel** is an object, user can set `maxWorkers`. Then, we use that number as the requested thread count for Tinypool (applied to `minThreads`/`maxThreads`). Depends on the user to use a valid integer.  It should be an integer > 0 (and ideally not larger than your CPU count).
 - If `parallel: true`  defaults to (number of cpus - 1) or fallback to 1 (as usual)

This lets users tune Tinypool concurrency per-loader, helping avoid OOMs on constrained machines or very large builds. Any feedback, concerns, or alternative approaches welcome — I’m happy to adjust the implementation. Thanks in advance for the review!

## Related links
- https://github.com/webpack/thread-loader/blob/v4.0.4/src/workerPools.js#L19
- https://github.com/webpack/terser-webpack-plugin/blob/v5.3.14/src/index.js#L131
- https://vitest.dev/guide/migration.html#pool-rework

## Checklist


packages/rspack/src/config/types.ts
```diff
@@ -872,7 +872,7 @@ export type RuleSetLoaderWithOptions = {
 
 	loader: RuleSetLoader;
 
-	parallel?: boolean;
+	parallel?: boolean | number;
```
chenjiahan: Thanks for your contribution!

I’d recommend changing the new option to a more extensible shape:

```js
parallel?: boolean | {  maxWorkers?: number }`
```

The term `worker` is more generic than `thread`. This avoids tying the public API to a specific implementation detail such as `tinypool` and gives us room to adapt the underlying parallel-execution model in the future.

chenjiahan: Ref: https://vitest.dev/guide/migration.html#pool-rework

Pablinho: Thanks @chenjiahan, just updated the PR with the comments.

PR #12281, opened 2025-11-25
feat: use __rspack_dev_server_uri to communicate to right dev server

stormslowly:
## Summary

This update depends on [Pull Request #53](https://github.com/web-infra-dev/rspack-dev-server/pull/53).

This change enhances the developer experience for users running the development server on a proxied site rather than on localhost.

For example, if the development server starts on a localhost **auto** port, and the user configures `lazyCompilation: true` (using the default settings), the development assets are proxied to a site like `https://a.com`.

### Before

With lazy compilation enabled, requests would be sent to `https://a.com/lazy-compilation-using-compilation-proxy....` These requests would fail, resulting in a blank page.

### After

Requests will now be sent to `http://127.0.0.1:8080/lazy-compilation-using-compilation-proxy....`, allowing lazy compilation to function as expected.

## Checklist

### DIFF
--- a/packages/rspack/hot/lazy-compilation-web.js
+++ b/packages/rspack/hot/lazy-compilation-web.js
@@ -7,6 +7,7 @@ if (typeof EventSource !== "function") {
 var urlBase = decodeURIComponent(__resourceQuery.slice(1));
 if (
 	!urlBase.startsWith("http") &&
+	!urlBase.startsWith("//") &&
 	typeof __rspack_dev_server_uri !== "undefined"
 ) {
 	urlBase = __rspack_dev_server_uri + urlBase;


stormslowly:
> This is not a reasonable design, because most higher-level tools (including Rsbuild and Modern.js) do not use `@rspack/dev-server`.

I agree that high-level tools should configure RSPack correctly. 

However, there is a key difference in this case. We need to know the final `hostname` and `port` of the development server. The hostname and port are determined by the dev server; for example, if the port is configured to `auto`.

For Lazy Compilation, if the `serverUrl` is not configured properly, the application will go blank. To ensure Lazy Compilation works as expected, it is crucial to direct active requests to the correct server.

I believe there are two ways to achieve this:

### Config Phase
Before the compiler starts running, configure the `serverUrl` with the correct hostname and port. Since the appropriate hostname and port come from the dev server, determining the right configuration at that moment may not be straightforward.

### Runtime Phase
During runtime, the hostname and port for active lazy compilation are determined. Therefore, Lazy Compilation needs a way to obtain the dev server information. A "magic variable" is one potential solution. Without the variable `__rspack_dev_server_uri`, Lazy Compilation will default to using the browser's URL hostname and port. If the variable exists, it will utilize the configured value.

For Rsbuild and Modern.js, they can either adapt to this magic variable or opt for the configuration phase solution.

PR #12305, opened 2025-11-26
fix: align async startup federation handler setup

ScriptedAlchemy:
- build remotes/consumes chunk mappings over all referenced chunks (parity with enhanced/webpack)
- async startup: seed handlers via __webpack_require__mf_startup_once() and run consumes/remotes ensure handlers for all mapped chunk ids before entry executes
- sync behavior unchanged; tests all pass except known flaky Example.test.js


ScriptedAlchemy:
Closed on request; changes merged locally into feature/async-startup-runtime-promise.

PR #12318, opened 2025-11-27
ci: add ecosystem CI based on rstack-ecosystem-ci

fi3ework:
## Summary

reusing the unified ecosystem CI repository at https://github.com/rspack-contrib/rstack-ecosystem-ci for Rspack will preserve the previous unconventional build process, which is unique to Rspack. after this, we can report the results from the Rspack ecosystem CI.

😅 i hope this workflow works correctly on the first try, but it may require some subsequent commits to work properly. after this workflow is working correctly, i will remove the old workflow.

## Checklist


chenjiahan:
🫡

PR #12346, opened 2025-12-02
chore(workflow): run diff CI on push events

yifancong:
## Summary
chore: change the diff ci yaml

**Workflow trigger improvements:**

* The CI workflow now also runs on `push` events to the `main` branch, in addition to pull request events (`opened`, `synchronize`, and `reopened`). This ensures that changes pushed directly to `main` are also checked by the workflow.
* The workflow trigger for pull requests now uses `reopened` instead of `closed`, so the workflow is triggered when a pull request is reopened rather than when it is closed.

## Checklist

### DIFF
--- a/.github/workflows/ci-diff.yml
+++ b/.github/workflows/ci-diff.yml
@@ -9,7 +9,7 @@ on:
       - main
 
 concurrency:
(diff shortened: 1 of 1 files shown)


yifancong:
> Could you add some PR description?

@chenjiahan  Done

PR #12348, opened 2025-12-02
feat: update resolver error messages to use colored output

chenjiahan:
## Summary

- add color utility functions (`red`, `yellow`, `cyan`) alongside `dim`
- update resolver error messages to use colored output for better visibility
- refactor dim function to return Display trait object instead of String

### Before

Screenshot 2025-12-02 at 15 18 42

### After

Screenshot 2025-12-02 at 15 15 25

## Checklist


stormslowly:
@codex

PR #12351, opened 2025-12-02
fix: should use sourceOrder to sort module outgoing in buildChunkGraph

### LINKED ISSUE #11148: [Bug]: sideEffects: false leads to incorrect CSS styles order
### System Info

  System:
    OS: macOS 14.0
    CPU: (8) arm64 Apple M1
    Memory: 99.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 9.15.2 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 138.0.7204.159
    Safari: 17.0
  npmPackages:
    @rspack/cli: ^1.3.15 => 1.4.9
    @rspack/core: ^1.3.15 => 1.4.9
    @rspack/plugin-react-refresh: ^1.4.3 => 1.4.3

### Details

When specifying sideEffects: false in package.json, rspack incorrectly generates the style order, while webpack processes them correctly.

Only in production build.

### Reproduce link

https://github.com/StaRenn/rspack-repro

### Reproduce Steps

1. `pnpm i`
2. `pnpm run build`
3. `pnpm run server:rspack` - Button "should be blue" is red
4. `pnpm run server:webpack` - Button "should be blue" is blue

Removing `sideEffects: false` in src/uiLib/package.json gives correct css order

JSerFeng:
## Summary

Fix #11148

Should use `sourceOrder` to sort module.dependencies when calculating chunk graph. This can keep the import order right.

```js
import {a} from './a'
import './b'
access(a); 
```

The order should be `a` then `b`

The order after side effect optimization would be:

```js
import './b'
access(a);
```

The order becomes `b` then `a`.

So use `sourceOrder` to sort all dependencies to make sure the first import statement be the first in the preOrderIndex

## Checklist

### DIFF
--- a/tests/rspack-test/configCases/css/css-order/__snapshot__/main.css.txt
+++ b/tests/rspack-test/configCases/css/css-order/__snapshot__/main.css.txt
@@ -1,1 +1,1 @@
-.teaser-module {  padding: 20px;  border: 1px solid #ddd;  border-radius: 8px;  margin: 16px;}.teaser-module {  background-color: orange;}.button-module {  padding: 8px 16px;  background-color: #007bff;  color: white;  border: none;  border-radius: 4px;}
+.button-module {  padding: 8px 16px;  background-color: #007bff;  color: white;  border: none;  border-radius: 4px;}.teaser-module {  padding: 20px;  border: 1px solid #ddd;  border-radius: 8px;  margin: 16px;}.teaser-module {  background-color: orange;}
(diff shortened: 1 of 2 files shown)


tests/rspack-test/configCases/css/css-order/__snapshot__/main.css.txt
```diff
@@ -1 +1 @@
-.teaser-module {  padding: 20px;  border: 1px solid #ddd;  border-radius: 8px;  margin: 16px;}.teaser-module {  background-color: orange;}.button-module {  padding: 8px 16px;  background-color: #007bff;  color: white;  border: none;  border-radius: 4px;}
\ No newline at end of file
+.button-module {  padding: 8px 16px;  background-color: #007bff;  color: white;  border: none;  border-radius: 4px;}.teaser-module {  padding: 20px;  border: 1px solid #ddd;  border-radius: 8px;  margin: 16px;}.teaser-module {  background-color: orange;}
```
JSerFeng, line 1: The same with webpack. previous snap is wrong

PR #12369, opened 2025-12-04
refactor: use swc_experimental on concatenated module 

CPunisher:
## Summary

Sadly there's no apparent performance improvement on my machine.

But we can see slightly improvement (16%) on analyze_module function on codspeed 

image

## Checklist

### DIFF
--- a/crates/rspack_core/src/concatenated_module.rs
+++ b/crates/rspack_core/src/concatenated_module.rs
@@ -3242,8 +3242,7 @@ fn collect_ident(ast: &Ast, semantic: &Semantic) -> Vec<NewConcatenatedModuleIde
   let mut ids = Vec::new();
   for (node_id, node) in ast.nodes() {
     if node.kind == NodeKind::Ident {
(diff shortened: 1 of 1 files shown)


tests/rspack-test/diagnosticsCases/module-parse-failed/concatenate_parse_module/stats.err
```diff
@@ -1,6 +1,6 @@
 ERROR in ./index.js + 1 modules
   × JavaScript parse error:
-  │ : Expected ',', got 'numeric literal (2, 2)'
+  │ : Expected ',', got 'numeric literal'
```
hardfist, line 3: why range removed here

CPunisher: This is not a range. It’s TokenValue actually. In the new parser I give up this kind of information in diagnostic in exchange for code simplification and performance. I think printing token value here is unnecessary because you can see the underlined source code in the diagnostic.

PR #12372, opened 2025-12-04
chore(ci): size limit refactor

stormslowly:
## Checklist


stormslowly:
`Build std` will introduce lots of `Dirty` to many (100+) crates while building with cache, then slow down the build process.

Issue #10919, opened 2025-07-06
[Bug]: ESModulesLinkingError while importing a re-exported variable from a different module


smeng9:
### System Info

  System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M2
    Memory: 95.41 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 23.10.0 - /usr/local/bin/node
    Yarn: 1.22.21 - /usr/local/bin/yarn
    npm: 10.9.2 - /usr/local/bin/npm
    pnpm: 8.12.0 - /usr/local/bin/pnpm
  Browsers:
    Safari: 17.5

### Details

We are using a package `unist-util-visit` (https://www.npmjs.com/package/unist-util-visit?activeTab=code) that internally reexports `CONTINUE, EXIT, SKIP` from its './lib/index.js' which is originally from `unist-util-visit-parents`

This seems is not resolved correctly.

➜  docusaurus git:(main) ✗ yarn build
[INFO] [en] Creating an optimized production build...
● Client ██████████████████████████████████████████████████ (93%) sealing after
● Server ██████████████████████████████████████████████████ (93%) sealing after [ERROR] Client bundle compiled with errors therefore further build is impossible.
× ESModulesLinkingError: export 'CONTINUE' (reexported as 'CONTINUE') was not found in './lib/index.js' (possible exports: visit)
   ╭─[2:0]
 1 │ // Note: types exported from `index.d.ts`
 2 │ export {CONTINUE, EXIT, SKIP, visit} from './lib/index.js'
   · ──────────────────────────────────────────────────────────
   ╰────

--------------------------

× ESModulesLinkingError: export 'EXIT' (reexported as 'EXIT') was not found in './lib/index.js' (possible exports: visit)
   ╭─[2:0]
 1 │ // Note: types exported from `index.d.ts`
 2 │ export {CONTINUE, EXIT, SKIP, visit} from './lib/index.js'
   · ──────────────────────────────────────────────────────────
   ╰────

--------------------------

× ESModulesLinkingError: export 'SKIP' (reexported as 'SKIP') was not found in './lib/index.js' (possible exports: visit)
   ╭─[2:0]
 1 │ // Note: types exported from `index.d.ts`
 2 │ export {CONTINUE, EXIT, SKIP, visit} from './lib/index.js'
   · ──────────────────────────────────────────────────────────
   ╰────

### Reproduce link

[...]

PR #12417, opened 2025-12-11
feat!: Add `pnp_manifest` option to resolver

smeng9:
## Summary

Fixes https://github.com/web-infra-dev/rspack/issues/10919

## Related links

Merge https://github.com/web-infra-dev/rspack-resolver/pull/141 first

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/options/raw_resolve.rs
+++ b/crates/rspack_binding_api/src/options/raw_resolve.rs
@@ -57,7 +57,7 @@ pub struct RawResolveOptions {
   pub roots: Option<Vec<String>>,
   pub pnp: Option<bool>,
   #[napi(ts_type = "string | false")]
-  pub pnp_manifest: Option<Either<String, ()>>,
+  pub pnp_manifest: Option<Either<String, bool>>,
 }
 
 fn normalize_alias(


crates/rspack_binding_api/src/rspack_resolver/mod.rs
```diff
@@ -213,6 +213,15 @@ impl ResolverFactory {
       symlinks: op.symlinks.unwrap_or(default.symlinks),
       builtin_modules: op.builtin_modules.unwrap_or(default.builtin_modules),
       enable_pnp: op.enable_pnp.unwrap_or_default(),
+      pnp_manifest: op.pnp_manifest.map(PathBuf::from).or_else(|| {
```
stormslowly: as We discussed in rspack-resolver's [PR](https://github.com/web-infra-dev/rspack-resolver/pull/141). We just map to PathBuf::from is enough, no needs for default value.

stormslowly:
@smeng9  We should add `pnpManifest` option to Rspack's resolve options to config resolver sticking to that manifest.
Beside, the document for resolve needs to be updated too.

crates/rspack_binding_api/src/rspack_resolver/mod.rs
smeng9: I have synchronized with napi in rspack-resolver

smeng9:
@stormslowly Since rspack-resolver does not have a formal document, I have updated the document of README.md in https://github.com/web-infra-dev/rspack-resolver/pull/141/commits/f9e2fcf57f538df6955b2b4f23c60309dd84466e

stormslowly:
@smeng9 rspack-resolver with pnpManifest option is published as [0.6.7](https://crates.io/crates/rspack_resolver/0.6.7), you can specifier that in the cargo.toml.

And we also need to expose this option in rspack's resolve options.
Could you plz help to add the options in rspack/core and update corresponding docs.

crates/rspack_core/src/resolver/resolver_impl.rs
```diff
@@ -324,6 +324,13 @@ fn to_rspack_resolver_options(
     builtin_modules: options.builtin_modules,
     imports_fields,
     enable_pnp: options.pnp.unwrap_or(false),
+    pnp_manifest: if options.pnp.unwrap_or(false) {
```
stormslowly: use `options.pnp_manifest.map(PathBuf::from)`

crates/rspack_core/Cargo.toml
```diff
@@ -85,6 +85,7 @@ winnow = { workspace = true }
 swc_experimental_ecma_ast      = { workspace = true }
 swc_experimental_ecma_parser   = { workspace = true }
 swc_experimental_ecma_semantic = { workspace = true }
+pnp.workspace = true
```
stormslowly: this line can be removed

crates/rspack_core/src/resolver/resolver_impl.rs
smeng9: I have updated to use options.pnp_manifest.map(PathBuf::from)

However I think providing the default value by calling pnp::find_closest_pnp_manifest_path if no pnp_manifest is provided in the options is still necessary. This ensures pnp global cache works out of the box without any additional configuration needed

crates/rspack_core/Cargo.toml
smeng9: Since we need pnp::find_closest_pnp_manifest_path function, we need the pnp dependency

crates/rspack_core/src/resolver/resolver_impl.rs
```diff
@@ -324,6 +324,15 @@ fn to_rspack_resolver_options(
     builtin_modules: options.builtin_modules,
     imports_fields,
     enable_pnp: options.pnp.unwrap_or(false),
+    pnp_manifest: options.pnp_manifest.map(PathBuf::from).or_else(|| {
+      if options.pnp.unwrap_or(false) {
+        std::env::current_dir()
+          .ok()
+          .and_then(|cwd| pnp::find_closest_pnp_manifest_path(&cwd))
```
stormslowly: As we discussed in the link provided (https://github.com/rstackjs/rspack-resolver/pull/141#discussion_r2641863739), we need to consider the implications of sticking to the current project PnP manifest. Doing so could potentially disrupt project references to other PnP-enabled projects.

In my opinion, if a user wants to utilize the global cache, they should explicitly configure their settings—in other words, stick to the manifest file. 

However, if we config the manifest for uesrs, they may encounter difficulties when working with projects that reference other Yarn projects.

stormslowly:
I am thinking that Yarn PnP's enableGlobalCache is enabled by default. https://yarnpkg.com/configuration/yarnrc#enableGlobalCache

So the PnP manifest should support global cache by default.
1. pnpManifest: null/undefined,   rspack fill it current path's PnP manifest file path
2. pnpManfiest: some string path, use the config manfiest
3. pnpManfiest: false           , support cross referenced enableGlobalCache disabled projects.

But case 1, will be a breaking change for current PnP users. 
I think We can introduce supporting GlobalCache enabled Pnp project by default in Rspack 2.0.

So what's your opinion @smeng9 

cc @LingyuCoder  @chenjiahan

smeng9:
Hi @stormslowly 

This merge request is a bug fix for incorrectly resolved `.pnp.cjs` in global cache, it will not introduce breaking change.
This is based on the assumption that there is no `.pnp.cjs` file in the global cache folder, and the user will only use the closest `.pnp.cjs` file from current path's PnP manifest file path

In rspack-resolve https://github.com/rstackjs/rspack-resolver/blob/main/src/lib.rs#L899C19-L899C49, we are already resolving `.pnp.cjs` from current path's PnP manifest file path. But if the file is in global folder, it does not have `.pnp.cjs`. We just lifted this resolving up from rspack-resolve to rspack, so the current folder is based on the project rather than the global cache.

Even if we have a yarn PnP project listed as another yarn PnP project's dependency, when performing a yarn install, the dependent project will be installed to global cache first, which strips away its `.pnp.cjs`

stormslowly:
Let me clarify my explanation with the following example: [GitHub Repository](https://github.com/stormslowly/mult_yarn_pnp_manifest).

```plaintext
.
├── app
│   ├── dist
│   │   ├── main.js
│   │   └── main.js.map
│   ├── index.js
│   ├── package.json
│   ├── README.md
│   ├── rspack.config.js
│   ├── .pnp.cjs
│   └── yarn.lock
├── lib
│   ├── index.js
│   ├── package.json
│   ├── README.md
│   ├── .pnp.cjs
│   └── yarn.lock
└── README.md
```

Both the `app` and `lib` directories enable Plug'n'Play (PnP) without a global cache. The `app/index.js` file requires `lib/index.js`, and `lib/index.js` requires the npm package `is-sorted`.

The `app` project works with the configuration `resolve: { pnp: true }`, because, in PnP mode, the resolver finds the PnP manifest from the current path. Specifically:

- It uses `app/.pnp.cjs` when resolving `lodash` from `app/index.js`.
- It uses `lib/.pnp.cjs` when resolving `is-sorted` from `lib/index.js`.

However, it fails when using the current Pull Request due to the configuration of the PnP manifest in the `rspack-resolver`, which causes the resolver to only reference the `app/.pnp.cjs` file. In other words, the resolution for `is-sorted` from `lib/index.js` still relies on `app/.pnp.cjs`, leading to the failure.

I hope this explanation is clearer.

```plaintext
$ yarn node /path/to/rspack/packages/rspack-cli/bin/rspack.js

ERROR in ../lib/index.js 1:26-36
  × Module not found: Can't resolve 'is-sorted' in '/Users/bytedance/git/problem/yarn_pnp_test/lib'
   ╭─[1:17]
 1 │ const isSorted = require("is-sorted");
   ·                  ────────────────────
 2 │
 3 │ module.exports = function (str) {
   ╰────

Rspack compiled with 1 error in 66 ms
```

smeng9:
I see the problem. If this is considered a breaking change, we can push it to 2.0

However I still think if app/index.js uses ../lib/index.js directly by referencing a single file out of it's package scope, which has hidden imports of `is-sorted`, then `is-sorted` should be declared in app's package.json's dependency, otherwise the require call is ambiguous and unsound.

smeng9:
If we are setting up the package to include files from another package, we should do the following https://github.com/stormslowly/mult_yarn_pnp_manifest/compare/master...smeng9:mult_yarn_pnp_manifest:master

stormslowly:
@smeng9, thank you very much for your understanding. 

If you need version 1.x of Rspack, you can simply remove the PnP find manifest in the Rspack resolve implementation. We expect to release it sometime next week.

The release plan for Rspack 2.0 is not yet clearly defined. If you are okay with Rspack 2.0, I will add the feature "Resolve Yarn PnP support for globalCache config by default" to this [discussion](https://github.com/web-infra-dev/rspack/discussions/9270). 

Additionally, I will adjust the option configuration according to this comment: https://github.com/web-infra-dev/rspack/pull/12417#issuecomment-3713784954.

smeng9:
sure sounds good

stormslowly:
@smeng9 
Do you have time to continue this PR?
If not, I can help

smeng9:
Hi @stormslowly 

I can still continue this PR, do we have anything else left besides resolving the conflict of Cargo.lock file?

stormslowly:
@smeng9 
I think we need to support these three kinds of config value for pnpManifest

- pnpManifest: `null`/`undefined`, rspack fills it with the current path's PnP manifest file path
- pnpManifest: some pnpManifest path, force rspack resolver stick to the pnp manifest file path
- pnpManifest: `false`, for support cross-referenced enableGlobalCache disabled projects.

smeng9:
Hi @stormslowly 

undefined / string / false options for pnpManifest should now be supported

stormslowly:
Thank you @smeng9

stormslowly (approved):

PR #12433, opened 2025-12-12
perf: remove ropey crate to reduce binary size

SyMind:
## Summary

This PR removes the ropey dependency and improves performance in two ways:

1. Replace ropey with `memchar`: Ropey is a rope data structure library designed for efficient large text modifications. However, Rspack only uses it for line and column offset calculations, not for actual text editing. By switching to `memchar`, a SIMD-accelerated string library, we achieve better performance for our specific use case.

2. Eliminate unnecessary string clone in `scan_dependencies`: During refactoring, I identified and removed an unnecessary string clone operation in the `scan_dependencies` function.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4502,7 +4502,6 @@ dependencies = [
  "fast-glob",
  "indexmap",
  "itertools 0.14.0",
- "memchr",
  "num-bigint",
  "once_cell",
  "rayon",
--- a/crates/rspack_plugin_javascript/Cargo.toml
+++ b/crates/rspack_plugin_javascript/Cargo.toml
@@ -18,7 +18,6 @@ itertools = { workspace = true }
 num-bigint = { workspace = true }
 once_cell = { workspace = true }
 rayon = { workspace = true }
-memchr = { workspace = true }
 regex = { workspace = true }
 regress = { workspace = true, features = ["pattern"] }
 rspack_cacheable = { workspace = true }


crates/rspack_plugin_json/src/lib.rs
```diff
@@ -84,8 +84,7 @@ impl ParserAndGenerator for JsonParserAndGenerator {
       .map_err(|e| {
         match e {
           UnexpectedCharacter { ch, line, column } => {
-            let rope = ropey::Rope::from_str(&source);
-            let line_offset = rope.try_line_to_byte(line - 1).expect("TODO:");
+            let line_offset = byte_line_column_to_offset(source.as_ref(), line, 0).expect("TODO:");
```
CPunisher: Please finish `TODO:`

crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -63,55 +62,90 @@ impl Debug for dyn SourceLocation {
...
 
-pub trait AsLoc {
-  fn as_loc(&self) -> &dyn SourceLocation;
-}
+    // Fast paths
+    if end_idx == 0 {
+      let p = SourcePosition { line: 1, column: 1 };
+      return Some((p, p));
+    }
+    if start_idx == end_idx {
```
CPunisher: What's the purpose of this if statement with empty body

crates/rspack_core/src/dependency/dependency_location.rs
```diff
@@ -63,55 +62,90 @@ impl Debug for dyn SourceLocation {
...
-      SourcePosition {
-        line: end_line + 1,
-        column: end_column,
-      },
-    ))
-  }
-}
-
 impl SourceLocation for &str {
   fn look_up_range_pos(&self, start: u32, end: u32) -> Option<(SourcePosition, SourcePosition)> {
```
CPunisher, line 66: Is it possible to use `byte_line_column_to_offset` here?

crates/rspack_plugin_javascript/src/magic_comment.rs
```diff
@@ -253,20 +271,54 @@ fn match_item_to_error_span(
...
+
+  // SAFETY: `match_start` and `match_end` are within `comment_text`.
+  let c_loc = byte_offset_to_location(comment_text, match_start, match_end);
 
   let Location { sl, sc, el, ec } = s_loc.merge_with_block_comment_location(&c_loc);
-  let start = s.line_to_byte(sl as usize) + sc as usize;
-  let end = s.line_to_byte(el as usize) + ec as usize;
+
+  // Compute absolute byte offsets from (line, column) for start and end in `source`.
+  let bytes = source.as_bytes();
```
CPunisher: Is it possible to use `byte_line_column_to_offset`?

crates/rspack_plugin_javascript/src/magic_comment.rs
```diff
@@ -220,17 +220,35 @@ impl Location {
...
+
+  // Count newlines before end to get end line
+  let mut el = 0usize;
+  let mut last_nl_pos_end = None;
+  for idx in memchr::memchr_iter(b'\n', &bytes[..end]) {
+    el += 1;
+    last_nl_pos_end = Some(idx);
+  }
+  let end_line_start_byte = last_nl_pos_end.map(|p| p + 1).unwrap_or(0);
+  let ec = end - end_line_start_byte;
```
CPunisher: There are multiple similar calculation spread over difference files. I'm not sure if their logic is similar enough to reuse the code. But at least I think we can try to put them together into one file / module.

crates/rspack_core/src/dependency/dependency_location.rs
SyMind: no, not same.

There is byte offset to utf16 line and column.

SyMind:
@CPunisher resolved all comments

PR #12449, opened 2025-12-15
fix: respect matchResource when generate nameForCondition

Austaras:
## Summary

In practice, this would cause vanilla-extract generated CSS become vendor chunk because it's generated by `node_modules/../vailla-extract/../extracted.js`

## Checklist


tests/rspack-test/configCases/chunks/matchResource/node_modules/foo-loader/index.js
```diff
@@ -0,0 +1,4 @@
+module.exports = content => {
+	console.log(content, 1111)
```
LingyuCoder: `console.log` should not be added unless necessary as they will contaminate the test logs.

PR #12465, opened 2025-12-16
fix: use RspackHash for HTML plugin content hash calculation

LingyuCoder:
## Summary

This PR fixes the content hash calculation in the HTML plugin. Previously, the plugin used `hash_for_source` utility function to generate content hashes. This change replaces it with `RspackHash` to ensure the hash calculation is consistent with the output configuration, which guarantees that generated HTML filenames correctly reflect content changes.

## Checklist

### DIFF
--- a/crates/rspack_util/src/lib.rs
+++ b/crates/rspack_util/src/lib.rs
@@ -24,9 +24,7 @@ pub mod test;
 pub mod tracing_preset;
 
 use std::{
-  collections::hash_map::DefaultHasher,
   future::Future,
-  hash::{Hash, Hasher},
   sync::LazyLock,
   time::{SystemTime, UNIX_EPOCH},
 };


crates/rspack_plugin_html/src/asset.rs
```diff
@@ -14,8 +15,8 @@ use rspack_core::{
   rspack_sources::{RawBufferSource, RawStringSource, SourceExt},
 };
 use rspack_error::{AnyhowResultToRspackResultExt, Result};
+use rspack_hash::RspackHash;
 use rspack_paths::Utf8PathBuf;
-use rspack_util::hash_for_source;
```
chenjiahan, line 18: The `hash_for_source` helper seems no longer used, should we remove it?

PR #12466, opened 2025-12-16
docs: fix rules[].use example

henryqdineen:
## Summary

```ts
  type RuleSetUse =
    | RuleSetUseItem[]
    | RuleSetUseItem
    | ((ctx: RawFuncUseCtx) => RuleSetUseItem[]);
```

When `use` is used as a function it should return `RuleSetUseItem[]` (an array), not a single `RuleSetUseItem`. My understanding, at least with webpack is that each item in the array is expected to have an `ident` set but even webpack does not document that well.

Thanks!

## Checklist

### DIFF
--- a/website/docs/en/config/module-rules.mdx
+++ b/website/docs/en/config/module-rules.mdx
@@ -718,16 +718,18 @@ export default {
       {
         test: /\.svg$/,
         type: 'asset',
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
LGTM, thank you!

PR #12469, opened 2025-12-16
perf(rstest): minor performance optimization

fi3ework:
## Summary

- ~~Hoist regex compilation to static LazyLock to avoid recompiling on every process_assets call~~
   > ❌ Size increased by 92.63KB from 48.19MB to 48.28MB (⬆️0.19%)
   
     LazyLock will increase about 92KB binary size, which is too large. since it's not a hot code, we can keep the current   code.
- Reduce string allocations in calc_mocked_target by using &str operations instead of to_string()
- Pre-allocate Vec capacity in mock_hoist_process_assets to avoid reallocations

## Checklist

### DIFF
--- a/crates/rspack_plugin_rstest/src/plugin.rs
+++ b/crates/rspack_plugin_rstest/src/plugin.rs
@@ -1,14 +1,7 @@
 use std::{
-  sync::{Arc, LazyLock},
+  sync::Arc,
   time::{Duration, Instant},
 };
-
-use regex::Regex;
-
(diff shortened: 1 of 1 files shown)


fi3ework:
> ❌ Size increased by 92.63KB from 48.19MB to 48.28MB (⬆️0.19%)

LazyLock will increase about 92KB binary size, which is too large. since it's not a hot code, we can keep the current code.

chenjiahan (approved):
LGTM 👍

PR #12483, opened 2025-12-17
feat: stabilize SubresourceIntegrityPlugin

LingyuCoder:
## Summary

This pull request stabilizes the SubresourceIntegrityPlugin by moving it from the experimental API to the stable API. The plugin is now exported directly from the main rspack package, making it available for production use without requiring the experiments flag.

The changes include:
- Exporting SubresourceIntegrityPlugin from the main exports
- Removing it from the experiments namespace
- Updating test configurations to use the stable API
- Updating documentation to reflect the stable status

This change improves the developer experience by making the SRI plugin more accessible and clearly indicating its stable status.

## Checklist

### DIFF
--- a/packages/rspack/src/exports.ts
+++ b/packages/rspack/src/exports.ts
@@ -373,6 +373,9 @@ interface Experiments {
 		cleanup: () => Promise<void>;
 	};
 	RemoveDuplicateModulesPlugin: typeof RemoveDuplicateModulesPlugin;
+	/**
+	 * @deprecated Use `rspack.SubresourceIntegrityPlugin` instead
+	 */
 	SubresourceIntegrityPlugin: typeof SubresourceIntegrityPlugin;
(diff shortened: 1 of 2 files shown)


packages/rspack/src/exports.ts
```diff
@@ -375,7 +377,6 @@ interface Experiments {
 	RsdoctorPlugin: typeof RsdoctorPlugin;
 	RstestPlugin: typeof RstestPlugin;
 	RslibPlugin: typeof RslibPlugin;
-	SubresourceIntegrityPlugin: typeof SubresourceIntegrityPlugin;
```
chenjiahan, line 378: Can we keep the `experiments.SubresourceIntegrityPlugin` for compatibility and mark it as deprecated?

chenjiahan (approved):

PR #12486, opened 2025-12-17
fix: exclude hidden dir from snapshot managedPaths default value

jerrykingxyz:
## Summary

Update the default managedPaths regex from /\/node_modules\// to /\/node_modules\/[^.]/ to exclude directories starting with a dot (e.g., .cache, .test) from being treated as managed paths in snapshot handling.

## Checklist

### DIFF
--- a/packages/rspack/src/config/normalization.ts
+++ b/packages/rspack/src/config/normalization.ts
@@ -411,7 +411,7 @@ export const getNormalizedRspackOptions = (
 							unmanagedPaths: nestedArray(snapshot.unmanagedPaths, p => [...p]),
 							managedPaths: optionalNestedArray(snapshot.managedPaths, p => [
 								...p
-							]) || [/\/node_modules\/[^.]/]
+							]) || [/[\\/]node_modules[\\/][^.]/]
 						},
 						storage: {
 							type: "filesystem",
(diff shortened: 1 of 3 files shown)


stormslowly:
Should we  consider `node_modules/.pnpm` ?

packages/rspack/src/config/normalization.ts
```diff
@@ -411,7 +411,7 @@ export const getNormalizedRspackOptions = (
 							unmanagedPaths: nestedArray(snapshot.unmanagedPaths, p => [...p]),
 							managedPaths: optionalNestedArray(snapshot.managedPaths, p => [
 								...p
-							]) || [/\/node_modules\//]
+							]) || [/\/node_modules\/[^.]/]
```
chenjiahan: Should we consider the windows path separator `[\\/]`

stormslowly: bingo，windows ci test failed.

jerrykingxyz:
> Should we consider `node_modules/.pnpm` ?

For pnpm, the directory is usually 
`<project_root>/node_modules/.pnpm/<lib>/node_modules/<lib>/package.json`
The node_modules in the second layer can be matched 

For yarn pnp, 
`<project_root>/.yarn/unplugged/<lib>/node_modules/<lib>/package.json`
also contains a node_modules that can be matched, so there is basically no problem

PR #12493, opened 2025-12-18
fix: relative resource path in source maps on Windows

SyMind:
## Summary

This PR fixes Windows-specific issues with relative resource paths in source maps by ensuring paths use forward slashes (URL-style) instead of backslashes (Windows filesystem style). Source maps require URL-style paths for cross-platform compatibility.

## Checklist

### DIFF
--- a/crates/rspack_plugin_devtool/src/module_filename_helpers.rs
+++ b/crates/rspack_plugin_devtool/src/module_filename_helpers.rs
@@ -3,6 +3,7 @@ use std::{
   path::Path,
 };
 
+use cow_utils::CowUtils;
 use rspack_core::{ChunkGraph, Compilation, OutputOptions, contextify};
 use rspack_error::Result;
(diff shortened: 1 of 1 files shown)


crates/rspack_plugin_devtool/src/module_filename_helpers.rs
```diff
@@ -56,13 +58,20 @@ fn resolve_relative_resource_path(
   };
 
   let Some(parent) = source_map_path.parent() else {
-    return Some(absolute_resource_path.to_string());
+    return Some(
+      absolute_resource_path
+        .normalize()
+        .to_string_lossy()
+        .cow_replace("\\", "/")
```
chenjiahan, line 65: Do we have a util to convert a path to posix path?

SyMind: currently not have

PR #12521, opened 2025-12-22
feat: splitChunks for EsmLibraryPlugin

JSerFeng:
### Merge dependencies

`optimization.splitChunks` for `EsmLibraryPlugin` should be separate from original `SplitChunksPlugin`, modules in web mode can be split into any chunks, as the execution is handled by `__webpack_require__` runtime. This does not correct for Esm output.

Esm executes once chunk is loaded. So any module matches a group, all sub dependencies of the module should also be matches the group to avoid cycle deps. If not, Imagine this: 

```
a -> b -> c

cacheGroup.test: /b/

Then b is extracted to another chunk, but leaves a and c still in the same chunk.

chunk1: [ac]
chunk2: [b]

chunk1 and chunk2 has cycle dependencies
```

So we should make sure that `b`'s dependency `c` is split into the new chunk as well.

### Ignore useless config

There are some configs of `optimization.splitChunks` are not supported. For example the `maxInitialRequests` and `maxSize`, that is used for improving chunk loading in browser, which is no need for library output. And it's also hard to implement, cause what module need to be split, and when it's split, what to do with its dependencies ? If not care, there must be a cycle, if care, there may be a huge chunk as all dependencies should also in the chunk. So for now we don't implement it

## Checklist


crates/rspack_plugin_split_chunks/src/plugin/mod.rs
```diff
@@ -59,7 +61,7 @@ impl SplitChunksPlugin {
     // Sort modules to ensure deterministic processing order
     all_modules.sort_unstable();
 
-    let module_sizes = Self::get_module_sizes(&all_modules, compilation);
+    let module_sizes = get_module_sizes(all_modules.par_iter().copied(), compilation);
```
hardfist, line 63: why copied needs par_iter?

hardfist:
@codex review

packages/rspack/src/rspackOptionsApply.ts
```diff
@@ -310,16 +311,19 @@ export class RspackOptionsApply {
           );
         }
 
+        enableLibSplitChunks = true;
         new EsmLibraryPlugin({
           preserveModules: options.output.library?.preserveModules,
+          splitChunks: options.optimization.splitChunks || undefined,
```
hardfist: why `|| undefined`  is needed here?

tests/rspack-test/esmOutputCases/interop/entry-exports-type-defaut/rspack.config.js
```diff
@@ -22,10 +22,6 @@ module.exports = {
   optimization: {
     splitChunks: {
       cacheGroups: {
-        cjs: {
```
hardfist, line 25: why change this configuration?

crates/rspack_plugin_split_chunks/src/plugin/mod.rs
JSerFeng: it receive a impl Iterator<ModuleIdentifier>, before it receive a &[ModuleIdentifier]

crates/rspack_binding_api/src/raw_options/raw_builtins/raw_esm_lib.rs
```diff
@@ -1,4 +1,19 @@
...
   pub preserve_modules: Option<String>,
+  pub split_chunks: Option<RawSplitChunksOptions<'a>>,
+}
+
+impl<'a> From<RawSplitChunksOptions<'a>> for Vec<CacheGroup> {
+  fn from(value: RawSplitChunksOptions<'a>) -> Self {
+    let mut groups = rspack_plugin_split_chunks::PluginOptions::from(value).cache_groups;
+
+    groups.sort_by(|a, b| a.priority.total_cmp(&b.priority));
+
```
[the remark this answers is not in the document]
JSerFeng: 👍

crates/rspack_plugin_esm_library/src/split_chunks.rs
```diff
@@ -0,0 +1,396 @@
...
+      let mut stack = get_module_deps(m, module_graph);
+      while let Some(m) = stack.pop() {
+        // if module is already in any group, skip
+        if modules_in_group.contains(&m) {
+          continue;
+        }
+
+        // if module is already in the group, skip
+        if !match_group.add_module(m) {
+          continue;
```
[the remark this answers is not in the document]
JSerFeng: 👍

JSerFeng:
@codex review

JSerFeng:
@codex review

crates/rspack_plugin_esm_library/src/split_chunks.rs
```diff
@@ -0,0 +1,402 @@
...
+    let Some(match_group) = group_modules.get_mut(&key) else {
+      continue;
+    };
+
+    let mut stack = match_group.modules.clone().into_iter().collect::<Vec<_>>();
+    while let Some(module_identifier) = stack.pop() {
+      for dep in get_module_deps(module_identifier, module_graph) {
+        if !modules_in_group.insert(dep) {
+          continue;
+        }
```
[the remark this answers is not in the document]
JSerFeng: 👎

tests/rspack-test/esmOutputCases/interop/entry-exports-type-defaut/rspack.config.js
JSerFeng: That was a mistake before, cjs is the same name with entry

crates/rspack_plugin_esm_library/src/split_chunks.rs
```diff
@@ -0,0 +1,402 @@
...
+                  .iter()
+                  .copied()
+                  .collect(),
+                cache_group_key: &group.key,
+              })
+              .await
+              {
+                Some((Either::Left(name), module_identifier))
+              } else {
+                Some((Either::Right(index), module_identifier))
```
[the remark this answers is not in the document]
JSerFeng: fixed

Issue #12413, opened 2025-12-10
[Bug]: CircularDependencyRspackPlugin does not work


n0th1ng-else:
### System Info

  System:
    OS: macOS 15.7.2
    CPU: (12) arm64 Apple M2 Pro
    Memory: 226.81 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.1 - /Users/sen/.nvm/versions/node/v22.21.1/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 10.9.4 - /Users/sen/.nvm/versions/node/v22.21.1/bin/npm
    pnpm: 10.25.0 - /Users/sen/Library/pnpm/pnpm
    bun: 1.2.9 - /Users/sen/.bun/bin/bun
    Deno: 2.5.6 - /Users/sen/.deno/bin/deno
  Browsers:
    Chrome: 143.0.7499.41
    Firefox: 146.0
    Safari: 26.1
  npmPackages:
    @rspack/cli: ^1.6.7 => 1.6.7
    @rspack/core: ^1.6.7 => 1.6.7

### Details

I am checking the **CircularDependencyRspackPlugin** plugin on different versions of RSPack. and at least from 1.4.10 until the latest one, the plugin does not work?

I have also checked https://github.com/kialo/rspack-circular-dependency-plugin and it work as expected.

Could you please take a look? thank you!

### Reproduce link

https://github.com/n0th1ng-else/rspack-repro-issues/tree/circular-deps-plugin

### Reproduce Steps

1. Clone the repo
2. install deps
3. Run `pnpm native:plugin`, observe `Process finished with exit code 0`
4. Run `pnpm external:plugin`, see `2 ERRORS in child compilations`

In my example there is a circular dependency: `src/utils/math/bar.js -> src/utils/math/foo.js -> src/utils/math/bar.js`

The expected result (**ideally without duplicates**):

```
ERROR in × Circular dependency detected:
  │ src/utils/math/bar.js -> src/utils/math/foo.js -> src/utils/math/bar.js



1 ERROR in child compilations (Use 'stats.children: true' resp. '--stats-children' for more details)
Rspack compiled with 4 errors in 89 ms
 ELIFECYCLE  Command failed with exit code 1.

Process finished with exit code 1
```

n0th1ng-else:
[...]

PR #12552, opened 2025-12-24
fix: remove allowAsyncCycles option and detect all dependencies of modules

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/12413

This pull request use `get_outgoing_connections` to get all connections for cycle detection. And removed the `allowAsyncCycles` which is never implemented.

To support `allowAsyncCycles`, we may need a big refactoring of `CircularDependencyRspackPlugin` so I think we should handle it in the future.

## Checklist


chenjiahan (approved):
LGTM 👍

PR #12564, opened 2025-12-26
test: cleanup diff comments

LingyuCoder:
## Summary

This PR cleans up outdated DIFF comments in test files that were previously used to mark differences from webpack. These comments are no longer needed as the tests have been updated to match rspack's behavior. Additionally, this PR adds the rspack.config.js file for the case-sensitive test case and removes the test.filter.js file that was previously used to skip this test.

## Checklist

### DIFF
--- a/tests/rspack-test/normalCases/errors/case-sensitive/test.filter.js
+++ b/tests/rspack-test/normalCases/errors/case-sensitive/test.filter.js
@@ -1,0 +1,1 @@
+module.exports = () => require("os").platform() !== "linux";


tests/rspack-test/normalCases/parsing/context/warnings.js
```diff
@@ -1,14 +1,5 @@
...
-	],
-	[/Critical dependency/
-		// CHANGE: not support moduleName
-		// , { moduleName: /templateLoaderIndirect\.js/ }
-	],
-	[/Critical dependency/
-		// CHANGE: not support moduleName
-		// , { moduleName: /templateLoaderIndirect\.js/ }
-	]
+	[/Critical dependency/, { moduleName: /templateLoader\.js/ }],
```
9aoy, line 2: `{ moduleName: /templateLoader\.js/ }` is a comment?

LingyuCoder: our stats factory supports `error.moduleName`, so this filter can be applied

PR #12565, opened 2025-12-26
docs: fix childAssets typo in ChunkGroup object

henryqdineen:
## Summary

I noticed the `StatsChunkGroup.children` prop is duplicated in the docs. It turns out that the second one should be `childAssets`

https://github.com/web-infra-dev/rspack/blob/0da65999fca4986897b134c2a9a472e0f555e78e/packages/rspack/src/stats/statsFactoryUtils.ts#L18-L25

Thanks!

## Checklist

### DIFF
--- a/website/docs/en/api/javascript-api/stats-json.mdx
+++ b/website/docs/en/api/javascript-api/stats-json.mdx
@@ -296,7 +296,7 @@ type StatsChunkGroup = {
     prefetch?: Array<StatsChunkGroup>;
   };
   // Assets of ordered children chunk groups, order by preload/prefetch
-  children?: {
+  childAssets?: {
     // preload assets
     preload?: Array<string>;
     // prefetch assets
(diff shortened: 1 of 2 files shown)


website/docs/en/api/javascript-api/stats-json.mdx
```diff
@@ -295,13 +295,6 @@ type StatsChunkGroup = {
     // prefetch children chunk groups
     prefetch?: Array<StatsChunkGroup>;
   };
-  // Assets of ordered children chunk groups, order by preload/prefetch
-  children?: {
```
chenjiahan, line 299: It seems like a typo, the `children` key should be `childAssets`, see:

https://github.com/web-infra-dev/rspack/blob/0da65999fca4986897b134c2a9a472e0f555e78e/packages/rspack/src/stats/statsFactoryUtils.ts#L18-L25

henryqdineen: Ah. Good catch. I will update ASAP.

henryqdineen: Fixed. Thanks!

chenjiahan (approved):
Thank you!

PR #12583, opened 2025-12-30
fix(cli): apply default lazy compilation value based on user config

stormslowly:
## Summary
The `compiler.options` is already set to default by rspack/core. We should load the user configuration in the CLI for future inspection.

## Related links

- https://github.com/web-infra-dev/rspack/pull/12580

## Checklist

### DIFF
--- a/packages/rspack-cli/src/commands/serve.ts
+++ b/packages/rspack-cli/src/commands/serve.ts
@@ -93,13 +93,11 @@ export class ServeCommand implements RspackCommand {
           !compiler.platform.electron &&
           !compiler.platform.webworker;
 
-        if (isWebAppOnly) {
-          if (userConfig.lazyCompilation === undefined) {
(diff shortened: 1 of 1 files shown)


packages/rspack-cli/src/commands/serve.ts
```diff
@@ -89,8 +93,8 @@ export class ServeCommand implements RspackCommand {
           !compiler.platform.electron &&
           !compiler.platform.webworker;
 
-        if (isWebAppOnly) {
-          compiler.options.lazyCompilation ??= {
+        if (isWebAppOnly && userConfig.lazyCompilation === undefined) {
+          compiler.options.lazyCompilation = {
```
chenjiahan, line 97: Can we change the default `lazyCompilation` in `buildConfig`? So we do not need to expose extra methods.

PR #12604, opened 2025-12-31
ci: only download bindings artifacts during release

stormslowly:
## Summary
1. Download bindings-* pattern artifacts, so new added artifacts will not break npm binding release.
2. Disable all fast-fail; test actions are flaky, fail fast results lots reruns.

Verified: https://github.com/web-infra-dev/rspack/actions/runs/20613663676/job/59204899567

## Checklist

### DIFF
--- a/.github/workflows/preview-commit.yml
+++ b/.github/workflows/preview-commit.yml
@@ -35,7 +35,7 @@ jobs:
     needs: [get-runner-labels, check-changed]
     if: needs.check-changed.outputs.changed == 'true'
     strategy:
-      fail-fast: true # for better utilize ci runners
+      fail-fast: false
       matrix:
         array:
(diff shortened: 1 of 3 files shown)


.github/workflows/preview-commit.yml
```diff
@@ -35,7 +35,7 @@ jobs:
     needs: [get-runner-labels, check-changed]
     if: needs.check-changed.outputs.changed == 'true'
     strategy:
-      fail-fast: true # for better utilize ci runners
+      fail-fast: false
```
chenjiahan, line 38: should we disable all fail-fast?

stormslowly: Yes，due to tests are flaky and the build-test  process takes so long time. If some job canceled due to other tests failure. the job that would success, will run again from scratch.

PR #12632, opened 2026-01-05
feat(rstest): add preserveNewUrl option to keep new URL untouched

fi3ework:
## Summary

- Add `preserveNewUrl` option to RstestPlugin (default: `false`)
- When enabled, matched new URL expression like `new URL("*.wasm", import.meta.url)` expressions are preserved in output, this is useful in testing case, especially for following code:

   ```js
   const path = new URL('a.wasm')
   const binary = fs.readFile(path)
   ```
- Non-matched assets continue to be processed normally

## Checklist


fi3ework:
@copilot fix the failed CI.

9aoy (approved):

Issue #12612, opened 2026-01-04
rspack@1.7: `compiler.webpack` is not a function


axetroy:
`compiler.webpack` has been remove after upgrade to rspack 1.7.x from 1.6.0

I've searched through release announcements

but couldn't find anything related. I think it was removed accidentally?

before rspack 1.6.0

```js
typeof compiler.webpack // 'function'
typeof compiler.rspack // 'function'
```

after rspack 1.7.0

```js
typeof compiler.webpack // 'object'
typeof compiler.rspack // 'object'
```

chenjiahan:
Related: https://github.com/web-infra-dev/rspack/pull/12371/files#diff-5eba14dafadd5b7ee447411919692ad06f841b6b93770e131cc6b364771f68b5R249-R257

@LingyuCoder cc

axetroy:
> Related: [#12371 (files)](https://github.com/web-infra-dev/rspack/pull/12371/files#diff-5eba14dafadd5b7ee447411919692ad06f841b6b93770e131cc6b364771f68b5R249-R257)
> 
> [@LingyuCoder](https://github.com/LingyuCoder) cc

it's a BREAKING CHANGE and it was not declared in the release announcements.

chenjiahan:
It seems like a bug and can be fixed.

PR #12636, opened 2026-01-06
fix: should export rspack function from `compiler.rspack`

LingyuCoder:
## Summary

fix https://github.com/web-infra-dev/rspack/issues/12612

The RuntimeGlobals should be different between `compiler.rspack` and the exports of `@rspack/core`. So we need to create a new exports object for each compiler and should not just assign the new RuntimeGlobals to exports to prevent overriding it.

## Checklist

### DIFF
--- a/packages/rspack/src/Compiler.ts
+++ b/packages/rspack/src/Compiler.ts
@@ -255,8 +255,7 @@ class Compiler {
     const compilerFn = function (...params: Parameters<typeof rspackFn>) {
       return rspackFn(...params);
     };
-    const compilerRspack = Object.assign(compilerFn, {
-      ...rspackExports,
+    const compilerRspack = Object.assign(compilerFn, rspackExports, {
       RuntimeGlobals: compilerRuntimeGlobals,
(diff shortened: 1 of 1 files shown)


packages/rspack/src/Compiler.ts
```diff
@@ -250,14 +252,17 @@ class Compiler {
...
-    this.rspack = {
-      ...rspack,
+    const compilerFn = function (...params: Parameters<typeof rspackFn>) {
+      return rspackFn(...params);
+    };
+    const compilerRspack = Object.assign(compilerFn, {
+      ...rspackExports,
       RuntimeGlobals: compilerRuntimeGlobals,
-    } as typeof rspack;
+    }) as unknown as typeof rspack;
```
chenjiahan: We can avoid creating a temp large object:

```js
const compilerRspack = Object.assign(
  compilerFn,
  rspackExports,
  { RuntimeGlobals: compilerRuntimeGlobals },
) as unknown as typeof rspack;
```

chenjiahan (approved):

PR #12643, opened 2026-01-06
fix: incremental rebuild no code generation entry panic

jerrykingxyz:
## Summary

Incremental rebuild at specified stages are disabled under the following conditions:

1. Manually disabled by the user. In this case, rspack will not move the artifacts from the previous compiler during rebuild.

2. Some plugins disable incremental builds at certain stages. For example, FlagDependencyUsagePlugin disables incremental builds for MODULES_HASHES. In this case, rspack will perform a full build based on the artifacts from the previous build, resulting in some content being retained and causing bugs. Please refer to the test cases for details.

https://github.com/web-infra-dev/rspack/blob/2ed157622959cf517c3d63d35448d77ccb182f85/crates/rspack_plugin_javascript/src/plugin/flag_dependency_usage_plugin.rs#L498

## Checklist

### DIFF
--- a/tests/rspack-test/watchCases/recover-from-error/delete-file/1/warnings.js
+++ b/tests/rspack-test/watchCases/recover-from-error/delete-file/1/warnings.js
@@ -1,1 +1,0 @@
-module.exports = [/FlagDependencyUsagePlugin/];
--- a/tests/rspack-test/watchCases/recover-from-error/delete-file/2/warnings.js
+++ b/tests/rspack-test/watchCases/recover-from-error/delete-file/2/warnings.js
@@ -1,1 +1,0 @@
-module.exports = [/FlagDependencyUsagePlugin/];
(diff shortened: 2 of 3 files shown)


crates/rspack_core/src/compilation/mod.rs
```diff
@@ -1850,6 +1851,7 @@ impl Compilation {
 
       modules
     } else {
+      self.cgm_hash_artifact.clear();
```
ahabhgk, line 1854: The clear should place at `disable_passes`, I don't know why the cleanup logic was removed in this PR: https://github.com/web-infra-dev/rspack/pull/12576/changes#diff-c43afdae69552b813a5f3557b972a46f3a40199f1abfc16ba0c5ace27af6618fR498

ahabhgk:
I guess after revert this https://github.com/web-infra-dev/rspack/pull/12576/changes#diff-c43afdae69552b813a5f3557b972a46f3a40199f1abfc16ba0c5ace27af6618fR498, the test will work as expected

PR #12657, opened 2026-01-07
fix(context-replacement): only apply changes when regex matches

### LINKED ISSUE #12656: [Bug]: Context Replacement Plugin applies changes even when the regex fails to match
### System Info

System:
    OS: macOS 26.1
    CPU: (14) arm64 Apple M4 Max
    Memory: 166.86 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.12.0 - /Users/agneymenon/.local/share/mise/installs/node/24.12.0/bin/node
    Yarn: 1.22.22 - /Users/agneymenon/.local/share/mise/installs/node/24.12.0/bin/yarn
    npm: 11.6.2 - /Users/agneymenon/.local/share/mise/installs/node/24.12.0/bin/npm
    pnpm: 10.26.2 - /Users/agneymenon/.local/share/mise/installs/node/24.12.0/bin/pnpm
    bun: 1.3.2 - /Users/agneymenon/.bun/bin/bun
    Watchman: 2025.05.26.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome Canary: 145.0.7618.0
    Safari: 26.1
  npmPackages:
    @rspack/cli: ^1.7.1 => 1.7.1 
    @rspack/core: ^1.7.1 => 1.7.1 

### Details

If `ContextReplacementPlugin` is added to a configuration, it applies changes to all contexts even if regex does not match. 

For example,

```
plugins: [
   new ContextReplacementPlugin(
      /^components$/,
      new RegExp(`src/components/.*`),
    ),
]
```

else where:
```
// This will not work anymore.
const context = require.context("./assets");
console.log(context.keys());
```

### Reproduce link

https://github.com/agneym/rspack-context-replacement-plugin-issue-reproduction

### Reproduce Steps

The reproduction repository has two configurations:

1. rspack.config.ts - minimal configuration
2. rspack-with-ctx-plugin - same configuration as above with context replacement plugin

I have commited builds with both of these for easy reproduction:

1. Normal build: https://github.com/agneym/rspack-context-replacement-plugin-issue-reproduction/blob/main/dist/main.js - Note that react.svg file is discovered from assets.
2. Ctx replacement plugin build - https://github.com/agneym/rspack-context-replacement-plugin-issue-reproduction/blob/main/dist-with-ctx-plugin/main.js - Above file is no longer discovered.

You can also run development servers to check the console logs in browser:

1. Normal dev server: `pnpm run dev`
2. Context replacement plugin dev server: `pnpm run dev:with-ctx-plugin`

agneym:
## Summary
- Fixed bug where `ContextReplacementPlugin` applied transformations to all `require.context()` calls regardless of regex match
- The regex check was incorrectly placed inside the `new_content_resource` conditional, causing `recursive`, `regExp`, and `critical` modifications to apply to non-matching contexts
- Now all transformations only apply when the configured regex successfully matches the request

## Test plan
- Added test case `context-replacement/non-matching-regex` that verifies non-matching contexts are not affected
- Existing context-replacement tests should continue to pass

Closes #12656

### DIFF
--- a/crates/rspack_plugin_module_replacement/src/context_replacement.rs
+++ b/crates/rspack_plugin_module_replacement/src/context_replacement.rs
@@ -43,10 +43,10 @@ impl ContextReplacementPlugin {
 
 #[plugin_hook(ContextModuleFactoryBeforeResolve for ContextReplacementPlugin)]
 async fn cmf_before_resolve(&self, mut result: BeforeResolveResult) -> Result<BeforeResolveResult> {
-  if let BeforeResolveResult::Data(data) = &mut result {
-    if self.resource_reg_exp.test(&data.request)
-      && let Some(new_content_resource) = &self.new_content_resource
-    {
+  if let BeforeResolveResult::Data(data) = &mut result
+    && self.resource_reg_exp.test(&data.request)
+  {
+    if let Some(new_content_resource) = &self.new_content_resource {
       data.request = new_content_resource.clone();
     }
     if let Some(new_content_recursive) = self.new_content_recursive {
--- a/tests/rspack-test/configCases/context-replacement/non-matching-regex/assets/file1.js
+++ b/tests/rspack-test/configCases/context-replacement/non-matching-regex/assets/file1.js
@@ -1,0 +1,1 @@
+module.exports = "asset1";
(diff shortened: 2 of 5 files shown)


SyMind (approved):

chenjiahan:
Thank you! ❤️

agneym:
@chenjiahan @SyMind rebased with main branch since some unexpected test failures

SyMind (approved):

PR #12663, opened 2026-01-07
perf(deps)!: mark mf runtime tools as optional dependency

ahabhgk:
## Summary

Remove dependency @module-federation/runtime-tools, mark it as peer dependency instead

## Checklist


packages/rspack/src/container/ModuleFederationPlugin.ts
```diff
@@ -292,9 +292,21 @@ function getPaths(options: ModuleFederationPluginOptions): RuntimePaths {
...
 
-  const runtimeToolsPath =
-    options.implementation ??
-    require.resolve('@module-federation/runtime-tools');
+  let runtimeToolsPath: string;
+  if (options.implementation) {
+    runtimeToolsPath = options.implementation;
+  } else {
+    try {
+      runtimeToolsPath = require.resolve('@module-federation/runtime-tools');
```
chenjiahan: Should we prioritize resolving this package from the user's context directory? We could set an option like `paths: [context, __dirname]`.

ahabhgk: `paths: [context]` should be good enough

Issue #12649, opened 2026-01-06
[Bug]: LazyCompilation causes ENAMETOOLONG


kwonoj:
### System Info

System:
    OS: macOS 26.2
    CPU: (12) arm64 Apple M3 Pro
    Memory: 189.67 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.15.1 - /Users/ohjoong.kwon/.proto/tools/node/20.15.1/bin/node
    Yarn: 4.5.1 - /Users/ohjoong.kwon/.proto/shims/yarn
    npm: 10.7.0 - /Users/ohjoong.kwon/.proto/tools/node/20.15.1/bin/npm
    pnpm: 10.4.0 - /Users/ohjoong.kwon/.proto/shims/pnpm
  Browsers:
    Chrome: 143.0.7499.170
    Safari: 26.2
  npmPackages:
    @rspack/cli: ^1.7.1 => 1.7.1 
    @rspack/core: ^1.7.1 => 1.7.1 
    @rspack/plugin-react-refresh: ^1.5.3 => 1.5.3 

### Details

With latest rspack which enables lazy compilation by default, some deeply nested path can cause ENAMETOOLONG error for the url.

It is possible to workaround by disabling lazy compilation but there's no way to opt in.

### Reproduce link

https://github.com/kwonoj/lazy-uri-long-repro

### Reproduce Steps

Clone / npm run dev.

stormslowly:
@kwonoj 

it seems that the using query string to hold module ids is not good for this case.
I will using some other ways to pass the ids.

thanks a lot

mattmoreira:
Related issue:
https://github.com/web-infra-dev/rspack/issues/9829

PR #12678, opened 2026-01-08
refactor(lazy-compilation): use POST request to transfer  ids of active modules

stormslowly:
## Summary

The original method of passing module IDs to the active state via URL query strings may encounter limitations related to browser/server constraints or URL length. Therefore, we are refactoring this process to pass those IDs through the POST body instead.

To handle potential bursts of requests when activating modules, we will implement a request queue during the activation process. The steps are as follows:

1. During the activation request for lazy modules, all active requests will be buffered in a `compiling` set.
2. After the Hot Module Replacement (HMR) for the previous request is completed, we will proceed to make the next active request, including all buffered module IDs. Then, we will return to step 1 to manage further requests.

## Related Links

Closes [Issue #12649](https://github.com/web-infra-dev/rspack/issues/12649).

## Checklist

### DIFF
--- a/packages/rspack/hot/lazy-compilation-node.js
+++ b/packages/rspack/hot/lazy-compilation-node.js
@@ -1,4 +1,79 @@
 var urlBase = decodeURIComponent(__resourceQuery.slice(1));
+var compiling = new Set();
+var errorHandlers = new Set();
+
+/** @type {import("http").ClientRequest | undefined} */
+var pendingRequest;
+/** @type {boolean} */
+var hasPendingUpdate = false;
+
+function sendRequest() {
+  if (compiling.size === 0) {
+    hasPendingUpdate = false;
+    return;
+  }
+
+  var modules = Array.from(compiling);
+  var data = modules.join('\n');
+
+  var httpModule = urlBase.startsWith('https')
+    ? require('https')
+    : require('http');
+
+  var request = httpModule.request(
+    urlBase,
+    {
+      method: 'POST',
+      agent: false,
+      headers: {
+        'Content-Type': 'text/plain',
+      },
+    },
+    function (res) {
+      pendingRequest = undefined;
+      if (res.statusCode < 200 || res.statusCode >= 300) {
+        var error = new Error(
+          'Problem communicating active modules to the server: HTTP ' +
+            res.statusCode,
+        );
+        errorHandlers.forEach(function (onError) {
+          onError(error);
+        });
+      }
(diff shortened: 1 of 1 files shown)


chenjiahan (commented):
PR title: `refactor` -> `refactor(lazy-compilation)`

packages/rspack/hot/lazy-compilation-web.js
```diff
@@ -1,44 +1,70 @@
...
+    var error = new Error('Problem communicating active modules to the server');
+    errorHandlers.forEach(function (onError) {
+      onError(error);
+    });
+  };
+
+  xhr.send(data);
+};
+
+var sendActiveRequest = function sendActiveRequest() {
```
chenjiahan:
```suggestion
function sendActiveRequest() {
```

packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
```diff
@@ -139,7 +130,6 @@ export const lazyCompilationMiddleware = (
   const activeModules: Set<string> = new Set();
 
   const options = {
-    // TODO: remove this when experiments.lazyCompilation is removed
```
chenjiahan, line 142: Why the TODOs removed in this PR

packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
```diff
@@ -173,24 +163,67 @@ function applyPlugin(
...
+    if (req.method === 'POST') {
+      try {
+        modules = await readModuleIdsFromBody(req);
+      } catch (err) {
+        logger.error('Failed to parse request body: ' + err);
+        res.writeHead(400);
+        res.end('Bad Request');
+        return;
+      }
+    } else {
```
chenjiahan: Should we handle non-POST requests?

stormslowly: If there are mismatches between the older lazyCompilation runtime code and the newer lazyCompilation middleware, I want them to be compatible.

stormslowly: lazyCompilation's runtime code and middleware all comes from `@rspack/core`, so that will be very rare case that mixing different versions. 
I will delete the non-post branch.

chenjiahan (approved):

PR #12686, opened 2026-01-09
feat: add `requireAlias` option to control require variable renaming

LingyuCoder:
## Summary

This PR introduces a new experimental parser option `requireAlias` for JavaScript modules, which controls whether the `require` identifier can be renamed to another variable.

By default, `requireAlias` is set to `true`, allowing users to rename `require` to custom variables (e.g., `var cjsRequire = require`). When set to `false`, the `require` identifier cannot be renamed, preserving its original behavior.

This option provides more flexibility for users who want to control how CommonJS imports are handled in their builds, especially in scenarios where preventing `require` renaming is necessary for compatibility or code generation purposes.

## Checklist

### DIFF
--- a/packages/rspack/src/config/types.ts
+++ b/packages/rspack/src/config/types.ts
@@ -1163,10 +1163,7 @@ export type JavascriptParserOptions = {
   overrideStrict?: 'strict' | 'non-strict';
 
   /**
-   * Control whether the CommonJS `require` function may be renamed to other identifiers
-   * (for example via aliasing or destructuring) when parsing modules.
-   * Set this to `false` to disallow renaming `require`, which can be useful for stricter
-   * analysis or when relying on `require` remaining a well-known identifier.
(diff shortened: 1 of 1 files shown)


packages/rspack/src/config/types.ts
```diff
@@ -1162,6 +1162,15 @@ export type JavascriptParserOptions = {
...
   overrideStrict?: 'strict' | 'non-strict';
 
+  /**
+   * Control whether the CommonJS `require` function may be renamed to other identifiers
+   * (for example via aliasing or destructuring) when parsing modules.
+   * Set this to `false` to disallow renaming `require`, which can be useful for stricter
+   * analysis or when relying on `require` remaining a well-known identifier.
+   * @default true
+   */
+  requireRename?: boolean;
```
chenjiahan: How about `parser.requireAlias`

PR #12689, opened 2026-01-09
Refactor: introduce transformImport to replace swc-plugin-import

JSerFeng:
## Summary

Introduce `rspackExperiments.transformImport`, it has the same capability as `rspackExperiments.import`, but with more easy-to-use API.

See output here, only the first item has specifier, the rest items are just render as normal `import '{{source}}'`

```
import { specifier } from "{{output[0]}}"
import "{{output[1]}}"
```

Example:

```js title="rspack.config.mjs"
export default {
  module: {
    rules: [
      {
        test: /\.js$/,
        loader: 'builtin:swc-loader',
        options: {
          rspackExperiments: {
            transformImport: [
              {
                source: 'lib',
                namedImport: false,
                output: [
                  'lib/{{ kebabCase filename }}',
                  'lib/{{ kebabCase filename }}/style.css',
                ],
              },
            ],
          },
        },
      },
    ],
  },
};
```

```js
import { MyButton, MyInput } from 'lib';

// 👇

import MyButton from 'lib/my-button';
import 'lib/my-button/style.css';
import MyInput from 'lib/my-input';
import 'lib/my-input/style.css';
```

## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/options.rs
+++ b/crates/rspack_loader_swc/src/options.rs
@@ -2,7 +2,7 @@ use rspack_cacheable::{
   cacheable,
   with::{AsRefStr, AsRefStrConverter},
 };
-use rspack_swc_plugin_import::{ImportOptions, RawImportOptions};
+use rspack_swc_plugin_import::{ImportOptions, RawImportOptions, TransformImportOptions};
 use serde::Deserialize;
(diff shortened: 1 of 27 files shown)


JSerFeng:
@chenjiahan @ityuany what do you think ?

PR #12696, opened 2026-01-12
fix: dynamic entry panic when disable build cache

SyMind:
## Summary

fix: dynamic entry panic when disable build cache

We should use `internal::try_dependency_by_id` to check if  dependency exist in dynamic entry plugin.

## Checklist

### DIFF
--- a/crates/rspack_plugin_dynamic_entry/src/lib.rs
+++ b/crates/rspack_plugin_dynamic_entry/src/lib.rs
@@ -1,10 +1,12 @@
 use std::sync::Arc;
 
+use atomic_refcell::AtomicRefCell;
 use derive_more::Debug;
-use futures::{future::BoxFuture, lock::Mutex};
+use futures::future::BoxFuture;
 use rspack_core::{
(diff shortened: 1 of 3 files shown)


crates/rspack_plugin_dynamic_entry/src/lib.rs
```diff
@@ -55,45 +57,64 @@ async fn make(&self, compilation: &mut Compilation) -> Result<()> {
...
   let decs = entry_fn().await?;
 
-  let mut imported_dependencies = self.imported_dependencies.lock().await;
-  let mut next_imported_dependencies: FxHashMap<Arc<str>, FxHashMap<EntryOptions, DependencyId>> =
-    Default::default();
+  if compilation
+    .incremental
+    .mutations_readable(IncrementalPasses::MAKE)
+  {
+    let mut imported_dependencies = self.imported_dependencies.borrow_mut();
```
quininer, line 64: sync locks that cross the await point appear dangerous?

SyMind: We can guarantee that `imported_dependencies` in plugin will never mutably borrow the contents concurrently with immutable borrows.

PR #12702, opened 2026-01-12
revert: "feat: add warning when requireAlias is disabled"

LingyuCoder:
Reverts web-infra-dev/rspack#12700


chenjiahan (approved):
Reverted because it may emit unexpected warnings when using libraries like moment.js

PR #12722, opened 2026-01-14
feat!: remove `experiments.topLevelAwait`

LingyuCoder:
## Summary

The Rspack [experiments.topLevelAwait](/config/experiments#experimentstoplevelawait) option was used to control support for top-level await, and it has always been enabled by default.
After careful observation, we found no real-world scenarios where disabling top-level await was necessary.

As a result, this option has been deprecated and is planned for removal in Rspack 2.0, at that point top-level await support can no longer be disabled.

## Checklist

### DIFF
--- a/crates/rspack/tests/snapshots/defaults__default_options.snap
+++ b/crates/rspack/tests/snapshots/defaults__default_options.snap
@@ -1514,7 +1514,6 @@ CompilerOptions {
                 MAKE | EMIT_ASSETS,
             ),
         },
-        top_level_await: true,
         css: false,
         lazy_barrel: true,
         defer_import: false,


website/docs/zh/config/deprecated-options.mdx
```diff
@@ -35,39 +35,6 @@ import { ApiMeta } from '../../../components/ApiMeta';
...
-};
-```
-
-:::warning
-该配置项已经废弃，将在 Rspack v2.0 中被移除。
-
-未来 `Top-level await` 特性将会始终开启，请从 Rspack 配置中移除此选项。
-:::
-
-## experiments.lazyCompilation
```
LingyuCoder, line 61: forgot to remove this doc in previous PR, so remove it in this

PR #12733, opened 2026-01-14
refactor!: drop CommonJS build and transition to pure ESM package

Timeless0911:
## Summary

- `@rspack/core` and `@rspack/cli` is now published as a pure ESM package, and its CommonJS build output has been removed.

  In Node.js v20.19.0 and later, the runtime natively supports loading ESM modules via `require()`. As a result, for most projects that use Rspack through its JavaScript API, this change should have no practical impact and does not require any code modifications.

  While we have ensured full compatibility for native Node.js usage, please note that tooling environments (such as TypeScript, test frameworks, or bundlers) might encounter breaking changes if you import `@rspack/core` using import syntax in an CommonJS project (`type: "module" not set`) before. You may need to adjust your code usage, `tsconfig.json`, or the configurations of your build/test tools to accommodate this change.

- `module.hot` has been replaced by `import.meta.webpackHot` and `hot/*.js` recommend to use named export in Rspack 2.0.
- tests has been updated to compatible with pure esm transition

## TODO

- `@rspack/browser` has not been refactored which will be done in later PR
- Remove default export in `hot/*.js`

## Checklist

### DIFF
--- a/packages/rspack/hot/dev-server.js
+++ b/packages/rspack/hot/dev-server.js
@@ -4,9 +4,9 @@
 */
 /* globals __webpack_hash__ */
 
-import log, { formatError } from './log.js';
-import hotEmitter from './emitter.js';
-import logApplyResult from './log-apply-result.js';
+import { log, formatError } from './log.js';
+import { emitter as hotEmitter } from './emitter.js';
+import { logApplyResult } from './log-apply-result.js';
 
 if (import.meta.webpackHot) {
   /** @type {undefined|string} */
--- a/packages/rspack/hot/emitter.js
+++ b/packages/rspack/hot/emitter.js
@@ -20,5 +20,6 @@ EventEmitter.prototype.emit = function (eventName) {
 
 var emitter = new EventEmitter();
 
+// TODO: remove default export when rspack-dev-server refactored
 export default emitter;
-export { emitter as 'module.exports' };
+export { emitter };
--- a/packages/rspack/hot/log-apply-result.js
+++ b/packages/rspack/hot/log-apply-result.js
@@ -3,7 +3,7 @@
 	Author Tobias Koppers @sokra
 */
 
-import log, { groupCollapsed, groupEnd } from './log.js';
+import { log, groupCollapsed, groupEnd } from './log.js';
 
 /**
  * @param {(string | number)[]} updatedModules updated modules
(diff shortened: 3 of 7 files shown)


packages/rspack/rslib.browser.config.mts
```diff
@@ -220,6 +220,12 @@ async function getModuleFederationRuntimeCode() {
   const { code: downgradedRuntime } = await swc.transform(runtime, {
     jsc: {
       target: 'es2015',
+      parser: {
+        syntax: 'ecmascript',
+      },
+    },
+    module: {
```
Timeless0911, line 228: Make `@rspack/browser` build success, pure esm refactor need to be done in later PR.

packages/rspack-cli/src/utils/rspackCore.ts
```diff
@@ -1,12 +1 @@
...
- * Currently, Rspack only provides a CJS bundle, so we use require to load it
- * for better startup performance.
- * https://github.com/nodejs/node/issues/59913
- */
-// @ts-expect-error can be removed after add `type: "module"` to package.json
-const require = createRequire(import.meta.url);
-const rspack: (typeof import('@rspack/core'))['rspack'] = require('@rspack/core');
-
-export { rspack };
+export { rspack } from '@rspack/core';
```
chenjiahan: this module can be removed

packages/rspack-cli/tsconfig.json
```diff
@@ -3,7 +3,9 @@
   "include": ["src"],
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src"
+    "rootDir": "src",
+    "module": "ES2022",
```
chenjiahan, line 7: ES2023

packages/rspack/hot/dev-server.js
```diff
@@ -3,15 +3,19 @@
 	Author Tobias Koppers @sokra
 */
 /* globals __webpack_hash__ */
-if (module.hot) {
+
+import log, { formatError } from './log.js';
+import hotEmitter from './emitter.js';
+import logApplyResult from './log-apply-result.js';
```
chenjiahan: Can we use named import instead of default import

packages/rspack/hot/lazy-compilation-node.js
```diff
@@ -1,19 +1,19 @@
+import http from 'http';
+import https from 'https';
```
chenjiahan: I prefer to lazy load these two modules to improve Node.js startup time

packages/rspack/rslib.browser.config.mts
chenjiahan: Can we add a `TODO`

Issue #10479, opened 2024-11-11
[Bug]: `require.resolve` was compiled incorrectly


chenjiahan:
### Version

```sh
Rslib 0.0.16
```

### Details

- Input:

```ts
// index.ts
import { require } from './require';

console.log(require.resolve('@rslib/core'));

// require.ts
import { createRequire } from 'module';

export const require = createRequire(import.meta.url);
```

- Output:

```js
import * as __WEBPACK_EXTERNAL_MODULE_module__ from "module";
const require_require = (0, __WEBPACK_EXTERNAL_MODULE_module__.createRequire)(import.meta.url);
console.log(require_require('@rslib/core'));
```

- Expected output:

```diff
import * as __WEBPACK_EXTERNAL_MODULE_module__ from "module";
const require_require = (0, __WEBPACK_EXTERNAL_MODULE_module__.createRequire)(import.meta.url);
- console.log(require_require('@rslib/core'));
+ console.log(require_require.resolve('@rslib/core'));
```

### Reproduce link

https://github.com/chenjiahan/rslib-repro-esm-require-resolve

### Reproduce Steps

1. pnpm i
2. pnpm build

LingyuCoder:
This issue was fixed by #12773. #14783 further fixes the case where `requireResolve` is set to `false`.

I also tested the original reproduction locally with the latest stable version, and the issue no longer reproduces. The generated output correctly preserves `require.resolve(...)`.

I think this issue can be closed now.

PR #12773, opened 2026-01-18
fix: `require.resolve()` replaced as `require()`

intellild:
## Summary
*`require.resolve()` and `require.resolveWeak` should not be replaced as `require()`
* if `require` is a local variable, should not be transpiled as commonjs import

## Related links

[#10479](https://github.com/web-infra-dev/rspack/issues/10479)

## Checklist


intellild:
It seems that the fix is incorrect, needs further investigation

LingyuCoder:
LGTM, thanks

LingyuCoder (approved):

chenjiahan:
Thank you!

PR #12780, opened 2026-01-19
feat!: default targets for loader/plugin derived by rspack target, part 2

ahabhgk:
## Summary
1. change `platforms` to `targets`, and it's no longer a `string[]` (`['chrome 100', 'chrome 101', 'chrome 102', ...]`),  it's a targets object now (`{ chrome: '100', safari: '7' }`) which is more compatible with other tools' targets
    - https://swc.rs/docs/configuration/compilation#envtargets
    - https://esbuild.github.io/api/#target
    - https://babeljs.io/docs/options#targets
    - https://lightningcss.dev/transpilation.html#browser-targets 
2. update docs
3. update create-rspack template

## Checklist

### DIFF
--- a/tests/rspack-test/statsOutputCases/dynamic-import/__snapshots__/stats.txt
+++ b/tests/rspack-test/statsOutputCases/dynamic-import/__snapshots__/stats.txt
@@ -1,4 +1,3 @@
-[baseline-browser-mapping] The data in this module is over two months old.  To ensure accurate Baseline data, please update: `npm i baseline-browser-mapping@latest -D`
 asset common.js 1.11 MiB [emitted] (name: common) (id hint: vendors)
 asset pages/home.js xx KiB [emitted] (name: pages/home)
 asset runtime.js xx KiB [emitted] (name: runtime)


packages/create-rspack/template-react-js/rspack.config.mjs
chenjiahan, line 1: The PR title starts with `docs` but it changed the code in `create-rspack`

packages/create-rspack/template-react-js/rspack.config.mjs
```diff
@@ -8,14 +8,12 @@ import { ReactRefreshRspackPlugin } from '@rspack/plugin-react-refresh';
...
 
-// Target browsers, see: https://github.com/browserslist/browserslist
-const targets = ['last 2 versions', '> 0.2%', 'not dead', 'Firefox ESR'];
-
 export default defineConfig({
   context: __dirname,
   entry: {
     main: './src/main.jsx',
   },
+  target: ['browserslist:defaults'],
```
chenjiahan: Should we change the default value? What is the motivation

ahabhgk: https://github.com/browserslist/browserslist#:~:text=defaults%3A%20Browserslist%E2%80%99s%20default%20browsers%20(%3E%200.5%25%2C%20last%202%20versions%2C%20Firefox%20ESR%2C%20not%20dead).

According to browserslist docs these two are the same, the change here is because I think defaults is shorter and cleaner, I'm ok to change it back

chenjiahan: They are not the same:

- before `['last 2 versions', '> 0.2%', 'not dead', 'Firefox ESR']` （Audience coverage: 89.9 %）
- after `> 0.5%, last 2 versions, Firefox ESR, not dead` (Audience coverage: 87.0 %, less than 80% coverage in China regions)

https://browsersl.ist/#q=defaults

Issue #12617, opened 2026-01-04
[Bug]:  svg-sprite-loader causes panic "should set in compilation first" with Rspack/Rsbuild


wxy-hh:
### System Info

System:
  OS: macOS 26
  CPU: M4
  Memory: 16
  Shell: zsh
Binaries:
  Node: v20.19.4
  pnpm: 10.26.1
npmPackages:
  @rsbuild/core: ^1.7.1
  @rsbuild/plugin-vue2: ^1.0.4
 svg-sprite-loader: ^6.0.11

### Details

@rsbuild/core从1.6.15升级到1.7.1后启动报错:Please file an issue on GitHub with the backtrace below: https://github.com/web-infra-dev/rspack/issues: panicked at crates/rspack_core/src/utils/deref_option.rs:52:26:

<img width="1780" height="676" alt="Image" src="https://github.com/user-attachments/assets/ecdc7d42-4ad3-4a74-af0d-dd060270e893" />

### Reproduce Steps

[rspack-svg-sprite-bug.zip](https://github.com/user-attachments/files/24420650/rspack-svg-sprite-bug.zip)

hardfist:
rspack not compatible with `svg-sprite-loader` and use `rspack-svg-sprite-loader` instead

PR #12782, opened 2026-01-19
chore: gracefully explain unsupported interfaces

martinjlowm:
## Summary

When Webpack plugins try to access hooks that don't exist in rspack, provide a clear error message instead of a cryptic message, e.g. "Cannot read properties of undefined (reading 'tap')" (attempting to use `pino-webpack-plugin` with rspack).

The error message explains:
- Which hook is not supported
- That this typically happens with webpack plugins
- Lists all available hooks for reference

This helps developers quickly identify webpack compatibility issues.

While I haven't tested it, this should hopefully catch https://github.com/web-infra-dev/rspack/issues/12617 as well way before the panic.

## Checklist

### DIFF
--- a/tests/rspack-test/hookCases/compilation#unsupportedHookError/basic/output.snap.txt
+++ b/tests/rspack-test/hookCases/compilation#unsupportedHookError/basic/output.snap.txt
@@ -1,0 +1,40 @@
+```js title=main.js
+(() => {
+var __webpack_modules__ = ({
+402(module) {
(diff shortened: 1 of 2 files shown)


ahabhgk (commented):
LGTM, need a test case to avoid regression

hardfist (approved):

hardfist:
@martinjlowm  Thanks for your contribution

PR #12817, opened 2026-01-22
refactor: align IncrementalPass with rspack hooks naming

hardfist:
## Summary
some of the IncrementalPass naming according to artifact but these artifact is actually not related to rspack pass directly, so change the naming to match rspack pass(which is nearly identical to hooks name)

## Checklist

### DIFF
--- a/website/docs/en/config/incremental.mdx
+++ b/website/docs/en/config/incremental.mdx
@@ -13,7 +13,7 @@ Two configuration ways are provided:
 
 1. Presets: including `boolean | 'none' | 'safe' | 'advance' | 'advance-silent'`
    - `false | 'none'`: Disable incremental, and it will not be enabled for any stage.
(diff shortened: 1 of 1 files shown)


website/docs/en/config/incremental.mdx
```diff
@@ -13,7 +13,7 @@ Two configuration ways are provided:
 
 1. Presets: including `boolean | 'none' | 'safe' | 'advance' | 'advance-silent'`
    - `false | 'none'`: Disable incremental, and it will not be enabled for any stage.
-   - `'safe'`: Enable incremental for `make` and `emitAssets` stages.
+   - `'safe'`: Enable incremental for the `buildModuleGraph` and `emitAssets` stage.
```
hardfist:
```suggestion
   - `'safe'`: Enable incremental for the `buildModuleGraph` and `emitAssets` stages.
```

PR #12839, opened 2026-01-26
fix: should not panic when access slate stats

SyMind:
## Summary

This PR ensures that accessing `Stats` instances from stale compilations in watch mode no longer causes panics, while still returning a valid stats JSON shape.

**Changes:**
- Guard `Stats` JSON generation against stale compilations by returning safe, empty `JsStatsCompilation` objects instead of calling into freed Rust-side structures.
- Short-circuit error and warning extraction in the stats factory when the given compilation is not the compiler’s `_lastCompilation`, returning empty collections for stale compilations.
- Add a dedicated `watchCases/compilation/access-stale-stats` test to verify that `stats.toJson()` and `stats.toJson({ all: true })` on stale stats remain shape-compatible and do not crash.

## Checklist

### DIFF
--- a/packages/rspack/src/Stats.ts
+++ b/packages/rspack/src/Stats.ts
@@ -100,6 +100,8 @@ export class Stats {
         if (statsCompilationMap.has(compilation)) {
           return statsCompilationMap.get(compilation)!;
         }
+        // return empty stats for stale compilation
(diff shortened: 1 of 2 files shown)


hardfist:
@copilot cherry-pick this to v1.x branch

hardfist:
@codex  cherry-pick this to v1.x branch

PR #12842, opened 2026-01-26
feat: add cache.portable configuration

jerrykingxyz:
## Summary

Add a `cache.portable` configuration to control whether it's enabled. It's disabled by default.

## Checklist

### DIFF
--- a/packages/rspack/etc/core.api.md
+++ b/packages/rspack/etc/core.api.md
@@ -687,22 +687,19 @@ export type CacheNormalized = boolean | {
 };
 
 // @public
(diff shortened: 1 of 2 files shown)


crates/node_binding/napi-binding.d.ts
```diff
@@ -1850,6 +1850,7 @@ export interface RawCacheOptionsPersistent {
   version?: string
   snapshot?: RawSnapshotOptions
   storage?: RawStorageOptions
+  portable?: boolean
```
chenjiahan, line 1853: Can we generate JSDoc for this new option

PR #12896, opened 2026-01-30
refactor: introduce PassExt trait

hardfist:
## Summary
currently Compilation holds too many fields which's only valid in specific PASS, by introudcing PASS struct(inspired by https://github.com/rust-lang/rust/tree/main/compiler/rustc_passes and https://doc.rust-lang.org/beta/nightly-rustc/rustc_mir_transform/pass_manager/trait.MirPass.html), we can move these temporary fields into PASS struct instead, PassExt also help us unify cache management and logging behavior between different pass implementation.
* support move pass only data into PASS other than compilation(like cache) which can reduce fields in compilation
* move ModuleGraphStaticArtifact to trainsient_cache cause it's only valid in current compilation

```rs
// OptimizeDependenciesPass.rs
struct OptimizeDependenciesPass();
impl PassExt for OptimizeDependenciesPass {
    async fn run_pass(&self,compilation: &mut Compilation){
          run_pass_impl(compilation).await
    }
}
// run_pass.rs
 let passes: Vec<Box<dyn PassExt>> = vec![
      // Phase 1: Build Module Graph
      Box::new(BuildModuleGraphPhasePass),
      // Phase 2: Seal & Optimize
      Box::new(SealPass),
      Box::new(OptimizeDependenciesPass),
      ...
    ];
run_passes(passes);
```

## Checklist

### DIFF
--- a/crates/rspack_core/src/compilation/after_process_assets/mod.rs
+++ b/crates/rspack_core/src/compilation/after_process_assets/mod.rs
@@ -1,0 +1,34 @@
+use async_trait::async_trait;
+
+use super::*;
+use crate::compilation::pass::PassExt;
+
+pub struct AfterProcessAssetsPass;
+
+#[async_trait]
+impl PassExt for AfterProcessAssetsPass {
+  fn name(&self) -> &'static str {
+    "after process assets"
+  }
+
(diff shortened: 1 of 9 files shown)


hardfist:
@codex review

PR #12902, opened 2026-01-30
feat: add support for readonly persistent cache

cellison-figma:
## Summary

This adds support for a `readonly` option for the persistent filesystem cache. The [equivalent webpack feature is documented here](https://webpack.js.org/configuration/cache/#cachereadonly).

A readonly cache supports CI workflows where branch builds load in a filesystem cache from the a recent ancestor commit on the trunk branch. In many cases, these branch builds happen too often and are too ephemeral to justify storing the build cache results. Adding a `readonly` option allows such builds to avoid I/O costs of writing the cache to disk.

## Checklist

### DIFF
--- a/tests/rspack-test/cacheCases/common/readonly/rspack.config.js
+++ b/tests/rspack-test/cacheCases/common/readonly/rspack.config.js
@@ -9,7 +9,7 @@ module.exports = {
   module: {
     rules: [
       {
-        test: /readonly\/file\.js/,
+        test: /file\.js$/,
         use: {
           loader: './loader.js',
           options: { count: 0 },


crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
```diff
@@ -87,6 +90,11 @@ impl BuildDeps {
   ///
   /// If any build dependencies have changed, this method will reset storage.
   pub async fn validate(&mut self) -> Result<()> {
+    if self.readonly {
+      tracing::info!("BuildDependencies: validation skipped because readonly mode is enabled");
```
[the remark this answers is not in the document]
cellison-figma: Perhaps this tracing info is not necessary

crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
```diff
@@ -87,6 +90,11 @@ impl BuildDeps {
   ///
   /// If any build dependencies have changed, this method will reset storage.
   pub async fn validate(&mut self) -> Result<()> {
+    if self.readonly {
```
jerrykingxyz: `self.readonly` should not be checked here. If the user's build dependencies have changed, the cache cannot be reused, even if readonly is enabled.

crates/rspack_core/src/cache/persistent/mod.rs
jerrykingxyz, line 81: We should add `#[cacheable(with=Skip)]` to readonly fields to ensure that the version does not change even if the user update the readonly value.

packages/rspack/src/config/normalization.ts
```diff
@@ -301,6 +301,7 @@ export const getNormalizedRspackOptions = (
           ),
         },
         portable: cache.portable,
+        readonly: cache.readonly || false,
```
jerrykingxyz: Remove `|| false` so that the default value is set by the Rust side.

crates/rspack_core/src/cache/persistent/mod.rs
```diff
@@ -126,6 +130,9 @@ impl PersistentCache {
   }
 
   async fn save(&mut self) {
+    if self.readonly {
```
jerrykingxyz: Remove it, we have already checked this in the after_compiler.

crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
jerrykingxyz: This reset is to prevent the persistent cache from reading dirty data. For readonly mode, I believe we should avoid reading data, but not clear the data. I removed this code and added a valid field to the persistent cache to reset the storage, which will be useful to you. https://github.com/web-infra-dev/rspack/commit/eef8745a75f7c8c02ac810c70e047ab51c972e4a

tests/e2e/cases/persistent-cache/readonly/index.js
```diff
@@ -0,0 +1,5 @@
+import { value } from './module.js';
```
jerrykingxyz: We can add a simple cachecase to test this. Here's one I've written; you can use cherrypick directly. https://github.com/web-infra-dev/rspack/commit/8836b14e56307cd1bb8300c85ccb99e383bce080

jerrykingxyz (commented):
You can rebase https://github.com/web-infra-dev/rspack/commits/jerry/readonly/ to get my modified code.

Please feel free to raise any questions in the PR; I will review them regularly.

tests/e2e/cases/persistent-cache/readonly/index.js
cellison-figma: Would you like me to remove this e2e test, or is it still valuable alongside the cachecase test?

cellison-figma:
@jerrykingxyz Thanks for your review and additional commits. Are there other pieces you'd like me to address or other pieces I may have missed?

tests/e2e/cases/persistent-cache/readonly/index.js
jerrykingxyz: E2E testing is too complex and is generally used when ordinary tests cannot cover. Remove it.

jerrykingxyz:
It appears some snapshots failed; you should run `npm run test -- -u` to update them. 
And the test case I added failed, you can reproduce and debug it by running `cd tests/rspack-test && npm run test:base -- -t "readonly"`.

PS. Don't forget to run `npm run build:cli:dev` to compile the local code before running the tests.

Ref: https://rspack.rs/contribute/development/testing

tests/rspack-test/cacheCases/common/readonly/rspack.config.js
```diff
@@ -0,0 +1,55 @@
...
+/** @type {import("@rspack/core").Configuration} */
+module.exports = {
+  context: __dirname,
+  cache: {
+    type: 'persistent',
+  },
+  module: {
+    rules: [
+      {
+        test: /readonly\/file\.js/,
```
jerrykingxyz: Windows testing failed; we should make this path compatible with Windows.

jerrykingxyz (approved):
@cellison-figma Thanks for your contribution.

PR #12913, opened 2026-02-02
chore(test): show cli test stdout/err when child proceess killed

stormslowly:
## Summary
For debugging why the rspack children process was killed during rspack-cli testing.

### background

When the `execa` children process is killed, the exitCode of the children process is [`undefined`](https://github.com/web-infra-dev/rspack/actions/runs/21556863904/job/62115370045?pr=12901#step:8:452).

## Checklist


stormslowly:
@codex

PR #12919, opened 2026-02-03
docs: add guide for React Server Components

SyMind:
## Summary

Adds React Server Components (RSC) documentation pages to the website (English + Chinese) and exposes them in the “tech” docs navigation.

## Checklist

### DIFF
--- a/website/docs/en/guide/tech/react.mdx
+++ b/website/docs/en/guide/tech/react.mdx
@@ -223,3 +223,7 @@ export default {
 ```
 
 For detailed usage of SVGR, please refer to [SVGR Documentation - webpack](https://react-svgr.com/docs/webpack/).
+
+## React Server Components
+
+If you need to use React Server Components, please refer to [Rspack - React Server Components](/guide/tech/rsc).
--- a/website/docs/zh/guide/tech/react.mdx
+++ b/website/docs/zh/guide/tech/react.mdx
@@ -221,3 +221,7 @@ export default {
 ```
 
 对于 SVGR 的详细用法，请参考 [SVGR 文档 - webpack](https://react-svgr.com/docs/webpack/)。
+
+## React 服务器组件
+
+如果你需要使用 React 服务器组件，请参考 [Rspack - React 服务器组件](/guide/tech/rsc)。


website/docs/en/guide/tech/rsc.mdx
chenjiahan, line 1: Can we add a link to this page in https://rspack.rs/guide/tech/react page

website/docs/en/guide/tech/rsc.mdx
```diff
@@ -0,0 +1,221 @@
...
+## How to use
+
+Rspack provides two solutions to support React Server Components:
+
+- **Using Rsbuild**: Rsbuild provides out-of-the-box support for React Server Components, allowing you to quickly create a React Server Component project.
+- **Manually Configuring Rspack**: You can refer to this document to manually add configurations related to React Server Components.
+
+## Example
+
+The [rspack-rsc-examples](https://github.com/SyMind/rspack-rsc-examples) repository contains full examples of building React Server Component applications using Rspack.
```
chenjiahan: Should we move this example repo to the https://github.com/rstackjs org?

website/docs/en/guide/tech/rsc.mdx
```diff
@@ -0,0 +1,221 @@
...
+
+# React Server Components
+
+Server Components are a new type of Component that renders ahead of time, before bundling, in an environment separate from your client app or SSR server. Rspack v2.0.0 and later have built-in full support for Server Components.
+
+## How to use
+
+Rspack provides two solutions to support React Server Components:
+
+- **Using Rsbuild**: Rsbuild provides out-of-the-box support for React Server Components, allowing you to quickly create a React Server Component project.
```
chenjiahan: Add a link to https://github.com/rstackjs/rsbuild-plugin-rsc

website/docs/en/guide/tech/rsc.mdx
```diff
@@ -0,0 +1,221 @@
...
+];
+```
+
+The loader identifies `"use client"` and `"use server"` directives. Only when the module belongs to the `Layers.rsc` layer, it uses `react-server-dom-rspack` to transform modules containing these directives into the Client References or Server References required by the React runtime.
+
+The RSC plugins also rely on the directive information identified by the loader to distinguish between Client Components and Server Components.
+
+### Configure layer
+
+In the Server Compiler, Rspack uses [layer](/config/experiments#experimentslayers) to identify modules that belong to the Server Component (RSC) environment.
```
chenjiahan: `config/experiments#experimentslayers` has been deprecated, use `/guide/features/layer`

website/docs/en/guide/tech/rsc.mdx
SyMind: I don't think it's necessary, as I've placed the React and RSC navigation items right next to each other.
image

website/docs/en/guide/tech/rsc.mdx
SyMind: done

website/docs/en/guide/tech/rsc.mdx
SyMind: done

nikhilsnayak:
It would be great to document the necessity of `'use server-entry'` directive since it is not a standard RSC directive.
It will be helpful for people coming from other frameworks like nextjs because nextjs aka turbopack doesn't implement this directive

SyMind:
@nikhilsnayak I have document the `'use server-entry'` directive. What else are you interested in?

nikhilsnayak:
From documentation what I understand is that, by marking the the application entry point as 'use server-entry', we get access to all the css and js assets required to bootstrap the application.

In case if there are multiple entry points to the application (example: nextjs supports multiple root layouts) then all the entry points must be marked with 'use server-entry' ?

Also, what happens if I don't mark the entry point with 'use server-entry' ?

In RSCs directives are optional and required only when using specific features
1. 'use client' - creates client referances
2. 'use server' - creates server references

Is 'use server-entry' mandatory if someone decides to use rspack's RSC implementation?

SyMind:
`"use server-entry"` is not mandatory. It is essentially a low-level helper provided by Rspack primarily for framework developers.

**What happens if you don't use it?** If you omit `"use server-entry"`, Rspack will not automatically inject `entryJsFiles` and `entryCssFiles` into your component. This means you would need to implement your own logic to maintain the mapping between Server Entries and Client Assets, and manually collect the CSS and JS resources required for the current route.

**Regarding Multiple Entry Points** (e.g., Next.js style) To implement a multi-entry architecture similar to Next.js (such as multiple root layouts) using Rspack's RSC plugin, the recommended approach is:

- Server Compiler: Use Dynamic Entry to add page entries programmatically. You should ensure that each entry component includes the `"use server-entry"` directive so that you can access the associated CSS and JS resources at runtime.
- Client Compiler: Add corresponding page entries via Dynamic Entry to match the server configuration.
- Optimization: Similar to Next.js, you can use the `dependOn` configuration to extract shared browser-side dependencies (like the React Runtime) into separate chunks.

**Additional Note** It is worth noting that Next.js does not treat "use client" as a code-splitting point by default. Rspack currently defaults to treating them as split points (aligning with the RSC RFC). While there is currently no configuration option exposed to toggle this behavior, implementing such a configuration would be very straightforward.

nikhilsnayak:
Thank you for the detailed explanation 🙏

> Additional Note It is worth noting that Next.js does not treat "use client" as a code-splitting point by default. Rspack currently defaults to treating them as split points (aligning with the RSC RFC). While there is currently no configuration option exposed to toggle this behavior, implementing such a configuration would be very straightforward.

Does this mean each module marked with 'use client' ends up as a seperate chunk?

SyMind:
> Does this mean each module marked with 'use client' ends up as a seperate chunk?

You can think of it that way. However, to be more precise, the final chunk structure is still controlled by Rspack's optimization strategy and can be adjusted via the `optimization.splitChunks` configuration.

website/docs/en/guide/tech/rsc.mdx
chenjiahan: The related docs should still be linked through references within the content itself, rather than relying on the site’s navigation. Because this is more friendly for AI and web crawlers.

PR #12921, opened 2026-02-03
feat(rstest): only handle global variables when globals is true

9aoy:
## Summary

only handle global variables when `globals` is true (`rstest --globals`).

https://rstest.rs/config/test/globals

```ts
rstest.mock('../src/foo')

const originalFoo = await rstest.importActual('../src/foo');

```

## Checklist

### DIFF
--- a/crates/rspack_plugin_rstest/src/parser_plugin.rs
+++ b/crates/rspack_plugin_rstest/src/parser_plugin.rs
@@ -37,8 +37,8 @@ enum ModulePathType {
   FileName,
 }
 
-#[derive(Debug, Default)]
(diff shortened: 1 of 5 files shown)


9aoy:
> Rstest plugin is per environment registered https://github.com/web-infra-dev/rstest/blob/2ec45b50bfaaed077ad16b5ef3208cc4fe255364/packages/core/src/core/plugins/basic.ts#L28. so i think it's fine to allow different config for each environment.

Sure — the `globals` option can be configured either via CLI options or directly in the test config, depending on needs.

PR #12922, opened 2026-02-03
feat: allow unknown exports to be concatenated

JSerFeng:
## Summary

Allow module whose exports info is unknown to be concatenated as well.

## Checklist


JSerFeng:
Closed as we can't concate module with unknown exports, findTarget can't know what export should use

PR #12942, opened 2026-02-04
feat: preserve tla expr for EsmLibrary

JSerFeng:
## Summary

Preserve top level await expr as is, support scope hoist async modules.

## Checklist


crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
```diff
@@ -1230,10 +1230,15 @@ impl JavascriptParser<'_> {
   }
 
   fn walk_await_expression(&mut self, expr: &AwaitExpr) {
-    if self.is_top_level_scope() {
+    let is_top_level_scope = self.is_top_level_scope();
+    if is_top_level_scope {
+      self.in_top_level_await = true;
```
ahabhgk: Why not just use `build_meta.has_top_level_await`

crates/rspack_core/src/compilation/build_chunk_graph/code_splitter.rs
```diff
@@ -1735,9 +1736,21 @@ Or do you want to use the entrypoints '{name}' and '{runtime}' independently on
...
         target.add_parent(chunk_group_ukey);
 
-        target_cgi
-          .available_modules_to_be_merged
-          .push(resulting_available_modules.clone());
+        if !self.ignore_top_level_await
+          && let Some(process_block) = &process_block
+          && let Some(block_id) = process_block.block.as_async()
+          && let Some(block) = compilation.get_module_graph().block_by_id(&block_id)
+          && block.in_top_level_await()
```
ahabhgk: What if use the `build_meta.has_top_level_await` of block's parent module, can we avoid the `in_top_level_await` thing?

crates/rspack_core/src/compilation/build_chunk_graph/pass.rs
```diff
@@ -23,7 +23,14 @@ impl PassExt for BuildChunkGraphPass {
     compilation.module_graph_cache_artifact.freeze();
     use_code_splitting_cache(compilation, |compilation| async {
       let start = logger.time("rebuild chunk graph");
-      build_chunk_graph(compilation)?;
+      let ignore_top_level_await = compilation
+        .plugin_driver
+        .build_chunk_graph_hooks
+        .ignore_top_level_await
```
ahabhgk: Can we avoid adding a new hook here?

crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
JSerFeng: has_top_level_await is a one time state, dependency and block can't know if current statement is in tla

crates/rspack_core/src/compilation/build_chunk_graph/code_splitter.rs
JSerFeng: Here is an example

```js
await import('a')
import('b')
```

`b` chunk can have normal available modules optimization while `a` doesn't

So I have to make this detection at block level

ahabhgk: then how do you handle: 
```js
const p = import("a");
// ...
await p;
```

JSerFeng:
@codex review

Issue #12097, opened 2025-11-05
[Bug]: import defer does not work as expected


chenjiahan:
### System Info

@rspack/core v1.6.1

### Details

After upgrading to @rspack/core v1.6.1, `import defer` is not working as expected; all modules that are imported with `defer` are imported at the top level:

<img width="1996" height="1544" alt="Image" src="https://github.com/user-attachments/assets/7750d3a2-c736-4219-9ae4-26c322eb2f34" />

### Reproduce link

https://github.com/web-infra-dev/rsbuild/pull/6490

### Reproduce Steps

1. git checkout import_defer_1103
2. pnpm i
3. cd packages/core
4. pnpm build

ahabhgk:
No solution to solve this for now, the output of external module (with `externalType: "module"`) need to:
1. synchronously load the external module by ESModule syntax
2. load but can't execute immediately

The only way that matches these conditions is defer import itself, so we can pass-through the syntax ([just like typescript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-9.html#support-for-import-defer:~:text=Note%20that%20import,and%20esnext.)), but it's not supported in any js engine AFAIK

chenjiahan:
Got it. Can we document this?

PR #12965, opened 2026-02-06
docs: add docs about using module external with defer import

ahabhgk:
## Summary

add docs about using module external with defer import

## Related links

resolve https://github.com/web-infra-dev/rspack/issues/12097

## Checklist

### DIFF
--- a/website/docs/en/config/experiments.mdx
+++ b/website/docs/en/config/experiments.mdx
@@ -245,12 +245,13 @@ import defer * as $ from 'jquery';
 It will generate:
 
 ```js
-import * as __WEBPACK_EXTERNAL_MODULE_jquery__ from 'jquery';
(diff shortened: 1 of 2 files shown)


website/docs/en/config/experiments.mdx
```diff
@@ -232,3 +232,25 @@ export default {
...
+For example, if `jquery` is configured as external:
+
+```js
+import defer * as $ from 'jquery';
+```
+
+It will generate:
+
+```js
+import * as __WEBPACK_EXTERNAL_MODULE_jquery__ from 'jquery';
```
chenjiahan: Rspack no longer emit `__WEBPACK_EXTERNAL_MODULE_` prefix, use `__rspack_external_` instead

PR #12982, opened 2026-02-08
fix(module-federation): encode runtime data URI and support named wrappers

BleedingDev:
## Summary
- Encode the Module Federation default runtime data URI entry in `getDefaultEntryRuntime`.
- Extend `Template.getFunctionContent` to strip both anonymous and named function wrappers.
- Add regression tests for both behaviors.

## Why
When running Module Federation with a strict Bun runtime, a raw runtime data URI request can fail with `Unhandled scheme`.

After encoding fix, another issue appears where `Template.getFunctionContent` only strips anonymous `function () {}` wrappers, but Bun can surface a named wrapper (`function moduleFederationDefaultRuntime_default() {}`), producing invalid extracted content.

## Changes
- `packages/rspack/src/container/ModuleFederationPlugin.ts`
  - Use `encodeURIComponent(content)` in the MF runtime data URI entry.
- `packages/rspack/src/Template.ts`
  - Broaden `FUNCTION_CONTENT_REGEX` to accept optional function names.
- `tests/rspack-test/Template.test.js`
  - Add coverage for named and anonymous function wrapper stripping.
- `tests/rspack-test/configCases/container-1-5/runtime-entry-data-uri-encoded/*`
  - Add a regression that asserts MF runtime entry data URI content is encoded.

## Validation
- `pnpm --filter @rspack/tests exec rstest run Template.test.js`
- `testFilter=runtime-entry-data-uri-encoded pnpm --filter @rspack/tests exec rstest run Config.part1.test.js`

Both pass locally.

## Maintainer Question
I tested this against Bun 1.3.8/1.3.9 in downstream MF repros and it resolves the failure there as well.

What testing strategy would you prefer for Bun compatibility going forward?
- Keep current Node-required CI and add Bun MF smoke checks as optional/nightly.
- Add Bun checks as required CI for selected MF suites.
- Keep Bun coverage only in downstream ecosystem CI.

I can adjust this PR to match your preferred approach.


2heal1:
Thanks for the Bun-related PR — really appreciated.

For now, MF hasn’t officially tested or supported Bun, so we’d prefer to rely on downstream CI for Bun coverage and keep this repo focused on Node/web compatibility.

chenjiahan (approved):
Thank you!

PR #12983, opened 2026-02-09
feat: rsdoctor support the side effects infos

yifancong:
## Summary
This PR adds side-effects information support to rsdoctor and wires it into the module analysis flow. As a result, it may affect module `optimization_bailout`, by changing it from a string to a struct.
**Optimization Bailout Refactor:**

- Introduced the `OptimizationBailoutItem` enum to represent optimization bailout reasons in a structured way, replacing the previous use of plain strings in `ModuleGraphModule` and related APIs. This includes a new display implementation for converting the enum to a string when needed. 
- Updated all relevant methods, data structures, and plugin logic to use `OptimizationBailoutItem` instead of `String`, including in `BuildResult`, `ModuleGraph`, and `NormalModule`. 
- Adjusted stats reporting to convert `OptimizationBailoutItem` instances to strings only at the stats/reporting boundary, maintaining structured data internally.

**Rsdoctor API Enhancements:**

- Added new structs and conversions for `RsdoctorSideEffectLocation` and `RsdoctorConnection`, and integrated them into the Rsdoctor API surface, allowing richer reporting of side effect locations and module graph connections. 
- Updated `JsRsdoctorModule` to include a `side_effects_locations` field, and added the corresponding conversion logic. 
-  Add tests for these.

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/compilation/mod.rs
+++ b/crates/rspack_binding_api/src/compilation/mod.rs
@@ -15,7 +15,7 @@ use napi_derive::napi;
 use rspack_collections::{DatabaseItem, IdentifierSet};
 use rspack_core::{
   BindingCell, BoxDependency, Compilation, CompilationId, EntryOptions, ExportsInfoArtifact,
-  FactorizeInfo, ModuleIdentifier, Reflector, rspack_sources::BoxSource,
+  FactorizeInfo, ModuleIdentifier, OptimizationBailoutItem, Reflector, rspack_sources::BoxSource,
 };
 use rspack_error::{Diagnostic, Severity, ToStringResultToRspackResultExt};
 use rspack_napi::napi::bindgen_prelude::*;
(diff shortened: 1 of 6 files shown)


crates/rspack_plugin_library/src/modern_module_library_plugin.rs
```diff
@@ -104,7 +104,7 @@ impl ModernModuleLibraryPlugin {
 
         reasons
           .iter()
-          .any(|r| r.contains("Module is an entry point"))
+          .any(|r| r.to_string().contains("Module is an entry point"))
```
LingyuCoder: can we implement `as_str()` and just get `&str` to prevent creating a new string

yifancong: It can’t be changed to `as_str` here because `r` is not a string and needs to be converted via `Display::fmt`.  OptimizationBailoutItem is enum type, OptimizationBailoutItem::SideEffects  don't have as_str api.
Instead, I changed it to directly use the `Message` from `OptimizationBailoutItem`, because when `OptimizationBailoutItem` is of the `SideEffects` struct type, it does not concatenate `"Module is an entry point"`. So here we can directly use `OptimizationBailoutItem`’s `message` value for the check.

yifancong:
> ## Rsdoctor Bundle Diff Analysis
> Found 5 projects in monorepo, 4 projects with changes.
> 
> 📊 Quick Summary
> Project	Total Size	Change
> react-10k	5.7 MB	+2.0 B (0.0%)
> react-1k	826.2 KB	+2.0 B (0.0%)
> react-5k	2.7 MB	+2.0 B (0.0%)
> rome	984.2 KB	0
> ui-components	2.3 MB	+56.0 B (0.0%)


image
Check the diff difference, which is a react product, and it is considered that it can be ignored.

PR #13000, opened 2026-02-09
fix: allow to export the same name of same local, fix entry splitting algorithm

JSerFeng:
## Summary

fix bugs for EsmLibraryPlugin.

1. Entry depends entry, should firstly split the module whose belonging chunks is most, then others

2. Even if we export chunk with strict signature, we can export the same name if the local is already exported with the same name

## Checklist

### DIFF
--- a/tests/rspack-test/esmOutputCases/basic/single-entry-split/__snapshots__/esm.snap.txt
+++ b/tests/rspack-test/esmOutputCases/basic/single-entry-split/__snapshots__/esm.snap.txt
@@ -1,10 +1,4 @@
 ```mjs title=bar.mjs
-
-export { v } from "./bar_js.mjs";
-
-```
-
-```mjs title=bar_js.mjs
 // ./bar.js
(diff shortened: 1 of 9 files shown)


JSerFeng (commented):
You can see the bar_js.mjs is missing in the output, as it reuse the entry chunk bar

tests/rspack-test/esmOutputCases/basic/single-entry-split/__snapshots__/esm.snap.txt
```diff
@@ -1,24 +1,12 @@
 ```mjs title=bar.mjs
-
-export { v } from "./bar_js.mjs";
-
-```
-
-```mjs title=bar_js.mjs
 // ./bar.js
```
JSerFeng, line 8: You can see the bar_js.mjs is missing in the output, as it reuses the entry chunk bar

SyMind (approved):

PR #13024, opened 2026-02-10
perf: Cache default context regexp for parser plugins

hardfist:
Summary:
- lazily instantiate the shared `^\.\/.*$` regexp so it can be reused without re-parsing in both `require.context` and import meta handling
- add `default_context_reg_exp` helper and update parser plugins to use it alongside the existing cleaning utilities
- ensure context module dependencies default to the shared regexp when no custom value is provided
- enable perf feature for regex
- add more regex clippy rules
- replace simple regex with starts_with

Testing:
- Not run (not requested)

### DIFF
--- a/crates/rspack_plugin_esm_library/src/render.rs
+++ b/crates/rspack_plugin_esm_library/src/render.rs
@@ -1,7 +1,4 @@
-use std::{
-  borrow::Cow,
-  sync::Arc,
-};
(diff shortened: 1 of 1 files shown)


hardfist:
@copilot fix cargo fmt for me

hardfist:
@copilot fix cargo shear for me

LingyuCoder (dismissed):

PR #13095, opened 2026-02-22
refactor: replace collect api with iterator api

hardfist:
## Summary
collect api like `modules` and `dependencies` cause lots unnecessary temporary HashMap which is slow, so we remove collect api to avoid introduce performance footgun in future. this is similar as 
- replace `module_graph_modules` and `dependencies` exposure with iterator-based helpers
- update binding and plugin callers to consume the new iterator APIs directly

## Testing
- Not run (not requested)


hardfist:
@codex review

chenjiahan (approved):

Issue #9353, opened 2025-02-18
[Feature]: Enable `stats.colors` by default in an environment that supports color output


chenjiahan:
### What problem does this feature solve?

Currently the [stats.colors](https://rspack.dev/config/stats#statscolors) is only enabled by default when using the Rspack CLI, see:

https://github.com/web-infra-dev/rspack/blob/92ddee29009c90ec35f04a69edf1f2ccb8beaf77/packages/rspack-cli/src/cli.ts#L202-L209

For users using using JavaScript API, `stats.colors` should be enabled by default in an environment that supports color output, we can move the above code to `@rspack/core` or to the Rust side.

### What does the proposed API of configuration look like?

/

PR #13161, opened 2026-03-02
feat(stats): enable stats.colors by default when environment supports

LingyuCoder:
## Summary

Enable `stats.colors` by default when the environment supports color output (TTY, `FORCE_COLOR`, or when `NO_COLOR` is not set), so that both JavaScript API and CLI users get colored stats without explicit config. Implements [issue #9353](https://github.com/web-infra-dev/rspack/issues/9353).

- **@rspack/core (JS):** Added `isStatsColorSupported()` using `picocolors`; when `stats.colors` is undefined, use it in `getRawStats()` and in `DefaultStatsPresetPlugin` so that Rust and stats output both get the env-based default.
- **Rust builder:** Default `stats.colors` via the `supports-color` crate (`Stream::Stdout`) instead of hardcoding `true`, so CI/non-TTY gets `false` by default.
- **CLI:** Removed the redundant `stats.colors` override; behavior is now handled in core.
- **Tests:** Rust snapshot filter for env-dependent `stats.colors`, new test `default_stats_colors_follows_environment`; new statsOutputCase `stats-colors-default` for the JS default path.

## Related links

- Close https://github.com/web-infra-dev/rspack/issues/9353

## Checklist

---

> [!NOTE]
> **Medium Risk**
> Changes default `stats.colors` behavior across both Rust and JS code paths, which may affect snapshot tests, CLI output formatting, and any tooling that parses stats output. Logic is straightforward but environment-dependent defaults can vary in CI/non-TTY contexts.
> 
> **Overview**
> **`stats.colors` now defaults to environment color support** (TTY/`FORCE_COLOR`/`NO_COLOR`) across the Rust builder and JS core.
> 
> Rust `CompilerOptionsBuilder` switches the default from `colors: true` to `supports-color` detection, while JS uses a new `isStatsColorSupported()` helper (backed by `picocolors`) and applies it both when adapting config (`getRawStats`) and in the default stats preset plugin. The CLI removes its previous implicit `stats.colors = true` override, and tests/snapshots are updated with env-dependent expectations plus a new stats output case.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit c0e79619a35a5aec57a60b3c0d35a6a1a2b18cc0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/rspack-cli/src/cli.ts
+++ b/packages/rspack-cli/src/cli.ts
@@ -15,7 +15,7 @@ import cac, { type CAC } from 'cac';
 import { BuildCommand } from './commands/build';
 import { PreviewCommand } from './commands/preview';
 import { ServeCommand } from './commands/serve';
-import type { RspackCLILogger } from './types';
+import type { RspackCLIColors, RspackCLILogger } from './types';
 import { loadExtendedConfig, loadRspackConfig } from './utils/loadConfig';
 import type {
   CommonOptions,
@@ -26,14 +26,6 @@ type Command = 'serve' | 'build';
 
 declare global {
   const RSPACK_CLI_VERSION: string;
-}
-
-interface RspackCLIColors {
-  isColorSupported: boolean;
-  red(text: string): string;
-  yellow(text: string): string;
-  cyan(text: string): string;
-  green(text: string): string;
 }
 
 function isEnvColorSupported(): boolean {


packages/rspack/package.json
```diff
@@ -64,7 +64,8 @@
     "webpack-sources": "3.3.4"
   },
   "dependencies": {
-    "@rspack/binding": "workspace:*"
+    "@rspack/binding": "workspace:*",
+    "picocolors": "^1.1.1"
```
chenjiahan: Can we avoid introducing a new dependency to core?

PR #13216, opened 2026-03-05
refactor: Move ModuleExecutor state into BuildModuleGraphArtifact

hardfist:
Summary
- store the `ModuleExecutor` inside `BuildModuleGraphArtifact` and add helpers on `Compilation` for safely swapping it while the artifact is stolen
- keep cache hooks and make/finish passes in sync by preserving the executor across artifact recovery and build stages
- update JS bindings to read the executor through the new accessors so it remains available during async module imports

Testing
- Not run (not requested)


hardfist:
Closed per author request.

PR #13219, opened 2026-03-05
fix: Fix stats artifact fallback warning test to use stale stats

hardfist:
## Summary
- ensure the stats-artifact-fallback-warning watch case only stores the stale stats provided by `compiler.hooks.done`
- reuse that stats object when printing warnings so the test stays consistent with runtime behavior

## Related links
- None.

## Checklist


hardfist:
@codex fix failing tests

PR #13234, opened 2026-03-06
refactor!: use `strictThisContextOnImports` to control ns obj as this

ahabhgk:
## Summary

Before `name` in `module.js` won't be tree shaken (whether `strictThisContextOnImports` enabled or not), this PR changes this:

- with `strictThisContextOnImports` disabled, `name` will be tree shaken, generate optimized code
- with `strictThisContextOnImports` enabled, `name` will not be tree shaken, generate correct code.

```js
// main.js
import * as m from "./reexport.js";
m.module.that().name

// reexport.js
export * as module from "./module.js";
console.log.bind(console) // bailout side effect optimization to keep the reexport.js in the output

// module.js
export function that() {
  return this;
}

export const name = "module.js"
```

PS: I was going to let `strictThisContextOnImports` also control the `return this` for commonjs, but failed, because:

- In commonjs this maybe more common, since in commonjs the top-level `this` represent `module`
- Json is the main reason, it's very common to write `import arr from "./arr.json"; arr.map(d => d + 1)`, the `map` or other methods on prototype is kind of implicitly accessing `this`

So for `ExportsType::DefaultOnly` and `ExportsType::DefaultWithNames` we will always make the tree shaking level minus one when calling exported function to generate the correct code

## Checklist

### DIFF
--- a/tests/rspack-test/esmOutputCases/basic/split-runtime-chunk/__snapshots__/esm.snap.txt
+++ b/tests/rspack-test/esmOutputCases/basic/split-runtime-chunk/__snapshots__/esm.snap.txt
@@ -7,7 +7,7 @@ const other = __webpack_require__("./other.js");
 
 
 it('should have correct value', () => {
-  expect((0,other.U)()).toBe(42)
+  expect((0,other.value)()).toBe(42)
 })
 
 ```
@@ -20,7 +20,7 @@ __webpack_require__.add({
   !*** ./other.js ***!
   \******************/
 (__unused_rspack_module, exports) {
-exports.U = () => 42
+exports.value = () => 42
 
 },
 });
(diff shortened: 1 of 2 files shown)


ahabhgk:
It's still WIP and unstable, we will have some different with webpack on this options for optimizing code better, once the PR is done and stable enough will add docs for it

chenjiahan:
Get 👌

crates/rspack_plugin_javascript/src/dependency/esm/esm_import_specifier_dependency.rs
```diff
@@ -268,7 +277,7 @@ impl Dependency for ESMImportSpecifierDependency {
       }
     }
 
-    if self.call && !self.direct_import && (namespace_object_as_context || ids.len() > 1) {
+    if namespace_object_as_context && self.call {
```
ahabhgk, line 280: This is the main change

PR #13243, opened 2026-03-06
perf: Optimize build module dependency handling

hardfist:
Summary
- pre‑allocate queues/dependency vectors in repair build for better reuse and cache behavior
- mark existing dependencies before running ProcessDependenciesTask so FactorizeTask doesn’t duplicate work
- streamline initialization of repair tasks with shared compiler/context state

Testing
- Not run (not requested)


hardfist:
Closed per author request.

PR #13322, opened 2026-03-12
refactor(split-chunks): speed up module group processing

LingyuCoder:
## Summary

This PR focuses on reducing `splitChunks` overhead in `rspack_plugin_split_chunks`, especially around module group preparation, winner selection, and repeated post-processing.

Key changes:
- Narrow invalidation after a winner is chosen by building an `AffectedModuleGroupIndex`, so we only revisit module groups that actually share both modules and chunks with the winner.
- Replace repeated linear best-group scans with a `BinaryHeap`-based `ModuleGroupQueue` plus lazy version invalidation, while keeping deterministic tie-breaking.
- Add a parallel fast path for static cache groups in `prepare_module_group_map` when `chunks` / `test` / `name` / `layer` do not require async or dynamic evaluation, avoiding the previous shared `DashMap` write contention.
- Reduce `minSize`-path clone overhead by returning borrowed `SplitChunkSizes` and collecting source-type modules into `Vec<ModuleIdentifier>` instead of cloning full sets.
- Keep deterministic ordering and fix review issues by:
  - comparing size reduction with `f64::total_cmp`
  - only removing queue versions after a successful `swap_remove`
  - requiring live keys when refreshing heap entries
  - using a stable tie-break in `ordered_module_identifiers`

Additional plumbing:
- add `has_custom_layer_filter` to detect whether the static fast path is safe
- align the same borrowed-size behavior in `rspack_plugin_esm_library`

Local validation:
- `cargo fmt --all --check`
- `cargo clippy -p rspack_plugin_split_chunks -- -D warnings`

## Related links

- PR: #13322
- Ecosystem benchmark / review discussion is in the PR thread

## Checklist

### DIFF
--- a/crates/rspack_plugin_split_chunks/src/plugin/module_group.rs
+++ b/crates/rspack_plugin_split_chunks/src/plugin/module_group.rs
@@ -140,7 +140,7 @@ struct ModuleGroupHeapEntry {
   chunks_len: usize,
   key: String,
   modules: Vec<ModuleIdentifier>,
-  size_reduce: f64,
+  size_reduce_bits: u64,
   version: u32,
 }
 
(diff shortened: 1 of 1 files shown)


LingyuCoder:
@codspeed explain the perf diff

PR #13384, opened 2026-03-17
feat(rsdoctor): add support for dependency connections

yifancong:
## Summary
This pull request introduces a new field to track the exports type of each module in the `RsdoctorModule` struct and ensures this information is properly populated when collecting modules. The changes also include necessary imports to support this feature.

Enhancements to module data tracking:

* Added a new `exports_type` field to the `RsdoctorModule` struct to store the exports type as a string.
* Updated the `collect_modules` function to populate the `exports_type` field based on the `BuildMetaExportsType` of each module.

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/rsdoctor.rs
+++ b/crates/rspack_binding_api/src/rsdoctor.rs
@@ -2,7 +2,6 @@ use std::collections::HashSet;
 
 use napi::Either;
 use napi_derive::napi;
-use rspack_core::BuildMetaExportsType;
 use rspack_plugin_rsdoctor::{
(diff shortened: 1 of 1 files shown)


crates/rspack_binding_api/src/rsdoctor.rs
```diff
@@ -60,6 +62,14 @@ impl From<RsdoctorModule> for JsRsdoctorModule {
         .into_iter()
         .map(|loc| loc.into())
         .collect::<Vec<_>>(),
+      exports_type: match value.exports_type {
+        BuildMetaExportsType::Namespace => "namespace",
```
LingyuCoder: should impl Display to convert this to string

yifancong: Done

PR #13388, opened 2026-03-17
refactor: Refactor ESMImportSpecifierDependency to share immutable data with Arc

hardfist:
## Summary
- move the immutable fields of `ESMImportSpecifierDependency` into a new `ESMImportSpecifierDependencyData`
- store that shared data behind `Arc` with `AsInner` to reduce clone cost
- update dependency and template accessors to read from the shared data while keeping mutable runtime state on the outer struct

## Testing
- `cargo fmt --all --check`
- `cargo lint`


hardfist:
Closed per author request.

PR #13397, opened 2026-03-18
chore: trigger empty PR for codspeed stability check

hardfist:
##Summary
- create an empty commit/PR to validate CI behavior on a tree-identical branch
- document the current CodSpeed gating behavior for this run

##Notes
- this branch contains an empty commit only; there are no file changes
- under the current `.github/workflows/ci.yml`, `check-changed` will mark `code_changed = false` for an empty PR
- because `test-linux` is skipped in that case, downstream CodSpeed-related jobs are also expected to be skipped
- this PR validates the stability of the empty-PR gating path, not an actual CodSpeed execution path


hardfist:
Update: the original empty-PR setup did not trigger CodSpeed because `changed_files = 0` and the CI workflow gated on `code_changed`. I pushed a follow-up commit (`8d24280`) adding a tiny non-doc marker file so the PR now enters the normal Linux build/bench path and can exercise CodSpeed.

PR #13403, opened 2026-03-18
perf(build_chunk_graph): optimize data structures in code splitter

JSerFeng:
## Summary

Optimize data structure operations in the `build_chunk_graph` code splitter:

- **Remove `num-bigint` `opt-level="s"` override**: BigUint bit operations (`.bit()`, `.set_bit()`, `&`, `|=`) used 1.2M+ times per build were compiled with size optimization instead of the global `opt-level=3`. This penalizes the hot path in `process_queue`.
- **Use bitmask for module-in-chunk check**: In `add_and_enter_entry_module`, replace `chunk_graph.is_module_in_chunk()` (2 HashMap lookups) with a single bit check on `mask_by_chunk` since the module ordinal is already computed.
- **`skipped_items`: O(n²) → O(n)**: Replace `shift_remove` loop with `retain()` in `process_outdated_chunk_group_info`.
- **`skipped_module_connections`: O(n²) → O(n)**: Replace `shift_remove_index` loop with `retain()`, collecting side effects during the pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -358,10 +358,6 @@ opt-level = "s"
 [profile.release.package.rspack_plugin_progress]
 opt-level = "s"
 
-[profile.release.package.num-bigint]
-opt-level = "s"
-
-
 [profile.release.package.swc_error_reporters]
 opt-level = "s"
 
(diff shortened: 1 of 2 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex is LinkedSet more suitable than IndexSet for `skipped_items` ?

PR #13406, opened 2026-03-18
fix(esm-lib): use package name for node_modules short name in dynamic imports

JSerFeng:
## Summary
- Fix `short_name_from_identifier` in `rspack_plugin_esm_library` to extract the package name from `node_modules` paths when the file is `index.js`. Previously, `node_modules/lib/dist/index.js` would return `dist` (the immediate parent directory) instead of `lib` (the actual package name).
- Scoped packages (`@scope/pkg`) are also handled correctly.
- Added test case `dyn-import-short-name-node-modules` to verify the fix.

## Changes
- `crates/rspack_plugin_esm_library/src/optimize_chunks.rs`: When stem is `index` and the path contains `node_modules/`, use `rfind("node_modules/")` to locate the last `node_modules` segment and extract the package name from the path after it.
- `tests/rspack-test/esmOutputCases/dynamic-import/dyn-import-short-name-node-modules/`: New test case with a `node_modules/my-lib/dist/index.js` structure that verifies the chunk is named `my-lib.mjs` instead of `dist.mjs`.

## Test plan
- [x] `npx rstest esm -u` — all 89 tests pass
- [x] New snapshot confirms chunk is named `my-lib.mjs` (package name) not `dist.mjs`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4497,6 +4497,7 @@ version = "0.100.0-beta.7"
 dependencies = [
  "async-trait",
  "atomic_refcell",
+ "cow-utils",
  "futures",
  "rayon",
  "regex",


JSerFeng:
@codex review

JSerFeng:
@codex review

PR #13413, opened 2026-03-19
feat: enable mangle exports for cjs reexport

ahabhgk:
## Summary

Now rspack support below cjs reexports:

- `module.exports = require("./m")`
- `[exports|module.exports|this].a.b = require("./m").a.b`

The first one is hard to support mangle exports because we have to rewrite the exports to something like this: `module.exports = { az: (temp = __webpack_require__(123)).abc, Oi: temp.def }`, but the second one is easy to support mangle exports, so for now we only mangle for the second case.

## Checklist


ahabhgk:
@codex review

PR #13418, opened 2026-03-20
fix(esm-library): deduplicate external module import identifiers in ESM output

JSerFeng:
## Summary
- Fix duplicate `import * as __rspack_external_X from "source"` statements in ESM library output when multiple external modules reference the same source
- Use `register_namespace_import` instead of `NormalInitFragment` in the `UsedNames + !safe_to_optimize` path so scope-hoisted externals consistently route through `raw_import_stmts`
- Filter out `ModuleExternal` init fragments in the render phase when their source is already handled by `raw_import_stmts`, covering the mixed scope-hoisting case (one external is `ModuleInfo::Concatenated`, another is `ModuleInfo::External`)

## Test plan
- [x] Added `dedup-external-imports` test: two ESM modules import from the same "fs" external with different specifiers (named vs namespace)
- [x] Added `dedup-external-imports-mixed` test: one ESM module (scope-hoisted) and one CJS module (non-scope-hoisted) both import from "fs"
- [x] Existing `multiple-externals` and `re-export-external` snapshots unchanged
- [x] All 5930+ ESM output tests pass (8 pre-existing failures from pnpm path changes)

### DIFF
--- a/crates/rspack_plugin_esm_library/src/link.rs
+++ b/crates/rspack_plugin_esm_library/src/link.rs
@@ -457,9 +457,10 @@ impl EsmLibraryPlugin {
                     .module_by_identifier(&symbol_binding.module)
                     .and_then(|m| m.as_external_module())
                   {
-                    let request = ext.user_request().to_string();
-                    let import_spec = chunk_link
-                      .raw_import_stmts
+                    let request = ext.get_request().primary().to_string();
+                    let (raw_import_stmts, used_names) =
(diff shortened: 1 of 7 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

PR #13419, opened 2026-03-20
fix(esm-lib): pull entry-only modules into shared chunks to prevent circular imports

JSerFeng:
## Summary
- Add `pull_module_into_non_entry_chunks` to move modules from entry chunks into shared (non-entry initial) chunks when they are referenced by the shared chunk's modules. This prevents circular `import` statements in ESM library output.
- Fix a `break` bug in `ensure_entry_exports` that only marked one entry chunk as dirty when multiple entries were affected.

## Problem
When `SplitChunksPlugin` extracts a shared module (e.g. `lib.js` with `export * from './a'`) into a shared chunk, tree-shaking may leave the re-exported modules (`a.js`, `b.js`) in their respective entry chunks. This caused:
- Circular chunk imports (`shared.mjs` → `entry1.mjs` → `shared.mjs`)
- Full `__webpack_require__` runtime injected into non-entry chunks
- Empty entry proxy chunks

## Fix
During `optimize_chunks`, before `ensure_entry_exports`, the new `pull_module_into_non_entry_chunks` function detects modules in entry chunks that are referenced by shared chunk modules and moves them into the shared chunk. This ensures clean ESM output where each entry imports only its needed exports from the shared chunk.

**Before:**
```
entry1.mjs → import "./src_lib_js.mjs"  (circular!)
src_lib_js.mjs → import "./entry1.mjs"  (circular!)
                 import "./src_b_js-src_entry2_js.mjs"
src_b_js-src_entry2_js.mjs → (full __webpack_require__ runtime)
```

**After:**
```
entry1.mjs → import { a } from "./src_lib_js.mjs"
entry2.mjs → import { b } from "./src_lib_js.mjs"
src_lib_js.mjs → export { a, b }  (clean ESM, no circular imports)
```

### DIFF
--- a/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
+++ b/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
@@ -48,11 +48,25 @@ pub(crate) fn pull_module_into_non_entry_chunks(compilation: &mut Compilation) {
       let module_graph = compilation.get_module_graph();
       let mut moves: Vec<(ModuleIdentifier, ChunkUkey, ChunkUkey)> = Vec::new();
 
-      for (chunk_ukey, chunk) in compilation.build_chunk_graph_artifact.chunk_by_ukey.iter() {
+      // Sort chunk keys for deterministic iteration order
+      let mut chunk_ukeys: Vec<_> = compilation
+        .build_chunk_graph_artifact
+        .chunk_by_ukey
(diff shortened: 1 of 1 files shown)


JSerFeng:
Closing in favor of a new PR with root-cause fix (marking entry deps exports as used instead of post-split chunk moving).

PR #13439, opened 2026-03-23
fix(rstest): resolve manual mock targets with resolver

9aoy:
## Summary

resolve manual mock targets with resolver.

Fix the case where a mocked relative request resolves through a directory entry file such as `index.js` / `index.ts`.

Before this change, `rs.mock("./common")` would resolve the mock target as `./__mocks__/common`, even when `./common` actually resolved to `./common/index.ts`. That caused rstest to miss mocks placed under `./common/__mocks__/index`.

After this change, when a relative mock request points to a real directory, rstest resolves the manual mock to the nested directory form:

- source: `./common`
- actual module: `./common/index.ts`
- manual mock: `./common/__mocks__/index`

## Related links

fix https://github.com/web-infra-dev/rstest/issues/613

## Checklist


fi3ework:
👍🏻

PR #13446, opened 2026-03-24
chore: add walltime support

hardfist:
## Summary
* add codspeed walltime support
* split memory | simulation into different jobs
* temporarily disable memory jobs(since it's only supported in github runner and we have limited github runner resource)

## Checklist


hardfist:
@codspeedbot explain regression in build_module_graph simulation

PR #13457, opened 2026-03-24
fix: use Symbol.for() for shared symbol identity in module bindings

tryggvigy:
## Summary

When `@rspack/core`'s JS wrapper is loaded more than once in the same Node.js process, the `identifier()` method on module objects returns `undefined`, crashing `importModule` with:

```
Error: Failed to convert JavaScript value `Undefined` into rust type `String`
```

### Root cause

The `define_symbols!` macro creates symbols using `Symbol()` (unique per call). All loads share the same native addon and thus the same `NormalModule.prototype`, but each load's JS wrapper runs separately and creates its own symbol instance. The last load's `Object.defineProperty` overwrites the prototype's `identifier()` method with one that reads its own symbol — but module objects created by earlier loads have a different one set on them. Since `Symbol("X") !== Symbol("X")`, the lookup returns `undefined`.

### How this can happen

On macOS (case-insensitive APFS), mismatched path casing (e.g. `/foo/MyProject/...` vs `/foo/myproject/...`) can cause Node.js's case-sensitive `require.cache` to load the same physical file under two different keys. In our case, the rspack CLI resolved `@rspack/core` via a lowercase real path while the `--config` file resolved it via an uppercase path inherited from a symlink — loading the package twice in the same process.

Any scenario where `@rspack/core` ends up loaded twice — symlinks, case-insensitive filesystems, unconventional module resolution — could trigger this.

### Fix

- Switch from `Symbol()` to `Symbol.for()` in `define_symbols!`, so all loads share the same global symbol instances. This affects all three symbols: `MODULE_IDENTIFIER_SYMBOL`, `MODULE_BUILD_INFO_SYMBOL`, and `COMPILATION_HOOKS_MAP_SYMBOL`.
- Add a defensive `?? ""` fallback in `importModule` for the identifier parameter.
- Bump napi feature from `napi7` to `napi9` to access `env.symbol_for()`.

### Risks

- **`Symbol.for()` uses the global symbol registry**, so any code could theoretically access these symbols. The names are specific enough that accidental collision is very unlikely; they could be namespaced (e.g. `rspack.MODULE_IDENTIFIER_SYMBOL`) if preferred.
- **napi7 → napi9** raises the minimum to Node-API 9 (Node.js 18.17+). Node 18 is already EOL, and napi features are purely additive.
- **The `?? ""` fallback** passes an empty string instead of crashing — worst case is slightly degraded debug info for `importModule`.

## Checklist

- Tests updated (or not required).
- Documentation updated (or not required).

### DIFF
--- a/crates/rspack_binding_api/src/define_symbols.rs
+++ b/crates/rspack_binding_api/src/define_symbols.rs
@@ -17,7 +17,7 @@ macro_rules! define_symbols {
       use napi::bindgen_prelude::JsObjectValue;
       $(
         let symbol = ::rspack_napi::OneShotRef::new(env.raw(), env.symbol_for($name)?)?;
-        exports.set_named_property($name, &symbol)?;
+        exports.set_named_property(stringify!($cell), &symbol)?;
         $cell.with(|once_cell| {
           once_cell.get_or_init(move || symbol);
         });


packages/rspack/src/loader-runner/index.ts
```diff
@@ -418,7 +418,7 @@ export async function runLoaders(
             options.layer,
             options.publicPath,
             options.baseUri,
-            context._module.identifier(),
+            context._module.identifier() ?? "",
```
tryggvigy: Might be overkill. I dont think we need this if we use `Symbol.for()`

SyMind:
I namespace all `define_symbols!` registry keys used with `Symbol.for(...)`.

`Symbol.for(...)` uses the global symbol registry, so generic keys increase the chance of collisions with other code running in the same JavaScript realm. Adding an `rspack.*` namespace makes the symbols self-descriptive and reduces that risk while keeping the cross-load sharing behavior that `Symbol.for(...)` is intended to provide.

SyMind (approved):

PR #13471, opened 2026-03-25
fix(runtime): invalidate ESM chunk cache in development SSR

### LINKED ISSUE #13304: [Bug] ESM SSR development mode - chunk cache not invalidated properly
# [Bug] ESM SSR development mode - chunk cache not invalidated properly

## Description

In ESM SSR development mode, when source files are modified, the server-rendered content doesn't reflect the changes, causing hydration mismatch errors. This is because rspack's internal chunk loading mechanism doesn't properly invalidate the cache.

## Steps to Reproduce

1. Create an SSR project with Modern.js using ESM mode (`type: "module"` in package.json)
2. Start the dev server
3. Modify a page component (e.g., change text in page.tsx)
4. Refresh the browser

## Expected Behavior

The server should render the updated content.

## Actual Behavior

The server renders stale content while the client has the new content, causing a hydration mismatch error:

```
Hydration failed because the server rendered text didn't match the client
+ ESM-CACHE-REPRODUCE
- Welcome to111
```

## Root Cause Analysis（maybe, but i not sure）

When rspack bundles the code, it uses code splitting - the main bundle (`index.js`) loads page chunks dynamically via `__webpack_require__.e("page")`.

The problem is:
1. The main bundle (`index.js`) can be reloaded using timestamp or Module._compile
2. But the chunk loading (`__webpack_require__.e`) has its own internal caching mechanism
3. This chunk cache is not properly invalidated in development mode

### Bundle Structure

```
dist/bundles/
├── index.js              # Main bundle (can be reloaded)
├── index-server-loaders.js  # Server loaders
└── page.js              # Page chunk (NOT reloaded - uses cached version)
```

### How it works

In the bundled `index.js`, there's code like:

```javascript
// This is rspack-generated code
__webpack_require__.e("page").then(() => {
  return __webpack_require__("./src/routes/page.tsx");
})
```

The `__webpack_require__.e` function is rspack's internal chunk loading mechanism, which has its own cache that's not controlled by external timestamp/cache-busting mechanisms.

## Environment

- rspack: latest
- Node.js: 20+
- Modern.js: V3

## Comparison with CJS mode

In CJS mode (`MODERN_LIB_FORMAT !== 'esm'`), this issue doesn't occur because:
- Uses `require()` which has a writable cache (`require.cache`)
- Can directly delete cache entries: `delete require.cache[filepath]`

In ESM mode:
- Uses `import()` which has a read-only cache (`import.meta.cache`)
- Cannot directly invalidate the cache

## Suggested Fix

The chunk loading mechanism (`__webpack_require__.e`) should properly handle cache invalidation in development mode, similar to how it handles HMR for client-side code.

## Related

- Similar to webpack issues with SSR and chunk loading
- Modern.js issue: web-infra-dev/modern.js#8373

GiveMe-A-Name:
## Summary
- replace coarse development cache-busting (`?t=${Date.now()}`) in ESM module chunk loading with per-chunk versioning based on `chunkId -> rendered chunk hash`
- keep cache invalidation scoped: only changed chunks get a new import URL, unchanged chunks can still hit cache
- keep production behavior unchanged (no dev cache-busting query emitted)
- update regression assertions to verify `?t=` is present in dev runtime, `Date.now` is not emitted, and no cache-busting query appears in production

## Issue
Closes #13304

## Verification
- `cd tests/rspack-test && pnpm run test -t "watchCases/runtime/module-dynamic-import-cache-bust"`
- `cd tests/rspack-test && pnpm run test -t "configCases/module/dev-module-dynamic-import-no-cache-bust"`

### DIFF
--- a/crates/rspack_plugin_runtime/src/runtime_module/module_chunk_loading.rs
+++ b/crates/rspack_plugin_runtime/src/runtime_module/module_chunk_loading.rs
@@ -297,12 +297,19 @@ impl RuntimeModule for ModuleChunkLoadingRuntimeModule {
       let body = if matches!(has_js_matcher, BooleanMatcher::Condition(false)) {
         "installedChunks[chunkId] = 0;".to_string()
       } else {
+        let cache_bust = if compilation.options.mode.is_development() {
+          " + \"?t=\" + Date.now()".to_string()
+        } else {
+          String::new()
+        };
+
         runtime_template.render(
           &self.template(TemplateId::WithLoading),
           Some(serde_json::json!({
             "_js_matcher": &has_js_matcher.render("chunkId"),
             "_import_function_name":&compilation.options.output.import_function_name,
             "_output_dir": &root_output_dir,
+            "_cache_bust": &cache_bust,
             "_match_fallback":    if matches!(has_js_matcher, BooleanMatcher::Condition(true)) {
               ""
             } else {
--- a/crates/rspack_plugin_runtime/src/runtime_module/module_chunk_loading.rs
+++ b/crates/rspack_plugin_runtime/src/runtime_module/module_chunk_loading.rs
@@ -337,10 +337,7 @@ impl RuntimeModule for ModuleChunkLoadingRuntimeModule {
 
           let fallback_hash =
             rspack_util::json_stringify_str(compilation.get_hash().unwrap_or("dev"));
-          format!(
-            " + \"?t=\" + ({}[chunkId] || {})",
-            chunk_hash_map, fallback_hash
-          )
+          format!(" + \"?t=\" + ({chunk_hash_map}[chunkId] || {fallback_hash})")
         } else {
           String::new()
         };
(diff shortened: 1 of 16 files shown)


JSerFeng:
@codex review

GiveMe-A-Name:
Follow-up explanation for the cache invalidation strategy change:

The original `?t=${Date.now()}` approach fixes stale ESM chunks, but it is too coarse in SSR dev because every dynamic `import()` gets a brand-new URL on every evaluation. That effectively disables chunk-level caching and can trigger unnecessary reload cascades.

This PR now uses a per-chunk version key:
- `?t=` is derived from `chunkId -> rendered chunk hash` (with a compilation hash fallback)
- unchanged chunks keep stable URLs and can still hit cache
- only changed chunks receive a new URL and are reloaded

So the behavior is still correct for stale-cache avoidance, but with a tighter invalidation scope aligned with the actual changed artifact.

GiveMe-A-Name:
I checked the failing CI job (`Test WASM / test / Test Node 24`) and the failure is unrelated to this PR changes.

Failing test:
- `packages/rspack-cli/tests/build/config/extends/extends.test.ts` line 78
- case: `function config with extends > should handle extends in function configs`
- error: `test timed out in 200000ms (no expect assertions completed)`

This is in rspack-cli extends test flow (WASM + Node 24), while this PR only touches runtime module chunk loading + rspack-test cases.

I have re-run failed jobs to verify whether it is transient.

If it fails again, I can prepare a separate follow-up PR for the rspack-cli timeout investigation/fix (keeping this PR focused on #13304).

chenjiahan:
@JSerFeng cc

PR #13472, opened 2026-03-25
fix(rsc): tree shaking client barrel

SyMind:
## Summary

This PR fixes export usage tracking for RSC client references when `referenced_exports` contains `*`.

In this case, `*` is an internal sentinel meaning the client reference should be treated as a whole-module reference instead of a narrowed set of named exports. Previously, this case was not handled in `client_reference_dependency`, which could cause the target client module's exports to be incorrectly tree-shaken away in real projects.

This change updates `get_referenced_exports` to return an exports-object reference when `*` is present, and keeps the existing precise export tracking for normal named-export cases. It also clarifies the related comments in English.

Additionally, this PR adds a regression test in `tests/rspack-test/configCases/rsc-plugin/tree-shaking-client-barrel` to cover a client barrel scenario where a server entry both imports a named export and side-effect imports the same client boundary. The test verifies that once the reference is upgraded to a whole-module reference, all exports from the client module are preserved.

## Checklist


SyMind:
@codex review

PR #13483, opened 2026-03-26
fix(esm-library): avoid duplicate __webpack_require__ export

JSerFeng:
## Summary

- avoid adding `__webpack_require__` to `chunk_link.exports` when module chunk loading will export it via runtime module
- record that export strategy in `ChunkLinkContext` during link and let render consume link metadata instead of scanning runtime modules
- add an `esmOutputCases` regression for `runtimeChunk: false` with a shared CJS chunk, which reproduces the duplicate export failure without the fix

## Related links

- https://github.com/rstackjs/rspack-dev-server/pull/166

## Checklist


JSerFeng:
@codex review

PR #13484, opened 2026-03-26
chore: try namespace runner

hardfist:
## Summary
* migrate bench runner to namespace to see whether it's more stable than github runner since github runner cause variance for simulation benchmark
* disable memory benchmark temporarily since it doesn't support libbpf well
* reduce bench cost from 20min -> 10min

## Checklist

### DIFF
--- a/.github/workflows/bench-rust.yml
+++ b/.github/workflows/bench-rust.yml
@@ -60,7 +60,7 @@ jobs:
         run: cargo codegen
 
       - name: Build benchmark (simulation,memory)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
     uses: ./.github/workflows/bench-rust.yml
     with:
       target: x86_64-unknown-linux-gnu
-      runner: '"ubuntu-22.04"'
+      runner: '"namespace-profile-linux-am64-large2;container.privileged=true;container.host-pid-namespace=true"'
```
deanjingshui, line 91: why we need container to use host's pid namespace?

PR #13487, opened 2026-03-26
ci: remove pnpm cache and simplify pnpm install steps

chenjiahan:
## Summary

This PR removes the custom pnpm cache flow from CI and simplifies dependency installation to plain `pnpm i`.

## Background

The current pnpm cache uses inconsistent cache keys.

Even if we fixed the cache keys, the cache adds extra CI logic, but it does not provide a clear and consistent win for these workflows.

A simple example from recent CI logs:

- a cache hit still spent about 5s restoring a ~279 MB cache, while `pnpm install` itself went from 9.5s to 5.4s
- in practice, most of the gain was offset by cache restore overhead

Given that tradeoff, the custom cache layer adds more complexity than value.

## Result

The CI install path is now more straightforward:

- less workflow indirection
- fewer custom actions to maintain


hardfist:
@codspeedbot analyze memory diff reason

PR #13490, opened 2026-03-26
perf(javascript): cache non-nested export target lookups

LingyuCoder:
## Summary
- optimize the non-nested export merge hot path in `FlagDependencyExportsPlugin`
- replace `par_bridge` with `into_par_iter` so Rayon can split the already-collected tasks directly with lower scheduling overhead
- cache `find_target_exports_info` results within each non-nested module merge task so repeated exports that resolve to the same target can reuse the same target-resolution result
- keep the cache scoped to a single module merge task with normalized target keys, which preserves deterministic behavior without cross-task synchronization or invalidation
- overall, this reduces repeated target resolution and small temporary allocations on the non-nested merge path without changing behavior


hardfist:
@codspeedbot explain perf diff

PR #13491, opened 2026-03-26
fix(hash): keep fullhash in sync with css-only content changes

### LINKED ISSUE #12866: [Bug]: [fullhash] does not update when CSS-only changes occur
### System Info

```
System:
    OS: macOS 26.2
    CPU: (16) arm64 Apple M4 Max
    Memory: 7.35 GB / 64.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 22.8.0 - /Users/v7rulnik/.local/state/fnm_multishells/48371_1769509735708/bin/node
    npm: 10.8.2 - /Users/v7rulnik/.local/state/fnm_multishells/48371_1769509735708/bin/npm
  Browsers:
    Chrome: 144.0.7559.97
    Safari: 26.2
  npmPackages:
    @rspack/cli: latest => 1.7.4
    @rspack/core: latest => 1.7.4
```

### Details

When CSS content changes but JavaScript remains unchanged, the [fullhash] placeholder does not update, while [contenthash] correctly updates.

  Current behavior:
  - CSS [contenthash] correctly updates when CSS changes
  - Compilation [fullhash] incorrectly stays the same

  Expected behavior:
  - Both [contenthash] and [fullhash] should update when CSS content changes

  This causes different build outputs to have identical fullhashes.

Webpack comparison: Webpack correctly updates [fullhash] when CSS changes. The reproduction repo includes both Rspack and Webpack tests demonstrating this difference.

### Reproduce link

https://github.com/7rulnik/rspack-fullhash-css-repro

### Reproduce Steps

1. Clone the repo: git clone https://github.com/7rulnik/rspack-fullhash-css-repro
  2. Run the reproduction script:
  cd rspack-fullhash-css-repro/rspack-test
  npm install
  ./reproduce.sh
  3. Observe the output:
    - Build 1 produces main.[fullhash].js and main.[contenthash].css
    - CSS is modified (color changes from red to green)
    - Build 2 produces files with the same fullhash but different contenthash

  Result:
```
  BUILD 1:
    [fullhash]:    15d7bc99b1b65226
    [contenthash]: 6fdaa5b099f97dd7

  BUILD 2 (CSS changed, JS unchanged):
    [fullhash]:    15d7bc99b1b65226  ❌ SAME (bug!)
    [contenthash]: 47fc5c9ca2bb0b36  ✅ CHANGED
```

  For comparison, run ./run-both.sh from the root to see Webpack correctly updating [fullhash] for the same scenario.

GiveMe-A-Name:
## Summary
- include chunk content hashes in compilation fullhash generation so css-only updates change 
- add a hash case () covering two builds that differ only in CSS content
- ensure emitted JS filename hash matches stats hash and changes when CSS content changes

Fixes #12866

### DIFF
--- a/crates/rspack_core/src/compilation/create_hash/mod.rs
+++ b/crates/rspack_core/src/compilation/create_hash/mod.rs
@@ -385,25 +385,6 @@ pub async fn create_hash(
   }
   logger.time_end(start);
 
-  fn source_type_sort_key(source_type: &SourceType) -> (u8, Option<&str>) {
-    match source_type {
-      SourceType::JavaScript => (0, None),
-      SourceType::Css => (1, None),
-      SourceType::CssUrl => (2, None),
-      SourceType::Wasm => (3, None),
-      SourceType::Asset => (4, None),
-      SourceType::Expose => (5, None),
-      SourceType::Remote => (6, None),
-      SourceType::ShareInit => (7, None),
-      SourceType::ConsumeShared => (8, None),
-      SourceType::ShareContainerShared => (9, None),
-      SourceType::Custom(name) => (10, Some(name.as_str())),
-      SourceType::Unknown => (11, None),
-      SourceType::CssImport => (12, None),
-      SourceType::Runtime => (13, None),
-    }
-  }
-
   // create full hash
   compilation
     .build_chunk_graph_artifact
@@ -417,7 +398,7 @@ pub async fn create_hash(
       if let Some(content_hashes) = chunk.content_hash(&compilation.chunk_hashes_artifact) {
         content_hashes
           .iter()
-          .sorted_unstable_by_key(|(source_type, _)| source_type_sort_key(source_type))
+          .sorted_unstable_by_key(|(source_type, _)| *source_type)
           .for_each(|(source_type, content_hash)| {
             source_type.hash(&mut compilation_hasher);
             content_hash.hash(&mut compilation_hasher);
--- a/crates/rspack_core/src/lib.rs
+++ b/crates/rspack_core/src/lib.rs
@@ -111,7 +111,7 @@ pub use rspack_sources;
 pub mod debug_info;
 
 #[cacheable]
-#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub enum SourceType {
   JavaScript,
   Css,


GiveMe-A-Name:
Triggered rerun for failed checks. Opened tracking issue for WASM Node24 CI flakiness: #13515 (includes run/job links and log signatures).

lint-staged.config.mjs
```diff
@@ -1,9 +1,34 @@
...
+  ];
+};
+
 export default {
   '*.rs': 'rustfmt',
   '*.{ts,tsx,js,mjs,yaml,yml}':
     'node ./node_modules/prettier/bin/prettier.cjs --write',
   '*.toml': 'pnpm exec taplo format',
-  '*.{ts,tsx,js,cts,cjs,mts,mjs}': 'pnpm run lint:js',
+  '*.{ts,tsx,js,cts,cjs,mts,mjs}': runJsLint,
```
CPunisher: why is this file changed?

crates/rspack_core/src/compilation/create_hash/mod.rs
```diff
@@ -385,15 +385,44 @@ pub async fn create_hash(
...
-    .for_each(|hash| {
-      hash.hash(&mut compilation_hasher);
+    .for_each(|chunk| {
+      if let Some(hash) = chunk.hash(&compilation.chunk_hashes_artifact) {
+        hash.hash(&mut compilation_hasher);
+      }
+      if let Some(content_hashes) = chunk.content_hash(&compilation.chunk_hashes_artifact) {
+        content_hashes
+          .iter()
+          .sorted_unstable_by_key(|(source_type, _)| source_type_sort_key(source_type))
```
CPunisher: why `source_type_sort_key` is needed here. Can we directly use `source_type.hash(&mut compilation_hasher)`?

lint-staged.config.mjs
GiveMe-A-Name: These changes are unrelated to the fullhash fix. I reverted `lint-staged.config.mjs` and `.gitignore` locally and will keep them out of this PR.

crates/rspack_core/src/compilation/create_hash/mod.rs
GiveMe-A-Name: Not directly. `source_type.hash(&mut compilation_hasher)` only makes the source type part of the hashed payload; it does not make the iteration order deterministic. `content_hashes` is map-backed, so we still need a stable sort before folding entries into `compilation_hasher`. The sort key is there to provide that deterministic order and to avoid display collisions such as `SourceType::Runtime` vs `SourceType::Custom("runtime")`. After the order is fixed, we still hash `source_type` together with `content_hash` so changes in the source-type set also affect the fullhash.

CPunisher: I see. But since `source_type` itself is sortable, is`content_hashes.iter().sorted().for_each` enough?

GiveMe-A-Name: Changed to derive `Ord`/`PartialOrd` on `SourceType` and sort directly by the key:

```rust
.sorted_unstable_by_key(|(source_type, _)| *source_type)
```

I didn't switch to `content_hashes.iter().sorted()` because that would sort the whole `(&SourceType, &RspackHashDigest)` tuple, while the intent here is to order entries only by `SourceType`. Keeping the sort keyed on `source_type` makes that explicit and avoids introducing ordering requirements for the digest type.

CPunisher (approved):

PR #13498, opened 2026-03-27
fix: improve pure comment matching

JSerFeng:
## Summary
- extend pure-notation handling to cover `new` expressions and parenthesized call/new expressions
- tighten pure comment parsing so only annotations at the start of a block comment are treated as pure
- add a tree-shaking fixture and snapshot covering `#__PURE__`, `@__PURE__`, and a non-annotation comment that merely mentions `#__PURE__`

## Why
The previous implementation only special-cased `CallExpr`, so cases like `/*#__PURE__*/ new Set()` were not recognized. While fixing that, the first regex relaxation was too broad and would have treated arbitrary comments containing `#__PURE__` as annotations.

## Impact
This keeps the new pure-notation support for `new` and parenthesized expressions, while avoiding false positives that could incorrectly drop side-effectful code.

## Validation
- `pnpm run build:js`
- `pnpm -C tests/rspack-test run testu -t pure_comments_new_expr`
- `cargo fmt --all --check`
- `cargo clippy -p rspack_plugin_javascript --all-targets -- -D warnings`

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
@@ -80,13 +80,11 @@ fn is_pure_call_expr(
   expr: &Expr,
   unresolved_ctxt: SyntaxContext,
   comments: Option<&dyn Comments>,
-  force_pure_flag: bool,
 ) -> bool {
   let Expr::Call(call_expr) = expr else {
     unreachable!();
   };
-  let pure_flag = force_pure_flag
-    || has_pure_comment(comments, expr.span().lo)
+  let pure_flag = has_pure_comment(comments, expr.span().lo)
     || has_pure_comment(comments, call_expr.callee.span().lo);
   if !pure_flag {
     !expr.may_have_side_effects(ExprCtx {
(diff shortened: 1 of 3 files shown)


JSerFeng:
@codex review

crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
```diff
@@ -101,14 +119,66 @@ fn is_pure_call_expr(
...
+  comments
+    .and_then(|comments| comments.get_leading(pos))
+    .is_some_and(|comment_list| {
+      comment_list
+        .iter()
+        .any(|comment| comment.kind == CommentKind::Block && PURE_COMMENTS.is_match(&comment.text))
     })
+}
+
+fn strip_parens(mut expr: &Expr) -> &Expr {
```
ahabhgk: Not needed because we already strip all parens when [the AST is parsed](https://github.com/web-infra-dev/rspack/blob/95fbc1910159e0576c29d8b5139c5930178f5a3b/crates/rspack_plugin_javascript/src/parser_and_generator/mod.rs#L254)

JSerFeng:
@codex review

PR #13514, opened 2026-03-29
fix(config): tighten rule loader/use typings

### LINKED ISSUE #10278: [Bug]: rspack shouldn't allow rules to be configured with the `use: <string>` shorthand while also providing options shorthand.
### System Info

```
  System:
    OS: macOS 14.7.5
    CPU: (10) arm64 Apple M2 Pro
    Memory: 227.81 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.9.0 - ~/.nvm/versions/node/v20.9.0/bin/node
    npm: 10.1.0 - ~/.nvm/versions/node/v20.9.0/bin/npm
    pnpm: 9.15.0 - ~/.nvm/versions/node/v20.9.0/bin/pnpm
    bun: 1.2.1 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 136.0.7103.49
    Safari: 17.6
```

### Details

Not sure if actual bug or feature request.

However, it seems that it should be illegal to mix the `Rule.use` shorthand with the `Rule.options` shorthand like so:
```js
{
        test: /\.[mc]?[jt]sx?$/,
        // This doesn't work with the options shorthand - the options are not applied to the loader.
        // If you change this to `loader: 'builtin:swc-loader'` it will work, and the options will
        // be applied to the loader correctly.
        use: 'builtin:swc-loader',
        options: {
          jsc: {
            parser: {
              syntax: 'typescript',
            },
          },
        },
      },
``` 

# Expected behavior
There should be a typescript error when using the `use: string` shorthand + options shorthand in the `rspack.config.js` file.

# Actual behavior
There is no typescript error when using shorthand + options in the `rspack.config.js` file.

### Reproduce link

https://github.com/ItamarGronich/rspack-use-options-repro

### Reproduce Steps

1. Clone the repo
2. Run `npm install`
3. Run `npm run build`
4. Observe the error message

Checkout comment in `./rspack.config.js` for more details on the issue.

kyungilcho:
## Summary

Tightens the `RuleSetRule` typing around `loader`, `use`, and top-level `options` so invalid shorthand combinations are rejected at type-check time.

It also stops mutating `rule.use` in the adapter and adds a type regression test to lock the behavior in.

## Related links

Closes #10278

## Checklist

### DIFF
--- a/website/docs/en/guide/features/builtin-swc-loader.mdx
+++ b/website/docs/en/guide/features/builtin-swc-loader.mdx
@@ -364,7 +364,7 @@ export default {
   module: {
     rules: [
       {
-        use: 'builtin:swc-loader',
+        loader: 'builtin:swc-loader',
         options: {
           transformImport: [
             {
@@ -394,7 +394,7 @@ export default {
   module: {
     rules: [
       {
-        use: 'builtin:swc-loader',
+        loader: 'builtin:swc-loader',
         options: {
           transformImport: [
             {
@@ -428,7 +428,7 @@ export default {
   module: {
     rules: [
       {
-        use: 'builtin:swc-loader',
+        loader: 'builtin:swc-loader',
         options: {
           transformImport: [
             {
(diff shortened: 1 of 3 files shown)


CPunisher:
Thanks! I remove the unnecessary tests and correct our docs.

PR #13521, opened 2026-03-30
chore: zigbuild for all linux targets

CPunisher:
## Summary

Enable `cargo-zigbuild` for all linux targets to lock the glibc version to 2.28.
Note that the glibc version is determined by node v20, which is the oldest version that rspack supports. And zig 0.14.1 also specifies glibc 2.28, see https://github.com/ziglang/zig/blob/0.14.1/lib/std/Target.zig#L473

## Checklist


CPunisher:
Close in favour of https://github.com/web-infra-dev/rspack/pull/13539

PR #13531, opened 2026-03-30
fix(types): allow partial jsc.parser config when detectSyntax is auto

chenjiahan:
## Summary

- allow `SwcLoaderOptions` with `detectSyntax: 'auto'` to accept partial `jsc.parser` options
- keep the existing `jsc.parser` typing unchanged when `detectSyntax` is `false`

## Related links

- None

## Checklist


CPunisher:
Should we update the docs?

chenjiahan:
@CPunisher This shouldn't require a doc update, as it's a type fix and the public type remain unchanged

PR #13548, opened 2026-03-31
feat(binding): add active-related APIs to ModuleGraphConnection

JSerFeng:
## Summary
- Add `active` getter, `conditional` getter, and `getActiveState(runtime)` method to `ModuleGraphConnection` binding, allowing JS plugins to inspect connection active state
- Add `is_conditional()` public accessor on core `ModuleGraphConnection`
- Add test case (`get-incoming-connections-active`) verifying active incoming connection properties

This is useful to debug side effects bailouts. For example we can visit all active incoming connections to know what module is included only by side effect connections.


JSerFeng:
@codex review

PR #13565, opened 2026-04-01
fix: limit tsfn lifecycle to prevent memory leak

SyMind:
## Summary

This PR fixes TSFN-related memory retention in the Node binding by making hook registration callbacks and function-valued options owned by the `JsCompiler` that created them, then releasing their active TSFN slots when that compiler is closed or dropped.

Previously these callbacks were materialized as plain long-lived `ThreadsafeFunction`s while parsing options, built-in plugin options, and JS hook registrations. Those TSFNs could keep JS closures alive after `compiler.close()`, and those closures could retain `Compiler` / `Compilation` wrappers captured from user code.

## What changed

- Added `CompilerScopedTsFnManager` and `CompilerScopedTsFnHandle`.
- Parse `RawOptions`, `BuiltinPlugin[]`, and `RegisterJsTaps` inside a compiler-local manager scope, so nested callback fields register cleanup with the current `JsCompiler`.
- Switched callback-valued raw options and built-in plugin options from direct `ThreadsafeFunction` fields to compiler-scoped handles.
- Made JS hook register callbacks compiler-scoped while keeping returned tap callbacks as ordinary TSFNs; tap caches continue to be cleared by hook cache cleanup.
- Keep the JS `RawOptions` object on `Compiler` so callbacks stay available for the native compiler lifetime.
- Release all compiler-scoped TSFN slots in `JsCompiler.close()` and when the native compiler wrapper is dropped. Calls after release now report a clear `compiler.close()` error.

## Why this fixes the leak

Compiler-scoped callbacks no longer survive past the native compiler lifetime. Hook tap callbacks that capture a specific compilation remain tied to the existing tap cache lifecycle, and option callbacks stay callable across repeated builds while the compiler is alive.

Together, this lets old `Compilation` wrappers be collected after later builds and lets `Compiler` wrappers be collected after `close()` once user references are gone.

## Validation

Added `compilerCases/tsfn-lifecycle` coverage for:

- hook closures that capture both compiler and compilation
- option callbacks that capture both compiler and compilation
- option callbacks across repeated builds with forced GC
- clear errors when running a closed compiler

## Checklist


SyMind:
@codex review

SyMind:
@codex review

SyMind:
@codex review

SyMind:
@codex review

hardfist (approved):

PR #13573, opened 2026-04-01
chore: change CodSpeed baseline branch from main to v1.x

stormslowly:
## Summary
- Change CI push trigger branches from `main`/`v2` to `v1.x`
- Update concurrency cancel-in-progress to protect `v1.x` branch

## Why
The `v1.x` branch is the active development branch for v1, but CodSpeed was still comparing PR benchmarks against `main`. This caused misleading performance regression detection since the baseline data came from a different branch. Switching the baseline to `v1.x` ensures CodSpeed compares against the correct reference point.

## Test plan
- [ ] Verify CodSpeed benchmark runs on push to `v1.x`
- [ ] Verify PR benchmark comparisons use `v1.x` as baseline

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,7 @@ on:
     types: [opened, synchronize]
   push:
     branches:
-      - main
-      - v2
+      - v1.x
     paths-ignore:
       - '**/*.md'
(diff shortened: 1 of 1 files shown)


stormslowly:
@codex review

PR #13579, opened 2026-04-02
perf: optimize concatenation plugin candidate analysis

LingyuCoder:
## Summary
This draft PR contains the completed portion of the concatenation-plugin stage 1 performance work for `ModuleConcatenationPlugin`.

It focuses on reducing repeated root-independent work during candidate discovery and on moving `get_imports` onto facts-backed data, while preserving current concatenation behavior.

## What changed
- added reusable concat facts scaffolding and conservative runtime reuse helpers
- moved root-independent bailout analysis into facts and kept bailout ordering aligned with the original logic
- split selection-time precheck facts from fuller per-connection facts so `select relevant modules` stays cheap
- rewired `get_imports` to consume facts-backed outgoing connections
- preserved side-effect-only ESM import behavior in the facts-backed `get_imports` path
- annotated the Rust concatenation benchmark as the primary stage-1 performance gate

## Why
Before these changes, the concatenation path repeated a lot of root-independent work during candidate discovery, and `get_imports` still depended on older tuple-packed cache data rather than the facts layer introduced for stage 1.

This PR is the first clean slice of that refactor.

## Validation
Validated during implementation with the existing local gates used for each batch:
- `pnpm run build:js`
- `cargo fmt --all --check`
- `cargo lint`
- `pnpm run test:rs`
- `pnpm --filter @rspack/binding test`
- targeted `rstest` slices for concatenate/scope-hoisting/externals cases
- `pnpm run bench:prepare`
- `cargo codspeed build -m simulation --profile codspeed -p rspack_benchmark --features codspeed`
- `pnpm run bench:rust`
- `pnpm --filter bench run bench`

## Notes
- This PR intentionally excludes the still-in-progress Task 4 work around `try_to_add` incoming-facts rewiring.
- During local verification, some `rstest` shards reported `Worker exited unexpectedly`; those were treated as accepted flaky failures per explicit user instruction and did not block this draft PR.

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs
@@ -923,7 +923,6 @@ impl ModuleConcatenationPlugin {
     let mut imports_cache = RuntimeIdentifierCache::<IdentifierIndexSet>::default();
 
     let module_graph = compilation.get_module_graph();
-    let module_graph_cache = &compilation.module_graph_cache_artifact;
     let module_static_cache = &compilation.module_static_cache;
     let compilation_context = &compilation.options.context;
     let cache_modules = relevant_modules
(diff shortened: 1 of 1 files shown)


LingyuCoder:
Closing this draft because the performance results regressed. Keeping the branch for reference.

PR #13583, opened 2026-04-02
chore: Fix wasm allocator assertions for emnapi 2 GiB memory limit

hardfist:
## Summary
- add a wasm-specific global allocator wrapper in `rspack_allocator`
- assert that `alloc`, `alloc_zeroed`, and `realloc` never return or extend into addresses at or above 2 GiB
- fail fast with a clear panic message when a wasm allocation exceeds emnapi's supported memory range

## Testing
- `cargo fmt --all --check`
- `cargo check -p rspack_binding_api --target wasm32-wasip1-threads`
- `cargo clippy -p rspack_allocator --target wasm32-wasip1-threads -- -D warnings`


CPunisher:
Close as stale

PR #13588, opened 2026-04-02
fix: cjs self reference mangle exports

ahabhgk:
## Summary

Fix mangle exports for cjs self reference dependency

## Checklist


ahabhgk:
@codex review

PR #13599, opened 2026-04-03
feat!: disable bundlerInfo force by default

LingyuCoder:
## Summary
- change `output.bundlerInfo.force` to default to `false` instead of forcing runtime module injection
- keep explicit `bundlerInfo.force = true` and array-based selective injection working
- update defaults snapshots and the English/Chinese docs to match the new behavior

## Why
`rspackChunk` is already enough for many runtime detection paths, so forcing bundler info into every build by default adds extra runtime output that no longer needs to be on by default.

### DIFF
--- a/website/docs/en/guide/migration/rspack_1.x.mdx
+++ b/website/docs/en/guide/migration/rspack_1.x.mdx
@@ -235,6 +235,20 @@ export default {
 };
 ```
 
+### Changed default value of output.bundlerInfo.force
+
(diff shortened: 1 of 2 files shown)


tests/e2e/browser-cases/basic-react/index.test.ts
```diff
@@ -3,11 +3,6 @@ import { test, expect } from '@/fixtures';
 test('@rspack/browser should bundle react app successfully', async ({
   page,
 }) => {
-  // There should be a long bundle result
```
LingyuCoder, line 6: Since the "Bundler Info" is turned off and the runtime module is removed, the size of the production output has changed. Using a size - based test for this test case is inherently unreasonable. It should be tested based on the content instead.

PR #13606, opened 2026-04-03
fix(esm_library): extract TLA shared modules to break circular dependency

JSerFeng:
## Summary
- When using ESM library output with top-level await (TLA), a circular dependency occurs: the main chunk `await import()`s the async chunk (because it has TLA), but the async chunk imports shared modules back from the main chunk via the "available modules" optimization.
- Adds `extract_tla_shared_modules()` in the `optimize_chunks` hook that scans async chunks' dependency graphs and extracts modules residing in ancestor chunks into separate chunks to break the cycle.
- Emits a warning when TLA is used with ESM library output, since bundled execution order may differ from the original source.

## Link

https://github.com/web-infra-dev/rslib/issues/1573

## Test plan
- [x] `cargo check -p rspack_plugin_esm_library` passes
- [x] `cargo test -p rspack_plugin_esm_library` — all 19 tests pass
- [ ] Manual verification with a TLA + shared dependency scenario

### DIFF
--- a/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
+++ b/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
@@ -119,13 +119,13 @@ pub(crate) fn extract_tla_shared_modules(compilation: &mut Compilation) -> bool 
 
       // Check if target is in an ancestor chunk of origin_async_chunk → circular dep
       for &target_chunk in target_chunks {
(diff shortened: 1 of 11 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

Timeless0911 (approved):

PR #13674, opened 2026-04-10
perf(code-splitting): precompute block module buckets

LingyuCoder:
## Summary

- precompute per-root block module buckets during `CodeSplitter::prepare`
- store grouped connections and block lists in shared slice-backed structures to reduce serial rebuild work in `get_block_modules`
- keep the serial phase focused on runtime-dependent `active_state` evaluation and queue decisions

## Why

`get_block_modules` was still doing runtime-independent bucket reconstruction inside the serial process-queue path. That work can be prepared in parallel ahead of time because the grouping from root module -> block -> target module + connection group does not depend on runtime.

This change moves that bucket-building step into `prepare`, which should reduce serial overhead in `extract_block_modules` / `get_block_modules` and cut some repeated vector cloning.

## Validation

- `cargo fmt --all --check`
- `CARGO_INCREMENTAL=0 cargo test -p rspack_core build_prepared_block_modules_groups_connections_by_block -- --nocapture`
- `CARGO_INCREMENTAL=0 cargo check -p rspack_core --tests`


LingyuCoder:
Closing this stacked PR for now because the benchmark regressed on rust@build_chunk_graph.

PR #13689, opened 2026-04-13
perf: trace hook interception only pays off once global tracing is already on

SyMind:
## Summary

`TraceHookPlugin` should only be registered when global tracing is enabled.

Eagerly registering `TraceHookPlugin` introduces measurable overhead. In the screenshot below, the blank gap on the Rust side during the make phase is caused by retrieving the `ContextModuleFactory` hook taps while the JavaScript thread is busy processing the Less loader, which blocks that hook collection work.

image

After this optimization, the Rust side of the make phase returns to normal.

image

## Checklist


SyMind:
@codex review

hardfist (approved):

PR #13690, opened 2026-04-13
feat(cli): use jiti to load typescript config

hardfist:
## Summary

- replace the `pirates`-based TS config register path with `jiti.import` and native loader handling in `rspack-cli`
- add `auto | jiti | native` config loader modes, keep `register` as a compatibility alias, and preserve external TS require hooks such as `ts-node/register`
- inline the ESM/package-type detection helpers into config loading and remove the old `crossImport` helper files
- add `jiti` as a runtime dependency, externalize it from the CLI bundle, and remove the unused `pirates` dependency
- expand config-loading coverage for CJS, ESM, native loader, Jiti loader, monorepo, and `ts-node/register` scenarios
- update English and Chinese CLI/config docs to describe the new loader behavior and defaults

## Testing

- `pnpm run build:js`
- `pnpm --filter @rspack/cli build`
- `pnpm --filter @rspack/cli test tests/build/config/config.test.ts`
- `pnpm --filter @rspack/cli test tests/preview/ts-config-file/index.test.ts`


packages/rspack-cli/prebundle.config.js
chenjiahan, line 1: Can be replaced with https://github.com/web-infra-dev/rsbuild/blob/main/packages/core/rslib.config.ts#L149-L160

chenjiahan (approved):

PR #13701, opened 2026-04-14
feat(tree-shaking): recognize pure global constructors and calls

JSerFeng:
## Summary

- Teach Rspack's side-effect detector about built-in pure globals so unused `new Set()`, `new Map()`, `new Uint8Array()`, `Object.keys(x)`, `Array.isArray(x)`, `String(x)`, etc. are now tree-shaken.
- Previously, SWC's `may_have_side_effects` only recognized `Date` as a pure callee and only empty-fn / pure-class expressions as pure `new` callees — so `export const cache = new Map()` was always retained even when `cache` was unused. Rolldown handles this via `oxc_ecmascript::side_effects::MayHaveSideEffects`, which maintains a pure-globals list; this PR mirrors that approach in Rspack.
- Shadowing-safe: the callee must resolve to an unresolved global (`ctxt == unresolved_ctxt`), so `const Set = sideEffect(); new Set()` is still kept. Arguments are recursively checked via the existing `are_pure_args` / `is_pure_call_args` helpers, so `new Set([impureArg()])` is still kept.

### Lists

- **`new` pure constructors**: `Set`, `Map`, `WeakSet`, `WeakMap`, `Object`, `Array`, `String`, `Number`, `Boolean`, `Date`, `ArrayBuffer`, `SharedArrayBuffer`, and all 11 TypedArrays (`Uint8Array`, `Int8Array`, `Uint8ClampedArray`, `Uint16Array`, `Int16Array`, `Uint32Array`, `Int32Array`, `Float32Array`, `Float64Array`, `BigInt64Array`, `BigUint64Array`).
- **Direct-call pure globals**: `Array`, `Object`, `String`, `Number`, `Boolean`, `Symbol`, `Date`.
- **Member-call pure globals**: `Object.{keys,values,entries,getOwnPropertyNames,getOwnPropertySymbols,getOwnPropertyDescriptor,getOwnPropertyDescriptors,getPrototypeOf,create,freeze,fromEntries,is,assign}`, `Array.{isArray,from,of}`.
- Conservatively omitted (conditional on args): `BigInt`, `RegExp`, `Error`/`TypeError`/…, `Promise`, `Reflect.*`, `JSON.*`. These can be added later with argument validation.

## Test plan

- [x] `cargo build -p rspack_plugin_javascript`
- [x] `cargo test -p rspack_plugin_javascript --lib side_effects`
- [ ] CI: new treeshaking case at `tests/rspack-test/treeShakingCases/global-pure-new/` (snapshot will be generated by CI — includes happy path, shadowed-binding guard, and impure-arg guard).
- [ ] CI: verify existing `tests/rspack-test/treeShakingCases/pure_comments_new_expr/` still passes (unchanged behavior: inner `new Set()` inside a user function is still tied to the function's lifetime).


JSerFeng:
Thanks for the reviews. All 7 points were valid — the first pass was too aggressive. Pushed 56d07be17c to address them:

**Tightened argument checks.** Added `are_args_trivially_safe` / `is_trivially_safe_expr` — args must be literal/template-with-no-exprs/unary-of-safe, or unresolved `undefined`/`NaN`/`Infinity`. Objects, arrays, and bare identifiers are no longer considered safe. This rules out `new Set([proxyArr])`, `new Uint8Array(objWithValueOf)`, `String(objWithToPrimitive)`, etc.

**Removed mutating / iterating member calls** from the allowlist:
- `Object.assign`, `Object.freeze`, `Object.create`, `Object.fromEntries` (mutate / run setters / run iterator)
- `Object.keys`, `Object.values`, `Object.entries`, `Object.getOwnPropertyNames`, `Object.getOwnPropertySymbols`, `Object.getOwnPropertyDescriptor(s)`, `Object.getPrototypeOf` (Proxy `ownKeys` / `get` traps)
- `Array.from` (iterator protocol + mapper fn)

**Kept member calls** limited to pure type/identity checks: `Array.isArray`, `Array.of`, `Object.is`, `Number.isInteger` / `isFinite` / `isNaN` / `isSafeInteger`.

**Direct-call split into two tiers:**
- `Boolean(x)` — pure regardless of args (only truthiness check, no coercion).
- `Array`, `Object`, `String`, `Number`, `Symbol`, `Date` — pure only when args are trivially safe.

**Shadow test fixed.** Moved the `const Set = impureSet();` / `const Map = impureSet();` into a separate `shadow.js` module and made the declarations shadow the real global names. Both `new Set()` and `new Map()` in `shadow.js` are correctly preserved because the identifiers have the local (not unresolved) ctxt.

**Verified locally with Node 22 / pnpm 10.26.2:**
- `cargo check -p rspack_plugin_javascript` ✓
- `rstest TreeShaking.test.js` — 84 passed ✓
- `rstest StatsOutput.test.js` — 152 passed ✓ (previously failing `runtime-specific-used-exports` snapshot regenerated — `Symbol()` is now correctly pure, so `module.js` with `export const identity = Symbol();` gets orphaned, a net improvement)
- New snapshot `tests/rspack-test/treeShakingCases/global-pure-new/__snapshots__/treeshaking.snap.txt` shows pure globals wrapped in `null &&` and impure/shadowed forms preserved.

JSerFeng:
@codex review

JSerFeng:
Good catch — that was a real soundness bug. Pushed a85c29313c.

The previous `TriviallySafeArgs` gate was too permissive for callees that throw based on argument *values*. Specifically, all of these were being incorrectly classified as pure (and could be tree-shaken silently dropping the throw):

- `new Set(1)` → throws TypeError (not iterable)
- `new Map("foo")` → throws TypeError (string iterates chars, not [k,v] pairs)
- `new Array(-1)` → throws RangeError
- `new Array(1.5)` → throws RangeError
- `new Uint8Array(-1)` / `new Uint8Array(1.5)` → throws RangeError
- `new Date(1n)` / `new Number(1n)` → throws TypeError (BigInt → Number)

**Fix:** split `ArgGate` into four variants by argument-shape requirement:

| Gate | Allowed args | Used by |
|------|-------------|---------|
| `AnyPureArgs` | any pure expression | `Boolean`, `Array.isArray`, `Object.is`, `Number.isXxx`, `Array.of` |
| `TriviallySafeArgs` | any literal (number/string/bool/null/regex/bigint) + `undefined`/`NaN`/`Infinity` | `String`, `Object`, `Symbol` (call only) |
| `NullishOrNoArgs` | `null`, `undefined`, `void <safe>`, or absent | `Set`, `Map`, `WeakSet`, `WeakMap`, `Array`, `Date`, `Number` |
| `NullishOrNonNegativeIntLiteral` | nullish or non-negative integer literal in `[0, 2^32)` (excluding BigInt, fractional, negative) | TypedArrays, `ArrayBuffer`, `SharedArrayBuffer` |

`Boolean` got promoted to `AnyPureArgs` in both positions — `ToBoolean` never throws and never invokes user code, so even `Boolean(myVar)` is pure.

**Verification.** Snapshot test extended with 9 new "MUST be kept" cases covering each throw category. Regenerated snapshot confirms:

- Pure forms (`new Set()`, `new Set(null)`, `new Uint8Array(16)`, `Boolean(marker)`, etc.) → wrapped in `null && (…)` ✓
- Throwing forms (`new Set(1)`, `new Array(-1)`, `new Uint8Array(1.5)`, `new Date(1n)`, etc.) → preserved verbatim ✓
- Impure-arg forms (`Boolean(impureArg())`, `new Set([impureArg()])`) → preserved ✓
- Shadow cases → preserved ✓

Unit tests bumped from 16 → 25, including direct arg-shape checks against synthetic `Lit::Num`/`Lit::Str`/`UnaryExpr` expressions.

JSerFeng:
The 14MB size delta is **not from this PR** — it's a CI infrastructure mismatch introduced by https://github.com/web-infra-dev/rspack/pull/13726 (\`d79172e7\`).

**What's happening:**

- The size-limit script fetches the baseline from \`rspack-ecosystem-benchmark/data\`. The most recent baseline is for \`b962256bb8\` at 51,084,280 bytes (51MB), uploaded 2026-04-17T03:27Z.
- That baseline is built by the external benchmark workflow with the **\`release\` profile** (\`lto = "fat"\`, \`codegen-units = 1\`, \`opt-level = 3\`, \`strip = true\`).
- After #13726, this PR's CI uses \`size-limit.yml\` which now downloads the artifact from \`Build Linux\` (built with the **\`ci\` profile** — \`lto = false\`, \`codegen-units = 256\`, \`opt-level = 2\`) and runs \`strip\` manually.
- After strip, my binary is 65,767,832 bytes (65MB). The 14MB delta is the cost of disabling LTO + lower codegen-units, not the ~600 lines of Rust this PR adds.

**Evidence this isn't the PR's code:**
- Cargo unit-test count went 16 → 25 (one new module, ~500 lines)
- Other PRs with this size-limit check passed because they ran *before* d79172e7 landed (they used the \`release\`-profile build path, which matched the benchmark baseline)

**Fix options (need maintainer action):**
1. Update the external benchmark workflow to also build with the \`ci\` profile so baselines match.
2. Or revert d79172e7 and keep the \`release\`-profile build in size-limit.yml.
3. Or override this check for now and re-run after benchmark data catches up with apples-to-apples profiles.

Happy to rebase or trigger a re-run once that's resolved.

JSerFeng:
The CodSpeed regression report has the **same root cause** as the size-limit failure — note its own footnote:

> No successful run was found on \`main\` (2bf2cd4) during the generation of this report, so **b962256 was used instead as the comparison base**. There might be some changes unrelated to this pull request in this report.

That's the same \`b962256bb8\` baseline already discussed for size-limit, built before:
- \`4bd8ba23ae\` (rust toolchain bump nightly-2025-11-13 → nightly-2026-04-16)
- \`d79172e7\` (CI profile change)

The two regressed benchmarks are tiny in absolute terms:
- \`persistent_cache_restore@basic-react-development\`: 25.6ms → 26.1ms (+0.5ms)
- \`create_chunk_assets\`: 3.6ms → 3.7ms (+0.1ms)

Both are well within toolchain/codegen noise from a one-month nightly bump, and they don't touch the parser hot path where this PR's code runs (which is in \`is_pure_new_expr\` / \`is_pure_call_expr\` during AST scan, not chunk-asset generation or cache restore).

Once a fresh \`main\` baseline is published (after the next benchmark run on the new toolchain/profile), both size-limit and codspeed should compare apples-to-apples.

Also worth noting: the Rsdoctor bundle-diff for this PR shows **all output bundles got slightly smaller** (-0.0% across popular-libs / react-1k/5k/10k / rome / ui-components), confirming the PR's actual effect is a minor net reduction in emitted JS, not growth.

JSerFeng:
@codex review

PR #10760, opened 2025-06-23
perf: parallel get concatenated modules final binding names

LingyuCoder:
## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #13707, opened 2026-04-14
fix: escape invalid chars anywhere in to_identifier_with_escaped

### LINKED ISSUE #12940: [Bug]: Importing a JSON file containing ! in a key results in a malformed js export statement
### System Info

  System:
    OS: Windows 11 10.0.26200
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-13800H
    Memory: 15.39 GB / 63.67 GB
  Binaries:
    Node: 22.19.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.12.0 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.9.3 - C:\Program Files\nodejs\npm.CMD
    bun: 1.3.2 - ~\.bun\bin\bun.EXE
  Browsers:
    Edge: Chromium (140.0.3485.54)
    Internet Explorer: 11.0.26100.7309

### Details

Trying to import a JSON file containing a key that starts with `!`, like "!top", results in a `JavaScript parse error: Unexpected token `!`. Expected identifier or string` error when minifying via SWC or using the bundled code in other projects. 

rslib (non bundled) generates the export key as `export { __webpack_exports___top as !top, ... }` instead of `export { __webpack_exports___top as "!top", ... }`.

I would expect rspack to use `"!top"` instead of `!top`.

### Reproduce link

[Link](https://github.com/Arnagos/rspack-repro-12940)

### Reproduce Steps

1. Import the relevant JSON file via `import fileName from './file-name.json';`
2. Make sure that minification is enabled and run rslib build with format configured to esm

Example JSON:
`{  "!top": [] }`

JSerFeng:
## Summary

- Fixes #12940. `to_identifier_with_escaped` only fixed up the first character, so pre-escaped input with invalid characters later in the string (e.g. JSON keys like `!top` or `with space` passed through `find_new_name`) leaked into JS identifier positions.
- Under `library: { type: 'modern-module' }` this produced unparseable output like `var _!top = ...` / `export { _!top as "!top" }` when a JSON module with non-identifier keys flowed through `EsmLibraryPlugin`.
- Fix delegates to `to_identifier` for the full escape; the `Cow::Borrowed` fast path keeps the perf win from #10760 for already-valid inputs.

## Test plan

- [x] `cargo test -p rspack_core --lib utils::identifier` — added unit tests covering `!top`, `_!top`, `with space`, `a.b`, plus the fast-path and first-char-only cases.
- [x] New `configCases/library/modern-module-json-special-keys` regression test builds a JSON module with `"!top"`, `"with space"`, `"a.b"`, `"normal"` through the `modern-module` library type and asserts the rendered bundle has no invalid `var <name> =` and uses string-quoted aliases for the non-identifier exports. Dynamic `import()` of the bundle verifies runtime exports match the JSON.
- [x] Extended `configCases/json/rspack-issue-10722` (existing modern-module JSON test) with `!top` and `with space` keys.
- [x] `pnpm run test:base -- -t "configCases/library"` — 320 passed.
- [x] `pnpm run test:base -- -t "configCases/json"` — 279 passed.
- [x] `pnpm run test:base -- -t "concatenate"` — 306 passed.


JSerFeng:
@codex review

ahabhgk (approved):

PR #13717, opened 2026-04-16
feat(config): update default performance budgets

chenjiahan:
## Summary

Update the default performance budgets to `300 * 1024` for `maxAssetSize` and `500 * 1024` for `maxEntrypointSize` so the defaults better fit modern applications and distinguish asset vs entrypoint limits.

## Checklist

### DIFF
--- a/packages/rspack/src/config/defaults.ts
+++ b/packages/rspack/src/config/defaults.ts
@@ -990,8 +990,8 @@ const applyPerformanceDefaults = (
   { production }: { production: boolean },
 ) => {
   if (performance === false) return;
-  D(performance, 'maxAssetSize', 250000);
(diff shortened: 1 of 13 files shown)


chenjiahan:
Addressed the review feedback in `1bd3bc9`.

- Updated the docs to render the evaluated byte defaults (`307200` / `512000`) instead of expression strings.
- Updated the `@default` values in `packages/rspack/src/config/types.ts` to the numeric byte values as well.
- Re-ran the relevant local checks, including Prettier on the touched files, `pnpm --dir website run check:spell`, and `pnpm run build:js`.

This should also resolve the failing `Prettier` CI job from the previous commit, which was complaining about the two docs files.

LingyuCoder (approved):

PR #13726, opened 2026-04-16
chore(ci): use ci profile for binary size-limit job to speed up build

stormslowly:
## Summary

- Switch the binary size-limit CI job from `release` profile to `ci` profile
- The `ci` profile (`opt-level=2`, `lto=false`, `codegen-units=256`) builds significantly faster than `release` (`opt-level=3`, `lto=true`, `codegen-units=1`)
- Size comparison remains valid since both PR head and base will use the same profile

## Test plan

- [ ] Verify the size-limit job completes successfully with the `ci` profile
- [ ] Confirm build time is reduced compared to previous runs with `release` profile

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
 
   size-limit:
     name: Binary Size Limit
-    needs: [check-changed]
+    needs: [check-changed, build-linux]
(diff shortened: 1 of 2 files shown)


CPunisher:
why not reuse the aritifact for testing?

stormslowly:
> why not reuse the aritifact for testing?

Nice shot, let me do it.

PR #13731, opened 2026-04-16
fix: rspack sources cache save and recovery

ahabhgk:
## Summary

Follow up bugfix for https://github.com/web-infra-dev/rspack/pull/13706

Detailed bug:

Inconsistencies after restoring the persistent cache during the `make` phase lead to a massive cache miss when the minimize persistent cache processes the chunks. Modules restored from the persistent cache differ from those freshly built without a persistent cache.

For example, during the first build (a cold build without persistent cache), the module hash is `xxx`, but during the second build (a hot build with persistent cache), the module hash becomes `yyy`. This difference in module hashes alters the chunk content (such as the `fullhash`), which causes the minimize cache to miss. However, during the third build (a hot build with persistent cache), the module hash remains `yyy` (as expected).

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -85,7 +85,7 @@ regex               = { version = "1.12.3", default-features = false, features =
 regex-syntax        = { version = "0.8.10", default-features = false, features = ["std"] }
 regress             = { version = "0.10.5", default-features = false, features = ["pattern"] }
 rspack_resolver     = { features = ["package_json_raw_json_api", "yarn_pnp"], version = "=0.8.0", default-features = false }
-rspack_sources      = { version = "=0.4.21", default-features = false }
+rspack_sources      = { git = "https://github.com/rstackjs/rspack-sources", branch = "expose-inner", default-features = false }
 rustc-hash          = { version = "2.1.2", default-features = false }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -85,7 +85,7 @@ regex               = { version = "1.12.3", default-features = false, features =
 regex-syntax        = { version = "0.8.10", default-features = false, features = ["std"] }
 regress             = { version = "0.10.5", default-features = false, features = ["pattern"] }
 rspack_resolver     = { features = ["package_json_raw_json_api", "yarn_pnp"], version = "=0.8.0", default-features = false }
-rspack_sources      = { version = "=0.4.21", default-features = false }
+rspack_sources      = { git = "https://github.com/rstackjs/rspack-sources", branch = "expose-inner", default-features = false }
```
jerrykingxyz: We should avoid using the Git source directly. If it is necessary to expose all fields, we could instead consider adding an `rkyv` feature to `rspack_sources`, thereby enabling internal support for `rkyv`.

ahabhgk: So, as long as the external data structure implements rkyv, it can be cached using rspack_cacheable?

jerrykingxyz (approved):

PR #13765, opened 2026-04-20
fix(rstest): no longer mark rs.mock module as async

fi3ework:
## Summary

- Remove the `JavascriptParserExt::handle_top_level_await` extension trait and its two call sites in `process_mock` (1-arg and 2-arg branches).
- They were introduced in #11050 to support async mock factories, but after the mock runtime was simplified in #11521 to plain synchronous `__webpack_require__.rstest_mock(...)` calls, there is no `await` in the emitted code — so `build_meta.has_top_level_await = true` only serves to pointlessly wrap the module in an async module wrapper and propagate the async requirement to every importer.
- Since async mock factories are no longer supported, drop this legacy plumbing.

## Test plan

- [x] Add two regression assertions in `tests/rspack-test/configCases/rstest/mock/test.js` that read the `mockFactory.js` (2-arg `rs.mock(id, factory)`) and `manualMock.js` (1-arg `rs.mock(id)`) bundle outputs and assert they do not contain `__rspack_async_done` — the identifier that only appears inside the `__webpack_require__.a(...)` async wrapper.
- [x] `cargo check -p rspack_plugin_rstest` passes.
- [x] `configCases/rstest/mock` passes locally.

### DIFF
--- a/crates/rspack_plugin_rstest/src/parser_plugin.rs
+++ b/crates/rspack_plugin_rstest/src/parser_plugin.rs
@@ -65,16 +65,6 @@ pub struct RstestParserPlugin {
   options: RstestParserPluginOptions,
 }
 
-trait JavascriptParserExt<'a> {
-  fn handle_top_level_await(&mut self);
-}
-
-impl<'a> JavascriptParserExt<'a> for JavascriptParser<'a> {
-  fn handle_top_level_await(&mut self) {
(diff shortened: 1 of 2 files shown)


fi3ework:
duplicated with https://github.com/web-infra-dev/rspack/pull/13764/.

PR #13774, opened 2026-04-21
perf: Replace parser hook dispatch SmallVecs with bitmasks

hardfist:
## Summary
- Replace `JavaScriptParserPluginDrive` hook dispatch storage with `u64` bitmasks
- Iterate plugin indices via bit scanning to avoid heap allocations in the common case
- Keep hook order stable and fail fast if more than 64 parser plugins are registered

## Testing
- `cargo fmt --all --check`


SyMind:
@codex review

SyMind (approved):

PR #13777, opened 2026-04-21
feat(cli): add --open / --no-open to rspack dev / serve

### LINKED ISSUE #13727: [Feature]: Add to `rspack dev` cli flags `--no-open` and `--open` to override `devServer.open`.

Useful during development in multiple browsers, when bunch of other tabs are open, when you want to open specific page (especially as package script) or when you only want to see how dev server starts for debugging purposes. 
Webpack [has](https://webpack.js.org/configuration/dev-server/#devserveropen) `--no-open`  and `--open` flags. While Vite [only](https://vite.dev/guide/cli) `--open` flag. 
  `rspack preview` [has](https://github.com/web-infra-dev/rspack/issues/13727) this flag or at least similar one.

Flags `--no-open` and `--open` override `devServer: { open: boolean | string | object | [string, object]}`.

 `rspack dev --no-open` sets `open: false`.
 `rspack dev --open value` where `value` is the same as can be passed to config. Multiple `--open` flags are treated as array of pages passed into `open`. Similar to Webpack.

[Example](https://webpack.js.org/configuration/dev-server/#devserveropen) from Webpack's docs. 1-to-1 recreation would be very nice feature to have.

mvanhorn:
## Summary

Adds `--open` and `--no-open` CLI flags to `rspack dev` (aka `serve`), so users can override `devServer.open` at the invocation level instead of editing `rspack.config.js` per environment.

Closes #13727

## Why

- `rspack preview` already has `--open` but `rspack dev` / `serve` did not, so package-script callers had no way to flip browser opening without editing the config.
- Webpack CLI has `--open` / `--no-open`, and Vite has `--open` - aligning with webpack here matches the existing "1-to-1 recreation" goal in the issue.

## Change

`packages/rspack-cli/src/commands/serve.ts`:

- `ServerOptions` gains `open?: boolean | string` (matches `DevServer.open` at its most common shape).
- New `.option('--open [value]', ...)` on the serve command. `cac` parses this as:
  - `--open`       -> `open: true`
  - `--open <url>` -> `open: '<url>'` (string)
  - `--no-open`    -> `open: false`
- After resolving `devServerOptions`, the CLI value (if defined) takes precedence over `devServer.open` from config. When the flag is absent, config wins, which keeps existing behavior unchanged.

Verified `cac`'s `--open [value]` / `--no-open` parse matrix against a minimal program:

```
--no-open      -> { open: false }
--open http://x -> { open: 'http://x' }
--open         -> { open: true }
```

## Scope

- The richer `devServer.open` shapes (object, [string, object], repeated-flag array) are not wired in - the string/boolean path covers the webpack-docs 1-to-1 example and almost all package.json use cases. Happy to add multi-open array semantics in a follow-up if the maintainers want it.
- `preview.ts` already has `--open` (boolean only). Not touched here.

## Testing

- `git diff` is localized to `serve.ts` and adds 9 lines.
- Pre-existing `tsc --noEmit` errors (all `Cannot find module '@rspack/core'`) reflect the monorepo's expected build-first step; no new errors are introduced in `serve.ts`.

### DIFF
--- a/packages/rspack-cli/src/commands/serve.ts
+++ b/packages/rspack-cli/src/commands/serve.ts
@@ -22,6 +22,7 @@ type ServerOptions = CommonOptionsForBuildAndServe & {
   hot?: boolean | 'only';
   port?: number | string;
   host?: string;
+  open?: boolean | string;
 };
 
 function normalizeHotOption(
@@ -47,7 +48,11 @@ export class ServeCommand implements RspackCommand {
     commonOptionsForBuildAndServe(commonOptions(command))
       .option('--hot [mode]', 'enables hot module replacement')
       .option('--port <port>', 'allows to specify a port to use')
-      .option('--host <host>', 'allows to specify a hostname to use');
+      .option('--host <host>', 'allows to specify a hostname to use')
+      .option(
+        '--open [value]',
+        'open browser on server start; pass --no-open to disable, or --open <url> to open a specific URL',
+      );
 
     command.action(
       cli.wrapAction(async (cliOptions: ServerOptions) => {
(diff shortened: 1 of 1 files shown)


chenjiahan (approved):
Thank you!

PR #13782, opened 2026-04-21
fix(test-tools): move wast-loader to dependencies

colinaaa:
## Summary
- move `wast-loader` from `devDependencies` to `dependencies` in `@rspack/test-tools`
- keep dependency ordering consistent and update `pnpm-lock.yaml` importer entries accordingly
- ensure runtime consumers of `setup-env` can resolve `wast-loader` without relying on dev-only install scope

https://github.com/web-infra-dev/rspack/blob/e7927e392dd277d1af9ad70361e90110d660c996/packages/rspack-test-tools/src/helper/setup-env.ts#L117

## Test Plan
- pnpm -C packages/rspack-test-tools build

### DIFF
--- a/packages/rspack-test-tools/package.json
+++ b/packages/rspack-test-tools/package.json
@@ -45,6 +45,7 @@
     "@babel/parser": "7.29.2",
     "@babel/traverse": "7.29.0",
     "@babel/types": "7.29.0",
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
👍

PR #13795, opened 2026-04-22
feat(watch): support mixed ignored patterns

### LINKED ISSUE #10596: [Feature]: `watchOptions.ignored` support `(string | RegExp)[]` type

As a framework developer, I would like to add some extra `watchOptions.ignored` configuration. If the original `watchOptions.ignored` is a RegExp, I cannot make the user configuration and my configuration work at the same time, because rspack does not support the `(string | RegExp)[]` type

https://rspack.rs/config/watch#watchoptionsignored

User's configuration:
```ts
export default {
  //...
  watchOptions: {
    ignored: /\.git/,
  },
};
```

After merged configuration:
```ts
export default {
  //...
  watchOptions: {
    ignored: [/\.git/,  '**/dist'],
  },
};
```

`watchOptions.ignored` support `(string | RegExp)[]` type.

```diff
- type ignored = RegExp | string | string[];
+ type ignored = RegExp | string | (string | RegExp)[];
```

Perfecto23:
## Summary

Extend `watchOptions.ignored` to accept `string | RegExp | (string | RegExp)[]`, so framework authors can merge user-supplied RegExp ignores with framework-appended glob ignores in a single array.

This is a Rspack extension for framework composition use cases (see #10596), not webpack `watchOptions.ignored` parity.

### What changed

- **Rust watcher data model** — new `FsWatcherIgnored::Mixed(Vec<FsWatcherIgnoredItem>)` variant that matches both glob paths and `RspackRegex` patterns in a single pass.
- **Glob parent-path matching** — `glob_match_path` now walks parent directories so that patterns like `**/.cache` correctly ignore child paths (e.g. `/project/.cache/file.js`). This applies to all glob-based `ignored` variants, not only mixed arrays.
- **Native watcher event filtering** — `Trigger::on_event` now checks ignored at the event entry point (via `to_string_lossy` for non-UTF-8 safety), preventing ignored sub-paths from triggering rebuilds through parent directory association.
- **NAPI binding** — `NativeWatcherOptions.ignored` extended to `string | (string | RegExp)[] | RegExp` with `Either4` dispatch.
- **Node/watchpack fallback** — mixed arrays are normalized into a predicate function: string globs are converted via `glob-to-regexp` (prebundled) and merged into one `RegExp`; user `RegExp` instances are tested individually with `lastIndex` reset to handle stateful `/g` `/y` flags.
- **RSC coordinator** — client compiler "watch nothing" override changed from `() => true` to `/.*/` to stay compatible with native watcher's function rejection.
- **Browser build** — added resolve alias for `glob-to-regexp` in `rslib.browser.config.ts`, consistent with the existing `watchpack` alias pattern.
- **Tests** — new `compilerCases/watch-options-ignored.js` covering both node and native watcher paths: verifies ignored glob/regex files do not trigger `invalid`/rebuild, and normal watched files still trigger rebuild.
- **Docs** — updated English and Chinese `watchOptions.ignored` docs with mixed array examples; softened glob description to "glob semantics" without referencing the internal implementation.

## Related links

Closes #10596

## Checklist

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -1779,9 +1779,9 @@ export interface NativeWatcherOptions {
   aggregateTimeout?: number
   /**
    * The ignored paths for the watcher.
-   * It can be a single path, an array of paths, or a regular expression.
+   * It can be a single path, a regular expression, or an array mixing paths and regular expressions.
    */
-  ignored?: string | string[] | RegExp
+  ignored?: string | (string | RegExp)[] | RegExp
 }
 
 export interface NodeFsStats {
--- a/crates/rspack_binding_api/src/native_watcher.rs
+++ b/crates/rspack_binding_api/src/native_watcher.rs
@@ -4,20 +4,32 @@ use std::{
   time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
-use napi::bindgen_prelude::*;
+use napi::{Either, bindgen_prelude::*};
 use napi_derive::*;
 use rspack_paths::ArcPath;
 use rspack_regex::RspackRegex;
-use rspack_watcher::{FsEventKind, FsWatcher, FsWatcherIgnored, FsWatcherOptions};
+use rspack_watcher::{
+  FsEventKind, FsWatcher, FsWatcherIgnored, FsWatcherIgnoredItem, FsWatcherOptions,
+};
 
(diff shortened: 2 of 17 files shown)


stormslowly:
due the options may be shared between watchpack and rspack watcher, so i will merge this after https://github.com/webpack/watchpack/pull/337 release.

BTW, let me fix the conflicts.

thanks a lot.

Perfecto23:
@stormslowly tks

PR #13803, opened 2026-04-22
fix(esm): rewrite evaluated in operator for externals

JSerFeng:
## Summary

- normalize `DefaultWithNamed` ids before `get_used_name` in `render_evaluated_in_operator`
- remove the external-only fallback and keep this path aligned with the actual usage-info model
- add regression coverage for the rslog-like repro and the concatenated externals case

## Related links

- fixes the `node_util is not defined` repro from `rstest/packages/vscode`
- related to `63ec093274df733e317071b076a4a8639860087b` (`fix(esm): deconflict external bindings in module output`)

## Checklist


JSerFeng:
@codex review

9aoy (approved):

PR #13804, opened 2026-04-22
Move cac from a devDependency to dependency of @rspack/cli as it's needed for the published types

dgoldstein0:
## Summary

cac is imported in the published types, so needs to be a dependency instead of a devDependency.  Otherwise usage of `import {defineConfig} from '@rspack/cli'` in typescript files can cause type errors when `cac` isn't installed or otherwise unable to be resolved.

usage in the published package:
```
    node_modules/@rspack/cli/dist/cli.d.ts:import { type CAC } from 'cac';
    node_modules/@rspack/cli/dist/utils/options.d.ts:import type { Command } from 'cac';
```

FYI this was also an issue in rspack 1.x as well

## Checklist

### DIFF
--- a/packages/rspack-cli/package.json
+++ b/packages/rspack-cli/package.json
@@ -33,13 +33,15 @@
     "prepare": "prebundle",
     "test": "rstest"
   },
+  "dependencies": {
+    "cac": "^7.0.0"
+  },
   "devDependencies": {
     "@discoveryjs/json-ext": "^0.5.7",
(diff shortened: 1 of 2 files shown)


chenjiahan:
Thank you for your contribution! I will bundle the dts files to fix this issue, see: https://github.com/web-infra-dev/rspack/pull/13807

dgoldstein0:
that works too, whatever ends up avoiding undeclared dependencies in the published package will solve this pain point.

PR #13805, opened 2026-04-23
fix(test-tools): support ESM default config exports

colinaaa:
## Summary

- Support ESM default-exported rspack.config.js files in @rspack/test-tools when configs are loaded through sync require().
- Keep CommonJS behavior unchanged by only unwrapping values that match the observed Node ESM namespace shape.
- Add focused configCases covering ESM object/factory configs and CommonJS object/factory regression cases with real default properties.

## Checklist


chenjiahan:
ESM default config exports are now supported by #15714 and #15761, so this PR can be closed as superseded.

Issue #13728, opened 2026-04-16
[Feature]: pureFunctions - support a function created by a factory


lukpsaxo:
### What problem does this feature solve?

Some functions are created by factories and are still pure e.g. reselect createSelector

### What does the proposed API of configuration look like?

[...]

PR #13810, opened 2026-04-23
feat(pure-functions): accept any top-level binding and trust consumer-side hints

JSerFeng:
## What

Relax the matching scope of `module.parser.javascript.pureFunctions`: in addition to the previously supported "locally defined exported functions", any top-level binding is now accepted — including `function`/`class`/`var` declarations, `import` specifiers, and `export { foo as bar }` aliases. Also adjust call-site resolution: when a name comes from the user's `pureFunctions` config, skip the deferred check against the source module and treat it as Direct.

Related: #13728

In addition:

- Update the existing `user-mark-pure-functions-undefined` test to drop `localOnly` / `importedOnly`, which are now valid.
- Add `user-mark-pure-functions-imported` to verify marking imports on the consumer side.
- Add `user-mark-pure-functions-local` to verify that non-exported local functions can be marked.
- Update the English and Chinese docs in sync (config/module, config/experiments, guide/optimization/tree-shaking).

## Why

The previous implementation only honored `pureFunctions` for functions exported by the module itself, which is too narrow:

- The most common real-world use case is marking calls into third-party libraries (e.g. `import { cva } from 'cva'`). The old behavior required configuring `pureFunctions` on the third-party source file, which is awkward when you can't (or don't want to) match by file path. In practice it degraded to the same expressiveness as a `/* #__NO_SIDE_EFFECTS__ */` annotation.
- Non-exported helpers inside a module (`function privateHelper() {...}`) couldn't be marked at all.
- webpack has no equivalent option, but the existing rspack restriction made this config noticeably less useful than users expect, and weaker than the intuition webpack users carry over from similar hint options.

After the change, users can either mark exports on the source module or assert "I claim this import call is pure" on the consumer side.

## How

Core change lives in `crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs`:

- Introduce `collect_pure_function_acceptable_names` that walks the program in two passes: pass 1 collects every top-level binding name (fn/class/var decls + import specifiers + the local name introduced by export decls); pass 2 fills in `export { x as y }` aliases and the `default` keyword (only when the default export is a fn/arrow). `collect_defined_configured_side_effects_free` uses this set for matching, leaving the "not found" warning path unchanged.
- Add a user-configured bypass in `resolve_explicit_side_effects_free_callee`: when the name appears in `parser.javascript_options.side_effects_free`, skip `try_extract_deferred_check` and resolve the binding directly to return Direct. This is what makes a consumer-side `pureFunctions: ['cva']` actually treat `cva()` as pure, instead of silently deferring to the source module's marker.
- Remove the now-unused helpers (`collect_exported_side_effects_free_refs`, `collect_top_level_function_like_local_refs`, `function_like_var_decl_names`, `insert_module_export_name`), eliminating a duplicate program walk.

Compatibility: the original "configure `pureFunctions` on the source module" workflow (including `export { impl as Public }` aliases and `default` exports) is preserved by the two-pass collection plus alias handling. `user-mark-pure-functions` / `user-mark-pure-functions-import-alias` keep their existing assertions.

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
@@ -13,9 +13,10 @@ use swc_core::{
   },
   ecma::{
     ast::{
-      ArrowExpr, BlockStmt, BlockStmtOrExpr, Class, ClassMember, Decl, Expr, ExportSpecifier,
-      ExprOrSpread, Function, ImportSpecifier, ModuleDecl, ModuleExportName, ModuleItem, Pat,
-      Program, PropName, Stmt, VarDecl, VarDeclKind, VarDeclOrExpr,
+      ArrayPat, ArrowExpr, AssignPat, AssignPatProp, BlockStmt, BlockStmtOrExpr, Class,
+      ClassMember, Decl, ExportSpecifier, Expr, ExprOrSpread, Function, ImportSpecifier,
+      KeyValuePatProp, ModuleDecl, ModuleExportName, ModuleItem, ObjectPat, ObjectPatProp, Pat,
+      Program, PropName, RestPat, Stmt, VarDecl, VarDeclKind, VarDeclOrExpr,
     },
     utils::{ExprCtx, ExprExt},
     visit::{Visit, VisitWith},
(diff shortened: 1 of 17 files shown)


JSerFeng:
@codex review

ahabhgk (approved):

PR #13817, opened 2026-04-23
refactor: remove prefetch functions for exports info

ahabhgk:
## Summary

prefetch functions is not necessary anymore, prefetch functions is used to avoid calling `&ModuleGraph` and `&mut ModuleGraph` at same time to enable parallel, since we move the exports info store from `module_graph` to `exports_info_artifact`, the `&mg / &mut mg` problem doesn't exists anymore, so the prefetch functions is also not needed anymore

## Checklist


ahabhgk:
@codex review

LingyuCoder (approved):

PR #13821, opened 2026-04-24
perf: avoid cloning ModuleInfo

LingyuCoder:
## Summary

Remove `Clone` from `ModuleInfo` and adjust concatenated module analysis to pass borrowed `ConcatenatedModuleInfo` values instead of cloning whole `ModuleInfo` entries.

The analysis pass now builds a list of concatenated module ids inline, skips external entries before spawning analysis work, and wraps the updated concatenated info back into `ModuleInfo` after analysis.

For `rspack_plugin_esm_library`, codegen now receives a lightweight snapshot of the module info map instead of cloning the full `IdentifierIndexMap<ModuleInfo>`.

## Validation

- `cargo check -p rspack_plugin_esm_library --lib`
- `cargo check -p rspack_core -p rspack_plugin_esm_library --lib`
- `cargo clippy -p rspack_core -p rspack_plugin_esm_library --lib -- -D warnings`
- `cargo fmt --all --check`


LingyuCoder:
@copilot resolve the merge conflicts in this pull request

Issue #13799, opened 2026-04-22
[Bug]: Failed to run the build with "pnpm build:start" - works fine in dev mode


abhidatta0:
### System Info

  System:
    OS: macOS 26.0
    CPU: (8) arm64 Apple M2
    Memory: 189.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.11.0 - /Users/abhirup/.nvm/versions/node/v24.11.0/bin/node
    npm: 11.6.1 - /Users/abhirup/.nvm/versions/node/v24.11.0/bin/npm
    pnpm: 10.24.0 - /Users/abhirup/.nvm/versions/node/v24.11.0/bin/pnpm
    bun: 1.3.1 - /Users/abhirup/.bun/bin/bun
    Watchman: 2025.09.15.00 - /opt/homebrew/bin/watchman
  Browsers:
    Chrome: 147.0.7727.102
    Firefox: 149.0.2
    Safari: 26.0
  npmPackages:
    @rspack/cli: ~1.2.0 => 1.2.8 
    @rspack/core: ~1.2.0 => 1.2.8 
    @rspack/plugin-react-refresh: ~1.0.1 => 1.0.3 

### Details

https://github.com/user-attachments/assets/792e19b7-68ab-4579-ab69-cbc5d3a5f441

As demonstrated in the video:  pnpm start (the dev mode ) works but when trying to run the prod build it fails.

**rspack.config.ts**

import * as path from "node:path";
import { defineConfig } from "@rspack/cli";
import { rspack } from "@rspack/core";
import * as RefreshPlugin from "@rspack/plugin-react-refresh";
import { ModuleFederationPlugin } from "@module-federation/enhanced/rspack";

import { mfConfig } from "./module-federation.config";

const isDev = process.env.NODE_ENV === "development";

// Target browsers, see: https://github.com/browserslist/browserslist
const targets = ["chrome >= 87", "edge >= 88", "firefox >= 78", "safari >= 14"];

export default defineConfig({
  context: __dirname,
  entry: {
    main: "./src/index.ts",
  },
  resolve: {
    extensions: ["...", ".ts", ".tsx", ".jsx"],
  },

  devServer: {
    port: 8081,
    historyApiFallback: true,
    watchFiles: [path.resolve(__dirname, "src")],
  },
  output: {
    // You need to set a unique value that is not equal to other applications
    uniqueName: "remote",
[...]

PR #13834, opened 2026-04-26
fix(cli): hint preview for missing default entry

Sylar-W:
## Summary

- Warn when `rspack serve` runs without a config and the default `./src` entry is missing.
- Point users who are trying to serve build output toward `rspack preview [dir]`.
- Add a regression test for the missing default entry case.

## Related links

Refs #13799

## Test Plan

- `NAPI_RS_NATIVE_LIBRARY_PATH="$PWD/node_modules/.pnpm/@rspack+binding-darwin-arm64@2.0.0/node_modules/@rspack/binding-darwin-arm64/rspack.darwin-arm64.node" pnpm --filter @rspack/cli test -- tests/serve/missing-default-entry/index.test.ts`
- `pnpm exec prettier --check packages/rspack-cli/src/cli.ts packages/rspack-cli/tests/serve/missing-default-entry/index.test.ts`
- `pnpm exec rslint packages/rspack-cli/src/cli.ts`
- `git diff --check`

## Checklist

### DIFF
--- a/packages/rspack-cli/src/cli.ts
+++ b/packages/rspack-cli/src/cli.ts
@@ -1,3 +1,4 @@
+import fs from 'node:fs';
 import path from 'node:path';
 import util from 'node:util';
 import type {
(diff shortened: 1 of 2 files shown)


chenjiahan (commented):
I'm a bit concerned that suggesting `rspack preview` here may be misleading.

`rspack serve` is intended for development, while `rspack preview` is for serving already-built output. When the default entry cannot be resolved, the right next step is usually to create/configure the entry, rather than switching to `preview`.

The existing `Module not found: Can't resolve ...` error already points to the actual issue. I think we should avoid adding the `preview` suggestion on the generic `rspack serve` error path.

Image

chenjiahan:
Closed. Thank you for contribution!

PR #13851, opened 2026-04-27
perf: reduce javascript parser heap allocations

hardfist:
## Summary

Reduce heap allocation overhead in the JavaScript dependency parser by avoiding eager allocations for result vectors that are usually empty, and by using stack-backed `SmallVec` storage for short-lived parser scope parameter lists.

## Details

- Initialize parser `errors`, `warning_diagnostics`, and async `blocks` with empty vectors so they allocate only when populated.
- Keep common IIFE parameter tracking on stack for functions with up to four parameters.
- Avoid a heap allocation for the optional single class-expression scope parameter.

## Validation

- `cargo fmt --all --check`

Note: local CodSpeed `memory` mode could not run because the memtrack process requires sudo in this environment.


hardfist:
@copilot resolve the merge conflicts in this pull request

PR #13854, opened 2026-04-27
perf(split-chunks): lazily precompute sorted modules for compare

LingyuCoder:
## Summary

Avoid repeatedly sorting module identifiers inside `compare_entries`, while only paying the sorting cost when the modules tie-break is actually reached.

This PR:
- makes `add_module` only update the module set during module-group preparation
- prepares `added` and the compare snapshot from the complete module set before min-size filtering and best-group selection
- reserves `removed` capacity based on the module count to avoid repeated growth during removals
- represents the compare snapshot as an enum state: unsorted before the first modules tie-break, sorted after it has been prepared
- fills the sorted state lazily only when `compare_entries` needs to compare modules
- keeps `remove_module` from maintaining the compare snapshot, since it is only used as a deterministic tie-break key
- changes `find_best_module_group` to compare entries by index so both candidate groups can be lazily prepared during comparison

## Checklist

### DIFF
--- a/crates/rspack_plugin_split_chunks/src/module_group.rs
+++ b/crates/rspack_plugin_split_chunks/src/module_group.rs
@@ -1,7 +1,6 @@
 use std::{cmp::Ordering, fmt};
 
 use derive_more::Debug;
-use itertools::Itertools;
 use rspack_collections::IdentifierSet;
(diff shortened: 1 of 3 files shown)


JSerFeng:
@codex review

JSerFeng (approved):

PR #13859, opened 2026-04-28
perf(split-chunks): batch split chunk runtime updates

LingyuCoder:
## Summary

- batch the data written to the new split chunk while iterating original chunks
- collect runtime names as `Ustr` values and build the new chunk `RuntimeSpec` once at the end
- avoid repeated runtime key updates during split chunk creation

## Related links

Stacked on #13854.

## Checklist


JSerFeng:
@codex review

JSerFeng (approved):

PR #13872, opened 2026-04-29
feat(rslib): support emit isolated declarations

Timeless0911:
## Summary

This PR adds Rslib support for emitting isolated declaration assets from modules transformed by `builtin:swc-loader`, so declaration output can be generated through the compilation asset pipeline and restored after persistent cache recovery.

- Adds `emitDts` options to `RslibPlugin` and emits `.d.ts` assets relative to `rootDir` and `declarationDir`.
- Uses SWC FastDts during loader transform, stores declaration output in module build info, and renders isolated declaration diagnostics through rspack errors.

## Checklist

### DIFF
--- a/crates/rspack_javascript_compiler/src/compiler/transform.rs
+++ b/crates/rspack_javascript_compiler/src/compiler/transform.rs
@@ -11,7 +11,7 @@ use anyhow::{Context, bail};
 use base64::prelude::*;
 use indoc::formatdoc;
 use rspack_error::Result;
-use rspack_util::{SpanExt, source_map::SourceMapKind, swc::minify_file_comments};
+use rspack_util::{source_map::SourceMapKind, swc::minify_file_comments};
 use swc_config::{is_module::IsModule, merge::Merge};
 pub use swc_core::base::config::Options as SwcOptions;
 use swc_core::{
@@ -29,6 +29,10 @@ use swc_core::{
   },
   ecma::{
     ast::{EsVersion, Pass, Program},
+    codegen::{
+      self, Emitter, Node,
+      text_writer::{self, WriteJs},
+    },
     parser::{
       Syntax, TsSyntax, parse_file_as_commonjs, parse_file_as_module, parse_file_as_program,
       parse_file_as_script,
@@ -41,7 +45,7 @@ use swc_typescript::fast_dts::FastDts;
 use url::Url;
 
 use super::{
-  IsolatedDtsDiagnostic, IsolatedDtsTransformOutput, JavaScriptCompiler, TransformOutput,
+  IsolatedDtsTransformOutput, JavaScriptCompiler, TransformOutput,
   stringify::{PrintOptions, SourceMapConfig},
 };
 
@@ -77,7 +81,6 @@ impl JavaScriptCompiler {
     program: &Program,
     filename: Arc<FileName>,
     unresolved_mark: Mark,
-    source_len: u32,
     target: EsVersion,
     comments: &SingleThreadedComments,
   ) -> Result<IsolatedDtsTransformOutput> {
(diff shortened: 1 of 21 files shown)


crates/rspack_javascript_compiler/src/error.rs
```diff
@@ -63,6 +69,46 @@ pub fn ecma_parse_error_deduped_to_rspack_error(
   )
 }
 
+pub fn render_pretty_span_diagnostic(
```
CPunisher: Why create this new diagnostic emitter? Can't we just reuse the existing one?

crates/rspack_loader_swc/src/isolated_dts.rs
```diff
@@ -0,0 +1,106 @@
...
+  build_info.extras.insert(
+    "rspack-swc-isolated-dts-emit".to_string(),
+    Value::Object(Map::from_iter([
+      ("resource_path".to_string(), Value::String(resource_path)),
+      ("code".to_string(), Value::String(code)),
+    ])),
+  );
+}
+
+pub(crate) fn handle_isolated_dts_diagnostics(
```
CPunisher: Another complex diagnostic emitter

crates/rspack_loader_swc/src/isolated_dts.rs
```diff
@@ -0,0 +1,106 @@
...
+use serde_json::{Map, Value};
+
+use crate::SWC_LOADER_IDENTIFIER;
+
+pub(crate) fn set_build_info(
+  build_info: &mut rspack_core::BuildInfo,
+  resource_path: String,
+  code: String,
+) {
+  build_info.extras.insert(
```
CPunisher: I think it's better to store the dts result to a dedicated field rather than `build_info.extras`

crates/rspack_javascript_compiler/src/compiler/transform.rs
```diff
@@ -70,6 +71,58 @@ impl JavaScriptCompiler {
...
+        .into_iter()
+        .map(|issue| IsolatedDtsDiagnostic {
+          message: issue.message.to_string(),
+          start: issue.range.span.real_lo(),
+          end: issue.range.span.real_hi(),
+        })
+        .collect();
+
+      let code = self
+        .print(
```
CPunisher: We'd better emit the code by creating a new `Emitter` here because `self.print` is too heavy in this case (for example, sourcemap generation).

crates/rspack_plugin_rslib/src/plugin.rs
```diff
@@ -31,11 +35,118 @@ use crate::{
...
+fn get_build_info(build_info: &BuildInfo) -> Option<SwcEmitDtsBuildInfo> {
+  let value = build_info.extras.get("rspack-swc-isolated-dts-emit")?;
+  let value = value.as_object()?;
+  Some(SwcEmitDtsBuildInfo {
+    resource_path: value.get("resource_path")?.as_str()?.to_string(),
+    code: value.get("code")?.as_str()?.to_string(),
+  })
+}
+
+fn emit_isolated_dts_asset(
```
CPunisher, line 51: I suspect the AI generated this kind of over complex path resolution. We'd better refine it.

crates/rspack_plugin_rslib/src/plugin.rs
```diff
@@ -31,11 +35,118 @@ use crate::{
...
+    path.to_path_buf()
+  } else {
+    normalize_joined_path(base.as_std_path().join(path.as_std_path()))
+      .to_string_lossy()
+      .to_string()
+      .into()
+  }
+}
+
+fn normalize_joined_path(path: PathBuf) -> PathBuf {
```
CPunisher: This is definitely unnecessary, and I never see it across the whole repo.

crates/rspack_loader_swc/src/lib.rs
```diff
@@ -125,8 +126,20 @@ impl SwcLoader {
...
+    let should_emit_isolated_dts = swc_options
+      .config
+      .jsc
+      .experimental
+      .emit_isolated_dts
+      .into_bool();
+    let isolated_dts_target = swc_options.config.jsc.target.unwrap_or(EsVersion::EsNext);
+    let source_len = source.len() as u32;
+    let dts_filename = filename.clone();
+    let dts_comments = comments.clone();
```
CPunisher: Can we move these declarations to the place where they are used?

CPunisher (approved):

chenjiahan (approved):

PR #13873, opened 2026-04-29
perf(split-chunks): optimize module chunk lookup

LingyuCoder:
## Summary

- store module chunks in the same order as the pre-sorted module list to avoid hash-map lookups during split chunks processing
- sort modules by precomputed identifier hash before falling back to the identifier for deterministic ordering
- simplify chunk combination grouping to iterate sorted vectors by size

## Validation

- cargo fmt --all
- git diff --check
- cargo check -p rspack_plugin_split_chunks


JSerFeng (approved):

JSerFeng:
@codex review

PR #13878, opened 2026-04-29
fix(rsc): scope server-entry CSS collection by entry

SyMind:
## Summary

- Fix RSC server-entry CSS ownership so each `use server-entry` only receives CSS emitted from its own async CSS block.
- Refactor component traversal to track the current server-entry owner explicitly, allowing shared server components to collect CSS per server entry while preventing cycles.
- Ensure nested `use server-entry` subtrees are isolated from their parent server-entry CSS/client module collection.
- Add regression coverage for shared server component CSS and nested server-entry CSS in the `server-css` config case.

## Checklist


SyMind:
@codex review

ahabhgk (approved):

PR #13955, opened 2026-05-08
perf: optimize build chunk graph traversal

LingyuCoder:
## Summary

- Use per-chunk module masks for hot module-in-chunk checks during chunk graph traversal.
- Replace BigUint-backed available-module sets with a fixed-width local bitset to reduce allocation and arithmetic overhead.
- Store single dependency connection lists inline instead of allocating an Arc<Vec<_>> for every list.
- Cache computed connection active states by runtime and connection list so repeated checks can reuse the result.

## Why

The build_chunk_graph profile shows time in repeated module membership checks, available-module set merging, dependency connection list allocation, and active-state evaluation. These changes keep the same traversal semantics while reducing the amount of generic big-integer work, heap allocation, and repeated connection-state computation on the hot path.

## Validation

- cargo check -p rspack_core


LingyuCoder:
Closing this PR because CI revealed semantic regressions and CodSpeed did not show the expected build_chunk_graph improvement. The optimization direction needs to be reworked from a clean baseline.

PR #13965, opened 2026-05-09
test: run type tests with rslint type-check

chenjiahan:
## Summary

This PR runs type tests through Rslint's `--type-check` path instead of the previous `tsc --noEmit` script, so CI can cover package and type-test tsconfigs in the faster lint job. It also removes the old `test:type` script and the `test:ci` dependency on that tsc-backed path.

## Related links

https://rslint.rs/guide/type-checking

## Checklist


chenjiahan:
Closing this PR because rslint --type-check needs built package declarations, but the lint CI job does not build packages. Adding a build step there would slow lint CI down

PR #13972, opened 2026-05-09
feat: track static esm namespace keys

JSerFeng:
## Summary

- add an ESM import parser tag for static primitive const property keys
- allow computed member chain evaluation only when the root is an ESM namespace import
- add a tree-shaking config case for `ns[key]` and `key in ns` preserving precise referenced exports

## Related links

N/A

## Checklist

## Testing

- `cargo fmt --all --check`
- `PATH=/Users/bytedance/.nvm/versions/node/v22.21.1/bin:$PATH pnpm --dir tests/rspack-test run test -t "esm-namespace-static-key"`


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
close as we need to implement this after new semantic analysis support

PR #13981, opened 2026-05-11
fix: align sync module rule resource matching

LingyuCoder:
## Summary

- make the sync module rule matcher check `resource` even when `test` is present
- add a config case covering rules that require both `test` and `resource` to match

## Root cause

The sync fast path used `else if` for `resource`, so a rule with both `test` and `resource` skipped the `resource` condition whenever `test` matched. The async matcher treats those conditions as conjunctive, so the two paths could select different rules.

## Checks

- `PATH=/Users/bytedance/.nvm/versions/node/v22.18.0/bin:$PATH pnpm run build:binding:dev`
- `PATH=/Users/bytedance/.nvm/versions/node/v22.18.0/bin:$PATH pnpm run test -t "configCases/rule-set/test-and-resource"`
- `cargo fmt --all --check`

### DIFF
--- a/crates/rspack_core/src/utils/module_rules.rs
+++ b/crates/rspack_core/src/utils/module_rules.rs
@@ -188,8 +188,18 @@ fn module_rule_matcher_sync<'a>(
     .unwrap_or_else(|| Utf8Path::new(""))
     .as_str();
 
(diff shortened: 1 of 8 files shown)


SyMind:
@codex review

SyMind (approved):

PR #14012, opened 2026-05-12
perf(deps): unify duplicate Rust dependencies to reduce binary size

intellild:
## Summary

- update the lockfile so the `url -> idna -> idna_adapter` dependency path uses `idna_adapter` 1.2.1
- remove the duplicate ICU 1.5.x dependency graph from `Cargo.lock`
- keep ICU-related crates unified on the existing 2.1.x graph used by SWC regexp support
- update the workspace `wasmparser` dependency from 0.222.0 to 0.235.0
- remove the duplicate `wasmparser` 0.222.0 entry from `Cargo.lock`
- remove the now-unneeded `icu_normalizer` and `wasmparser` duplicate-dependency skips from `deny.toml`

## Root Cause

`idna_adapter` 1.2.0 still resolved to `icu_normalizer`/`icu_properties` 1.5.x, while `swc_ecma_compat_regexp` already used `icu_properties` 2.1.x. Updating `idna_adapter` allowed Cargo to resolve both paths through the same ICU 2.1.x family.

`rspack_plugin_wasm` still used workspace `wasmparser` 0.222.0, while the Wasmtime dependency graph already used 0.235.0. Updating the workspace dependency lets Cargo use the same `wasmparser` version for both paths, so the cargo-deny skip is no longer needed.

## Validation

- `cargo tree -i icu_provider` shows only `icu_provider v2.1.1`
- `cargo tree -d | grep -E "^icu_| icu_|^idna_adapter| idna_adapter"`
- `cargo check -p rspack_javascript_compiler --locked`
- `cargo check -p rspack_plugin_wasm --locked`
- `cargo tree -d --locked --all-features --target all` no longer reports duplicate `wasmparser`

Not run: `cargo deny check bans`, because `cargo-deny` is not installed in this environment.

---
_by OpenAI Codex_


intellild:
@codex review

hardfist (approved):

PR #14013, opened 2026-05-12
perf(regex): prefer native regex fallback

LingyuCoder:
## Summary

- try Rust native regex via `RegexBuilder` before falling back to `regress`
- keep the existing `EndWith` fast path ahead of the native-regex attempt
- add a `.*` match-all fast path in `Algo::new` so the default wrapped context regexp does not need backend regex compilation
- align flag handling with Turbopack for `d/g/i/m/s/u`, while keeping sticky `y` on `regress` for correctness outside the `.*` fast path
- normalize escaped JS regex slashes (`\/`) before constructing Rust regex
- port relevant Turbopack behavior tests for simple native regex, negative lookahead fallback, invalid regex, and the `.*` fast path

## Validation

- `cargo test -p rspack_regex`
- `cargo check -p rspack --lib`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,7 +79,7 @@ proc-macro2         = { version = "1.0.106", default-features = false }
 prost               = { version = "0.13", default-features = false }
 quote               = { version = "1.0.45", default-features = false }
 rayon               = { version = "1.12.0", default-features = false }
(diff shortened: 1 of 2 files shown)


LingyuCoder:
@copilot remove "unicode-case" feature of regex crate

hardfist (approved):

PR #14014, opened 2026-05-12
perf: cache reserved name atom set

LingyuCoder:
## Summary

- Add a shared `LazyLock<FxHashSet<Atom>>` for reserved JavaScript names in `rspack_core::reserved_names`.
- Replace repeated `RESERVED_NAMES.iter().map(...).collect()` calls with clones of the cached atom set.
- Add a focused unit test to keep the cached atom set aligned with the source reserved-name list.

## Impact

This avoids rebuilding the reserved-name atom set at multiple hot call sites while preserving the existing mutable `HashSet` behavior at each use site.

## Validation

- `cargo test -p rspack_core reserved_names_atom_set_matches_reserved_names`
- `cargo check -p rspack_plugin_javascript -p rspack_plugin_esm_library`
- `cargo fmt --all --check`
- `git diff --check`


LingyuCoder:
@codex review

JSerFeng (approved):

PR #14018, opened 2026-05-12
fix(rsc): skip client entry mismatch without injections

SyMind:
## Summary

- Skip RSC client entry mismatch diagnostics when an entry has no client compiler injections.
- Avoid mutating missing client entries by replacing the final unwrap with a guarded lookup.
- This supports multi-entry RSC setups where an RSC-only service entry has no client components or CSS and does not need a matching client compiler entry.

## Checklist


SyMind:
@codex review

ahabhgk (approved):

PR #14037, opened 2026-05-14
fix(rslib): emit type-only isolated dts dependencies

Timeless0911:
## Summary

This PR fixes RslibPlugin isolated dts output for type-only dependencies. When an entry declaration re-exports types from a TS file that is not part of the module graph, RslibPlugin now follows generated declaration references, resolves them with the bundler resolver, generates isolated declarations for the referenced TS sources, and emits the corresponding `.d.ts` assets.

## Related links

https://github.com/web-infra-dev/rspack/pull/13872

## Checklist

### DIFF
--- a/crates/rspack_plugin_rslib/src/isolated_dts.rs
+++ b/crates/rspack_plugin_rslib/src/isolated_dts.rs
@@ -12,11 +12,10 @@ use rspack_javascript_compiler::JavaScriptCompiler;
 use rspack_paths::{Utf8Path, Utf8PathBuf};
 use rspack_util::node_path::NodePath;
 use swc_core::{
-  common::{FileName, SourceMap, comments::SingleThreadedComments},
+  common::FileName,
   ecma::{
-    ast::{EsVersion, ExportAll, ImportDecl, NamedExport, TsExternalModuleRef, TsImportType},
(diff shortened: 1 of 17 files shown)


crates/rspack_plugin_rslib/src/isolated_dts.rs
```diff
@@ -0,0 +1,227 @@
...
+      )?;
+      handle_isolated_dts_diagnostics(&resource_path, dts_output.diagnostics)?;
+
+      let dts = IsolatedDts {
+        resource_path: resource_path.as_str().to_string(),
+        code: dts_output.code,
+        references: dts_output.references,
+      };
+      queue.push_back(dts.clone());
+      outputs.push(dts);
```
CPunisher: For `queue`, only `(dts.resource_path, dts.references)` is needed. The code cloning can be avoid.

For `outputs`, only `(dts.resource_path, dts.code)` is needed. The references cloning can be avoid.

So it could be:
```rs
queue.push_back((resource_path.as_str().to_string(), dts_output.references));
outputs.push((resource_path.as_str().to_string(), dts_output.code));
```

CPunisher (dismissed):

Timeless0911:

CPunisher (approved):

PR #14041, opened 2026-05-14
perf: remove unnecessary async_trait usage

hardfist:
## Summary

- replace compilation pass `async_trait` dispatch with `impl Future` static dispatch and direct pass sequencing
- remove redundant `#[async_trait]` usage from static async traits and non-async runtime helper traits
- drop the unused `async-trait` dependency from `rspack_parallel`

## Why

`async-trait` is useful when an async trait must be used through `dyn` dispatch, but it works by erasing each async method into a boxed future. That boxing can allocate on the heap for every call. The traits changed here are statically dispatched, or do not have async methods, so using native `impl Future` / plain trait methods avoids those unnecessary allocations while keeping `async_trait` in places that still need object-safe async trait calls.

Related Rust blog: [Async fn and return-position impl Trait in traits](https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits/).

## Validation

- `cargo fmt --all`
- `cargo check -p rspack_core -p rspack_parallel -p rspack_plugin_lazy_compilation -p rspack_binding_api`
- `cargo check -p rspack_benchmark --benches`
- pre-commit `lint-staged` ran `rustfmt` and `pnpm exec taplo format` during commit

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4257,7 +4257,6 @@ dependencies = [
 name = "rspack_parallel"
 version = "0.100.3"
 dependencies = [
- "async-trait",
  "rayon",
  "rspack_tasks",
  "tokio",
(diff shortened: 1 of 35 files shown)


hardfist:

hardfist:
@codspeedbot fix this regression again

adriencaccia:
Hey @hardfist, sorry for this error. We are fixing it right now!

hardfist:
@codspeedbot fix this regression again

stormslowly (approved):

PR #14045, opened 2026-05-15
feat: optimize generated getter for const export 

ahabhgk:
## Summary

- Optimize ESM `export const` codegen by emitting value bindings in `__webpack_require__.d` instead of generated getter functions when circular-module information is available and the current module is not circular.
- Keep getter bindings for non-const exports and circular modules, and place circular modules' `__webpack_require__.d` calls at the top while non-circular modules emit them at the end of the module body.
- Update the `__webpack_require__.d` runtime binding format to support flat definitions like `'name', 0, value` for value exports and `'name', () => value` for getter exports.
- Enable `CircularModulesInfoPlugin` only for production/default-production mode, and store circular-module info as `Option<IdentifierSet>` so an empty set still means the plugin ran and the optimization can apply.
- Update related runtime interop handling and snapshots for the new export-definition format.

## Related links

- Inspired by https://github.com/vercel/next.js/pull/82760

## Checklist

### DIFF
--- a/crates/rspack_binding_api/src/plugins/interceptor.rs
+++ b/crates/rspack_binding_api/src/plugins/interceptor.rs
@@ -1296,7 +1296,7 @@ impl CompilationOptimizeModules for CompilationOptimizeModulesTap {
   async fn run(
     &self,
     _compilation: &Compilation,
-    _circular_modules: &mut IdentifierSet,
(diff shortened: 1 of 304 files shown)


ahabhgk:
@codex review

hardfist (approved):

PR #14046, opened 2026-05-15
fix(hmr): hash only emitted bytes in code generation result

### LINKED ISSUE #11635: [Bug]: Child HMR not working when parent component has HMR
### Version

```sh
System:
    OS: Linux 6.6 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (6) x64 Intel(R) Core(TM) i5-9400F CPU @ 2.90GHz
    Memory: 13.55 GB / 15.59 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  npmPackages:
    @rsbuild/core: ^1.5.4 => 1.5.6 
    @rsbuild/plugin-vue2: ^1.0.4 => 1.0.4
```

### Details

1. ​​trigger the parent component's HMR.​
2. changing the child component, HMR is not working.

### Reproduce link

[yuhengshen/rsbuild-test-hmr](https://github.com/yuhengshen/rsbuild-test-hmr)

### Reproduce Steps

-

stormslowly:
## Summary

 Rspack routed through `Source::hash`, which for `SourceMapSource` also folds in the wrapped `source_map` — including its `sourcesContent` snapshot of the loader-input file.

For files split into multiple sub-modules (Vue SFCs, anything emitting `?type=...` virtual modules), this means changing one block shifts the codegen hash of every sibling sub-module even when their emitted bytes are unchanged. The HMR plugin then ships byte-identical modules in hot-update chunks, and anything keyed off "this module changed" misfires.

| | rspack (before) | rspack (after) |
|---|---|---|
| Hash inputs | full `Source::hash` (incl. `source_map.sourcesContent`) + `chunk_init_fragments` + `runtime_requirements` | `source.buffer()` + `chunk_init_fragments` + `runtime_requirements` |
| Stable across same-bytes rebuild? | yes | no | yes |

## Side effect: fixes #11635

Vue 2 `<script setup>` parent/child HMR breakage is downstream of this divergence — editing a Vue template made the byte-identical script-setup sub-module ride into hot-update.js, which bubbled into `api.reload` and broke `record.Ctor` alignment. With this patch the script-setup stays out of the update and the HMR runtime takes the `api.rerender` path instead.

## Test plan

- [ ] CI green
- [ ] no regression in `tests/rspack-test/hotCases/`

Local repro: `@rsbuild/core@2.0.6` + `@rsbuild/plugin-vue2@1.1.1` Vue 2.7 minimal project, Playwright-driven edit sequence `[app, app, child, child, app, child]`. Before patch: final child template stays on the previous value. After patch: updates correctly, matching webpack 5.106 + `vue-loader@15.11.1`.

### DIFF
--- a/crates/rspack_core/src/artifacts/code_generation_results.rs
+++ b/crates/rspack_core/src/artifacts/code_generation_results.rs
@@ -167,7 +167,16 @@ impl CodeGenerationResult {
     let mut hasher = RspackHash::with_salt(hash_function, hash_salt);
     for (source_type, source) in self.inner.as_ref() {
       source_type.hash(&mut hasher);
-      source.hash(&mut hasher);
+      // Hash only the emitted bytes, matching webpack's
+      // `CodeGenerationResults.getHash` which feeds `source.updateHash(hash)`
+      // and never mixes source-map metadata into the digest.
+      // `Source::hash` for `SourceMapSource` also folds in `source_map`
+      // (including the `sourcesContent` snapshot of the loader-input file).
+      // For multi-block sources like Vue SFCs that means changing one block
+      // shifts the codegen hash of every sibling sub-module even when their
+      // emitted bytes are identical, which causes HMR to ship unchanged
+      // modules in hot-update chunks and breaks downstream HMR flows.
+      source.buffer().hash(&mut hasher);
     }
     self.chunk_init_fragments.hash(&mut hasher);
     self.runtime_requirements.hash(&mut hasher);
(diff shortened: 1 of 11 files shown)


stormslowly:
Closing — the framing of this PR ("align with webpack's hash inputs") is factually wrong. webpack-sources `SourceMapSource.updateHash` also hashes the source map buffer, so the divergence I cited doesn't exist. Re-investigating the real reason why webpack doesn't include byte-identical sub-modules in hot-update chunks before re-filing.

PR #14048, opened 2026-05-15
perf: optimize mangle exports plugin

LingyuCoder:
### What changed

- Collapse the mangleability pre-scan into one export-info pass that also derives whether non-provided exports must be avoided.
- Carry export names directly alongside mangleable export ids so deterministic id assignment and sorting no longer build or query a per-exports-info name map.
- Replace short-name regex checks with direct ASCII byte predicates for generated mangle names.

### Why this should improve performance

`MangleExportsPlugin` runs over exports info, which can be much higher cardinality than chunks, chunk groups, or runtimes. The previous path repeatedly scanned export collections and allocated a temporary name lookup map for every mangled exports info, then performed map lookups inside sorting and deterministic id assignment.

The new path keeps the same observable ordering and assignment behavior while reducing repeated traversal, per-export lookup work, and temporary hash-map allocation on the export-info hot path. The expected benefit is lower CPU time and fewer transient allocations when compilations contain many modules or nested exports with mangleable names.

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/mangle_exports_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/mangle_exports_plugin.rs
@@ -1,8 +1,7 @@
-use std::{borrow::Cow, sync::LazyLock};
+use std::borrow::Cow;
 
 use itertools::Itertools;
 use rayon::prelude::*;
-use regex::Regex;
 use rspack_core::{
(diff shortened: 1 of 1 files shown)


LingyuCoder:
@codex review

hardfist:

hardfist (approved):

PR #14049, opened 2026-05-15
fix(stats): preserve sub-millisecond precision in logger time entries

stormslowly:
## Why

`stats.toJson({ logging: 'verbose' })` reported every per-phase timing as an integer ms because the napi serializer truncated the nanosecond remainder via `u64` integer division. Short stages like `Compilation.seal` / `optimize tree` / `optimize modules` were always `0 ms` even when they actually took tens of µs.

## What

Compute elapsed ms as `f64` and format with `ryu_js::Buffer`, so messages become e.g. `seal: 0.019542 ms`. The webpack-compat parsing path and existing `[\d.]+ ms` regex consumers already support decimals — nothing else needs to change.

### DIFF
--- a/crates/rspack_binding_api/src/stats.rs
+++ b/crates/rspack_binding_api/src/stats.rs
@@ -16,7 +16,7 @@ use rspack_napi::napi::{
   Either,
   bindgen_prelude::{Buffer, Result, SharedReference, ToNapiValue},
 };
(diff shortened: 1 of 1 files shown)


stormslowly:
@codex

hardfist (approved):

PR #14068, opened 2026-05-18
fix(rsc): skip manifest runtime for worker entries

SyMind:
## Summary

- Skip generating the RSC manifest runtime for worker entries, since they can appear as named entrypoints in the chunk graph but are not top-level RSC compilation entries.
- Fix the `RSC entry state not found` failure when an RSC client component creates a module worker.
- Add a regression config case covering an RSC server/client build with a client-created worker entry.

## Checklist


ahabhgk (approved):

SyMind:
@codex review

PR #14105, opened 2026-05-21
ci(release-canary): add targets and dry_run workflow inputs

stormslowly:
## How to use
Just ask your agent like
```txt
publish current commit for all macos target by release canary workflow 
// or
publish current commit for all linux target with dry-run mode by release canary workflow
```

## Why

Canary release currently builds bindings for all 10 targets and always publishes to npm. That wastes CI capacity when only a subset is needed (testing a single platform fix, debugging the workflow, etc.) and offers no way to rehearse the pipeline end-to-end without actually shipping packages.

## What

Two new `workflow_dispatch` inputs on `release-canary.yml`:

- `targets` (string, default `all`) — comma-separated list of target triples to build, e.g. `x86_64-apple-darwin,aarch64-apple-darwin`. A new `prepare` job filters the target list and emits the matrix via `fromJSON`, so unselected targets aren't scheduled at all (no skipped placeholder jobs).
- `dry_run` (boolean, default `false`) — threads `--dry-run` into `./x publish snapshot --tag latest`, so `pnpm publish` simulates without pushing anything to npm.

A new **List binding sizes** step in the release job walks `artifacts/bindings-*/*.{node,wasm}` and writes a markdown table (binding name, size) to `$GITHUB_STEP_SUMMARY`, surfacing each artifact's size on the run page for every run — useful both for dry-run rehearsals and for spotting size regressions.

The scheduled (cron) run is unchanged: defaults resolve to `targets='all'`, `dry_run=false`.

Naming uses `dry_run` (snake_case) to match the convention in `release.yml` / `reusable-release-*.yml`.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -25,6 +25,16 @@ on:
           - ci
           - debug
         description: 'profiling means release with debug info for profiling, ci means release with debug info and faster build time'
+      targets:
+        required: false
+        type: string
+        default: 'all'
+        description: 'Comma-separated targets to build (e.g. x86_64-apple-darwin,aarch64-apple-darwin), or "all" to build every target'
+      dry_run:
+        required: false
+        type: boolean
+        default: false
+        description: 'Build only — skip the actual npm publish and report binding sizes'
 
 permissions:
   # To publish packages with provenance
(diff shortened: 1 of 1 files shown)


stormslowly:
Canary Release is also a formally published version of rspack.
So it's not acceptable to have a partition target release.

PR #14106, opened 2026-05-21
perf: optimize find graph roots

intellild:
## Summary

Optimize `find_graph_roots` by storing the temporary graph in dense `Vec`-backed node and cycle indexes instead of per-call `NodeUkey`/`CycleUkey` maps. This removes atomic key generation and turns internal node lookups into direct index access.

The dependency collection path now uses a push-style callback so callers do not need to allocate and return a dependency `Vec`. Dependencies are sorted and deduplicated once inside `find_graph_roots`, and the DFS stack now tracks dependency indexes instead of cloning each node's dependency list.

This targets the chunk root module hot path, where work scales with modules and outgoing graph edges. Expected benefits are fewer temporary allocations, less HashMap/HashSet churn, no dependency-list clones during traversal, and better locality while preserving stable root selection.

Checked with `cargo check -p rspack_core`, `cargo test -p rspack_core find_graph_roots`, `pnpm run build:binding:dev`, `pnpm run build:js`, and `pnpm run test:unit`.

## Checklist

---
_by OpenAI Codex_

### DIFF
--- a/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
+++ b/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
@@ -734,14 +734,12 @@ impl ChunkGraph {
     exports_info_artifact: &ExportsInfoArtifact,
   ) -> Vec<ModuleIdentifier> {
     let cgc = self.expect_chunk_graph_chunk(chunk);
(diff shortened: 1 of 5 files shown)


intellild:

intellild:

hardfist (approved):

PR #14124, opened 2026-05-22
perf(javascript): reduce parser allocations

hardfist:
## Summary
most module don't have errors and warnings, so eagerly allocation for errors and warnings is not necessary.
- Replace eager diagnostic-vector allocation in `JavaScriptParser::new` with empty vectors for parser warnings and errors.
- Keep dependency, block, presentational dependency, and parser plugin vector preallocation unchanged.

## Why

Valgrind DHAT analysis of the `build_module_graph` benchmark showed `JavaScriptParser::new` runs on a high-cardinality module parsing path. Warning and error vectors are usually empty on successful parses, so allocating them lazily avoids unnecessary per-module heap allocation.

## Expected impact

This should slightly reduce build-module-graph heap allocation on successful parser paths without changing parser behavior or emitted output.

Checked with `cargo check -p rspack_plugin_javascript`.


hardfist:

stormslowly (approved):

PR #14126, opened 2026-05-22
perf: cache SWC build input config

hardfist:
### What changed

Moved the SWC `build_as_input` flow into `rspack_javascript_compiler` for the normal non-plugin path and added a bounded cache for the options/config-derived build input configuration. The parse closure still runs for every transform, so parsed ASTs and source-specific work are not cached.

The implementation keeps SWC plugin and hidden jest cases on the upstream SWC path, and mirrors the module resolver/import rewrite/minifier/comment preservation behavior needed by the local path.

### Why this should improve performance

`builtin:swc-loader` repeatedly transforms many modules with the same loader options. Caching the stable configuration construction avoids repeating the merge/adjust/source-map normalization work for every module while preserving per-file parsing and pass execution.

### Hot path and data cardinality

This targets the module transform path, which scales with the number of modules passing through `builtin:swc-loader`. The cache key is based on options/config plus the small file-shape data needed for SWC parser adjustment, rather than source content or AST data.

### Expected benefit

Expected CPU and allocation reduction on builds with many JavaScript/TypeScript modules sharing the same SWC loader options, with no output change because parsing and mutable pass construction still happen per transform.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5887,6 +5887,7 @@ dependencies = [
  "swc_ecma_transforms_compat",
  "swc_ecma_transforms_module",
  "swc_ecma_transforms_optimization",
+ "swc_ecma_transforms_proposal",
  "swc_ecma_transforms_react",
  "swc_ecma_transforms_typescript",
  "swc_ecma_utils",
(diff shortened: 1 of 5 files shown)


hardfist:

PR #14134, opened 2026-05-23
fix(types): align SWC unsafe function option

haakjunlee:
## Summary

Align the TypeScript definitions for SWC/Terser compress options with the runtime option name accepted by Rspack.

`unsafe_Function` builds successfully with both `builtin:swc-loader` and `SwcJsMinimizerRspackPlugin`, while the previous `unsafe_function` type points users to an option shape that fails during compiler creation.

## Related links

Repro branch: https://github.com/haakjunlee/rspack-repro/tree/type-error

## Checklist

Validation:

- `corepack pnpm@10.33.4 -C ../rspack --filter @rspack/core build`
- `corepack pnpm@10.33.4 -C ../rspack run lint-ci:js`
- Verified against the repro project with local `@rspack/core` tarball: `unsafe_Function` builds, `unsafe_function` reproduces the original option error.


chenjiahan (approved):
Thank you!

PR #14140, opened 2026-05-25
perf: reduce sourcemap plugin overhead

LingyuCoder:
## What changed

- Cache each asset's source map path and file URL in `SourceMapDevToolPlugin` and reuse it while building, deduplicating, and rendering source names.
- Skip eval sourcemap module condition matching entirely when `test`, `include`, and `exclude` are all unset.
- Reduce temporary allocation in source filename duplicate replacement by using one count/position map instead of two maps plus per-name index vectors.

## Why this should improve performance

Sourcemap generation scales with emitted assets and each asset's source entries. Relative source names previously rebuilt the same source-map file URL for each source entry and carried cloned path buffers through later phases. The new per-asset context moves that stable work to one preparation point and shares it across source-name generation, fallback deduplication, and final rendering.

Eval sourcemap rendering is module-level work. Most devtool configurations do not set `test`, `include`, or `exclude`, so the new fast path avoids per-module asset-condition dispatch and root-module lookups in the common case.

Duplicate source-name handling can run over many source entries. Storing only `(count, position)` avoids allocating vectors of indexes for each repeated filename while preserving output order and duplicate suffix behavior.

## Hot path and data cardinality

The affected paths are `SourceMapDevToolPlugin::collect_tasks`, `deduplicate_source_names`, `create_mapped_asset`, `EvalSourceMapDevToolPlugin::render_module_content`, and `ModuleFilenameHelpers::replace_duplicates`.

The cardinality risk is sources per asset and modules rendered for eval sourcemaps. These are larger than chunks/runtimes, so repeated URL conversion, path cloning, module filtering, and temporary container allocation compound quickly in sourcemap-heavy builds.

## Expected benefit

Expected benefits are lower CPU from fewer repeated URL/path computations and condition checks, plus lower allocation pressure during source-name deduplication. Observable sourcemap output should stay unchanged.

### DIFF
--- a/crates/rspack_plugin_devtool/src/eval_source_map_dev_tool_plugin.rs
+++ b/crates/rspack_plugin_devtool/src/eval_source_map_dev_tool_plugin.rs
@@ -112,27 +112,30 @@ async fn render_module_content(
     .code_generation_results
     .get_hash(&module.identifier(), Some(chunk.runtime()))
     .expect("should have codegen results hash in process assets");
-  let condition_object = AssetConditionsObject {
-    test: self.test.as_ref(),
-    include: self.include.as_ref(),
-    exclude: self.exclude.as_ref(),
-  };
 
-  if module
-    .as_normal_module()
-    .is_some_and(|m| !match_object(&condition_object, m.resource_resolved_data().resource()))
-  {
-    return Ok(());
-  }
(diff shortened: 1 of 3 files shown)


LingyuCoder:
Closing because the benchmark runs did not show a meaningful performance improvement.

PR #14142, opened 2026-05-25
fix(esm-library): split runtime chunk for async chunks

### LINKED ISSUE #1652: bug: invalid ESM chunk imports non-exported __webpack_require__ with Promise.all dynamic imports

JSerFeng:
## Summary

Split runtime chunk for entry who has dependent async chunks.

Fix https://github.com/web-infra-dev/rslib/issues/1652

## Checklist

### DIFF
--- a/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
+++ b/crates/rspack_plugin_esm_library/src/optimize_chunks.rs
@@ -263,6 +263,7 @@ pub(crate) fn extract_tla_shared_modules(compilation: &mut Compilation) -> bool 
 
 /// Ensure that all entry chunks only export the exports used by other chunks,
 /// this requires no other chunks depend on the entry chunk to get exports
+/// or the entry chunk's runtime from async chunks.
 ///
 /// for example entryA -> a -> b => c -> a
 /// entry chunk: a, b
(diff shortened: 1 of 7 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex review

Timeless0911 (approved):

PR #14152, opened 2026-05-26
fix(esm-library): reuse inline export bindings

JSerFeng:
## Summary

- reuse same-chunk hoisted internal bindings for inlined ESM library exports instead of materializing duplicate inline const declarations
- re-export reused bindings from entry chunks when the reusable binding lives in another chunk
- add regression coverage for inline re-export with a live local binding

## Testing

- `pnpm run build:js`
- `cd tests/rspack-test && pnpm run test -t "esmOutput"`
- `cargo fmt --all --check`
- `git diff --check`


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

ahabhgk (approved):

PR #14158, opened 2026-05-26
perf(core): cache resolved module rule options

intellild:
## Summary

- Add `ResolvedModuleRules` caching keyed by the matched module rule effect pointer sequence.
- Keep rule identity compiler-local and cache-only, without adding rule id fields to the option model.
- Pre-resolve global generator options into shared `Arc<GeneratorOptions>`, matching the existing parser option cache shape.
- Pass shared `Arc<ResolvedParserAndGeneratorOptions>` into registered parser/generator builders instead of separate transient option refs.
- Keep `ResolvedParserAndGeneratorOptions` as only `Option<Arc<ParserOptions>>` and `Option<Arc<GeneratorOptions>>`; CSS options are represented by the existing enum variants.
- Normalize CSS base options into `CssModule` variants during option resolution, so CSS parser/generator can borrow typed data from the shared enum options without duplicating state.
- Store module generator options as shared read-only `Arc<GeneratorOptions>`.
- Move asset module build state into `AssetBuildInfo`, and store preserveModules per-module asset filename overrides there instead of mutating generator options.

## Why

Parser and generator options were merged during normal module creation. Large builds can create many modules that match the same rule sequence, so the same parser/generator option merge work and allocations were repeated. This moves that stable work behind a cached resolved-rule object and shares immutable option results between modules.

The registered parser/generator builder path also needed to consume the resolved options object directly. Otherwise plugins such as CSS still had to convert or clone parser/generator options per module after the rule cache had already resolved them.

Rule effects are owned by immutable `CompilerOptions` held by `NormalModuleFactory`, so their addresses are stable for the factory cache lifetime and can serve as compiler-local rule identities.

Asset filename overrides from preserveModules are build-derived module state, not rule generator configuration. Keeping them in `AssetBuildInfo` lets generator options stay immutable and removes the copy-on-write mutation path.

## Validation

- `cargo test -p rspack_core normal_module_factory --lib`
- `cargo test -p rspack_plugin_css --lib`
- `cargo test -p rspack_plugin_asset -p rspack_plugin_json -p rspack_plugin_javascript -p rspack_plugin_wasm --lib`
- `cargo test -p rspack_plugin_css -p rspack_plugin_asset -p rspack_plugin_json -p rspack_plugin_javascript -p rspack_plugin_wasm --lib`
- `cargo test -p rspack_plugin_asset --lib`
- `cargo test -p rspack_plugin_rslib --lib`
- `cargo test -p rspack_plugin_esm_library --lib`
- `pnpm run format:rs`
- `pnpm run test:rs`

---
_by OpenAI Codex_


intellild:

intellild:

intellild:

intellild:
@codex review

intellild:
@codex review

LingyuCoder (approved):

PR #14161, opened 2026-05-26
perf(core): box large NormalModule fields

hardfist:
## Summary

Box the largest cold/heavy `NormalModule` fields instead of constructing `NormalModule` with a custom unsafe boxed initializer:

- `build_info: Box<BuildInfo>`
- `match_resource: Option<Box<ResourceData>>`
- `cached_source_sizes: Box<SourceSizeCache>`

This keeps the normal `NormalModule::new(...).boxed()` path while shrinking the inline `NormalModule` layout. Local DWARF size for `NormalModule` drops from 1648 bytes to 488 bytes, so the `Box::new(self)` move copies much less data.

In the local `rust@build_module_graph` Callgrind comparison, memcpy estimated cycles dropped from 277,349,885 to 263,056,346 (-5.15%), and memcpy share dropped from 8.2578% to 7.9334%. The `NormalModule as ModuleExt>::boxed` memcpy caller dropped from 29,735,845 to 5,438,939 estimated cycles (-81.71%).

## Checklist

Checks used:

- `cargo fmt --all --check`
- `cargo codspeed build -m simulation --profile codspeed -p rspack_benchmark`
- `codspeed run -m simulation -- cargo codspeed run -m simulation -- build_module_graph`

### DIFF
--- a/crates/rspack_binding_api/src/modules/normal_module.rs
+++ b/crates/rspack_binding_api/src/modules/normal_module.rs
@@ -78,8 +78,9 @@ impl NormalModule {
             query,
             fragment,
           } = parse_resource(&val).expect("Should parse resource");
-          *module.match_resource_mut() =
-            Some(ResourceData::new_with_path(val, path, query, fragment));
+          module.set_match_resource(Some(ResourceData::new_with_path(
+            val, path, query, fragment,
+          )));
         }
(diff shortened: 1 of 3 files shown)


hardfist:
close since  it seems bring no gain

PR #14165, opened 2026-05-27
fix(cli): block prototype-pollution payloads in --env parsing

stormslowly:
## Why

Harden `rspack-cli` `--env` parsing against prototype-pollution-style key paths so that:

- malicious dotted keys cannot mutate `Object.prototype` (and from there leak into every plain object created later in the build)
- reserved `RSPACK_*` build flags written by the CLI can't be overridden through the prototype chain

## What

`packages/rspack-cli/src/utils/options.ts`

- Block dotted `--env` paths whose segments hit a small dangerous-key denylist (lodash `_.set` hardening pattern).
- `setBuiltinEnvArg` now uses `Object.prototype.hasOwnProperty.call` instead of `in`, so a polluted prototype can't mask a legitimate write.

`packages/rspack-cli/tests/utils/options.test.ts` (new)

- Regression coverage for the hardened parser plus the existing nested-key and plain-object-shape happy paths.

Single-package change, no public API impact — the `env` object passed to user config functions keeps its plain `{}` shape.

## Credits

Reported privately via responsible disclosure by the security research team from the University of Sydney, focusing on detecting open source software vulnerabilities:

- Liyi Zhou — https://lzhou1110.github.io/
- Ziyue Wang — https://zyy0530.github.io/
- Strick — https://str1ckl4nd.github.io/
- Maurice — https://maurice.busystar.org/
- Chenchen Yu — https://7thparkk.github.io/

Thanks for the careful write-up and the heads-up before public disclosure.

### DIFF
--- a/packages/rspack-cli/src/utils/options.ts
+++ b/packages/rspack-cli/src/utils/options.ts
@@ -125,9 +125,10 @@ function setBuiltinEnvArg(
   value: unknown,
 ) {
   const envName = `RSPACK_${envNameSuffix}`;
-  // `Object.hasOwn` (not `in`) so that a polluted `Object.prototype` cannot
-  // hide a legitimate write of the reserved RSPACK_* flag.
-  if (!Object.hasOwn(env, envName)) {
+  // `hasOwnProperty.call` (not `in`) so that a polluted `Object.prototype`
+  // cannot hide a legitimate write of the reserved RSPACK_* flag.
+  // (`Object.hasOwn` would be cleaner but requires ES2022.)
+  if (!Object.prototype.hasOwnProperty.call(env, envName)) {
     env[envName] = value;
   }
 }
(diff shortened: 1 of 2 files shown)


stormslowly:
Addressed both review comments in 3b44f09:

- **Codex (P2)**: reverted root + intermediate nodes to plain `{}`. The dangerous-key denylist already fences off pollution at the parser entrance, so the `env` object handed to user config functions keeps its expected shape (`env.hasOwnProperty(...)`, `env instanceof Object`).
- **Copilot**: swapped `Object.hasOwn` for `Object.prototype.hasOwnProperty.call` to stay within the repo's ES2021 target.

Added a regression test asserting the plain-object shape for benign inputs. All 6 unit tests green locally.

chenjiahan (approved):

Issue #12444, opened 2025-12-14
[Bug]: lazyConmpilation doesn't work with changing dynamic entries


stormslowly:
### System Info

  System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 101.41 MB / 36.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 22.20.0 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/node
    npm: 10.9.3 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/npm
    pnpm: 10.24.0 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/pnpm
  Browsers:
    Chrome: 143.0.7499.42
    Safari: 18.6

### Details

LazyCompilation doesn't working properly when dynamic entries changed.

### Reproduce Steps

case https://github.com/web-infra-dev/rspack/blob/2c28bcf00d106808016170e5e4d6bb59c2aba1ec/tests/e2e/cases/make/remove-dynamic-entry/index.test.ts failed when `entries: true` enabled

PR #14170, opened 2026-05-27
fix(lazy-compilation): clean up stale entry proxies on dynamic entry removal

stormslowly:
## Why

`lazyCompilation.entries: true` together with a dynamic `entry` function crashes the incremental rebuild with `Module not found` when one of the entries is removed (issue #12444).

The failure looks like one bug but is three independent issues across three crates that only compound on this code path:

- `LazyCompilationPlugin::compiler_make` called `invalid()` on every active proxy each compile — including proxies whose underlying entry source had been deleted on disk. The follow-up force-rebuild then fails to resolve the deleted file.
- `DynamicEntryPlugin::make` only seeded `imported_dependencies` on the Hot (incremental) branch. The first compile is always Cold, so the next compile saw every entry dep as new and reallocated dep ids — breaking the dep → module connection that other plugins rely on.
- `cutout`'s `RemovedFiles` branch revoked the entry dep's connection before `BuildEntryAndClean` could read it. For entry deps the connection is the only edge back to the target module, so the orphan proxy was never cleaned up and a failed `LazyCompilationDependency` stayed stuck in `make_failed_dependencies`.

All three fixes are required for the cleanup chain to run end-to-end.

### Core fixes
- `rspack_plugin_lazy_compilation/src/plugin.rs` — `compiler_make` decides stale purely from the live module_graph: drop the proxy from `active_modules` **only when its sole incoming edge is an entry dep that no longer belongs to any current entry**. Any other shape (multiple incomings, non-entry dep, or still-current entry dep) keeps the proxy active.
- `rspack_plugin_lazy_compilation/src/module.rs` — `is_entry` field, accessor and constructor parameter removed; the dep kind is no longer cached on the proxy because the proxy identifier doesn't encode it (the same resource can be both an entry and a lazy `import()` target).
- `rspack_plugin_dynamic_entry/src/lib.rs` — also seed `imported_dependencies` on the Cold branch so subsequent Hot rebuilds reuse the same dep ids.
- `rspack_core/.../cutout/mod.rs` — snapshot the targets of orphan-style entry deps into `fix_issuers.need_check_modules` **before** `revoke_dependency` deletes the connection. The snapshot/revoke order is load-bearing; a comment in the source warns future refactors not to reorder it.

### Test plan

Three e2e cases under `tests/e2e/cases/lazy-compilation/`:

| case | scenario |
|---|---|
| `dynamic-entry-remove` | two entries on separate HTML pages, delete one entry source, assert `stats.errors == []` |
| `dynamic-entry-import-shared` | shared file factorized as `import()` first then also referenced as a dynamic entry; remove the import(), then remove the entry — both stages must finish cleanly |
| `dynamic-entry-removed-with-import-alive` | shared file is both an entry and an `import()` target; remove only the dynamic entry, then prove the proxy stays active by reloading main and watching the import() resolve again |

Existing `cases/lazy-compilation/*` and `cases/make/remove-dynamic-entry` continue to pass. Reverting any one of the three rust fixes individually reproduces the original failure, confirming each fix is load-bearing.

Refs #12444

### DIFF
--- a/crates/rspack_core/src/compilation/build_module_graph/graph_updater/cutout/mod.rs
+++ b/crates/rspack_core/src/compilation/build_module_graph/graph_updater/cutout/mod.rs
@@ -90,6 +90,29 @@ impl Cutout {
       .fix_build_meta
       .analyze_force_build_modules(artifact, &force_build_modules);
 
+    // Snapshot the targets of entry-like dependencies before they get revoked.
+    // For entry deps (no parent module, no parent block) the connection is the
+    // only edge back to their target module. After `revoke_dependency(force=false)`
+    // below removes that connection, later `BuildEntryAndClean` can no longer
+    // find the target via the module graph, so it would miss running an orphan
+    // check on the target. Add those targets to `need_check_modules` up-front
+    // so `fix_issuers` can revoke them if they end up without any incoming.
+    {
+      let module_graph = artifact.get_module_graph();
+      for dep_id in &force_build_deps {
+        if module_graph.get_parent_module(dep_id).is_some()
+          || module_graph.get_parent_block(dep_id).is_some()
+        {
+          continue;
+        }
+        if let Some(con) = module_graph.connection_by_dependency_id(dep_id) {
+          self
+            .fix_issuers
+            .add_need_check_module(*con.module_identifier());
+        }
+      }
+    }
+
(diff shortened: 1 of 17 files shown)


ahabhgk:
@codex review

ahabhgk (approved):

PR #14173, opened 2026-05-27
revert: upgrade swc_core from 66 to 67

JSerFeng:
## Summary

- Revert `chore: upgrade swc_core from 66 to 67 (#14135)`.
- Restore the SWC dependency set to the previous `swc_core 66` versions.
- Regenerate the workspace SWC version constant back to `66.0.2`.

## Why

The `swc_core 67` upgrade changed decorator helper output for 2022-03 decorators. Downstream ecosystem projects can still resolve `@swc/helpers@0.5.21` from their existing lockfiles, which is incompatible with the generated `_apply_decs_2203_r` helper import and keeps Ecosystem CI failing even after raising Rspack's peer dependency minimum.

This revert unblocks the release while downstream lockfiles / helper version handling are updated separately.

## Validation

- `git diff --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5703,9 +5703,9 @@ checksum = "b7401a30af6cb5818bb64852270bb722533397edcfc7344954a38f420819ece2"
 
 [[package]]
 name = "swc"
-version = "65.0.0"
+version = "64.0.0"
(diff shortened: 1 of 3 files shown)


CPunisher (approved):

JSerFeng:
duplicated, already reverted

PR #14174, opened 2026-05-27
fix: derive modern-module require external type from target

JSerFeng:
## What

- Thread a `fallbackType` option through `ExternalsPlugin` so `externalsType: 'modern-module'` can choose the final CommonJS external variant before rendering.
- Use `node-commonjs` for Node-like targets and `commonjs` for other targets when a `modern-module` external is reached through CommonJS `require()`.
- Update the web modern-module external test and EN/ZH externals docs to describe the target-dependent behavior.

## Why

`modern-module` already resolves static ESM imports as `module` and dynamic imports as `module-import`, but require-style dependencies were collapsed to one CommonJS rendering path. That made web-like targets emit `createRequire`, even though the target decision is known earlier from the normalized target presets.

## How

- Store the resolved CommonJS fallback type in dependency metadata as `ExternalTypeEnum::CommonJs(ExternalType)`.
- Choose the fallback in the JS options apply path from `externalsPresets`, and in the Rust builder path from target properties.
- Keep rendering logic simple: `ExternalModule::resolve_external_type()` only consumes the already-resolved fallback type.

## Validation

- `corepack pnpm run build:js`
- `corepack pnpm --dir tests/rspack-test run test -- -t modern-module-explicit-web`
- `corepack pnpm --dir tests/rspack-test run test -- -t modern-module-explicit-node`
- `corepack pnpm --dir tests/rspack-test exec rstest esm`
- `cargo check -p rspack`
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -2262,6 +2262,7 @@ export interface RawExternalItemFnResult {
 
 export interface RawExternalsPluginOptions {
   type: string
+  fallbackType?: string
   externals: (string | RegExp | Record<string, string | boolean | string[] | Record<string, string[]>> | ((...args: any[]) => any))[]
   placeInInitial: boolean
 }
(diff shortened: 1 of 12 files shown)


Timeless0911 (approved):
LGTM

PR #14177, opened 2026-05-27
fix: compat es5 for define property getters runtime

ahabhgk:
## Summary

`{ [kind]: defs[key] }` is not compatible to es5

## Checklist


chenjiahan (commented):
Can we generate less runtime code for es6+ targets?

ahabhgk:
@codex review

JSerFeng (approved):

PR #14178, opened 2026-05-27
fix: resolve external type in esm rendering

JSerFeng:
## What

Use `resolve_external_type()` when ESM library and Rslib external rendering need to decide whether an external should be treated as a module external.

This updates the ESM library issuer disambiguation path, the ESM link import path, and Rslib's star re-export external cutout/render path. It also adds an ESM output regression case for `export * from "externals"` with `externalsType: "modern-module"` and `RslibPlugin`.

## Why

`get_external_type()` only reflects the configured external type, so `modern-module` and `module-import` can be classified incorrectly without considering the actual dependency kind. Static ESM externals should resolve to `module`, while dynamic imports and CommonJS dependencies should keep their resolved behavior.

## How

- Replace module-like checks based on `get_external_type().starts_with("module")` with `resolve_external_type() == "module"`.
- Add `esmOutputCases/externals/modern-module-star-reexport-rslib` to verify the output preserves `export * from "externals"`.
- Refresh the affected `module-import` namespace collision snapshot after the resolved-type behavior changes the async chunk query hash.

## Validation

- `cargo fmt --all --check`
- `corepack pnpm --dir tests/rspack-test run test -t "modern-module-star-reexport-rslib"`
- `corepack pnpm --dir tests/rspack-test run test -t "esmOutputCases/externals"`

### DIFF
--- a/crates/rspack_plugin_esm_library/src/plugin.rs
+++ b/crates/rspack_plugin_esm_library/src/plugin.rs
@@ -757,6 +757,8 @@ async fn after_factorize(
   // Check if this is an external module using the existing downcast helper
   if let Some(external_module) = module.as_external_module_mut()
     && external_module.resolve_external_type() == "module"
+    && (external_module.get_external_type() != "modern-module"
+      || data.dependencies.first().is_some_and(is_esm_dep_like))
   {
     // If there's an issuer, append it to the module id
     if let Some(issuer_identifier) = &data.issuer_identifier {
--- a/crates/rspack_plugin_esm_library/src/plugin.rs
+++ b/crates/rspack_plugin_esm_library/src/plugin.rs
@@ -756,9 +756,7 @@ async fn after_factorize(
 ) -> Result<()> {
   // Check if this is an external module using the existing downcast helper
   if let Some(external_module) = module.as_external_module_mut()
-    && (external_module.get_external_type().starts_with("module")
-      || (external_module.get_external_type() == "modern-module"
-        && data.dependencies.first().is_some_and(is_esm_dep_like)))
+    && external_module.resolve_external_type() == "module"
   {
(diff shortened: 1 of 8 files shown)


JSerFeng:
@codex review

crates/rspack_plugin_rslib/src/dyn_import_external.rs
```diff
@@ -268,7 +267,8 @@ impl DependencyTemplate for ExportImportedDependencyTemplate {
       && let Some(module) = module
         .and_then(|mid| mg.module_by_identifier(mid))
         .and_then(|m| {
-          m.as_external_module().filter(|&m| m.get_external_type().starts_with("module"))
+          m.as_external_module()
+            .filter(|&m| m.resolve_external_type() == "module")
```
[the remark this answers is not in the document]
JSerFeng: This is on purpose

crates/rspack_plugin_esm_library/src/plugin.rs
```diff
@@ -756,9 +756,7 @@ async fn after_factorize(
 ) -> Result<()> {
   // Check if this is an external module using the existing downcast helper
   if let Some(external_module) = module.as_external_module_mut()
-    && (external_module.get_external_type().starts_with("module")
-      || (external_module.get_external_type() == "modern-module"
-        && data.dependencies.first().is_some_and(is_esm_dep_like)))
```
Timeless0911, line 761: `data.dependencies.first().is_some_and(is_esm_dep_like)))`  is no longer needed?

JSerFeng: good catch

JSerFeng:
@codex review

Timeless0911 (approved):

PR #14180, opened 2026-05-27
fix: respect resolved external type in ESM rendering

JSerFeng:
## What

- Gate ESM dynamic import external passthrough on `ExternalModule::resolve_external_type() == "import"`.
- Keep Rslib dynamic-import external cutout broad for non-ESM output, but in `output.module` only cut out externals that resolve to `import`.
- Prevent ESM link-time raw star re-exports from being emitted for externals that do not resolve to `module`.
- Add ESM output regression coverage for commonjs dynamic import externals and commonjs star re-export externals.

## Why

Raw ESM `import(...)` and `export * from ...` output is only correct for the matching resolved external type. If a user forces a CommonJS-style external, preserving the raw ESM form bypasses the intended CJS external rendering path and can emit the wrong runtime shape.

## How

- Check `resolve_external_type()` before bypassing normal dynamic import rendering.
- Pass `output.module` into Rslib dynamic external cutout so CJS library compact output keeps its existing behavior while ESM output preserves non-`import` externals for normal rendering.
- Filter ESM `raw_star_exports` to resolved `module` externals only.
- Update affected ESM snapshots around module star re-export optimization.

## Checks

- `cargo fmt --all --check`
- `git diff --check`
- `pnpm --dir tests/rspack-test run test -t "configCases/rslib/compact-external-dynamic-import-runtime"`
- `pnpm --dir tests/rspack-test exec rstest EsmOutput.test.js`

### DIFF
--- a/crates/rspack_plugin_esm_library/src/dependency/dyn_import.rs
+++ b/crates/rspack_plugin_esm_library/src/dependency/dyn_import.rs
@@ -3,8 +3,9 @@ use std::{borrow::Cow, sync::Arc};
 use atomic_refcell::AtomicRefCell;
 use rspack_collections::IdentifierMap;
 use rspack_core::{
-  Dependency, DependencyId, DependencyTemplate, ExportsType, FakeNamespaceObjectMode, ModuleGraph,
-  ModuleReferenceOptions, RuntimeGlobals, TemplateContext, get_exports_type,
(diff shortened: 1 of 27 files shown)


JSerFeng:
@codex review

Timeless0911 (approved):

PR #14185, opened 2026-05-27
feat(binding): print backtrace on native crashes (SIGSEGV/SIGBUS/SIGILL)

stormslowly:
## Why

Today, when the released `.node` binding hits a fatal native signal (stack overflow, UB in a C dep, unaligned access, …), users see only:

```
Illegal instruction: 4   npm run build
```

The Rust panic hook (`panic::install_panic_handler`) only fires for `panic!` — it can't catch raw signals. That means maintainers triaging issue reports get no stack, no faulting address, nothing actionable.

## What

Install an async-signal-safe handler at module load that, on SIGSEGV / SIGBUS / SIGILL:

- prints the signal name + faulting address (`siginfo_t::si_addr`)
- prints a raw-address backtrace via `backtrace::trace` (deferred symbolication — resolve later with `atos -o <binary> <addr>` or `addr2line`)
- re-raises with the default handler so the OS still produces a crash dump (`~/Library/Logs/DiagnosticReports/` on macOS, `coredumpctl` on Linux)

Implementation notes:

- Runs on a dedicated 64 KiB alt stack via `sigaltstack`, so a stack-overflow signal can still execute the handler.
- All writes go through `libc::write` directly — no `println!`, no allocations, no locks.
- A re-entry guard restores the default handler if the handler itself faults.
- WASM and non-Unix targets are no-ops (Windows has its own SEH model via napi-rs).

### Before

```
Illegal instruction: 4   npm run build
```

### After

```
========================================
rspack native crash: signal=SIGILL (likely Rust trap from stack-overflow guard)  fault_addr=0x16f603ff0
This is a bug in rspack (or a native dep). Please file an issue: https://github.com/web-infra-dev/rspack/issues
Raw backtrace (resolve with `atos -o <binary> <addr>`):
  #0  ip=0x10c2e1234  sp=0x16f604010
  #1  ip=0x10c2e5678  sp=0x16f604080
  ...
========================================
```

`cargo check` passes under both dev and release profiles.


stormslowly:
Addressed all three review comments in e772b30:

- **Copilot — per-thread sigaltstack**: split installer into a process-wide `sigaction` registration and an idempotent `install_alt_stack_for_current_thread()` that leaks one 64 KiB buffer per thread. Wired into the tokio runtime builder via `on_thread_start`, so every worker now gets its own alt stack.
- **Copilot — workspace deps**: moved `backtrace` and `libc` to `[workspace.dependencies]`; the crate consumes them via `workspace = true`.
- **Typos CI**: spelled out "Structured Exception Handling" so the Windows-only branch doesn't trip the `SEH`/`SHE` rule.

`cargo check` passes under both dev and release profiles.

PR #14186, opened 2026-05-27
perf: reduce sourcemap benchmark overhead

hardfist:
## Summary

- Keep this PR scoped to Rspack-side performance changes only; the temporary third-party crate patches were reverted.
- Add `simdutf8` fast paths for loader/source buffer UTF-8 conversion.
- Reduce overhead in identifier escaping while preserving the invalid-character escape behavior.
- Add a fast path for the default devtool module filename template.

## Benchmark

Target benchmark: `threejs-production-sourcemap`  
Measurement mode: CodSpeed CPU simulation with `RAYON_NUM_THREADS=1` for local comparability. Rayon scheduling cost was intentionally not optimized because it is distorted under Valgrind/CodSpeed.  
Primary metrics: accesses and estimated cycles.

| Scope | Accesses Before | Accesses After | Accesses Delta | Estimated Cycles Before | Estimated Cycles After | Estimated Cycles Delta |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| Rspack-only changes | 1,184,875,619 | 1,171,950,653 | -12,924,966 (-1.091%) | 1,359,247,384 | 1,347,093,638 | -12,153,746 (-0.894%) |

Raw local artifact: `optimization-artifacts/codspeed/20260528-144134-rspack-only-rerun/metrics.txt`

## Validation

- `cargo lint`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5085,7 +5085,8 @@ dependencies = [
 [[package]]
 name = "rspack_sources"
 version = "0.4.23"
-source = "git+https://github.com/rstackjs/rspack-sources?branch=perf%2Foptimize-mappings-encoder#0c054acc75ccc1dbfeaacfa1d126ad689a0ce938"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "81a72d10c90a000749a73245046be8071ecdd1ce6d21673ff17fe1da8eed6e8d"
 dependencies = [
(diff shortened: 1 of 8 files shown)


hardfist:
@codspeedbot explain the performance gain in bundle@threejs-production-sourcemap case

stormslowly (approved):

PR #14191, opened 2026-05-28
fix(pure-functions): keep deferred local dependencies

### LINKED ISSUE #14188: [Bug]: `pureFunctions` nullifies binding still read by side-effectful expression
### System Info

  System:
    OS: macOS 26.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 302.03 MB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.17.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 11.6.3 - /usr/local/bin/npm
    pnpm: 9.15.4 - /usr/local/bin/pnpm
    Deno: 2.5.6 - /opt/homebrew/bin/deno
    Watchman: 20230430.125247.0 - /usr/local/bin/watchman
  npmPackages:
    @rspack/cli: ^2.0.4 => 2.0.4 
    @rspack/core: ^2.0.4 => 2.0.4 

### Details

With `experiments.pureFunctions` enabled, rspack incorrectly nullifies a binding that is detected as an unused pure expression, but is actually still used as a parameter to a side-effecting function.

### Reproduce link

https://github.com/jfirebaugh/rspack-pure-functions-bug

### Reproduce Steps

Clone the above repro and run `npm install && npm test`.

JSerFeng:
## What

- Propagate deferred pure checks through InnerGraph top-level symbol dependencies.
- Add a regression case for a retained deferred-impure expression reading a local binding that would otherwise be replaced by an unused pure expression.

## Why

When `experiments.pureFunctions` was enabled, InnerGraph could treat `const a = foo` as an unused pure declarator even though a retained expression like `const b = sideEffect(a)` still reads `a`. If `sideEffect` later resolves as impure through a deferred pure check, `b` is kept but `a` was still rendered as `null && foo`, causing runtime failures such as `fn is not a function`.

## How

- Keep `TopLevel` dependencies in the usage set when the referenced symbol has deferred pure checks.
- Merge deferred pure checks from referenced top-level symbols into the current symbol's `UsedByExports`, so the local binding is preserved when the dependent expression is retained.

## Validation

- `pnpm run build:js`
- `cd tests/rspack-test && pnpm run test -t "deferred-pure-keeps-local-dependency"`
- `cd tests/rspack-test && pnpm run test -t "user-mark-pure-functions"`
- `cd tests/rspack-test && pnpm run test -t "tree-shaking"`
- `pnpm exec cargo fmt --all --check`
- `git diff --check`
- Verified the original `jfirebaugh/rspack-pure-functions-bug` repro with the local patched Rspack output now runs successfully and prints `undefined`.

Closes #14188

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/inner_graph/plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/inner_graph/plugin.rs
@@ -114,11 +114,14 @@ impl InnerGraphParserPlugin {
               InnerGraphMapSetValue::Str(v) => {
                 new_set.insert(InnerGraphMapSetValue::Str(v));
               }
-              InnerGraphMapSetValue::TopLevel(v) => {
-                if v == *key {
+              InnerGraphMapSetValue::TopLevel(dep_symbol) => {
+                if dep_symbol == *key {
                   continue;
                 }
-                match state.inner_graph.get(&v) {
+                if deferred_pure_checks_by_symbol.contains_key(&dep_symbol) {
+                  new_set.insert(InnerGraphMapSetValue::TopLevel(dep_symbol));
+                }
+                match state.inner_graph.get(&dep_symbol) {
                   Some(InnerGraphMapValue::True) => {
                     set_is_true = true;
                     break;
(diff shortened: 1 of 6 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

SyMind (approved):

PR #14203, opened 2026-05-29
test(module-federation): add lazyCompilation entries case

### LINKED ISSUE #12443: [Bug]: lazycompilation with entries enable doesn't working with MF
### System Info

```
  System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M3 Pro
    Memory: 101.41 MB / 36.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 22.20.0 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/node
    npm: 10.9.3 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/npm
    pnpm: 10.24.0 - /Users/bytedance/.nvm/versions/node/v22.20.0/bin/pnpm
  Browsers:
    Chrome: 143.0.7499.42
    Safari: 18.6
```

### Details

after Enable `{lazyCompilation: { entries: true} }`, [the case ](https://github.com/web-infra-dev/rspack/blob/main/tests/e2e/cases/module-federation/container-entry-split-chunks/index.test.ts)failed 

### Reproduce link

_No response_

### Reproduce Steps

https://github.com/web-infra-dev/rspack/blob/main/tests/e2e/cases/module-federation/container-entry-split-chunks/index.test.ts
with config `{lazyCompilation: { entries: true} }`

stormslowly:
## Why

`lazyCompilation: { entries: true }` did not work with the `container-entry-split-chunks` Module Federation case (#12443).

That case is a **self-referential MF build**: `remoteEntry` and `main` run on the same page, each emitting its own runtime that defines the same `self["rspackHotUpdate"]` global (derived from a single `output.uniqueName`). The later-loaded `remoteEntry` clobbers the host's. lazyCompilation `entries` activates the entry **over HMR**, so the activation update is routed to the wrong runtime and is lost — the remote never renders.

## What

Add a new e2e case `container-entry-split-chunks-lazy` (a copy of `container-entry-split-chunks`) that exercises the lazy + MF combination:

- `lazyCompilation: { entries: true }`
- `optimization.runtimeChunk: 'single'` — one shared runtime keeps a single HMR global, so host and container no longer clash and lazy entry activation works.

The original `container-entry-split-chunks` case is left **unchanged**. Test-only change; no engine change.

Closes #12443


stormslowly:
@codex review it

hardfist (approved):

PR #14210, opened 2026-05-29
fix(rspack_watcher): scan after registering the disk watcher to close a startup race

stormslowly:
## Why
The native-watcher test `nativeWatcherCases-watchCases/change-event/metadata > should compile step 1` flakily timed out at 30s on Linux CI (e.g. [run 26564410186](https://github.com/web-infra-dev/rspack/actions/runs/26564410186/job/78257139614)).

`FsWatcher::wait_for_event` ran the startup **scan** (which catches changes whose `mtime > start_time`) **before** `disk_watcher.watch()`, and the scan runs in spawned tasks — so it could snapshot a file's mtime *before* the inotify/FSEvents watch was registered. A change landing in that window was seen by **neither** path:

- the scan read the *pre-change* mtime → `mtime > start_time` was false → no event;
- the watch wasn't active yet → the live event was never delivered.

→ no rebuild → step 1 hangs to timeout. (The case touches `index.js` via an async `fs.utimes` in the compile `done` hook; under CI load that write lands inside the gap.)

## What
`notify`'s `watch()` registers the kernel watch **synchronously**, so moving the scan to run **after** it returns closes the gap: a change before the watch is active is on disk by scan time (caught by the scan), a change after it is active is caught by the live watcher. Same order webpack/watchpack uses (watch, then scan). One file, no extra syscalls, no timers.

## Evidence
Cross-platform verification with a pre-fix control, replicating the original load environment — each cell loops the **full `nativeWatcher` suite** (`retry: 0`) × 30 iterations × 4 shards, built from both the pre-fix parent `3b38926018` and this fix ([run 26733042615](https://github.com/web-infra-dev/rspack/actions/runs/26733042615)):

| Platform | Pre-fix (control) | Post-fix |
|---|---|---|
| linux-x64 / node 20 | **reproduced** the flake (iteration 5) | **120/120 green** |
| mac-arm64 / node 22 | not reproduced (0/120) | 120/120 green |
| win-x64 / node 22 | not reproduced (0/120) | 120/120 green |

- **linux-x64**: pre-fix reproduces under load, post-fix is green → the fix resolves the race.
- **mac-arm64 / win-x64**: the pre-fix binding never reproduced in 120 iterations each, so this race is effectively **Linux-specific** (the inotify arm-timing window doesn't bite the FSEvents / ReadDirectoryChangesW backends the same way); post-fix green confirms the reorder is **safe / non-regressing** there.


stormslowly:
@codex review

hardfist (approved):

PR #14218, opened 2026-05-29
fix(css): preserve import render conditions

intellild:
## Summary

- preserve inherited/current CSS render conditions across CSS import module creation
- include render conditions in CSS request and inline/style dedupe keys so different conditional imports create independent modules
- align the existing css/import test case sources with webpack coverage for nested media/supports/layer imports

## Related links

N/A

## Checklist

Validation:

- `cargo check -p rspack_plugin_css`
- `cargo test -p rspack_plugin_css css_source_builder`
- `cd tests/rspack-test && pnpm run test -t "configCases/css/import"`

---
_by OpenAI Codex_


intellild:

packages/rspack/src/config/defaults.ts
```diff
@@ -1242,20 +1242,27 @@ const getResolveDefaults = ({
     },
   };
 
-  resolveOptions.byDependency!['css-import'] = {
+  const styleConditions = [];
+
+  styleConditions.push('webpack');
```
ahabhgk: If I remember correctly this is intented to not align with webpack, cc @chenjiahan

chenjiahan: `webpack` should be removed, see https://rspack.rs/guide/migration/rspack_1.x#css-resolve-conditions

intellild:
@codex review

ahabhgk (dismissed):

ahabhgk (approved):

PR #14240, opened 2026-06-01
chore(dep): css-module-lexer 0.0.16 → 0.0.15

stormslowly:
## Why

Revert the `css-module-lexer` bump (`0.0.16` → `0.0.15`) to verify whether the upgrade introduced a regression. Draft PR so CI can validate the downgrade end-to-end.

| | version |
|---|---|
| before | `0.0.16` |
| after | `0.0.15` |

`cargo check -p rspack_plugin_css` passes locally — no API changes required.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1126,9 +1126,9 @@ dependencies = [
 
 [[package]]
 name = "css-module-lexer"
-version = "0.0.16"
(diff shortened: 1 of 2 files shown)


hardfist:
@codspeedbot explain this walltime regression

PR #14249, opened 2026-06-02
perf: concat source with capacity

SyMind:
## Summary

- Add `ConcatSource::with_capacity` and make `ConcatSource::new` construct directly from pre-boxed sources, reducing repeated `add` calls and intermediate reallocations.
- Preallocate `ConcatSource` / `Vec` capacity in hot rendering paths, including concatenated module code generation, init fragments, JavaScript runtime rendering, chunk formats, CSS modules, banners, and source map comments.
- Avoid unnecessary source materialization in several checks by using `size()` and reuse boxed namespace object sources when appending to concat output.

## Checklist

### DIFF
--- a/crates/rspack_core/src/concatenated_module.rs
+++ b/crates/rspack_core/src/concatenated_module.rs
@@ -1591,7 +1591,6 @@ impl Module for ConcatenatedModule {
       });
     }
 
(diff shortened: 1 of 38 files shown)


SyMind:
@codex review

PR #14250, opened 2026-06-02
chore: scale threejs walltime benchmark input

hardfist:
## Summary

- Generate a local `threejs-10x` benchmark fixture during `pnpm run bench:prepare` by copying the upstream threejs source input ten times and importing each copy from a generated entry.
- Add a separate Cargo benchmark target, `walltime`, for the threejs walltime bundle benchmarks so they are not registered in the existing `benches` target.
- Point the Rust walltime CI job at `--bench walltime`.
- Run the `threejs-10x` walltime benchmark with native filesystem output instead of the regular bundle benchmark's in-memory output filesystem.
- Disable compare-before-emit for the native output path and clean native output after each measured walltime iteration, so iterations do not switch into a warmed skip-write path and missing `dist` does not panic before emit.

## Why

The walltime benchmark needs a larger threejs input scale to make the measured workload more representative and stable without increasing or mixing into the default simulation benchmark workload.

## Validation

- `node ./scripts/bench/setup.mjs`
- Verified generated fixture scale locally: 10 copied inputs and generated entry/config/package exist.
- `cargo fmt --package rspack_benchmark`
- `cargo check -p rspack_benchmark --bench walltime`
- `cargo check -p rspack_benchmark --bench benches`
- Commit hooks ran `rustfmt` for staged Rust files.

### DIFF
--- a/.github/workflows/bench-rust.yml
+++ b/.github/workflows/bench-rust.yml
@@ -21,10 +21,6 @@ on:
         required: false
         type: string
         default: ''
-      enable_threejs_10x: # Register the scaled threejs 10x benchmark cases
-        required: false
-        type: boolean
-        default: false
       ref: # Git reference to checkout
         required: false
         type: string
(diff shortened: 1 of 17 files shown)


CPunisher:
We'd better add `threejs-10x` to https://github.com/rstackjs/rspack-benchcases instead of constructing it in the script

stormslowly (dismissed):

CPunisher (approved):

PR #14263, opened 2026-06-03
perf: reduce compilation hash overhead

LingyuCoder:
## Summary

Reduce compilation hash overhead in the existing create-hash path without moving work out of CodSpeed measurement.

- In `Chunk::update_hash`, keep the common no-runtime-module path on the existing `get_ordered_chunk_modules_identifier` helper, preserving the previous ordered-module behavior.
- For chunks with runtime modules, collect module identifiers directly, filter the small runtime-module list before sorting, then sort the remaining normal module identifiers by `as_str()`. This preserves the previous hash input order while avoiding runtime-module items in the normal-module sort path only when runtime modules exist.
- In `Chunk::update_hash`, skip ordered runtime-module hashing entirely for chunks that have no runtime modules.
- For chunks with runtime modules, use the small runtime-module `Vec` for filtering instead of building an `IdentifierSet`; runtime modules are normally few and there is no threshold/cache.
- Skip the per-chunk `dependent_full_hash` scan when the hook has no taps/interceptors.
- Before hashing runtime modules in non-runtime chunks, skip the `rspack_parallel::scope` entirely when none of those chunks have runtime modules. This avoids the previous zero-spawn scope path.
- Before hashing runtime modules in each runtime chunk, skip the `rspack_parallel::scope` when that runtime chunk has no runtime modules.
- When spawning non-runtime chunk hashing tasks, pass `&SharedPluginDriver` through `rspack_parallel::scope` instead of cloning the shared plugin driver once per chunk task.
- Consume runtime chunk `referenced_by` lists with `std::mem::take` instead of cloning them while topologically processing runtime chunks.
- For full-hash aggregation, sort `ChunkUkey` keys directly and read `ChunkHashesArtifact` once per key instead of sorting `Chunk` values and performing separate hash/content-hash artifact lookups through `Chunk` accessors.
- In runtime-module code generation, skip the empty-runtime-module path and reserve the runtime source map / generated-module set before filling them.
- Preserve deterministic chunk/module order, missing-hash behavior, runtime-module hashing, entry hashing, JS/CSS content-hash behavior, and error-path state updates.

## Scope Notes

- Final diff touches only `crates/rspack_core/src/chunk.rs` and `crates/rspack_core/src/compilation/create_hash/mod.rs`.
- No `chunk_graph_chunk.rs` / `chunk_graph_module.rs` changes.
- No JS/CSS plugin changes remain.
- No separate `ChunkHashingArtifact`.
- No `ChunkHashesArtifact::prepare_ordered_modules` cache or ordered-module precomputation remains.
- No unexplained threshold and no content-hash folding fast paths.
- No CodSpeed benchmark timing manipulation remains.

## Target Benchmarks

- `rust@create_chunk_hashes`: target >= 10%
- `rust@create_full_hash`: target >= 10%

## Validation

- CI-only validation per optimization workflow; no local benchmark/build/test run.
- Local pre-push checks: `cargo fmt --package rspack_core` and `git diff --check`.
- Current pushed head: `965040bc49b0079079cbf93c81cc9fe278466406`.
- Subagent correctness review for this head found no correctness issues. It confirmed the new runtime-module codegen empty-path and reserve calls preserve stale-source clearing, error behavior, and `code_generated_modules` semantics; it also confirmed the underlying map/set reserve calls are valid for `IdentifierMap`/`IdentifierSet`.
- CI/CodSpeed pending for current head.

Previous CodSpeed result for `ad1730b` did not meet the target: `rust@create_full_hash -2.06%` and no >=10% `rust@create_chunk_hashes` result was reported. This head keeps the scoped create-hash changes and adds a measured-path runtime-module codegen allocation fast path for the next CodSpeed run.

### DIFF
--- a/crates/rspack_core/src/chunk.rs
+++ b/crates/rspack_core/src/chunk.rs
@@ -2,7 +2,6 @@ use std::{cmp::Ordering, fmt::Debug, hash::Hash};
 
 use itertools::Itertools;
 use rayon::prelude::*;
-use rspack_collections::IdentifierSet;
 use rspack_error::Diagnostic;
 use rspack_hash::{RspackHash, RspackHashDigest};
 use rspack_util::fx_hash::{FxIndexMap, FxIndexSet};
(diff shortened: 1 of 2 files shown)


LingyuCoder:
@copilot review

Please review the latest head `965040bc49b0079079cbf93c81cc9fe278466406`, focusing on correctness of the scoped create-hash changes and runtime-module codegen fast path.

PR #14083, opened 2026-05-19
fix(watcher): re-send updated file deps to native watcher (#12904)

stormslowly:
Closes #12904.

## Why

After lazy compilation activates a module, the proxy is force-revoked and re-factorized within one build cycle. The lazy module's source path ends up in `IncrementalInfo.updated` (removed then re-added), which is filtered out of `addedFileDependencies`. The native watcher's incremental update protocol never sees it, and subsequent FSEvents are silently dropped by `DependencyFinder`.

## What

- **JsDependencies (binding)** — `added_file_dependencies` / `added_context_dependencies` / `added_missing_dependencies` / `added_build_dependencies` now merge `updated` into `added` so the JS-side native watcher's incremental `(added, removed)` protocol re-sees paths that went through a remove+add cycle. `PathTracker.add` is idempotent on the native side, so re-emitting tracked paths is safe.
- **Compilation::*_dependencies** — tuple entries stay disjoint (added/updated/removed). Persistent cache (`PersistentCache::after_compile`) already does `file_added.chain(file_updated)` and is unaffected — no double snapshot work.
- Fix three pre-existing copy-paste bugs in `context_dependencies()` / `missing_dependencies()` / `build_dependencies()` where the per-cycle additions were chaining `&self.file_dependencies` instead of the matching collection.
- Add a unit test in `file_counter` documenting the remove+add cycle that triggered the bug.

## Notes

Minimal local reproduction was not achievable (tried current main + v1.6.6 with concurrent trigger+modify, parent-rebuild, warm persistent cache, burst patterns — all PASS). This patch is the most plausible fix based on tracing the FileCounter / PathTracker flow that the user's instrumentation pointed at. Verification via canary in the reporter's project.

Issue #12904, opened 2026-01-31
[Bug]: nativeWatcher does not detect file changes for lazy-compiled modules


T9-Forever:
### System Info

- **@rspack/core**: 1.6.x
- **@rsbuild/core**: 1.6.6
- **OS**: macOS 14.x
- **Node.js**: 20.x
- **Project size**: ~4700 file dependencies, ~10600 missing dependencies

### Steps to Reproduce
1. Enable both lazyCompilation and nativeWatcher
2. Start dev server: npm run dev
3. Open browser, navigate to a lazy-loaded route (triggers lazy compilation)
4. Modify a file in that lazy-loaded module
5. Expected: HMR update, page reflects changes
6. Actual: Nothing happens, no rebuild triggered

### Workaround
Disable either `lazyCompilation` OR `nativeWatcher`:
- **lazyCompilation: false** + **nativeWatcher: true** → Works ✅
- **lazyCompilation: true** + **nativeWatcher: false** → Works ✅

#### Step 1: Add Debug Logs to Native Watcher (Rust)
Modified `rspack/crates/rspack_watcher/src/trigger.rs`:
```rust
pub fn on_event(&self, path: &ArcPath, kind: FsEventKind) {
  let finder = self.finder();
  let associated_event = finder.find_associated_event(path, kind);

  if associated_event.is_empty() {
    eprintln!("[NativeWatcher] ⚠️ FILE CHANGE NOT DETECTED: {:?} {}", kind, path.display());
    eprintln!("[NativeWatcher]    Registered: files={}, dirs={}, missing={}",
      finder.files.len(), finder.directories.len(), finder.missing.len());
  } else {
    eprintln!("[NativeWatcher] ✅ File change detected: {:?} {} -> {} events",
      kind, path.display(), associated_event.len());
  }

  self.trigger_events(associated_event);
}
```

#### Step 2: Add Debug Logs to FileCounter (Rust)
Modified `rspack/crates/rspack_core/src/utils/file_counter/mod.rs`:
```rust
pub fn add_files(&mut self, resource_id: &ResourceId, paths: &ArcPathSet) {
  for path in paths {
    let list = self.inner.entry(path.clone()).or_default();
    let path_str = path.to_string_lossy();

    // Track specific file for debugging
    if path_str.contains("PersonaSetupEntry") {
      if list.is_empty() {
        eprintln!("[FileCounter] 🎯 NEW PersonaSetupEntry file added: {}", path_str);
      } else {
[...]

PR #14271, opened 2026-06-04
fix(watcher): carry dependency deltas across coalesced rebuilds

stormslowly:
## Why

`lazyCompilation` + `experiments.nativeWatcher`: edits to lazy-compiled modules are not detected (#12904).

When a build finishes while `watching.invalid` is already set — a **coalesced rebuild**, which lazy activation routinely produces via programmatic `invalidate()` — `Watching._done` takes the `#go()` early-return and skips the `watch()` delivery. The native watcher is delta-based, so the skipped build's `added` paths are lost: the next make's `reset_temporary_data` clears them and they are never watched.

Both workarounds in the issue avoid it: `nativeWatcher: false` (watchpack gets the full set every watch) and `lazyCompilation: false` (no programmatic `invalidate()` to coalesce).

## What

Carry each skipped build's `(added, removed)` deltas forward and merge them into the next delivered `watch()`. Non-coalesced builds and watchpack are unaffected.

## Notes

Root-cause alternative to the symptom patch in #14083 (re-sending `updated` deps): a path lost via the coalescing skip never becomes `updated`, so re-sending `updated` does not recover it.


hardfist:
@codspeed explain the walltime diff

hardfist:
@codspeed explain the walltime diff

stormslowly:
Superseded by #14279 — pushed to an origin branch instead of a fork.

PR #14286, opened 2026-06-05
fix(tree-shaking): handle nested pure function calls

JSerFeng:
## Summary

Fix `experiments.pureFunctions` purity analysis so a marked pure call can be removed when its arguments contain nested expression containers that are themselves pure, including cases like `a([b()])` and `a([b(0, foo())])`.

The root cause was that `is_pure_expression` only recognized direct calls/imports/classes/functions and delegated other expressions to SWC side-effect checks, so arrays/objects/conditionals/binary/unary/optional calls containing user-marked pure calls were treated too conservatively.

Add an ESM output regression case under `esmOutputCases/basic/pure-functions-expression-nesting` with snapshot guards that fail if the `a` or `b` markers remain in generated output.

## Related links

None.

## Checklist

## Validation

- `git diff --cached --check`
- `source ~/.nvm/nvm.sh && nvm use 22 >/dev/null && cd tests/rspack-test && pnpm run test -t "esmOutputCases/basic/pure-functions-expression-nesting"`
- `source ~/.nvm/nvm.sh && nvm use 22 >/dev/null && cd tests/rspack-test && npx rstest esm`
- `CARGO_BUILD_JOBS=1 cargo check -p rspack_plugin_javascript`
- `CARGO_BUILD_JOBS=1 cargo fmt --all --check`


JSerFeng:
@codex ui

JSerFeng:
@codex review

JSerFeng:
see #14375

PR #14296, opened 2026-06-05
feat(cache): add maxAge and maxGenerations for persistent cache cleanup

### LINKED ISSUE #14275: [Feature]: Add retention limits for persistent filesystem cache

Rspack's persistent filesystem cache can grow without a practical short-term bound when cache versions change frequently.

The filesystem storage currently hard-codes a seven-day expiry:

https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/cache/persistent/storage/mod.rs

The JavaScript storage configuration exposes only `type` and `directory`:

https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/config/types.ts

This means every cache version created during those seven days is retained, regardless of total byte size or number of versions. Expiration only helps after a version has aged past the fixed TTL.

In a large multi-compiler Rsbuild application, we observed:

- approximately 5.17 GB per cache generation across browser, server, and worker compilers
- five generations created in less than 14 hours
- 23.8 GB retained in the persistent cache directory
- all generations were too recent for the hard-coded seven-day cleanup

The cache entries were valid serialized module-graph data rather than one malformed file. Configuration iteration alone was enough to create new version directories.

Related issues:

- #9745 requested configurable expiry, but was closed for missing context and the option was not implemented.
- #9405 / #9517 added `maxGenerations` for the in-memory cache only, not persistent filesystem storage.

A configurable age limit would expose behavior that already exists internally. A version-count bound would prevent multiple large same-day generations from consuming unbounded disk space. A byte-size bound would provide the strongest guarantee when generation sizes vary significantly.

For example:

```ts
const config = {
  cache: {
    type: 'persistent',
    storage: {
      type: 'filesystem',
      directory: './node_modules/.cache/rspack',
      maxAge: 24 * 60 * 60 * 1000,
      maxVersions: 2,
      maxSize: 10 * 1024 * 1024 * 1024,
    },
  },
};
```

The minimum useful implementation could expose:

- `maxAge`: configure the existing internal expiry instead of hard-coding seven days.
- `maxVersions`: retain the newest/accessed N versions using the access timestamps already stored in `_meta`.

`maxSize` could be added either at the same time or separately if calculating directory sizes during cleanup is considered too expensive. `maxVersions` alone would solve the common configuration-churn case without requiring a full directory-size scan.

The exact option names and whether these live on `cache` or `cache.storage` are flexible; the important behavior is allowing applications to place a deterministic bound on persistent-cache retention.

matthewdavis-oai:
## Summary

- add `cache.storage.maxVersions` for persistent filesystem caches, with positive integer validation across the JavaScript and Rust APIs
- prefix retained cache versions with a portable-aware compiler scope so multi-compiler builds sharing a directory do not count each other's versions
- extend the existing two-column `_meta` cleanup path to remove least-recently-accessed versions after a successful DB save
- use physical version directories as the inventory so cleanup recovers from concurrent compilers overwriting shared `_meta` updates
- keep retention disabled at zero runtime cost when `maxVersions` is not configured, with no lock, heartbeat, or second metadata index
- document the option in English and Chinese and add validation, portability, compiler-isolation, strict-limit, and concurrent-metadata tests

## Related links

Closes #14275.

## Validation

- `pnpm run build:cli:dev`
- `pnpm run test -t "cacheCases/portable/basic"` in `tests/rspack-test`
- `pnpm run test -t "compilerCases/persistent-cache-max-versions"` in `tests/rspack-test`
- `cargo test -p rspack_storage --lib -- --skip filesystem::db::transaction::lock::state::tests::test_state_lock`
- `cargo test -p rspack_core retention_scope_uses_typed_compiler_identity --lib`
- `cargo test -p rspack_binding_api rejects_zero_max_versions --lib`
- `cargo clippy -p rspack_storage -p rspack_core -p rspack_binding_api --lib -- -D warnings`
- `pnpm run lint:js`
- `cargo fmt --all --check`
- `git diff --check`

`test_state_lock` shells out to `ps`, which is unavailable in the sandbox used for local validation. All other `rspack_storage` tests passed.

## Checklist


chenjiahan:
Thank you! @ahabhgk please take a look

ahabhgk:
Generally looks good, I will change some code and make some adjust on this PR, then we are ready to merge

ahabhgk:
@codex review

ahabhgk:
@codex review

ahabhgk:
@codex review

ahabhgk:
@codex review

ahabhgk (dismissed):

ahabhgk:
@matthewdavis-oai Here are some changes I made:

1. This configuration item is now called `maxGenerations` instead of `maxVersions`.
2. Expiration time can also be configured through `maxAge`.
3. Cache cleanup isolation between different compilers is no longer achieved through `retention_scope`. Now, it's uniformly isolated using `storage.directory`. Different compilers should be configured with different `storage.directory` to isolate the cache. rspack will by default configure different `storage.directory` based on `compiler.name + mode` (and also `compilerIndex` for multi-compiler configurations). So the recommended isolation method is to configure different [`name`](https://rspack.rs/config/other-options#name) for different compiler.

ahabhgk (approved):

website/docs/en/config/cache.mdx
```diff
@@ -214,30 +216,75 @@ Specifies paths within `snapshot.managedPaths` that should not be treated as man
...
+    },
+  },
+};
+```
+
+#### cache.storage.maxGenerations
+
+- **Type:** `number` (positive integer)
+
+- **Default:** No generation count limit; `maxAge` cleanup still applies
```
chenjiahan, line 272: @ahabhgk should we set a default maxGenerations?

ahabhgk: Nope for now, we can adjust the default in the future

PR #14311, opened 2026-06-08
fix(watcher): exclude ignored subtrees from watch events, not just registration

stormslowly:
## Why

`ignored` patterns were only consulted at **registration** time, so filesystem events for paths inside an ignored directory still flowed into the watcher and got dispatched.

On macOS/Windows the watcher subscribes to a single **recursive root**, so the OS delivers events for ignored subtrees too (e.g. the consumer's build output dir `dist/.rstest-temp/`). Such a path isn't a registered file, so `find_associated_event` bubbles up to the nearest registered ancestor directory and emits a `[dir](Change)` batch that is **indistinguishable from a real source change** — a noise rebuild that races with the user's intended write and flakes watch-mode tests (e.g. rstest).

### Before / After — event for `dist/.rstest-temp/foo.mjs` (ignored)

| | Result |
|---|---|
| Before | dispatched as `[<registered ancestor dir>](Change)` → noise rebuild |
| After  | dropped at ingress → no event |

## What

- Add `FsWatcherIgnored::matches_with_ancestors` — walks the path and every ancestor, because directory-style globs like `**/dist/.rstest-temp` match the directory entry but **not** the files inside it.
- Apply it at the event ingress (`Trigger::on_event`) via `PathManager::is_ignored_path`, so ignored paths never produce dispatched events.
- Reuse the same predicate at registration, closing the identical directory-style-glob gap there too.

Single source of truth for "is this path in ignored scope", shared by registration and the event ingress. Mirrors watchpack/webpack, where the build output dir is excluded from the watched tree end-to-end.

## Test

- `cargo test -p rspack_watcher` — 16 unit + 1 integration, all green.
- New tests: ancestor-walk catches files inside a matched dir / keeps unrelated paths / `None` short-circuits; `on_event` drops events inside an ignored subtree.


stormslowly:
@codex code review

PR #14324, opened 2026-06-08
chore: bind walltime bench to dedicated CPUs

hardfist:
## Summary

Bind the CodSpeed walltime Rust benchmark run commands to CPUs `300-315` using `taskset`.

## Why

This keeps the walltime benchmark workload on a dedicated CPU range for more controlled measurements on the self-hosted runner.

## Impact

Only the `walltime` CodSpeed benchmark path in `.github/workflows/bench-rust.yml` is affected. Simulation benchmarks are unchanged.

## Validation

- Commit hooks ran lint-staged/prettier for the workflow file during commit.
- Not run: full CI locally, because this is a GitHub Actions workflow-only change.

### DIFF
--- a/.github/workflows/bench-rust.yml
+++ b/.github/workflows/bench-rust.yml
@@ -122,9 +122,9 @@ jobs:
               if [ -n "${{ inputs.bench_filter }}" ]; then
                 bench_args+=("${{ inputs.bench_filter }}")
               fi
-              RSPACK_BENCHCASES_DIR=$PWD/.bench/rspack-benchcases cargo codspeed run "${bench_args[@]}"
+              taskset -c 300-315 env RSPACK_BENCHCASES_DIR=$PWD/.bench/rspack-benchcases cargo codspeed run "${bench_args[@]}"
             else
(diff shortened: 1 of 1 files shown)


.github/workflows/bench-rust.yml
```diff
@@ -122,9 +122,9 @@ jobs:
               if [ -n "${{ inputs.bench_filter }}" ]; then
                 bench_args+=("${{ inputs.bench_filter }}")
               fi
-              RSPACK_BENCHCASES_DIR=$PWD/.bench/rspack-benchcases cargo codspeed run "${bench_args[@]}"
+              taskset -c 300-315 env RSPACK_BENCHCASES_DIR=$PWD/.bench/rspack-benchcases cargo codspeed run "${bench_args[@]}"
```
deanjingshui, line 125: 1. 跑bench前加一行命令：sync; echo 1 > /proc/sys/vm/drop_caches
2. 我把逻辑核关掉了，这里锁核的2行命令改一下：taskset -c 176-191 <你的命令>

PR #14325, opened 2026-06-08
perf(core): avoid parsing concatenated module refs

nnnnoel:
## Summary

- Store concatenation module reference options in `ConcatenatedModuleInfo` when references are created.
- Replace hot-path JSON/hex module reference parsing with a direct lookup during concatenated module code generation.
- Keep the existing `ConcatenationScope.refs` data for ESM library linking compatibility.
- Add `.codegraph/` to `.gitignore`.

## Module reference format

Before this change, `create_module_reference` encoded the target export information into the generated identifier itself:

```text
__rspack_module_ref{module_index}_{hex(json(ids))}[_call][_directImport][_deferredImport][_asiSafe0|_asiSafe1]__._
```

The replacement pass later had to recover the same `ModuleReferenceOptions` by parsing the identifier, hex-decoding the export payload, JSON-deserializing `ids`, and then parsing each flag suffix.

After this change, the generated reference only carries a compact identity:

```text
__rspack_module_ref{module_index}_{local_ref_index}__._
```

The actual `ModuleReferenceOptions` are stored structurally when the reference is created:

- `current_module.module_references` stores the options by the generated identifier without the trailing `._`.
- `ConcatenationScope.refs` is still populated with the full generated reference string for the existing ESM library linking path.

So `ids`, `call`, `direct_import`, `deferred_import`, `asi_safe`, and the resolved target module index are no longer serialized into the identifier. They remain in `ModuleReferenceOptions` and are read back directly during concatenated module code generation.

## Algorithmic rationale

The previous algorithm treated the generated identifier as both an identity and a serialized payload. That made every replacement in `concatenate_module_code_generation` pay the cost of reconstructing data that was already known at reference creation time:

1. parse the module reference string,
2. decode hex,
3. deserialize JSON into a fresh `ids` vector,
4. parse boolean flag suffixes,
5. convert parsed strings back into `Atom`s before calling `get_final_name`.

This is unnecessary work on the hot path. The new algorithm separates identity from metadata:

1. `create_module_reference` assigns a per-current-module `local_ref_index` and emits a short stable identifier.
2. The full `ModuleReferenceOptions` are stored once in the current `ConcatenatedModuleInfo`.
3. The replacement pass performs a direct lookup from the generated symbol to the stored options.
4. `ids.clone()` happens only once, immediately before `get_final_name`, because `get_final_name` still owns the export-name vector.

The lookup is safe because these module reference symbols are generated by `ConcatenationScope::create_module_reference`; they are not user-authored identifiers. The existing `refs` map remains unchanged at the API level for the ESM library linker, so this avoids changing the cross-plugin contract while removing JSON/hex parsing from the core replacement path.

Benchmark (`BENCH_MODE=walltime cargo codspeed run -m walltime -p rspack_benchmark --bench benches rust@concatenate_module_code_generation`):

- Baseline: `9.4164 ms`
- After run 1: `9.2599 ms`
- After run 2: `9.2028 ms`

## Related links

N/A

## Checklist

Verification:

- `cargo test -p rspack_core concatenation_scope`
- `cargo test -p rspack_plugin_esm_library`
- `pnpm run build:js`
- `pnpm run test -t "concatenate"`
- `pnpm run test -t "scope-hoisting"`
- `pnpm run test -t "esmOutputCases/dynamic-import"`
- `pnpm run test -t "esmOutputCases/externals"`
- `pnpm run test -t "defer-import"`
- `cargo fmt --all --check`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -198,5 +198,8 @@ rstack-examples/
 .aider
 .codex
 
+# ignore known coding support frameworks
+.codegraph/
+
 # rstest event report files
 /*-report.txt
--- a/crates/rspack_core/src/concatenated_module.rs
+++ b/crates/rspack_core/src/concatenated_module.rs
@@ -1441,7 +1441,7 @@ impl Module for ConcatenatedModule {
         let mut refs = vec![];
         for reference in info.global_scope_ident.iter() {
           let name = &reference.id.sym;
-          if !ConcatenationScope::is_module_reference(name.as_str()) {
+          if !ConcatenationScope::has_module_reference_prefix(name.as_str()) {
             continue;
           }
           if let Some(match_info) = info.module_references.get(name) {
@@ -1452,6 +1452,19 @@ impl Module for ConcatenatedModule {
               match_info,
               build_meta.strict_esm_module,
             ));
+          } else if let Some(stripped_name) =
+            ConcatenationScope::strip_module_reference_property_access_suffix(name.as_str())
+          {
+            let stripped_name = Atom::from(stripped_name);
+            if let Some(match_info) = info.module_references.get(&stripped_name) {
+              let referenced_info_id = &references_info[match_info.index].0;
+              refs.push((
+                reference.clone(),
+                referenced_info_id,
+                match_info,
+                build_meta.strict_esm_module,
+              ));
+            }
           }
         }
 
--- a/crates/rspack_core/src/concatenated_module.rs
+++ b/crates/rspack_core/src/concatenated_module.rs
@@ -50,11 +50,12 @@ use crate::{
   ExportsArgument, ExportsInfoArtifact, ExportsType, FactoryMeta, ImportedByDeferModulesArtifact,
   InitFragment, InitFragmentStage, LibIdentOptions, Module, ModuleArgument,
   ModuleCodeGenerationContext, ModuleGraph, ModuleGraphCacheArtifact, ModuleGraphConnection,
-  ModuleIdentifier, ModuleLayer, ModuleStaticCache, ModuleType, NAMESPACE_OBJECT_EXPORT,
-  ParserOptions, Resolve, RuntimeCondition, RuntimeGlobals, RuntimeSpec, SideEffectsStateArtifact,
-  SourceType, URLStaticMode, UsageState, UsedName, UsedNameItem, escape_identifier, fast_set,
-  filter_runtime, find_target, get_runtime_key, impl_source_map_config, merge_runtime_condition,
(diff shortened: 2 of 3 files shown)


nnnnoel:
I took another look at the reported regressions because the changed hot path should reduce work rather than add enough work to explain a regression of this size.

I added temporary instrumentation around the module reference lookup path and checked whether the suffix-strip fallback path is actually exercised. In the relevant benchmark paths, it was not:

```text
rust@concatenate_module_code_generation
prefix_skips=1,112
direct_hits=4,270,080
fallback_candidates=0
fallback_hits=0
lookup_misses=0

bundle@threejs-10x-production-sourcemap
prefix_skips=2,849,815
direct_hits=2,450,650
fallback_candidates=0
fallback_hits=0
lookup_misses=0

rust@create_full_hash
prefix_skips=1
direct_hits=3,840
fallback_candidates=0
fallback_hits=0
lookup_misses=0
```

`bundle@threejs-10x-development` did not emit any module-reference lookup stats in this instrumentation run, so it does not appear to exercise this lookup path in the measured execution.

So the suffix-strip fallback is not on the regressed path. The hot path is still either a fast prefix reject for normal identifiers or a direct `module_references` hit for generated module reference identifiers.

Given that the CodSpeed report also compares different CPU environments (`AMD EPYC 9V74 80-Core Processor` vs `AMD EPYC 7763 64-Core Processor`), I do not think the small unrelated regressions are causally explained by this PR. The directly targeted benchmark, `rust@concatenate_module_code_generation`, is the one that exercises the changed path and shows the expected direction.

hardfist:
just ignore walltime bench which is flaky

nnnnoel:
[rust@create_chunk_hashes](https://app.codspeed.io/web-infra-dev/rspack/branches/nnnnoel%3Aperf%2Fconcatenate-module-codegen-lookup?uri=xtask%2Fbenchmark%2Fstages%2Fcreate_chunk_hashes.rs%3A%3Astage%3A%3Abench%3A%3Arust%40create_chunk_hashes&runnerMode=Simulation&utm_source=github&utm_medium=comment-v2&utm_content=benchmark) in local machine

base: 458.35 µs ~ 460.89 µs
head: 458.46 µs ~ 468.88 µs
change: [-1.6750% -0.4635% +0.7250%], p = 0.45

nnnnoel:
@codspeedbot explain why this is slower

nnnnoel:
@codspeedbot could you re-run the benchmark?

nnnnoel:
I'll find out another way.

PR #14327, opened 2026-06-08
fix: correct mf manifest expose assets

2heal1:
## Summary

Fix Module Federation manifest expose asset collection so each expose uses its own entry chunk instead of falling back to the first chunk in the async block group.

Add a regression case that keeps multiple expose entries separated while still allowing shared sync assets to appear in both exposes.

## Root cause

The manifest plugin previously recorded the first named chunk from each expose chunk group. In complex chunk groups, that first chunk can belong to another expose or a shared chunk, so the manifest can assign unrelated expose entry files to the current expose.

## Validation

- `cargo fmt -p rspack_plugin_mf`
- `cargo check -p rspack_plugin_mf`
- `pnpm --filter @rspack/test-tools build`
- `pnpm run build:js`
- `env testFilter=/Users/bytedance/outter/rspack/tests/rspack-test/configCases/container-1-5/manifest-multiple-exposes pnpm --dir tests/rspack-test test:base Config.part1.test.js`
- Verified the reported app manifest no longer includes `__federation_expose_AdgroupEditModule` under `CreativeEditModule` after rebuilding with the local Rspack package.

### DIFF
--- a/.github/actions/artifact/download/action.yml
+++ b/.github/actions/artifact/download/action.yml
@@ -18,7 +18,7 @@ runs:
   using: composite
   steps:
     - name: Download artifact from github
-      uses: actions/download-artifact@65a9edc5881444af0b9093a5e628f2fe47ea3b2e # v4.1.7
+      uses: actions/download-artifact@v4.1.7
       if: ${{ runner.environment == 'github-hosted' || inputs.force-use-github == 'true' }}
       with:
         name: ${{ inputs.name }}
--- a/.github/actions/artifact/download/action.yml
+++ b/.github/actions/artifact/download/action.yml
@@ -18,7 +18,7 @@ runs:
   using: composite
   steps:
     - name: Download artifact from github
-      uses: actions/download-artifact@v4.1.7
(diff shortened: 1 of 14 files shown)


2heal1:
CI note: I checked the branch against the latest `origin/v1.x`; this is not caused by the PR branch being outdated.

The latest `v1.x` still uses version-tagged references in the shared artifact composite actions:

- `actions/upload-artifact@v4`
- `actions/download-artifact@v4.1.7`

Current CI rejects those unpinned references and fails during artifact upload/download. I previously touched a wider set of workflow files, but that broader change has been reverted. The remaining `.github` changes only pin the two shared artifact actions that are actually used by this PR's CI.

Keeping this in the PR for now so CI can pass on `v1.x`. If `v1.x` fixes these shared artifact actions separately, we can drop these two lines from this PR later.

2heal1:
Main cherry-pick PR: #14339

2heal1:
@codex  review

ahabhgk (approved):

PR #14329, opened 2026-06-09
fix(watcher): filter ignored-subtree events that emit a spurious folder change

stormslowly:
## Why

On the macOS/Windows **recursive-root** watch the OS delivers an event for every path under the project — mostly deep `node_modules` / build output nobody registered. Such an event isn't a registered file, so `find_associated_event` bubbles it up to a registered parent and emits a `[dir](Change)` indistinguishable from a real source change — a spurious rebuild that flakes watch-mode tests. And `ignored` was only checked at registration, so these events still reached `Trigger::on_event` unfiltered.

## What

Filter ignored events the way watchpack does, cheaply enough to run per-event:

- Port watchpack's `glob-to-regexp` (webpack/watchpack#312) faithfully, then append `(?:$|/)` so a directory match also covers its subtree.
- Fold all patterns into one precompiled `regex::Regex` and classify each event with a single `is_match` — `O(1)` per event, no ancestor walk. Compiled once per watcher, reused across the rebuild loop.
- Apply the filter at event ingress *and* at registration (absolutized first, so both test the same absolute paths events report).

Matching normalizes `\`→`/`, so it is separator-agnostic on Windows. watchpack's `ignored`-option test cases are ported alongside.


stormslowly:
@codspeedbot explain why regression

hardfist (approved):

Issue #11913, opened 2025-10-17
[Feature]: generage css hash tail from file path


denisx:
### What problem does this feature solve?

Is It possible generage css hash tail from file path? It will help reduce gz/br version of files, and make less css names length.

Article https://dev.to/denisx/reduce-bundle-size-via-one-letter-css-classname-hash-strategy-10g6

webpack pr https://github.com/webpack/css-loader/pull/1181, https://github.com/webpack/css-loader/pull/1410

Mentions https://x.com/markdalgleish/status/1385545921186828290

### What does the proposed API of configuration look like?

localIdentName / hashStrategy: "minimal-subset"

PR #14334, opened 2026-06-09
fix(css): include local name in hash-only local idents

### LINKED ISSUE #14333: [Bug]: #11913 changes css module [hash] behavior
### System Info

-

### Details

This is a breaking change for css module users and should not be introduced silently in a minor release.

### Reproduce link

[repro](https://github.com/Austaras/repro-css-module)

### Reproduce Steps

[hash] used to be decided by export name, but now it's only decided by file path while [fullhash] remains the same.

intellild:
## Summary

Fix CSS module local ident hashing so `[hash]` uses a per-local hash when `localIdentName` does not include `[local]`. This aligns the hash input with css-loader behavior and prevents different local names in the same module from sharing the same hash-only class name.

Updated the existing CSS local ident snapshots, including the builtins hash case, to cover the corrected output.

## Related links

- Fixes #14333
- See also #11913 

## Checklist

## Validation

- `env CI=true pnpm run build:binding:dev`
- `pnpm --filter @rspack/tests test:base -- -t "css/local-ident-name|builtins/css-modules-local-ident-name-hash"`

---
_by OpenAI Codex_


Austaras:
> This aligns the hash input with css-loader behavior

This seems good

ahabhgk (approved):

PR #14335, opened 2026-06-09
perf: replace scope parent-chain lookup with scoped symbol table

hardfist:
### What changed

`ScopeInfoDB` in the JavaScript parser previously kept one `FxHashMap<Atom, VariableInfoId>` per scope and resolved a variable by probing the map of every scope along the parent chain, inserting tombstones into the starting scope as a negative cache on misses.

Since the parser enters and exits scopes in strict stack order and always reads/writes through the innermost active scope, this PR replaces the per-scope maps with a classic scoped symbol table:

- a single global `FxHashMap<Atom, SmallVec<[Binding; 2]>>` mapping each name to a stack of bindings, innermost last
- lookup is one hash probe (`stack.last()`) instead of one probe per enclosing scope
- each scope records the names it binds; `exit_scope` (called at the three scope-restore sites in `walk.rs`) pops them
- negative-cache tombstone insertion on lookup misses is removed entirely, as misses are already O(1)
- `debug_assert`s enforce the stack-discipline invariant

### Why it should improve performance

Variable resolution is on the parse hot path and scales with identifier occurrences (dependency-level cardinality). On the `cases/all` benchmark, callgrind showed ~2.9M lookups performing 8.2M per-scope hash probes plus tombstone-insert churn, ~2.1% of all executed instructions. With the binding-stack table each lookup costs a single probe regardless of scope depth.

### Measured effect (callgrind, cases/all)

- total executed instructions: 39.15B -> 38.46B (-1.75%)
- scope lookup cost: ~824M Ir (2.1%) -> ~500M Ir (1.3%); the former top hotspot `HashMap<Atom, VariableInfoId>::get` no longer appears in the profile

## Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/parser/mod.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/parser/mod.rs
@@ -909,9 +909,8 @@ impl<'parser> JavascriptParser<'parser> {
 
   pub fn get_all_variables_from_current_scope(
     &self,
-  ) -> impl Iterator<Item = (&str, &VariableInfoId)> {
-    let scope = self.definitions_db.expect_get_scope(self.definitions);
-    scope.variables()
+  ) -> impl Iterator<Item = (&str, VariableInfoId)> {
+    self.definitions_db.scope_variables(self.definitions)
   }
 
   pub fn define_variable(&mut self, name: Atom) {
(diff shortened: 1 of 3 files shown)


hardfist:

CPunisher (approved):

PR #14337, opened 2026-06-10
perf: verify rstackjs/rspack-resolver#275 (std::fs in FileSystemOs) via CodSpeed

stormslowly:
## Why

Verification PR for rstackjs/rspack-resolver#275, which replaces `tokio::fs` with `std::fs` in `FileSystemOs` and showed **-22.9% aggregate estimated cycles** on the resolver's own benches.

This PR patches `rspack_resolver` to that branch's head (`b9eb9b8`) via `[patch.crates-io]` so rspack's CodSpeed benches measure the end-to-end impact inside rspack builds.

**Not intended to be merged** — once verified, the resolver PR should land and be released, then rspack bumps the version normally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,9 +176,9 @@ checksum = "dde20b3d026af13f561bdd0f15edf01fc734f0dafcedbaf42bba506a9517f223"
 
 [[package]]
 name = "arc-swap"
-version = "1.8.2"
+version = "1.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f9f3647c145568cec02c42054e07bdf9a5a698e15b466fb2341bfc393cd24aa5"
+checksum = "6a3a1fd6f75306b68087b831f025c712524bcb19aad54e557b1129cfa0a2b207"
 dependencies = [
  "rustversion",
 ]
@@ -602,7 +602,7 @@ dependencies = [
  "semver",
  "serde",
  "serde_json",
- "thiserror 2.0.17",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


stormslowly:
## Verification result: no measurable impact on rspack builds

Local callgrind (Linux x86_64, CodSpeed simulation flags, `RAYON_NUM_THREADS=1`) comparing base `2ab2da29fa` vs this branch across all `benches` scenarios: **+0.17% total estimated cycles** — flat within noise. Largest resolver-heavy case `build_module_graph`: +0.40%. CodSpeed CI agrees ("will not alter performance"; the +5.3% on `js@external getResolve` is the one path that does benefit, see below).

### Why the resolver-side -23% does not transfer

rspack never instantiates the resolver's default `FileSystemOs` for builds — it injects its own filesystem via `ResolverGeneric::new_with_file_system(BoxFS)` (`crates/rspack_core/src/resolver/boxfs.rs`), so rstackjs/rspack-resolver#275's change is dead code on the hot build path.

`FileSystemOs` IS used by:
- the standalone napi resolver API (`crates/rspack_binding_api/src/rspack_resolver/mod.rs`) — matches the `js@external getResolve` improvement
- the persistent-cache build-deps helper

Additionally, callgrind shows resolver fs machinery (`BoxFS` `meta`/`realpath` futures) at only ~0.02–0.04% of instructions in rspack's Rust benches, so even porting the same std::fs change into `rspack_fs::NativeFileSystem` would not move these benchmarks.

**Conclusion:** rstackjs/rspack-resolver#275 is worth landing for resolver-standalone users and the napi resolve API, but brings no measurable gain to rspack core builds. Closing this verification PR is fine once recorded.

PR #14354, opened 2026-06-11
perf: avoid double UTF-8 scan when relativizing identifier paths

stormslowly:
## Why

`make_paths_relative` / `make_paths_absolute` convert `PathBuf` results back to strings via `to_string_lossy()`, which validates UTF-8 **and** walks the bytes in chunks — even though these paths originate from valid UTF-8 `&str` inputs. Callgrind shows the lossy chunk walk plus revalidation (`Utf8Chunks::next` + `from_utf8`) at ~11% of the `create_named_module_ids` stage benchmark, with similar shares in `create_module_ids`.

This replaces the lossy conversions with a single-validation `into_os_string().into_string()` (keeping a lossy fallback for non-UTF-8 paths), which is behaviorally identical.

Expected: improvement on the module-id stage benchmarks; CodSpeed will verify.


stormslowly:
@codex code review

PR #14355, opened 2026-06-11
perf: add ASCII fast path for dependency location column computation

stormslowly:
### Why

`DependencyLocationAdvancer::advance_pos` computes UTF-16 column counts with `encode_utf16().count()` on every segment, decoding every char, even though the vast majority of source slices are pure ASCII (callgrind: ~1.7% of `build_module_graph`, ~1.9% of `scan_dependencies`). For ASCII the UTF-16 code-unit count equals the byte length, and `is_ascii()` is a SIMD-accelerated byte scan, so this adds an ASCII fast path to both column computations. This matters most for minified one-line vendor sources where segments are long and pure ASCII. Behavior is unchanged for non-ASCII input.

Expectation: small but broad improvement across parse-heavy benchmarks; CodSpeed will verify.


stormslowly:
@codex review

hardfist (approved):

PR #14357, opened 2026-06-11
perf: verify rstackjs/rspack-resolver#278 (reuse parent realpath buffer) via CodSpeed

stormslowly:
## Why

Verification PR for rstackjs/rspack-resolver#278, which makes `realpath_uncached`'s non-symlink branch reuse the owned parent realpath buffer (`push` the final segment) instead of allocating a fresh path via `strip_prefix` + `normalize_with`. This path is hot during node_modules resolution with `symlinks: true` (the default).

This PR patches `rspack_resolver` to rstackjs/rspack-resolver@99becacf via `[patch.crates-io]` to measure the impact on rspack benchmarks through CodSpeed. Not intended to be merged.

Same methodology as #14337.

## Results (run 6a2a70cf vs main@ed6f627)

CodSpeed headline is "+3%", but the two flagged improvements (`js@collect imported identifiers` +3.07%, `sources@cached_source_hash` +2.94%) do not touch the resolver — treat them as noise.

The benchmarks that actually exercise resolution moved in the right direction but stayed below the significance threshold (Simulation mode, instruction counts):

| Benchmark | BASE | HEAD | Delta |
| --- | --- | --- | --- |
| `bundle@basic-react-development` | 288 ms | 282.5 ms | -1.9% |
| `rust@build_module_graph` | 875.7 ms | 872.9 ms | -0.3% |
| `bundle@threejs-production-sourcemap` | 343.2 ms | 342.5 ms | -0.2% |
| `bundle@basic-react-production-sourcemap` | 250.1 ms | 249.6 ms | -0.2% |
| `bundle@threejs-development` | 229.9 ms | 229.5 ms | -0.2% |

Conclusion: the resolver-level win (several percent inside `resolve`, per the callgrind data on rstackjs/rspack-resolver#278) dilutes to ~0.2–2% on whole-build rspack benchmarks — consistently positive, but within CodSpeed's unchanged band. The direct measurement remains the resolver repo's own CodSpeed run on #278.


stormslowly:

PR #13864, opened 2026-04-28
feat: support source phase import for WebAssembly modules

magic-akari:
**Summary**

This PR implements support for the TC39 [source-phase-imports proposal](https://github.com/tc39/proposal-source-phase-imports) for WebAssembly modules, enabling developers to import WebAssembly modules at the source phase using `import.source()` syntax.

**Motivation:**

Source phase imports allow importing a WebAssembly module as a `WebAssembly.Module` object rather than an instantiated module. This enables several important use cases:

- **Multiple instantiation**: The same compiled WASM module can be instantiated multiple times with different imports
- **Worker sharing**: Compiled modules can be efficiently shared across Web Workers using `postMessage`

**Example usage:**

```javascript
// Import as compiled WebAssembly.Module
const wasmModule = await import.source("./module.wasm");

// Instantiate multiple times with different imports
const instance1 = await WebAssembly.instantiate(wasmModule, imports1);
const instance2 = await WebAssembly.instantiate(wasmModule, imports2);

// Share with workers
worker.postMessage(wasmModule);
```

**Related:**

- TC39 Proposal: https://github.com/tc39/proposal-source-phase-imports

## Related links

- webpack/webpack#20364

## Checklist

- [x] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #14364, opened 2026-06-11
fix: remove webpackSource and webpackDefer magic comments

ahabhgk:
## Summary

In https://github.com/web-infra-dev/rspack/pull/13864 we add `/* webpackSource: true */` and `/* webpackDefer: true */` magic comments support. However, after further consideration, we have decided to remove support for these two magic comments. Users should directly use `import.source()` and `import.defer()` instead of magic comments.

## Checklist


chenjiahan (commented):
Can we provide more context in the PR description?

chenjiahan (approved):

PR #14366, opened 2026-06-11
fix: report diagnostics instead of panicking on unknown context modes

dr4gan0x:
## Summary

Passing an invalid `mode` to `import.meta.webpackContext` currently panics and takes the whole build down:

```js
const ctx = import.meta.webpackContext("./dir", { mode: "lazi" }); // panic: unknown context mode
```

`require.context("./dir", false, /\.js$/, "lazi")` has the opposite problem: it silently falls back to `sync`, so a simple typo eagerly bundles the whole directory without any hint that something went wrong. Same story for `import(/* webpackMode: "lazi" */ "./a")`, which quietly turns into `lazy` while the other magic comments (`webpackFetchPriority` etc.) already warn on invalid values. There were a couple of `TODO should give warning` comments sitting next to this code for a while, this PR picks them up.

What changed:

- invalid `webpackMode` values now produce the usual "Magic comments parse failed" warning and fall back to lazy, same shape as the existing `webpackFetchPriority` validation and same outcome as webpack ([ImportParserPlugin](https://github.com/webpack/webpack/blob/main/lib/dependencies/ImportParserPlugin.js))
- invalid modes passed to `require.context` / `import.meta.webpackContext` now emit a traceable error pointing at the offending span, matching the `Unsupported mode "..." in context` error webpack throws from [ContextModule](https://github.com/webpack/webpack/blob/main/lib/ContextModule.js) during code generation
- the panicking `From<&str> for ContextMode` impl is gone; `try_convert_str_to_context_mode` is now the single string entry point and callers are responsible for diagnostics

Since webpack fails the build for unknown context modes I kept those as errors, but happy to downgrade them to warnings if that's preferred.

Tests: a unit test for the magic comment parsing, two new diagnostics cases, and a regenerated `magic_comment_warning` snapshot — that fixture already contained `webpackMode: "unknown"`, it just never warned before.

## Related links

- webpack's `webpackMode` validation: https://github.com/webpack/webpack/blob/main/lib/dependencies/ImportParserPlugin.js
- webpack's `Unsupported mode` error: https://github.com/webpack/webpack/blob/main/lib/ContextModule.js

## Checklist


crates/rspack_plugin_javascript/src/magic_comment.rs
```diff
@@ -1042,6 +1044,19 @@ mod tests_extract_magic_comment_object {
...
+
+    assert_eq!(comments.get_mode(), None);
+    assert_eq!(warnings.len(), 1);
+    assert!(
+      warnings[0]
+        .message
+        .contains(r#"`webpackMode` expected "lazy", "lazy-once", "eager" or "weak""#)
+    );
+  }
+
```
intellild: Please remove this test, `rspack-test` already cover the case

intellild (changes requested):

intellild:
Thank you for contribution. The main branch implementation has changed a lot so I port the branch to #15361

PR #14368, opened 2026-06-11
feat: support MF SSR remote cache clearing

2heal1:
## Summary

Add Module Federation cache clearing for Node SSR. After a remote is updated, the MF runtime can invalidate the remote and the host modules whose cached exports would otherwise keep serving the old remote, without restarting the SSR process.

This PR requires the matching `@module-federation/runtime-tools` pre-release because that package owns the public `clearCache` / forced remote registration flow.

### Compile-time dependency metadata

`RemoteRuntimeModule` now emits reverse dependency metadata alongside the existing remote-loading data:

- `remoteKeyToRemoteModuleIds`: remote key -> remote module IDs.
- `remoteKeyToExternalModuleIds`: remote key -> external/container-entry module IDs.
- `remoteModuleIdToConsumerModuleIds`: remote module -> direct host consumers.
- `consumerModuleIdToParentModuleIds`: consumer -> its direct importers.
- `remoteKeyToChunkIds`: remote key -> chunks that load it.

The metadata is generated from the compilation `ModuleGraph` and `ChunkGraph`, rather than from HMR state:

1. For each `RemoteModule`, obtain its `remote_key`, module ID, and external dependency module ID.
2. Read the remote module's incoming connections. Each connection's `original_module_identifier` is a direct consumer.
3. For every direct consumer, read its incoming connections to obtain parents/importers.
4. Include every referenced chunk, so async consumers and their remote chunks are represented.
5. Sort and deduplicate every generated ID list for stable runtime metadata.

Using the normal module graph means this works in production SSR builds where HMR's `module.parents` is unavailable.

### Runtime invalidation

The runtime-tools cache-clear flow uses those indexes to identify the smallest invalidation set:

1. Resolve the requested remote to its remote key.
2. Invalidate its remote and external module state.
3. Find direct consumers from `remoteModuleIdToConsumerModuleIds`.
4. Walk `consumerModuleIdToParentModuleIds` upward, invalidating cached host-module exports so a later SSR request re-executes the consumer chain and reloads the remote.
5. Leave module factories intact for local consumers; only remove their executed module-cache entries.

This is necessary because clearing only the remote module is insufficient: a previously executed page/route module can remain in `__webpack_require__.c` and return old exports without touching the remote loading path.

### Node chunk safety

Node `require` and `readFileVm` chunk runtimes expose controlled `chunkCacheControls`:

- target-specific clear/snapshot/restore operations for `installedChunks`;
- CommonJS `require.cache` cleanup for affected filesystem chunks;
- per-chunk generations.

A clear invalidates the target chunk generation before clearing it. An older async chunk load checks that generation before installing factories or marking the chunk loaded, so it cannot repopulate the cache after a newer request has begun.

### Provider-side cleanup

Container entries expose `__webpack_clear_cache__`, which clears the provider bundle's module cache. This ensures a reloaded remote does not reuse its own cached exposed-module exports.

## Behavior and limits

- Existing in-flight SSR requests are allowed to complete.
- The operation invalidates old state; it does not preload or validate the replacement remote.
- Shared instances are retained by default to avoid splitting process-wide singletons.
- For native ESM remotes, callers should use versioned entry URLs because Node cannot reliably evict a same-URL ESM module from its native cache.

## Tests

- Runtime metadata covers remote, external module, consumer, parent, and async chunk mappings.
- SSR cache clearing covers multiple exposes and consumer routes.
- Covers force-registering a replacement remote entry.
- Covers pending/rejected remote `get`, stale async chunk completion, repeated clearing, CommonJS cache cleanup, large payload release, and shared-module retention.
- Verified locally with:
  - `pnpm run build:cli:dev:wasm`
  - `pnpm run test -t "serialCases/container/mf-ssr-clear-cache"`
  - `pnpm run lint:js`


2heal1:
@codex review

PR #14381, opened 2026-06-12
perf(core): reduce memcpy overhead

hardfist:
## What changed

- Avoid wrapping concatenation analysis code generation output in `CachedSource` so source materialization does less redundant copying.
- Box `CodeGenerationResult` at the async scope return boundary before it is collected on the main path.
- Keep the earlier concatenated module async move reduction in the same PR branch.

## Why

Callgrind on the threejs production bundle showed `memcpy`/`memmove` cost coming partly from large code generation results crossing async boundaries and from redundant source wrappers in module concatenation analysis. These changes keep behavior the same while reducing unnecessary movement/copying on that path.

## Impact

Local callgrind validation on threejs production showed the current PR branch reducing `__memcpy_avx_unaligned_erms` Ir from `24,390,572` to `24,027,278`, a `1.4895%` reduction. This is below the original 3% target, but it is the smallest positive set verified so far; broader attempted async boxing and init-fragment fast paths were measured and not kept.

## Checks

- `cargo check -p rspack_core -p rspack_plugin_javascript`
- Local callgrind harness for threejs production bundle

### DIFF
--- a/crates/rspack_core/src/compilation/code_generation/mod.rs
+++ b/crates/rspack_core/src/compilation/code_generation/mod.rs
@@ -223,7 +223,7 @@ pub(crate) async fn code_generation_modules(
           }
         }
 
-        (job.module, job.runtimes, codegen_res)
+        (job.module, job.runtimes, codegen_res.map(Box::new))
       })
     })
   })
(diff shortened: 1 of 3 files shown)


hardfist:
@explain why threejs-production-sourcemap case becomes faster

hardfist:
@codspeed explain perf gain

stormslowly (dismissed):

stormslowly (approved):

PR #14384, opened 2026-06-12
refactor: split content hash from std Hash

intellild:
## Background

Rspack currently has many output hash calculations built on top of Rust's standard `Hash` trait. This creates two problems:

- The behavior does not always match webpack. For example, webpack skips optional fields when they are not provided, while derived/std `Hash` may still encode the option state.
- `Hash` is designed for `HashMap`/`HashSet` keys, not for webpack-compatible content hash. Its requirements are different, and relying on Rust/std hashing behavior can make output hashes sensitive to implementation changes.

This PR separates these two concerns. Data structures that participate in output/content hashing should implement the Rspack content-hash trait (`RspackContentHash`) and be written through `RspackHash`. Data structures that only need to be map/set keys should continue to implement the standard `Hash` trait. This lets us align content hash behavior with webpack while keeping the freedom to optimize `Hash` implementations independently later without affecting output hashes.

## Changes

- Route output/content hash inputs through `RspackContentHash` instead of using standard-library `Hash` directly.
- Preserve existing internal deterministic hashing behavior for non-content-hash paths such as module graph hashing, split-chunks max-size grouping, and generated external identifiers.
- Keep `RspackHash` focused on output/content hash calculation rather than reintroducing it as a standard `Hasher`.
- Update hash-only snapshots for the new content hash results.
- Add coverage for updating visible outer scope bindings.

## Validation

- `pnpm run test:unit`
- `pnpm run test:rs`

---
_by OpenAI Codex_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4152,11 +4152,18 @@ name = "rspack_hash"
 version = "0.101.1"
 dependencies = [
  "base64-simd 0.8.0",
+ "itoa",
+ "lightningcss",
  "md4",
  "rspack_cacheable",
+ "rspack_collections",
+ "rspack_macros",
  "rspack_util",
+ "serde",
  "sha2",
+ "simd-json",
  "smol_str",
+ "swc_config",
  "xxhash-rust",
 ]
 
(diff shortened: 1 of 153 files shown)


intellild:
blocked by #14423

intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

intellild:

hardfist (approved):

PR #14403, opened 2026-06-13
perf(ci): split wasm release build into parallel wasi/browser jobs

stormslowly:
## Why

The `wasm32-wasip1-threads` release build is the long pole of the release wall-clock: one job builds both variants (wasi + browser) serially, ~26m cold.

## What

- **Split the wasm release build into two parallel jobs** (wasi + browser) via `binding_postfix`, one variant per runner. Downstream adapted for the split artifacts: `scripts/build-npm.cjs` skips the browser-only artifact, `rslib.browser.config.ts` reads `rspack.browser.wasm` from the `-browser` artifact, and the `@rspack/browser` e2e job downloads it (the redundant browser test job is skipped). CI builds keep the single-artifact path.
- `release-canary.yml`: same split, and pin the wasm targets to `profile=release` so a non-release canary still publishes the real two-variant build instead of the CI copy shortcut.
- Build **all** release targets on self-hosted runners (Windows/Mac were github-hosted), force release artifacts to build from scratch (no Rust cache), and include the target in build/test job names.

## Effect

- wasm build, cold same-conditions: **~26m → ~19m**.
- Whole release build wall-clock: **~32m → ~24m**.

Validated by canary publishes (`profile=release`/`profiling`/`debug`) and a `test=true` dry-run (wasm Node + browser e2e green).


stormslowly:
Superseded by #14414 — the branch was renamed to `release/2.0.9-nightly`, which auto-closed this PR. Same changes continue there.

PR #14404, opened 2026-06-14
fix(core): support watcher events on native WatchFileSystem

### LINKED ISSUE #11545: [Bug]: Native watcher typecheck issue
### System Info

System:
    OS: Linux 6.14 Ubuntu 24.04.2 LTS 24.04.2 LTS (Noble Numbat)
    CPU: (24) x64 Intel(R) Core(TM) Ultra 9 285K
    Memory: 86.70 GB / 125.25 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 22.17.0 - ~/.nvm/versions/node/v22.17.0/bin/node
    Yarn: 4.6.0 - ~/.nvm/versions/node/v22.17.0/bin/yarn
    npm: 10.9.2 - ~/.nvm/versions/node/v22.17.0/bin/npm
  Browsers:
    Chrome: 138.0.7204.157
  npmPackages:
    @rspack/cli: 1.5.1 => 1.5.1 
    @rspack/core: 1.5.1 => 1.5.1

### Details

When the `experiments.nativeWatcher` flag is enabled, TypeScript error checking is broken.

### Reproduce link

https://github.com/adubrouski/rspack-native-watcher-typecheck-issue-repro

### Reproduce Steps

1. Clone the repo
2. Install the dependencies (`yarn install`)
3. Run `yarn run dev:rspack`
4. You should see one TypeScript error in the terminal. Change 
   ```typescript
   const variable: string = 123;
   ``` 
   to
   ```typescript
   const variable: number = 123;
   ``` 
   in `./src/index.ts`. The TypeScript error will be fixed, but the same error will still be shown in the terminal.
5. Restart `yarn run dev:rspack`. The error will no longer be shown. Now change 
   ```typescript
   const variable: number = 123;
   ``` 
   to 
   ```typescript
   const variable: string = 123;
   ``` 
   in `./src/index.ts`. There should be a TypeScript error again, but it does not appear in the terminal.
6. Restart `yarn run dev:rspack`. You should now see one TypeScript error in the terminal.

> If we change experiments.nativeWatcher to `false` - steps 5 and 6 will not reproduce.

# Expected behavior

- After fixing the error (step 4), the error message should disappear from the terminal.
- After introducing a new error (step 5), the error message should immediately appear in the terminal without requiring a restart.

# Expected behavior

- After fixing the error (step 4), the error message still remains in the terminal.
- After introducing a new error (step 5), the error message does not appear until the process is restarted.

stormslowly:
## Why

Fixes #11545. With `experiments.nativeWatcher` enabled, `ts-checker-rspack-plugin`'s typecheck goes stale — after fixing a type error the message persists, and a newly introduced error only shows up after a restart.

The plugin observes file changes through watchpack-private APIs on `compiler.watchFileSystem.watcher` (`on('change')` / `on('remove')` / `_onChange` / `_onRemove`). `NativeWatchFileSystem` exposes no `watcher`, so those listeners never attach and the checker never learns which files changed.

Before / after — unmodified ts-checker, `nativeWatcher` on, edit a file to fix the error:

| | result |
| --- | --- |
| before | error stays (stale) |
| after | error clears |

## What

- Add a standard `on` / `once` / `emit` event API (`change` / `remove` / `aggregated`) to `WatchFileSystem`, implemented for both the native and node watch file systems. `on` / `once` listeners are stored on an internal emitter so they survive the watcher being replaced each watch cycle.
- Expose a watchpack-compatible `watcher` shim on `NativeWatchFileSystem` that proxies `on` / `once` / `_onChange` / `_onRemove` to the native watcher, so an **unmodified** ts-checker keeps working under `nativeWatcher`. The shim is recreated per `watch()` cycle (mirroring watchpack) so consumers that re-attach each cycle don't accumulate listeners. Collections such as `fileWatchers` / `directoryWatchers` are intentionally not implemented; plugins that iterate them should migrate to `emit`.
- The native watcher now delivers per-file `change` / `remove` undelayed events.
- Classify a watched path that no longer exists as `remove` instead of `change`, so deletions report `remove` on every platform (macOS FSEvents reports an unlink as a rename, which previously surfaced as `change`).

## Usage (for plugin authors)

Observe or inject watch events through the standard `on` / `once` / `emit` API on `compiler.watchFileSystem`. It behaves the same under the node and native watchers, so there is no need to reach into watchpack/native internals.

`ts-checker-rspack-plugin` is the motivating example: it tracks changed/removed files to re-run the type check. Previously it reached into the watchpack-private `watcher`, which only exists on the node watcher:

```ts
// before — only the node (watchpack) watcher has `.watcher`
const { watcher } = compiler.watchFileSystem;
watcher.on("change", file => changed.add(file));
watcher.on("remove", file => removed.add(file));
```

The standard API does the same and keeps working under `experiments.nativeWatcher`:

```ts
// after — works with both the node and native watchers
compiler.watchFileSystem.on?.("change", file => changed.add(file));
compiler.watchFileSystem.on?.("remove", file => removed.add(file));
```

Events:

| event | listener args | meaning |
| --- | --- | --- |
| `change` | `(filename, mtime)` | a watched file changed |
| `remove` | `(filename)` | a watched file was removed |
| `aggregated` | `(changes, removals)` | batched change/removal sets for one rebuild |

A plugin that owns its files (e.g. virtual modules) can inject an event with `emit`:

```ts
compiler.watchFileSystem.emit?.("change", filename, Date.now());
```

The `watcher` shim keeps the current ts-checker working unchanged; migrating to `on` is the forward-compatible path.

## Tests

- `watchCases/change-event/watch-fs-events` asserts that both `watchFileSystem.on('change')` and `watchFileSystem.watcher.on('change')` report an edit, under both the node and native watchers.
- `rspack_watcher` test `should_emit_remove_when_a_watched_file_is_deleted` deterministically covers that deleting a watched file is reported as `remove` (native fs remove events are asynchronous and hard to assert reliably from a watch case).


stormslowly:
Thanks for the review — addressed the points:

1. **NodeWatchFileSystem `on`/`once` lost across cycles** — `on`/`once` now register on an internal long-lived `EventEmitter`; each cycle's fresh Watchpack forwards its `change`/`remove` to it, so listeners survive watcher replacement and can be attached before the first `watch()`.
2. **Native `watcher` shim accumulating listeners** — the shim is now recreated on every `watch()` call (mirroring watchpack's per-cycle instance), so consumers that re-attach each cycle don't pile up listeners. The stable `on`/`once` API is backed by the separate long-lived emitter.
3. & 4. **`emit` mtime** — documented that the native pipeline re-stamps events with their arrival time and cannot carry a caller-provided `mtime`; the parameter is kept for parity with the node implementation. The misleading comment was corrected.

Also: deletions are now classified as `remove` instead of `change` (macOS FSEvents reports an unlink as a rename), covered by a deterministic `rspack_watcher` Rust test, since asserting async native `remove` events from a watch case is flaky.

hardfist (approved):

PR #14421, opened 2026-06-15
perf: reduce build chunk graph preparation work

LingyuCoder:
## What changed

- Skip prepared connection-map entries for modules with no effective outgoing build-chunk-graph connections.
- Avoid sorting prepared ordered dependencies when their `source_order` values are already monotonic.
- Use a shared empty block-module result for leaf modules and avoid populating runtime cache entries for them.
- Let block extraction handle modules with blocks but no prepared connection map.

## Why

Leaf modules and already ordered dependency lists are common in build chunk graph preparation. Avoiding empty allocations, cache writes, and unnecessary sorts reduces work in the preparation and traversal paths.


LingyuCoder:

JSerFeng (approved):

PR #14422, opened 2026-06-15
perf: optimize module concatenation traversal

LingyuCoder:
## Summary

This PR ports the module concatenation traversal optimization from #14195, scoped to `crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs`.

The change reduces repeated work while walking module concatenation candidates:

- Cache outgoing connection metadata in `CachedOutgoingConnection`, including the target module identifier, whether referenced exports are statically known, and the active state for the module runtime.
- Cache incoming connection metadata in `CachedIncomingConnection`, including the active state and whether the dependency is ESM-like, so later traversal does not repeatedly query the module graph and dependency type.
- Store cached import and success results as `Arc<[ModuleIdentifier]>`, avoiding repeated cloning of vectors/sets when cache hits are reused during recursive traversal.
- Reuse cached module chunk sets when checking whether candidate modules are in the same chunks as the root, avoiding repeated `chunk_graph` lookups for hot-path checks.
- Split the incoming-connection traversal into a fast path for the common non-multi-runtime case and only build the detailed active connection list when runtime-condition validation is actually needed.

The runtime-condition validation is moved into the multi-runtime branch as part of that split. The benefit is not that running this check earlier is inherently faster; it lets the common path avoid preparing data that only the runtime-condition check needs. For single-runtime or no-runtime cases, the traversal can track simple booleans such as whether an active connection exists and whether it is non-ESM, instead of allocating and retaining per-origin active connection lists. Only the multi-runtime path builds those detailed lists and then performs the runtime-condition validation.

These changes improve performance because module concatenation repeatedly visits related modules and checks the same connection, runtime, chunk, and dependency metadata. Moving those derived values into per-module caches turns repeated graph/dependency lookups and intermediate collection allocations into simple field reads, and avoids constructing data structures for checks that are unnecessary in the common path.

## Behavior and correctness notes

This PR changes the internal traversal structure, but it is intended to preserve the module-concatenation eligibility semantics:

- Active-connection checks are still evaluated against the requested runtime. The cached `active` value is only reused when the cached module runtime is equal to, a subset of, or a superset of the queried runtime in cases where that implication is valid; otherwise the code falls back to `connection.is_active` / `connection.is_target_active`.
- Same-chunk validation remains equivalent: the previous `chunk_graph.is_module_in_chunk` checks are replaced with subset checks against cached `HashSet<ChunkUkey>` values cloned from `chunk_graph.get_module_chunks`.
- ESM-like dependency validation remains equivalent: dependency type classification is cached on `CachedIncomingConnection` and later reused instead of repeatedly resolving the dependency from the module graph.
- Import de-duplication is preserved with the `seen` set before storing cached `Arc<[ModuleIdentifier]>` results, so repeated ESM dependencies to the same target do not change the candidate set.
- Non-ESM bailout diagnostics are filtered back through `is_connection_active_in_runtime`, so the diagnostic only reports active unsupported incoming connections for the queried runtime.

One observable difference is the bailout reason priority in multi-runtime cases. Runtime-condition validation now happens inside the multi-runtime traversal branch before the later different-chunk / non-ESM bailout formatting. If a candidate violates multiple constraints at once, the selected bailout reason may differ from before, but the candidate still bails out and is not concatenated. This is a diagnostic-order difference rather than a generated-code semantics change.

Validation performed locally on the PR head:

```bash
cargo check -p rspack_plugin_javascript
```

This passed successfully. Full correctness should still rely on the existing module-concatenation regression suite, especially multi-runtime, runtime-condition, same-chunk, and non-ESM bailout cases.

### DIFF
--- a/crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs
@@ -6,7 +6,7 @@ use rspack_collections::{
   Identifiable, IdentifierDashMap, IdentifierIndexSet, IdentifierMap, IdentifierSet,
 };
 use rspack_core::{
-  BoxDependency, BoxModule, Compilation, CompilationOptimizeChunkModules, DependencyId,
+  BoxDependency, BoxModule, ChunkUkey, Compilation, CompilationOptimizeChunkModules, DependencyId,
   DependencyType, ExportProvided, ExportsInfoArtifact, ExtendedReferencedExport, GetTargetResult,
   ImportedByDeferModulesArtifact, LibIdentOptions, Logger, ModuleGraph, ModuleGraphCacheArtifact,
   ModuleGraphConnection, ModuleGraphModule, ModuleIdentifier, OptimizationBailoutItem, Plugin,
@@ -20,7 +20,7 @@ use rspack_core::{
 use rspack_error::{Result, ToStringResultToRspackResultExt};
 use rspack_hook::{plugin, plugin_hook};
 use rspack_util::itoa;
-use rustc_hash::FxHashMap as HashMap;
+use rustc_hash::{FxHashMap as HashMap, FxHashSet as HashSet};
 
 fn format_bailout_reason(msg: &str) -> String {
   format!("ModuleConcatenation bailout: {msg}")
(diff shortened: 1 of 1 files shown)


LingyuCoder:

LingyuCoder:

LingyuCoder:

JSerFeng:
@codex review

JSerFeng (approved):

PR #14425, opened 2026-06-15
fix(pure-functions): keep deferred pure import callees

### LINKED ISSUE #14387: [Bug]: `experiments.pureFunctions` removes an export from the export map while keeping a call to it, emitting `undefined(...)`
### System Info

```
System:
  OS: Windows
Binaries:
  Node: 24.15.0
  npm: 11.12.1
npmPackages:
  @rspack/cli: 2.0.8 => 2.0.8
  @rspack/core: 2.0.8 => 2.0.8
```

### Details

With `mode: 'production'` and `experiments.pureFunctions: true`, rspack can drop a function from a module's export map while keeping a cross-module call to it in an importing module. The emitted code looks like this:

```js
const mergeStore = /* unused export .actionFactory */ undefined('Store/MERGE');
```

(or `(void 0)(...)` with the default minimizer), which throws `TypeError: undefined is not a function` the moment the module evaluates. There is no build-time warning or error; the build reports success. The emitted source module is annotated `// UNUSED EXPORTS: ...`, yet a retained call in another module still targets that export.

The trigger requires a specific cross-module shape: a module has an unused export whose initializer calls an imported function, alongside an effectively-used sibling export, and the unused export's only importer is reachable only through a dynamic `import()` where just `.default` is accessed. The full conditions, and the nearby variations that do not trigger, are documented in the [repro README](https://github.com/ArgonAlex/rspack-pure-functions-severed-export#module-graph). The whole reproduction is 5 small modules, ~25 lines of code, no loaders or plugins.

**Versions:** reproduces on @rspack/core 2.0.0, 2.0.5, 2.0.6, 2.0.7, and 2.0.8 (latest), so it appears to exist for as long as the experiment has been available. With `pureFunctions: false` the output is correct on all of them. Note that Rsbuild sets `experiments.pureFunctions: true` unconditionally as of v2.0.12 (web-infra-dev/rsbuild#7824, cc @chenjiahan as the author of that change), so downstream users hit this by upgrading Rsbuild rather than by opting in. We hit it in a large production SSR bundle (~13 MB); the reproduction is minimized from that.

**Related issues:** #14188 is the single-module sibling (a local binding nullified to `(null)()` while still used by a side-effectful call), fixed by #14191 in 2.0.6. The cross-module case here, where the severed binding is an imported export, is not covered by that fix and still reproduces on 2.0.8. #13728 is the inverse symptom (factory-created functions not tree-shaken at all; dead code remains, no crash).

### Reproduce link

https://github.com/ArgonAlex/rspack-pure-functions-severed-export

### Reproduce Steps

```sh
git clone https://github.com/ArgonAlex/rspack-pure-functions-severed-export
cd rspack-pure-functions-severed-export
npm install
npm run build    # compiles successfully, no warnings
npm run check    # TypeError: undefined is not a function
grep "unused export" dist/bundle.js
# const mergeStore = /* unused export .actionFactory */ undefined('Store/MERGE');
```

Flip `experiments.pureFunctions` to `false` in `rspack.config.js` and rebuild: the export is kept, the call resolves, and `npm run check` runs cleanly.

JSerFeng:
## Summary

Fixes a `experiments.pureFunctions` tree-shaking case where an imported callee could be rendered as `undefined` even though a retained top-level initializer still calls it.

Deferred pure checks now share one resolver for `usedExports` connection checks and side-effects analysis. The resolver keeps deferred pure callees active only when the referenced export resolves impure, and it falls back to the original resolved module when module concatenation redirects the dependency to a module without per-export side-effect metadata.

The regression also covers an unused pure helper imported from the same shared module, so pure deferred imports are not retained just because another export from that module is active.

## Related links

Closes #14387

## Validation

- `pnpm run build:js`
- `cargo fmt --all --check`
- `git diff --check`
- `pnpm --dir tests/rspack-test run test -t "configCases/tree-shaking/pure-functions-severed-export"`
- `pnpm --dir tests/rspack-test run test -t "configCases/tree-shaking"`
- External repro from #14387 compiled and ran with the local binding

## Checklist

### DIFF
--- a/crates/rspack_core/src/exports/utils.rs
+++ b/crates/rspack_core/src/exports/utils.rs
@@ -191,9 +191,9 @@ pub enum UsageState {
 #[cacheable]
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct UsedByExports {
-  pub condition: UsedByExportsCondition,
+  condition: UsedByExportsCondition,
   #[cacheable(with=AsVec)]
-  pub deferred_pure_checks: Vec<UsedByExportsDeferredPureCheck>,
+  deferred_pure_checks: Vec<UsedByExportsDeferredPureCheck>,
 }
 
 impl UsedByExports {
@@ -217,6 +217,23 @@ impl UsedByExports {
   ) -> Self {
     self.deferred_pure_checks = deferred_pure_checks;
     self
+  }
+
+  pub fn condition(&self) -> &UsedByExportsCondition {
+    &self.condition
+  }
+
+  pub fn deferred_pure_checks(&self) -> &[UsedByExportsDeferredPureCheck] {
+    &self.deferred_pure_checks
+  }
+
+  pub fn has_deferred_pure_checks(&self) -> bool {
+    !self.deferred_pure_checks.is_empty()
+  }
+
+  pub fn is_false_without_deferred_pure_checks(&self) -> bool {
+    matches!(&self.condition, UsedByExportsCondition::Bool(false))
+      && !self.has_deferred_pure_checks()
   }
 }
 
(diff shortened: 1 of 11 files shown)


JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

JSerFeng:
@codex review

ahabhgk (approved):

PR #14428, opened 2026-06-15
chore(deps): bump swc_ecma_minifier to 55.0.2 for template-literal minify fix

pkasarda:
## Summary

Bumps **`swc_ecma_minifier` 55.0.0 → 55.0.2** in `Cargo.lock`, which pulls the template-literal minification fix from swc-project/swc#11939. rspack's minify path (`rspack_javascript_compiler`) depends on `swc_ecma_minifier` directly, so this is sufficient — no `swc_core` facade bump needed (that would also drag in 68.1.x's react-compiler crates).

Also bumps the shared swc crates required by 55.0.2 to their compatible patch versions: `swc_common` 23.0.1, `swc_ecma_codegen` 28.0.1, `swc_ecma_parser` 41.1.0, `swc_ecma_transforms_base` 44.0.2. **Lock-only** — no `Cargo.toml` changes (all within existing `^` ranges).

## The bug it fixes

The SWC minifier drops a character when constant-folding the `+`-concatenation of two template literals that each contain an interpolation. `concat_tpl` merged the boundary quasis' `raw` but left `cooked` stale, and downstream passes evaluate the template to a string via `cooked`.

Repro on current rspack (`swc_ecma_minifier` 55.0.0) — a production build minifying:

```js
const html = `<b>${1}</b>` + `<i>${2}</i>`;
```

emits:

```js
const html = "<b>1<i>2</i>";   // wrong — the `</b>` is dropped (expected "<b>1</b><i>2</i>")
```

Fixed in `swc_ecma_minifier` 55.0.2.

Upstream fix: https://github.com/swc-project/swc/pull/11939

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5878,9 +5878,9 @@ dependencies = [
 
 [[package]]
 name = "swc_common"
-version = "23.0.0"
+version = "23.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6ebd5f732c45e88fa1c69cf2ac65bf24e2578bfd428b74348d3346ee137b8c50"
+checksum = "353daaf5e6c2ad90efd4ff52270173074ded88fdba5a40f302c7c3a9b1c037fb"
 dependencies = [
  "anyhow",
  "ast_node",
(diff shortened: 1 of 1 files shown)


CPunisher (approved):

CPunisher:
Thanks!

PR #14434, opened 2026-06-16
perf(loader-runner): cache esm loader imports

hardfist:
## Summary

Cache ESM loader imports in `loadLoader` so repeated loader executions for the same loader path do not repeatedly enter Node's dynamic `import()` path.
* before
image

* after
image

## Details

The CPU profile showed `loadLoader` spending most of its inclusive time below `importModuleDynamicallyCallback`, with the actual cost coming from Node ESM resolve/hooks and `MessageChannel`/`postMessage` work. Node caches ESM module evaluation, but repeated dynamic imports can still pay the resolver and loader hook overhead before reaching that cache.

This change keeps Rspack's lazy loader loading behavior, but adds a small cache in front of ESM imports:

- caches the loaded ESM loader module by loader type and path
- reuses an in-flight import promise when multiple modules hit the same loader concurrently
- caches the `pathToFileURL(...).toString()` result
- clears the in-flight promise on import failure so later calls can retry

CJS loader behavior is left unchanged.

## Validation

- `./node_modules/.bin/prettier packages/rspack/src/loader-runner/loadLoader.ts --check`
- `cd packages/rspack && ./node_modules/.bin/rslib build`


chenjiahan (approved):

hardfist:
@codspeedbot explain the walltime regression

PR #14436, opened 2026-06-16
perf(sources): speed up source map tokenization

nnnnoel:
## Summary
- Use an ASCII fast path during `OriginalSource` chunk streaming without doing extra work in `OriginalSource::new`.
- Route potential-token boundary lookup through a helper that keeps the existing `memchr` path for short spans and uses a single-pass SWAR scan for longer spans.
- Add token splitting coverage for boundary priority and separator handling.

## Why
CPU profiling showed source map generation spending time in `PotentialTokens::next`, `OriginalSourceChunks::stream`, and repeated UTF-16/ASCII column work. This keeps `OriginalSource` construction cheap, preserves the common short-token path on `memchr`, and reduces repeated scans where the source map streaming path can use a fast ASCII hint.

## Validation
- `pnpm run format:rs`
- `git diff --check`
- `cargo test -p rspack_sources`
- `cargo bench -p rspack_benchmark --bench rspack_sources --profile profiling -- 'sources@(complex_replace_source_map|repetitive_react_components_map)' --warm-up-time 1 --measurement-time 3 --sample-size 20`

## Benchmark
Local Criterion wall-time benchmark. `origin/main` was measured at `0d39755010`; the final branch was measured at `6f38abd534`. Delta is computed from the median time, where negative is faster.

| Benchmark | `origin/main` | Final branch | Delta |
|---|---:|---:|---:|
| `sources@complex_replace_source_map` | `680.97 us` | `660.52 us` | `-3.00%` |
| `sources@complex_replace_source_map_cached_source_stream_chunks` | `366.31 us` | `360.41 us` | `-1.61%` |
| `sources@repetitive_react_components_map` | `141.30 us` | `138.43 us` | `-2.03%` |

### DIFF
--- a/crates/rspack_sources/src/original_source.rs
+++ b/crates/rspack_sources/src/original_source.rs
@@ -39,14 +39,18 @@ use crate::{
 pub struct OriginalSource {
   value: Box<str>,
   name: Box<str>,
+  is_ascii: bool,
 }
 
 impl OriginalSource {
   /// Create a [OriginalSource].
   pub fn new(value: impl Into<Box<str>>, name: impl Into<Box<str>>) -> Self {
+    let value = value.into();
+    let is_ascii = value.is_ascii();
     Self {
-      value: value.into(),
+      value,
       name: name.into(),
+      is_ascii,
     }
   }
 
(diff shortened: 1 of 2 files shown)


crates/rspack_sources/src/original_source.rs
```diff
@@ -39,14 +39,18 @@ use crate::{
...
   value: Box<str>,
   name: Box<str>,
+  is_ascii: bool,
 }
 
 impl OriginalSource {
   /// Create a [OriginalSource].
   pub fn new(value: impl Into<Box<str>>, name: impl Into<Box<str>>) -> Self {
+    let value = value.into();
+    let is_ascii = value.is_ascii();
```
SyMind: From the perspective of rspack_sources design, no computation should be performed during the construction of OriginalSource.

nnnnoel:
Can i know how to compare about codspeed simulation mode bench in local machine? (macOS, ARM64, Tahoe 26.3)

nnnnoel:
My performance optimization perspective seems wrong.
I'll find out to optimize CPU cost.

PR #14438, opened 2026-06-16
perf: optimize source map json serialization

nnnnoel:
### What changed

This replaces the generic `simd_json::to_writer` path in `SourceMap::to_json` with a dedicated JSON writer for the known source map shape. It keeps the existing serialized field order and skip behavior, including omitting `sourcesContent` when all entries are empty.

String fields now use a raw UTF-8 fast path when no JSON escaping is needed, and fall back to `json-escape-simd` only for strings containing quotes, backslashes, or control characters. The implementation also writes numeric fields and `ignoreList` directly into the output buffer.

### Why

`SourceMap::to_json` is a serialization hot path for generated source maps. The previous generic serializer paid repeated per-field serde dispatch and string serialization overhead even though the output schema is fixed. The source map arrays and mappings string can be large, so avoiding extra generic serialization work and keeping writes streaming into one preallocated buffer reduces CPU overhead in this path.

### Expected impact

The main benefit is lower CPU cost while preserving output compatibility with the serde serializer. The latest local Criterion measurement for `rspack_sources/sources@source_map_to_json` was:

```text
time: [2.4835 ms 2.5159 ms 2.5644 ms]
```

This benchmark covers the focused source map JSON serialization case.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5182,6 +5182,7 @@ dependencies = [
 name = "rspack_sources"
 version = "0.100.8"
 dependencies = [
+ "json-escape-simd",
  "memchr",
  "regex",
  "rkyv 0.8.16",
--- a/crates/rspack_sources/Cargo.toml
+++ b/crates/rspack_sources/Cargo.toml
@@ -17,6 +17,7 @@ codspeed         = []
 rspack_cacheable = ["dep:rkyv"]
 
 [dependencies]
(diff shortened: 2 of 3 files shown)


SyMind:
@codex review

nnnnoel:
My performance optimization perspective seems wrong.
I'll find out to optimize CPU cost.

PR #14440, opened 2026-06-16
chore(watcher): instrument find_common_root panic (diagnostic, do not merge)

stormslowly:
## Why

`WatcherRootAnalyzer::find_common_root` panics on Windows with `assert!(path.exists())` → `Path should exist` in `NativeWatcher.part3` (e.g. [run 27600712729](https://github.com/web-infra-dev/rspack/actions/runs/27600712729/job/81608093792)), crashing the test worker (`exit 0xC0000409`).

This is a **pre-existing flaky panic** in the native watcher (the assert landed with the watcher itself in #13987) and is **unrelated to the runtime changes** — it just surfaced on this branch's Windows CI. The non-determinism comes from `find_root` picking an arbitrary `DashMap` node, so the watch root it returns can be a path that does not exist on disk.

## What

**Diagnostic only — do not merge.** Temporary `eprintln` instrumentation with no behavior change (the `assert` is kept on purpose so it still panics). Goal: capture the real `PathTree` state from CI and confirm the actual case instead of guessing.

- **per-cycle summary** (emitted only when abnormal): count of nodes missing on disk + orphan subtree roots → tells us whether the tree became a *forest*
- **full tree dump** right before the panic: every node's `exists` / `is_dir` / `children` / `parent_in_tree`
- **entry classification**: `depth==0` = `find_root` result (forest / missing root) vs `depth>0` = descend step (TOCTOU delete)

Grep CI logs for `[WATCHER_ROOT_DEBUG]`.


stormslowly:
Diagnostic complete. Root cause confirmed: a latent orphan-forest bug in PathTree::remove_path (cross-set path migration on the shared tree). The real fix lands in #14448. Closing this do-not-merge diagnostic branch.

PR #14445, opened 2026-06-16
perf(sources): cache generated source info

nnnnoel:
## Summary

- Replace `get_generated_source_info` with a single `memchr` newline scan and only compute the UTF-16 width for the final line.
- Cache `GeneratedInfo` on `SourceMapSource` so repeated final-source sourcemap generation does not rescan the generated source each time.
- Thread the cached generated-source info through source-map streaming while keeping `SourceMapSource` equality and hashing based only on observable source data.

## Profiling

Before:

Before sourcemap profile

After:

After sourcemap profile

- Baseline `sample`: `get_generated_source_info` had 1291 top-of-stack samples.
- After the cache: `get_generated_source_info` no longer appears in the `>= 5` top-of-stack list.
- Remaining hot samples move to sourcemap decoding, map generation, and serialization work.

## Related links

N/A

## Checklist

### DIFF
--- a/crates/rspack_sources/src/source_map_source.rs
+++ b/crates/rspack_sources/src/source_map_source.rs
@@ -1,14 +1,14 @@
 use std::{
   borrow::Cow,
   hash::{Hash, Hasher},
-  sync::Arc,
+  sync::{Arc, OnceLock},
 };
 
 use crate::{
   MapOptions, Source, SourceMap, SourceValue,
   helpers::{
-    Chunks, StreamChunks, TextSpan, get_map, stream_chunks_of_combined_source_map,
-    stream_chunks_of_source_map,
+    Chunks, GeneratedInfo, StreamChunks, TextSpan, get_generated_source_info, get_map,
+    stream_chunks_of_combined_source_map, stream_chunks_of_source_map_with_generated_info,
   },
   object_pool::ObjectPool,
 };
@@ -59,7 +59,6 @@ impl<V, N> From<WithoutOriginalOptions<V, N>> for SourceMapSourceOptions<V, N> {
 /// source map for the original source.
 ///
 /// - [webpack-sources docs](https://github.com/webpack/webpack-sources/#sourcemapsource).
-#[derive(Eq)]
 pub struct SourceMapSource {
   value: Box<str>,
   name: Box<str>,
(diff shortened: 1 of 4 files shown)


crates/rspack_sources/src/source_map_source.rs
```diff
@@ -59,14 +59,14 @@ impl<V, N> From<WithoutOriginalOptions<V, N>> for SourceMapSourceOptions<V, N> {
...
 /// - [webpack-sources docs](https://github.com/webpack/webpack-sources/#sourcemapsource).
-#[derive(Eq)]
 pub struct SourceMapSource {
   value: Box<str>,
   name: Box<str>,
   source_map: SourceMap<'static>,
   original_source: Option<Box<str>>,
   inner_source_map: Option<SourceMap<'static>>,
   remove_original_source: bool,
+  generated_info: OnceLock<GeneratedInfo>,
```
SyMind: We should avoid storing any computation-related data within SourceMapSource; this aligns with the architectural design principles of rspack-sources.

Please refer to the implementation of CachedSource. The intended approach is to wrap nodes involving potentially expensive computations with CachedSource at the Rspack level, which caches the results of source and map to prevent redundant calculations.

nnnnoel: Thanks for the clarification. I agree that `SourceMapSource` should not store computation-related cache state.

I re-ran the comparison.

```sh
target/release/deps/rspack_sources-* \
  --bench rspack_sources/sources@concat_generate_string \
  --sample-size 20 \
  --warm-up-time 2 \
  --measurement-time 5
```

| benchmark | origin/main | current PR | revised patch |
| --- | ---: | ---: | ---: |
| `sources@concat_generate_string` | 36.262 µs | 28.113 µs (-22.47%) | 34.382 µs (-5.18%) |
| `sources@concat_generate_string_with_cache` | 14.842 µs | 14.799 µs (-0.29%) | 14.909 µs (+0.45%) |

The direct `concat_generate_string` benchmark benefits a lot from the current PR because it repeatedly calls `map()` on an unwrapped `SourceMapSource`, so the internal `OnceLock<GeneratedInfo>` avoids repeated generated-line/column scans.

However, `concat_generate_string_with_cache` models the intended Rspack-level usage with `CachedSource`. In that case, the full map result is cached at the wrapper boundary, and the additional `SourceMapSource`-level cache does not provide a meaningful benefit.

So I think the better tradeoff is to remove `OnceLock<GeneratedInfo>` from `SourceMapSource`, keep `SourceMapSource` as a pure source-map node, and retain only the stateless `get_generated_source_info` fast path.

nnnnoel:
@codspeedbot explain this regression

PR #14448, opened 2026-06-16
fix(rspack_watcher): prevent watch-root panic from orphaned path tree

stormslowly:
## Why

On recovery a dependency path migrates between sets within a single `analyze` cycle: e.g. the unresolved `node_modules/some-module` resolves to a real directory, so it leaves `missingDependencies` and enters the directory/context set. It then appears in one set's `added` **and** another set's `removed` at the same time.

`analyze` applied each set's delta independently to a single **shared** `path_tree`, so the stray `removed` deleted a node the other set still referenced — disconnecting the tree into a forest of orphan subtree roots. `find_root` (which picks an arbitrary `DashMap` node via `iter().next()`) could then land on a non-existent orphan, which `assert!(path.exists())` turned into a panic. macOS/Windows only (Linux uses `WatcherDirectoriesAnalyzer`), flaky.

Confirmed on Windows CI via diagnostic instrumentation; reproduced locally on macOS with the real `rspack-resolver` path sets for `recover-from-error/missing-module`.

## What

- **`analyze`**: union the three sets' `added` / `removed`, then update the tree with `adds − removes` (net added) and `removes − adds` (net removed). A migrating path is in both `adds` and `removes`, so it cancels out — the shared tree neither re-adds nor deletes it, and stays connected.
- **`remove_path`**: detach from the **parent's** child set. The old code removed `path` from its *own* set (a no-op), leaving a stale child reference on the parent.

## Verification

`cargo test -p rspack_watcher` 28/28, clippy clean. A new end-to-end test replays the real two-phase migration through `PathManager` + `analyze` and asserts it does not panic and the migrated node is preserved.

### DIFF
--- a/crates/rspack_watcher/src/analyzer/root.rs
+++ b/crates/rspack_watcher/src/analyzer/root.rs
@@ -22,11 +22,18 @@ impl Analyzer for WatcherRootAnalyzer {
     let (_, added_directories, removed_directories) = path_accessor.directories();
     let (_, added_missing, removed_missing) = path_accessor.missing();
 
-    self.path_tree.update_paths(added_files, removed_files);
-    self
-      .path_tree
-      .update_paths(added_directories, removed_directories);
-    self.path_tree.update_paths(added_missing, removed_missing);
+    // files / directories / missing share one `path_tree`. On recovery a path
+    // can migrate between sets within a single cycle (e.g. the unresolved
+    // dependency `node_modules/some-module` becomes a real directory): it then
+    // shows up in one set's `added` and another set's `removed`. Applying each
+    // set's delta independently lets the stray `removed` delete a node the other
+    // set still references — orphaning its children and disconnecting the tree.
+    // Merge the three sets first, then cancel migrating paths out of both nets.
+    let adds = union3(added_files, added_directories, added_missing);
+    let removes = union3(removed_files, removed_directories, removed_missing);
+    let added = difference(&adds, &removes);
+    let removed = difference(&removes, &adds);
+    self.path_tree.update_paths(&added, &removed);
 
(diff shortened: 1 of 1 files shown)


stormslowly:
Thanks @copilot. The branch was revised after this review — I dropped the defensive approach (nearest-existing-ancestor fallback + children-preservation guard) in favor of fixing the root cause, so all three comments now point at code that no longer exists (the threads are marked outdated). Mapping each:

1. **`find_common_root_recursive` could return a file** — the `nearest_existing_ancestor` fallback is gone and the original `assert!(path.exists())` is restored. (The file-as-root edge is pre-existing on `main` and unchanged by this PR.)
2. **"cleaned up once its last child is removed" / pending deletion** — that comment and the children-preservation guard are removed. `remove_path` now unconditionally removes the node and detaches it from its **parent's** child set.
3. **test relying on a possibly-existing `__does_not_exist__` dir** — that test is removed. The new regression tests build the tree explicitly via `PathManager`/`add_path` and don't depend on disk (non)existence of a fixed path.

**Root cause:** files / directories / missing share one `PathTree`, but `analyze` applied each set's added/removed independently. On recovery a path migrates across sets within a single cycle (e.g. `node_modules/some-module` missing→directory), so the stray `removed` deletes a node another set still references → orphan forest → `find_root` can land on a non-existent node. The fix unions the three sets and cancels migrating paths (`adds−removes` / `removes−adds`) so the tree stays connected; the original fail-fast assert is kept.

Verified: `cargo test -p rspack_watcher` green, clippy clean; Windows CI shows 0 panic and orphan count 0 across 169 cycles.

hardfist (approved):

PR #14455, opened 2026-06-17
perf(sources): compact cow

SyMind:
## Checklist


Cargo.toml
```diff
@@ -24,6 +24,7 @@ async-recursion = { version = "1.1.1", default-features = false }
 async-trait     = { version = "0.1.89", default-features = false }
 atomic_refcell  = { version = "0.1.14", default-features = false }
 base64-simd     = { version = "0.8.0", default-features = false, features = ["alloc"] }
+beef            = { version = "0.5.2", default-features = false, features = ["impl_serde"] }
```
quininer, line 27: it make sense? https://github.com/maciejhirsz/beef/issues/57

SyMind: I used beef::lean::Cow, it is two words in 64 platform

SyMind: But it doesn't seem very effective.

PR #14456, opened 2026-06-17
perf(core): optimize resource parsing

chenjiahan:
## Summary

Centralizes path/query/fragment resource parsing in `util/identifier`, removes the duplicate `loader-runner` helper export, and adds a fast path for plain resource paths so the common case avoids the regexp.

It also narrows parser return objects by dropping unused `resource` fields from `parseResource` and `parseResourceWithoutFragment`.

## Performance data

- No cache object: `55.991 ms` -> `16.939 ms` median (`3.31x`, `69.7%` faster).
- Cache object with all misses: `92.901 ms` -> `52.901 ms` median (`1.76x`, `43.1%` faster).
- Benchmark project `dev:rsbuild` cpuprofile target self time: about `48.247 ms` -> `0 ms` in filtered samples


hardfist (approved):

chenjiahan:
@codspeedbot analyze this regression

PR #14462, opened 2026-06-17
fix(test): isolate source for fixture-writing config cases

stormslowly:
## Why

`Test WASM` flakily failed `Can't resolve ./star*/a.js` in `configCases/code-generation/path-ends-with-star`.

Root cause (proved with logging in this PR's history): `configCases` is walked by **two parallel test processes** — `Config.part1` and `RuntimeModeConfig.part1`. They isolate **output** (unique `dist`) but compile **in-place from the shared source dir**. The fixture creates/removes `star*` in that shared dir, so whichever suite finished first ran its `done`-hook `fs.rmSync(star*)` and deleted the dir while the other was still building it. wasm only because it is ~10x slower, widening the window.

## What

Isolate the **input** too, opt-in via `test.config.js { isolateSource: true }`:

```
js/<suite>/<case>/
  ├─ src/    ← context (per-suite copy of the case dir; fixtures write here)
  └─ dist/   ← output.path
```

`getCompileSource`/`getCompileDist` on the test context return the copy paths when `isolateSource` is set, and fall back to `getSource`/`getDist` otherwise — so the ~1780 read-only config cases are byte-identical and pay no copy cost.

`path-ends-with-star` opts in; its original create+remove fixture is now safe because each suite owns its copy.

## Verified

- local: `Config.part1` + `RuntimeModeConfig.part1` full run — 1035 passed, 0 regression; on-disk layout confirmed (`<case>/src` copy + `<case>/dist` output per suite).
- the diagnostic + 150ms-widen commits in history reproduced the race deterministically and confirmed the earlier fix; this PR replaces that with the general isolation mechanism.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,8 +128,6 @@ webpack-dist
 .pnpm-store/
 
 !tests/rspack-test/configCases/target
-# Runtime-created `*` fixture for the path-ends-with-star case
-tests/rspack-test/configCases/code-generation/path-ends-with-star/star\*/
 tests/e2e/temp
 packages/*/tests/js
 tests/rspack-test/js
--- a/.gitignore
+++ b/.gitignore
@@ -128,6 +128,8 @@ webpack-dist
 .pnpm-store/
 
 !tests/rspack-test/configCases/target
+# Runtime-created `*` fixture for the path-ends-with-star case
+tests/rspack-test/configCases/code-generation/path-ends-with-star/star\*/
 tests/e2e/temp
 packages/*/tests/js
 tests/rspack-test/js
--- a/crates/rspack_core/src/resolver/boxfs.rs
+++ b/crates/rspack_core/src/resolver/boxfs.rs
@@ -27,7 +27,23 @@ fn star_diag_path(path: &std::path::Path) -> bool {
     .is_some_and(|p| p.contains("path-ends-with-star"))
 }
 
-fn star_diag_meta(op: &str, path: &std::path::Path, res: &rspack_fs::Result<rspack_fs::FileMetadata>) {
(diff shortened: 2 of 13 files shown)


stormslowly:
## Root cause found (not a wasm fs bug — a test fixture race)

The `star*` dir is created/removed in the **shared source dir**, but this case is executed **in parallel by two separate test files**: `Config.part1` and `RuntimeModeConfig.part1` (both walk `configCases`, and `code-generation` ∈ `[a-d]`). One suite's `done`-hook `fs.rmSync(star*)` deletes the dir while the other is still building it.

Reproduced deterministically by adding a 150ms sleep before the `star*` fs calls. Smoking gun (two different PIDs, same dir):

```
pid=5137 created {dirExists:true, fileExists:true}
pid=5346 created {dirExists:true, fileExists:true}
pid=5137 afterCompile {hasStarError:false}
pid=5137 done -> cleanup {dirExists:true}        # 5137 rmSync deletes star*
pid=5346 afterCompile {hasStarError:true, dirExists:false, fileExists:false,
                       starDirEntries: READDIR_ERR ENOENT}
pid=5346 STAR-ERROR: ... failed to read .../star*/a.js
[STAR-DIAG] metadata(.../star*) -> Err{kind:NotFound, errno:Some(44)}   # WASI ENOENT
```

wasm only because it is ~10x slower, widening the remove-during-build window.

**Fix**: create the fixture idempotently and never delete it mid-run (matches the existing `schemes/file-url` / `require/module-require` fixtures, which create into the source dir and don't destructively clean up). Verifying now with the widen still in place — it should pass.

hardfist (approved):

PR #14464, opened 2026-06-17
perf(core): lazily allocate build info assets

LingyuCoder:
## Summary

This changes `BuildInfo.assets` from an eagerly allocated `BindingCell<HashMap<...>>` into a small lazy wrapper. Most modules never emit build assets, so `BuildInfo::default()` no longer creates the `HashMap`, `Box`, and binding reflector state for that cold path.

The wrapper keeps the existing map-like read/write surface used by module asset collection, concatenated modules, executor asset forwarding, stats/benchmark reads, and JS `buildInfo.assets` access.

## Performance Rationale

`BuildInfo` is embedded in every module, so per-module default allocation is paid at module cardinality. The previous `assets: BindingCell<HashMap<...>>` default path allocated backing storage even for modules with no assets. The new representation stores `None` until Rust inserts module assets.

For JS consumers, reading empty `buildInfo.assets` returns an empty object without mutating the module, while modules that actually have assets still expose the existing native asset reflector.

### DIFF
--- a/crates/rspack_binding_api/src/build_info.rs
+++ b/crates/rspack_binding_api/src/build_info.rs
@@ -1,7 +1,7 @@
 use std::{cell::RefCell, sync::LazyLock};
 
 use napi::{
-  Env, JsString, JsValue, Property, PropertyAttributes, Unknown,
+  Env, JsString, JsValue, NapiRaw, Property, PropertyAttributes, Unknown,
   bindgen_prelude::{
     Array, FromNapiMutRef, FromNapiValue, JsObjectValue, Object, ToNapiValue, WeakReference,
   },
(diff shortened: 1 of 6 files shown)


LingyuCoder:

LingyuCoder:
Updated this PR to also lazily allocate several low-usage `BuildInfo` container fields:

- `value_dependencies`
- `extras`
- `deferred_pure_checks`

These now follow the same pattern as `assets`: default `BuildInfo` keeps them unallocated, read-only empty paths stay allocation-free, and backing storage is created only on insertion/replacement.

Fresh local validation after the update:

- `cargo test -p rspack_core build_info_assets_are_created_lazily`
- `cargo check -p rspack_plugin_javascript -p rspack_plugin_rslib`
- `cargo fmt --all`
- `pnpm run test:unit`
- `cargo lint`

PR #14467, opened 2026-06-17
test: remove flaky file-dependencies-huge config case

stormslowly:
## Why

`Config.part2` / `RuntimeModeConfig.part2` intermittently fail with
`afterAll hook timed out in 10000ms` — recurs on `main`, all assertions pass,
only the per-case `compiler.close()` cleanup hangs.

`configCases/dependencies/file-dependencies-huge` builds **1,000,000** file deps
synchronously in a `done` hook (~8s on the CI runner). rstest runs **one test
file per worker process in parallel**, so that ~8s CPU burn starves a *concurrent*
worker's event loop for **>10s**; when that worker is a config shard mid-`afterAll`
`compiler.close()`, the napi close callback can't be delivered within rstest's
default 10s `hookTimeout`. The case already needed a WASM skip (#14399) for the
same freeze.

We remove the case rather than keep shrinking it: the regression it guarded (the
`push(...data)` max-call-stack overflow) stays defended in source — both
`MergeCaller.push` and `createFakeCompilationDependencies.addAll` iterate instead
of spreading, with an explicit comment — so dropping this heavyweight, repeatedly
flaky end-to-end case outweighs re-guarding that low-risk, well-documented path.

Full investigation, instrumentation method and CI log evidence (8s freeze →
10.75s cross-process event-loop stall, with dose–response):
https://github.com/web-infra-dev/rspack/pull/14467#issuecomment-4732972569


stormslowly:
## Root cause: `Config.part2` / `RuntimeModeConfig.part2` `afterAll hook timed out in 10000ms`

**TL;DR** — Not a `compiler.close()` bug, and unrelated to any source change. `configCases/dependencies/file-dependencies-huge` monopolizes a CPU core for ~8s (synchronously, in its `done` hook). Because rstest runs **one test file per worker process in parallel**, that ~8s burn **starves a *concurrent* worker's event loop for >10s** on the few-core CI runner. When the starved victim is a `*Config.part2` worker sitting in its `afterAll` `compiler.close()` cleanup, the native (napi) close callback can't be delivered within rstest's default **10s `hookTimeout`** → the hook times out. All assertions pass; only cleanup is starved.

This is the **same case and the same freeze mechanism** as the already-known WASM flake fixed by #14399 — the difference is WASM starves *in-process* (single-threaded WASI), while native starves a *sibling process*. The WASM fix only skipped the case under `WASM`, so the **native path still runs the 1,000,000-dep version**.

### Log links
- Instrumented evidence run, Test Linux Node 20 — **attempt 1** (the `[HUGE]` + `[LAG] 10753ms` smoking-gun timeline below): https://github.com/web-infra-dev/rspack/actions/runs/27673815751/job/81847174585
- Same run, **attempt 2** (dose–response, sub-threshold): https://github.com/web-infra-dev/rspack/actions/runs/27673815751/job/81866732103
- **`main` recurrence** (`Config.part2` afterAll timeout, no source change): https://github.com/web-infra-dev/rspack/actions/runs/27613789763/job/81645807215
- First observed on #14461's CI (CI-YAML-only PR): https://github.com/web-infra-dev/rspack/actions/runs/27671260660/job/81838251238

### Symptom
- Fails only in the per-case cleanup hook: `creator.ts` `afterAll` → `tester.resume()` → `context.closeCompiler()` → `Compiler.close()` (awaits a napi close callback). The 10s is rstest's default `hookTimeout` (the suite only sets `testTimeout: 60s`).
- `Tests … passed` — every assertion passes; only the `afterAll` hook times out.
- **Recurs on `main`**, e.g. run [`81645807215`](https://github.com/web-infra-dev/rspack/actions/runs/27613789763/job/81645807215) (Config.part2 only), independent of any PR. The PR where it was first noticed (#14461) only edits CI YAML.

### Method
Temporary instrumentation on this branch (not for merge):
- `[HUGE]` — phase timestamps around the `file-dependencies-huge` `done` hook (measures the real synchronous freeze on the runner).
- `[LAG]` — a per-process event-loop lag sampler (detects cross-process starvation).
- `[CLOSEDIAG]` — a 7s watchdog that snapshots active handles/requests + rss/heap for any slow `compiler.close()`, before the 10s timeout.

### Evidence ([Test Linux Node 20, run `27673815751`](https://github.com/web-infra-dev/rspack/actions/runs/27673815751/job/81847174585))
Real freeze on the CI runner is **~8s**, not the ~1s a fast dev box suggests (building the 1M array alone is ~5s, amplified by GC/memory pressure from multiple 8GB-heap workers):
```
[HUGE] end pid=4214 build=4858ms addAll=687ms materialize=569ms toEqual=1730ms total=7844ms
[HUGE] end pid=6580 build=5909ms ... toEqual=730ms total=8180ms
```
Smoking-gun timeline — a **passing** run dancing right at the 10s edge:
```
08:11:49.879  [HUGE] start pid=4214                         # file-dependencies-huge (Config.part1 worker)
08:11:57.723  [HUGE] end   pid=4214 total=7844ms            # ~8s synchronous CPU burn
08:11:59.433  ✓ Config.part1.test.js
08:12:02.145  [LAG] pid=5163 event-loop stalled 10753ms     # a DIFFERENT worker, starved 10.75s across the freeze
08:12:04.153  ✓ Config.part2.test.js                        # finishes ~2s after the stall clears
```
`pid 4214` = the part1 worker running the huge case; `pid 5163` = a concurrent worker whose loop was starved **10,753 ms** — past the 10s threshold. It happened not to be mid-`close()` this run, so it passed by a hair.

**Dose–response** confirms causality — a [second attempt](https://github.com/web-infra-dev/rspack/actions/runs/27673815751/job/81866732103) had shorter huge freezes and correspondingly sub-threshold stalls (no failure):
```
[HUGE] total=4936ms / 6602ms   →   [LAG] max cross-process stall 7.7–9.2s  (< 10s, safe)
```

### Why it's always `part2`
`part2` (e-o, ~590 cases) is the largest config shard, so its `afterAll` runs the most sequential `compiler.close()` calls and has the longest shutdown window — the widest overlap with the huge-case freeze. `part1` contains the huge case but its own freeze ends before its (shorter) `afterAll`; `part3` rarely aligns in time.

### The 3rd link is true by construction
`afterAll` `await`s `close()`'s napi callback, which is delivered via the worker's libuv loop. If that loop is stalled >10s, the callback cannot be delivered and the 10s `hookTimeout` necessarily fires. The instrumentation directly measured the first two links (8s freeze → 10.75s cross-process stall); the third follows from how the timeout works.

### Suggested fix
The case only exists to verify that adding more file deps than the engine's `push(...data)` argument cap (~65536) doesn't throw `Maximum call stack size exceeded` (see its own comment). `HUGE_DEPS_COUNT = 1_000_000` is far beyond that and is the sole CPU monopolizer.
- **Reduce `HUGE_DEPS_COUNT` to ~`100_000`** (still ≫ 65536, preserves the intent) → freeze drops to <1s, cross-process stall <2s, flake gone.
- Optionally lighten the `expect(...).toEqual(...)` over the full 1M array (the most expensive sync step) to a length + spot check.

This mirrors #14399 but fixes the native path that was left running 1M.

stormslowly:
miss-leading

PR #14471, opened 2026-06-17
chore(ci): use mini runner for ecosystem ci dispatcher

stormslowly:
## Summary

- Run the `ecosystem_ci_per_commit` job on `vars.CI_LINUX_MINI_RUNNER` instead of always using `ubuntu-latest`.
- Keep the existing `ubuntu-latest` fallback when the repository variable is not configured.

## Why

This job mostly triggers and polls ecosystem CI. Moving it to the existing mini runner pool reduces GitHub-hosted runner usage for main-branch CI without changing the external ecosystem CI behavior.


stormslowly:
Hold on; we need docker for convictional/trigger-workflow-and-wait.

PR #14522, opened 2026-06-20
chore(ci): generate rspack_workspace versions via build.rs instead of a cargo codegen step

stormslowly:
## Why

The native binding exposes `EXPECTED_RSPACK_CORE_VERSION` for the JS-side napi version check (`packages/rspack/src/util/bindingVersionCheck.ts`). Its value comes from `crates/rspack_workspace/src/generated.rs`, which until now was produced by a dedicated `cargo codegen` (xtask) step wired into **7 places across 5 workflows**.

That step compiles the whole `xtask` binary (clap + deps) just to regenerate one already-committed file. In the binding-build jobs `xtask` is not otherwise compiled, so every target/profile paid a full `xtask` build for nothing.

## What

- Move the generation into `crates/rspack_workspace/build.rs`. It regenerates the committed `src/generated.rs` from the root `Cargo.toml` + `package.json` during the crate's normal compile — `rspack_workspace` is already in every binding/build dependency tree, so this adds no extra crate to build.
  - Writes **only inside the rspack repo** (guards on the root `package.json` / `[workspace]`), **only when content changes**, and is fully best-effort. A published or vendored consumer therefore never attempts to write the read-only registry source dir.
- `generated.rs` stays committed, so the workspace versions remain visible in version control; `build.rs` keeps it in sync. The existing binding-build `Diff artifact` check now also catches a stale version file for free.
- Remove the redundant `cargo codegen` steps (7×, 5 workflows), the `xtask codegen` subcommand, and the `cargo codegen` alias. `./x version` regenerates the file via `cargo check -p rspack_workspace`.

**Before → after (binding-build job):** full `xtask` (clap) compile + run → a trivial `build.rs` file read inside a compile that already happens.

No behavior change: the napi version-check path and the three generated version constants are unchanged.


stormslowly:
### Note for reviewers: CI catches a stale/hand-edited `generated.rs`

`crates/rspack_workspace/src/generated.rs` stays committed (so the versions remain visible in VCS), and is regenerated by the new `build.rs`. If someone hand-edits it — or bumps a version in `package.json` / `Cargo.toml` without regenerating — CI catches the drift: in the binding-build job `build.rs` regenerates the correct content during the build (workspace crates are never restored from `rust-cache`, so it always runs cold in CI), and the existing `Diff artifact` step then fails because `git diff` is no longer clean.

Verified end-to-end by temporarily committing a wrong value (`swc_core = "0.0.0-CI-DRIFT-TEST"`): the `Build Linux` job failed at the `Diff artifact` step —

https://github.com/web-infra-dev/rspack/actions/runs/27899581597/job/82557375749

The test commit has been reverted.

stormslowly:
Thanks for the review — addressed all three in `crates/rspack_workspace/build.rs` (644b6e2ff5):

1. **Silent stale constants when `generate()` returns `None`** — when the workspace files exist (we're inside the rspack repo) but versions can't be read, `build.rs` now emits a `cargo::warning` instead of silently leaving the committed constants stale.
2. **`expect(..)` on write contradicts "best-effort"** — a write failure (e.g. read-only tree) is now reported via `cargo::warning` and the build continues using the committed `generated.rs`, instead of panicking.
3. **Manual edits not reasserted** — added `cargo::rerun-if-changed=src/generated.rs`. Verified locally that a hand-edit is corrected on the next build, and that consecutive no-op builds stay `Fresh` (no rebuild loop — the content-compare guard prevents churn).

stormslowly:
Now, CI freshness uses content hash. I don't need to optimize this now.

PR #14531, opened 2026-06-22
test: probe watcher FS_ACCURACY across macOS / Linux / Windows

stormslowly:
Draft / measurement only — not for merge.

## Why

We want to know the actual filesystem mtime accuracy (`FS_ACCURACY`) that the watcher converges to on each OS, for **both** watcher implementations:

- **native watcher** — `rspack_watcher`'s real `safe_time` / `ensure_fs_accuracy` ladder (`crates/rspack_watcher/src/time_info.rs`)
- **watchpack** — the real `watchpack` package, read via `getTimeInfoEntries().accuracy`

## How

`.github/workflows/fs-accuracy-probe.yml` runs a matrix over `ubuntu-latest`, `macos-latest`, `windows-latest`. Each job writes ~50 real files spread over ~1.1s of wall clock and reports the converged accuracy plus the empirical min mtime delta (ground truth).

- native probe: an `#[ignore]`d test `probe_fs_accuracy` (run in isolation so the process-global `FS_ACCURACY` ratchet stays pristine)
- watchpack probe: `scripts/fs-accuracy-probe/probe-watchpack.mjs`

Read results by grepping `FS_ACCURACY_PROBE` in each job log. Trigger via the workflow_dispatch button or by pushing to this branch.

## Note

watchpack feeds `+stats.mtime` (integer ms) into its ladder, so its `% 1` branch never fires in practice — the effective floor is **10ms**, same as the native watcher. Local macOS run: both converge to **10ms** (empirical min delta ~21ms).

### DIFF
--- a/.github/workflows/fs-accuracy-probe.yml
+++ b/.github/workflows/fs-accuracy-probe.yml
@@ -1,0 +1,51 @@
+name: FS Accuracy Probe
+
+# Measures the filesystem mtime accuracy (`FS_ACCURACY`) that each watcher
+# converges to, on macOS / Linux / Windows, for both implementations:
+#   - native watcher: rspack_watcher's real `safe_time`/`ensure_fs_accuracy`
+#   - watchpack: the real `watchpack` package via `getTimeInfoEntries().accuracy`
+# Look for `FS_ACCURACY_PROBE` lines in each job log.
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/fs-accuracy-probe.yml'
+      - 'scripts/fs-accuracy-probe/**'
+      - 'crates/rspack_watcher/src/time_info.rs'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  probe:
+    name: Probe (${{ matrix.os }})
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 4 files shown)


stormslowly:
## Results — FS_ACCURACY across OSes (CI run 27939200275, all green)

| OS | native watcher | watchpack | empirical min mtime delta |
|----|:---:|:---:|:---:|
| Linux (ubuntu-latest) | **10 ms** | **10 ms** | 21 ms |
| macOS (macos-latest) | **10 ms** | **10 ms** | 22–23 ms |
| Windows (windows-latest) | **10 ms** | **10 ms** | 21 ms |

### Conclusion

1. **All three OSes × both implementations converge to 10 ms** — no platform or implementation divergence.
2. **10 ms is the real-world floor of both ladders.** The native (Rust) ladder works on integer ms = {10,100,1000,2000}. watchpack has a `% 1 → 1ms` branch, but it feeds `+stats.mtime` (integer ms from `Date.getTime()`), so that branch never fires → effective floor is also 10 ms.
3. **Node `stat.mtime` and Rust `SystemTime` agree per platform.** Empirical min mtime delta is ~21–23 ms everywhere (≫ the unreachable <10 ms tier, ≪ 100 ms), so the ladder stably lands on 10 ms. Windows native even shows `non_mult_10=50` (NTFS mtime finer than 10 ms), but the ladder still locks at 10 ms.

> First Windows attempt failed on a transient crates.io download flake (`curl [16] HTTP2 framing layer / Connection reset`) while compiling `unicode-segmentation`; `rerun --failed` passed. Not code-related.

PR #14534, opened 2026-06-22
feat: make persistent cache max generations default to 3

ahabhgk:
## Summary
- Make filesystem cache `maxAge` and `maxGenerations` explicit defaults across binding, core, and storage layers
- Treat `Infinity` as a disable flag in JS config normalization and validate the raw storage limits there
- Update cache docs and default-case expectations to reflect the new defaults

## Testing
- Not run (not requested)


chenjiahan (dismissed):
👍

hardfist (approved):

PR #14537, opened 2026-06-22
perf(module-concat): prune readable identifier work

LingyuCoder:
## Summary

- Store module concatenation bailout details as structured data so readable identifiers are only prepared for warnings that are actually emitted.
- Replace the eager readable identifier warmup for all concatenation candidates with a narrower parallel warmup for selected roots and emitted warnings.


LingyuCoder:

JSerFeng:
@codex review

LingyuCoder:
@codspeedbot explain the regression of mangle exports plugin

JSerFeng (approved):

Issue #14556, opened 2026-06-24
[Tracking]: webpack compatibility gaps in Rspack


intellild:
This issue tracks Rspack compatibility gaps that have been explicitly recommended for implementation.

Last updated: 2026-08-27 18:50:11 CST

## Recommended to implement in Rspack

_No features are currently marked as recommended._

intellild:
Last updated: 2026-08-27 17:16:27 CST

## Completed webpack-compatibility features

Use this comment to record webpack features that were previously missing in Rspack but have since been completed.

### Completed

- [x] `#__NO_SIDE_EFFECTS__` / `@__NO_SIDE_EFFECTS__` pure annotations - Rspack has parser support and coverage for explicit side-effect-free annotations. Source: Rspack implementation https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs and tests `tests/rspack-test/configCases/tree-shaking/no-side-effects-notation`, `tests/rspack-test/watchCases/side-effects/explicit-side-effects-free-incremental`.
- [x] CommonJS `const mod = require("...")` member usage as referenced exports - Rspack records referenced specifiers on `CommonJsRequireDependency`, covering the same compatibility area as webpack PR #21003. Source: webpack v5.107.1 release notes https://github.com/webpack/webpack/releases/tag/v5.107.1 and Rspack implementation https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs.
- [x] `RealContentHashPlugin` asset hash update hooks - Rspack now exposes and invokes the real-content-hash update hook, closing the hook portion of the built-in plugin compatibility gap. Source: Rspack PR #14660 https://github.com/web-infra-dev/rspack/pull/14660.
- [x] Persistent cache storage option alignment - Rspack now aligns `cache.name` and filesystem storage `directory`/`location` options with webpack. Source: Rspack PR #15049 https://github.com/web-infra-dev/rspack/pull/15049.

- [x] `import.meta.glob({ caseSensitive })` - Rspack now supports case-sensitive and case-insensitive glob matching, preserves filesystem casing, validates static boolean options, and exposes the public type/docs. Source: Rspack PR #14811 https://github.com/web-infra-dev/rspack/pull/14811.
[...]

PR #14557, opened 2026-06-24
chore: add webpack gap tracking skill

intellild:
## Summary

- Add a repository skill for tracking Rspack compatibility gaps against webpack.
- Record the tracking issue and placeholder comment IDs for missing features, completed features, and missing tests.
- Include update rules for timestamps, preserving existing issue/comment format, native CSS routing, and deduplication.

## Validation

- Not run; documentation-only skill change.

---
_by OpenAI Codex_

### DIFF
--- a/.agents/skills/rspack-webpack-gap-tracker/SKILL.md
+++ b/.agents/skills/rspack-webpack-gap-tracker/SKILL.md
@@ -1,0 +1,75 @@
+---
+name: rspack-webpack-gap-tracker
+description: Use when tracking, auditing, or updating Rspack compatibility gaps against webpack, including missing webpack features, completed compatibility features, and missing webpack test coverage.
+---
+
+# Rspack Webpack Gap Tracker
+
+## Tracking Targets
+
+- Main tracking issue: https://github.com/web-infra-dev/rspack/issues/14556
+- Completed feature comment ID: `4785525873`
+- Missing test comment ID: `4785526309`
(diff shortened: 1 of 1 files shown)


chenjiahan (commented):
Not all webpack-features should be ported to Rspack. Can we have the agent evaluate whether a webpack feature is suitable for Rspack?

Then split the “Missing features” section into categories such as recommended to implement, pending evaluation, and not recommended to implement.

intellild:
> Not all webpack-features should be ported to Rspack. Can we have the agent evaluate whether a webpack feature is suitable for Rspack?
> 

I think we need some rules, for example if a plugin can be implemented without native rust and won't affect performance, place it in rstack orgranization is better.

chenjiahan:
Yeah, can we add some basic rules to the skill?

intellild (commented):
Addressed in `f2209066d3`.

- Added recommendation classification rules to the skill: recommended / pending evaluation / not recommended for Rspack.
- Added the requested decision rules: existing Rspack config/plugin gaps are recommended, JS-only/plugin-compatible gaps without native/perf impact should prefer third-party or rstackjs packages, and future updates should check Rsbuild/rstackjs for existing solutions.
- Updated #14556 to use the same three-category `Missing features` format and added recommendation reasons to the existing entries.

---
_by OpenAI Codex_

chenjiahan (approved):

PR #14573, opened 2026-06-25
perf: refactor rspack sources cache

SyMind:
## Checklist


SyMind:

PR #14580, opened 2026-06-25
fix(css): skip re-applying unchanged CSS during HMR

### LINKED ISSUE #7951: [Bug]: Dynamic imports forces too much reloads, some times full page reload in dev mode
### Version

```sh
System:
    OS: Windows 11 10.0.26200
    CPU: (24) x64 AMD Ryzen 9 3900XT 12-Core Processor           
    Memory: 12.21 GB / 31.92 GB
  Browsers:
    Chrome: 149.0.7827.115
    Edge: Chromium (149.0.4022.69)
    Firefox: 151.0.4 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.26100.8115
  npmPackages:
    @rsbuild/core: ^2.0.14 => 2.0.15 
    @rsbuild/plugin-babel: ^1.2.1 => 1.2.1 
    @rsbuild/plugin-react: ^2.0.1 => 2.1.0
```

### Details

After staring dev server, when dynamic import is called for first time, it can update full page styles, or reload full page. On large progect full page reload always happens. In reproduction demo, sometimes you can randomly hit situation, where all styles are missing and alert is showing.

https://github.com/user-attachments/assets/b3872e94-18e1-424a-af02-3a5b793dad73

https://github.com/user-attachments/assets/3442cc27-b62d-4778-bc9d-2896f8381941

### Reproduce link

https://github.com/user-attachments/files/29136452/dynamic-import.zip

### Reproduce Steps

[dynamic-import.zip](https://github.com/user-attachments/files/29136452/dynamic-import.zip)

1. unzip
2. npm install
3. npm run dev
4. click button

### LINKED ISSUE #11377: [Bug]: fast refresh causes style flash
### Version

```sh
System:
  OS: macOS 14.7.7
  CPU: (10) arm64 Apple M1 Pro
  Memory: 83.83 MB / 16.00 GB
  Shell: 5.9 - /bin/zsh
Browsers:
  Chrome: 139.0.7258.67
  Safari: 18.6
  Safari Technology Preview: 18.4
npmPackages:
  @rsbuild/core: 1.4.9 => 1.4.9
  @rsbuild/plugin-react: ^1.3.5 => 1.3.5
```

### Details

Rsbuild `1.4.9` and onward causes style flash during fast refresh / hmr.
However, on `1.4.8` changes would not cause a style flash.

See linked CodeSandbox on 1.4.9 with issue. 
If you revert to 1.4.8, issue no longer occurs.

### Reproduce link

https://codesandbox.io/p/devbox/rsbuild-hmr-phmdwv

### Reproduce Steps

1. Install `@rsbuild/core@1.4.9` and react + plugin
2. Run `rsbuild dev`
3. Modify `App.tsx` and observe style flash

### Screenshots

1.4.9 always loads `index.css` on _any_ file change.

1.4.8 does not have this behavior.

stormslowly:
## Why

The CSS HMR handler (`CssExtractRspackPlugin`) re-applies the stylesheet of **every** chunk in the hot-update `c` list. The runtime chunk is always in `c` (its full hash changes on every rebuild), so the entry stylesheet is removed and re-inserted on **every** update — even when its content never changed.

The old swap order makes this visible: `dispose` removes the old `<link>` **before** `apply` enables the new one, so there is a window where the page has **zero** active stylesheets. The most visible case is the first lazy-compiled dynamic `import()` (delivered as an HMR update): the import resolves right inside that window, and anything that forces a paint there (e.g. a synchronous `alert()`) freezes a fully unstyled page on screen — reported in web-infra-dev/rsbuild#7951, related to #11377.

Additionally, several updated chunks can map to one extracted stylesheet (e.g. a fixed `filename` without `[name]` plus a `splitChunks` style chunk). The handler processed the file once per chunk: it double-fetched the CSS and, since `findStylesheet` only ever returns the first match, leaked one duplicated `<link>` per update.

### Before / After (first dynamic import, CSS unchanged)

| | Before | After |
|---|---|---|
| entry `<link>` | removed + re-inserted (flash) | untouched |
| active stylesheets during swap | 1 → **0** → 1 | 1 → 2 → 1 (no unstyled gap) |
| `<link>` count over repeated edits | grows 1→2→3 | stays 1 |

## What

Runtime-only change in `css_loading_with_hmr.ejs`, aligning the extract_css HMR runtime with the built-in CSS runtime (`rspack_plugin_css`) and webpack's native CSS loading runtime:

- Compare the reloaded stylesheet's content (`cssRules` text) with the existing one after it loads; if identical, drop the probe `<link>` and keep the existing one untouched.
- When content did change, swap in a safe order: enable the new sheet (`sheet.disabled = false`) first, then remove the old one.
- De-duplicate stylesheets by href so each extracted file is processed once per update.

Added an e2e case (`cases/css/shared-stylesheet-dedup`) where one update lists several chunks for one stylesheet; it asserts updates still apply and the `<link>` count stays at 1 (fails on the previous runtime with `Expected: 1, Received: 2`).

## Issues

- resolve #11377
- resolve web-infra-dev/rsbuild#7951

### DIFF
--- a/crates/rspack_plugin_extract_css/src/runtime/css_loading_with_hmr.ejs
+++ b/crates/rspack_plugin_extract_css/src/runtime/css_loading_with_hmr.ejs
@@ -1,25 +1,31 @@
 var oldTags = [];
 var newTags = [];
+var cssTextKey = function (link) {
+	return Array.from(link.sheet.cssRules, function (r) { return r.cssText; }).join();
+}
 var applyHandler = function (options) {
 	return {
-		dispose: function () {
+		dispose: function () {},
+		apply: function () {
+			for (var i = 0; i < newTags.length; i++) newTags[i].sheet.disabled = false;
 			for (var i = 0; i < oldTags.length; i++) {
 				var oldTag = oldTags[i];
 				if (oldTag.parentNode) oldTag.parentNode.removeChild(oldTag);
 			}
 			oldTags.length = 0;
-		},
-		apply: function () {
-			for (var i = 0; i < newTags.length; i++) newTags[i].rel = "stylesheet";
 			newTags.length = 0;
 		}
 	}
 }
 <%- HMR_DOWNLOAD_UPDATE_HANDLERS %>.miniCss = function (chunkIds, removedChunks, removedModules, promises, applyHandlers, updatedModulesList) {
 	applyHandlers.push(applyHandler);
+	var seen = {};
 	chunkIds.forEach(function (chunkId) {
 		var href = <%- REQUIRE_SCOPE %>.miniCssF(chunkId);
+		if (!href) return;
 		var fullhref = <%- PUBLIC_PATH %> + href;
+		if (seen[fullhref]) return;
+		seen[fullhref] = true;
 		var oldTag = findStylesheet(href, fullhref);
 		if (!oldTag) return;
 		promises.push(new Promise(function (resolve, reject) {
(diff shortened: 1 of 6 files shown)


chenjiahan (approved):
LGTM 👍

PR #14615, opened 2026-06-29
fix(rslib): preserve import phase in dynamic import external rendering

elecmonkey:
## Summary

`render_dyn_import_external_module` in `rspack_plugin_rslib` hardcoded `import(...)` as the callee for dynamic import externals, ignoring the dependency's `ImportPhase`. This caused `import.source("./foo.wasm")` targeting an external module to be incorrectly rendered as `import("./foo.wasm")` in the output, breaking source-phase semantics.

The fix reads `import_dep.get_phase()` and selects the correct callee:
- `Evaluation` → `import(...)`
- `Source` → `import.source(...)`
- `Defer` → `import.defer(...)`

## Checklist


Timeless0911 (approved):

Timeless0911:
canary version: https://github.com/web-infra-dev/rspack/actions/runs/28358518330

PR #14621, opened 2026-06-29
perf: build rspack source maps directly

intellild:
## Summary

- Build SWC print sourcemaps directly as `rspack_sources::SourceMap` instead of routing through `swc_common::SourceMap::build_source_map` and `swc_sourcemap`.
- Parse transform input sourcemaps into `rspack_sources::SourceMap<'static>` for inline, file, and user-provided maps.
- Compose input sourcemaps with `SourceMapSource` while preserving cheap-map line behavior, UTF-16 columns, names, source contents, and ignore-list handling.

## Validation

- `cargo check -p rspack_javascript_compiler`
- `pnpm run build:js`
- `pnpm run test --project base -t source-map/loader-source-map`
- `pnpm run test --project base -t source-map/cheap`
- `pnpm run test --project base -t swc-api`
- `pnpm run test --project base -t source-map/source-map`
- `pnpm run test --project base -t source-map/inline`
- `pnpm run test --project base -t source-map/source-map-ignore-list`
- `pnpm run test --project base -t source-map/eval-nosources-source-map`

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

crates/rspack_sources/src/helpers.rs
```diff
@@ -371,6 +371,37 @@ pub fn utf16_len(s: &str) -> usize {
...
+/// JavaScript source maps use zero-based UTF-16 code unit columns.
+#[inline]
+pub fn utf8_column_to_utf16_column(line: &str, utf8_column: usize) -> Option<usize> {
+  line.get(..utf8_column).map(utf16_len)
+}
+
+/// Convert a source-map column within a line to a UTF-8 byte column.
+///
+/// Returns `None` when the UTF-16 column falls inside a surrogate pair.
+pub fn utf16_column_to_utf8_column(line: &str, utf16_column: usize) -> Option<usize> {
```
SyMind: I don't find where use this fn.

crates/rspack_sources/src/source.rs
```diff
@@ -455,6 +455,8 @@ impl SourceMap<'static> {
 
   /// Create a [SourceMap] from bytes.
   pub fn from_bytes(mut bytes: Vec<u8>) -> Result<Self> {
+    strip_source_map_garbage_header(&mut bytes)?;
```
SyMind: After checking ECMA-426, I think garbage-header stripping should not be part of the default `SourceMap::from_json` / `SourceMap::from_bytes` parser.

The spec describes this behavior in `FetchSourceMap`: when a source map is retrieved over HTTP(S), the consumer may strip a leading `)]}'...` line before JSON parsing. `ParseSourceMap` itself still expects a normal JSON source map.

crates/rspack_sources/src/helpers.rs
intellild: Thanks for catching this. You are right, this helper is not used anywhere, so I removed it together with the public re-export and the now-unneeded test assertions.

---
_by OpenAI Codex_

crates/rspack_sources/src/source.rs
intellild: Thanks for checking the spec and pointing this out. I agree that ECMA-426 describes this in the fetch path rather than the pure parse algorithm.

For this codebase, though, `SourceMap::from_json` / `from_bytes` are the shared entry points used by the places that consume incoming source maps. Moving the garbage-header handling out would require each caller to remember to duplicate the same compatibility logic, which feels easier to regress.

There are also practical compatibility reasons to keep this centralized: upstream loaders can pass through or generate input source maps with this garbage header, and SWC already accepts this form as well (see the earlier discussion: https://github.com/web-infra-dev/rspack/pull/14621#discussion_r3496329285). Keeping the behavior in the common parser lets Rspack continue accepting those input maps consistently.

---
_by OpenAI Codex_

crates/rspack_javascript_compiler/src/compiler/stringify.rs
```diff
@@ -190,6 +161,259 @@ impl JavaScriptCompiler {
...
+      original_source: None,
+      inner_source_map: Some(input_source_map),
+      remove_original_source: true,
+    }))
+    .map_static(
+      &ObjectPool::default(),
+      &MapOptions::new(config.source_map_kind.emit_columns()),
+    )
+    .map(|mut source_map| {
+      restore_ignore_list(&mut source_map, &ignored_sources);
```
SyMind: I think this can be removed. In Rspack, the source map `ignoreList` should be controlled by the `ignoreList` option of `SourceMapDevToolPlugin`.

crates/rspack_sources/src/source.rs
SyMind: As a bundler, Rspack mainly generates and composes source maps. Source maps passed from loaders, plugins, inline maps, or `inputSourceMap` should already be valid JSON.

As bundlers, webpack, Rolldown, and Turbopack do not handle this case either.

intellild: I checked `swc`'s sourcemap is forked from sentry. Removed garbage header support.

intellild:
@codex review

SyMind (approved):

PR #14625, opened 2026-06-29
perf: validate side effects artifact clone on Rome benchmark

hardfist:
Temporary validation PR for #14622 against the Rome benchmark base from #14624.

This keeps the Rome benchmark case in the base branch and applies only the perf commit from #14622 on top, so CodSpeed can report whether the optimization affects `rome-ts`.

Validation target:
- Base: `codex/rome-ts-benchmark-case` / #14624
- Head: `codex/verify-rome-side-effects-perf`
- Perf commit under test: `42801754c1f1dd2d2e28c5ee4264cd6145a6b328`

No code changes beyond cherry-picking the #14622 perf commit.


hardfist:

JSerFeng (approved):

PR #14633, opened 2026-06-30
perf(hook): reduce hook macro expansion size and optimize performance

intellild:
## What changed

- Add shared `HookCommon` metadata/stage storage in `rspack_hook` so generated hooks share name, base tap stages, and interceptor count handling.
- Cache base tap stages at `tap()` registration time and keep base taps sorted by stage, preserving original insertion order for equal stages.
- Add primitive stage helpers, including sorted stage-index iteration and `HookTapIndex`, so interceptor slow paths sort and merge `i32`/`usize` data instead of boxed tap trait objects.
- Keep no-interceptor hook calls on a simple fast path that directly iterates registered taps without stage collection or sorting.
- Update `plugin_hook` so wrapper `Deref` is only generated for `stage = self...` expressions; constant/static stage expressions no longer need dynamic wrapper access.
- Change `used_stages()` to return a sorted, deduped `Vec<i32>` by using the already sorted base stage list, avoiding the old `Vec -> HashSet -> Vec -> sort` path in the binding interceptor register flow.
- Add macro tests for base tap stage ordering, interceptor additional tap ordering, and dynamic `self` stage behavior.

## Applied optimizations

- Reduces repeated generic macro expansion by moving common hook state and metadata into non-generic `HookCommon`.
- Avoids repeatedly calling `stage()` on base tap trait objects during hook calls; base stages are captured once when taps are registered.
- Avoids sorting base taps at call time when there are no interceptors, which is the common hot path.
- For interceptor-created additional taps, sorts only lightweight stage indices and merges them with the already sorted base tap queue.
- Removes unnecessary `Deref` generation for plugin hook wrappers whose stage expression does not reference `self`.
- Removes the `rustc-hash` dependency from `rspack_hook` after `used_stages()` no longer needs `FxHashSet`.

## Why

This reduces emitted hook-related code size and lowers per-hook-call overhead while preserving dynamic stage behavior for interceptor-created taps and `stage = self...` plugin hooks.

## Scope note

The tracing release-off experiment is not included in this branch. It was only used locally for size comparison.

## Checks

- `cargo test -p rspack_macros_test`
- `cargo check -p rspack_hook -p rspack_macros -p rspack_core`
- `cargo check -p rspack_hook -p rspack_macros -p rspack_binding_api`
- `cargo shear`
- `cargo fmt --all --check`
- `git diff --check`

Release `rspack_core` rlib size comparison:

- baseline: `63,032,150` bytes
- this branch: `58,476,106` bytes
- this branch plus local-only tracing release off experiment: `58,353,084` bytes

---
_by OpenAI Codex_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4053,7 +4053,6 @@ dependencies = [
  "async-trait",
  "rspack_error",
  "rspack_macros",
- "rustc-hash",
  "tracing",
 ]
 
--- a/crates/rspack_hook/Cargo.toml
+++ b/crates/rspack_hook/Cargo.toml
@@ -11,7 +11,6 @@ version.workspace = true
 async-trait   = { workspace = true }
 rspack_error  = { workspace = true }
 rspack_macros = { workspace = true }
-rustc-hash    = { workspace = true }
 tracing       = { workspace = true }
 
 [lints]
--- a/crates/rspack_hook/src/lib.rs
+++ b/crates/rspack_hook/src/lib.rs
@@ -1,6 +1,5 @@
 use async_trait::async_trait;
 use rspack_error::Result;
-use rustc_hash::FxHashSet;
 
 pub struct HookMetadata {
   pub name: &'static str,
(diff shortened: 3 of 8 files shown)


intellild:

hardfist (approved):

PR #14636, opened 2026-06-30
perf: erase js tap storage types to reduce binary size

intellild:
## Summary

Refactors JS plugin tap registration storage so the register/cache layer stores dyn threadsafe functions instead of monomorphized hook argument and return types.

The concrete hook argument and return types are restored at each generated tap call boundary, keeping hook execution typed while reducing repeated generic expansion across register storage types.

## Validation

- `cargo fmt --all --check`
- `cargo check -p rspack_napi -p rspack_binding_api`

---
_by OpenAI Codex_


intellild:
@codex review

hardfist (approved):

PR #14641, opened 2026-07-01
feat: add name/value accessors to ESMExportSpecifierDependency

yashwanth195:
## Summary

ESMExportSpecifierDependency is a public struct, but its name and value fields are private with no accessors, so code outside rspack_plugin_javascript cannot read them:

- name: the public export name (for example, Checkbox in export { CheckboxShim as Checkbox })
- value: the local binding, equivalent to dependency.id in webpack (for example, CheckboxShim)

This PR adds two borrowing accessors, name() and value(), exposing those fields.

**Why:** Consumers that reconstruct a module's export map, for example to emit a library manifest, need both the public export name and the local binding to correctly resolve aliased exports. Without accessors, external code cannot distinguish the two and may end up dropping or mislabeling aliased exports. This follows the same accessor pattern already used by sibling types such as *ESMImportSpecifierDependency::name()

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/dependency/esm/esm_export_specifier_dependency.rs
+++ b/crates/rspack_plugin_javascript/src/dependency/esm/esm_export_specifier_dependency.rs
@@ -46,6 +46,14 @@ impl ESMExportSpecifierDependency {
       id: DependencyId::new(),
     }
   }
+
+  pub fn name(&self) -> &Atom {
+    &self.name
+  }
+
+  pub fn value(&self) -> &Atom {
+    &self.value
+  }
 }
 
 #[cacheable_dyn]


SyMind:
@yashwanth195 Could you share more context about your use case, ideally with a small example of how you plan to use these accessors? Are you building a custom Rspack Rust binding?

yashwanth195:
Yes, we're building a custom native Rspack plugin in Rust (N-API binding), porting an internal Webpack plugin (`LibraryProducerPlugin`) that emits a library manifest.
For each named ESM export we need both the public export name and the local binding. For example:

```js
export { CheckboxShim as Checkbox };
```
In this case:
- `Checkbox` becomes the manifest key.
- `CheckboxShim` is used to build the redirected export ID.

This matches the original Webpack implementation that we have, which reads `dependency.name` and `dependency.id`.
In Rspack, the equivalent information exists on `ESMExportSpecifierDependency`, but is stored in the private `name` and `value` fields with no public accessors. As a result, a plugin living in a separate crate cannot read them. Aliased exports therefore get dropped or mislabeled because we cannot distinguish the public export name from the local binding.

Proposed change is purely additive and mirrors patterns already used by other dependency types:

```rust
pub fn name(&self) -> &Atom {
    &self.name
}

pub fn value(&self) -> &Atom {
    &self.value
}
```
This would:
- Expose the information already stored on the dependency.
- Avoid any API behavior changes.
- Avoid any serialization changes.
- Enable external plugins to correctly reconstruct ESM export mappings.
- Preserve aliased exports during manifest generation.

We currently carry this as a local patch and would love to drop it once supported upstream.

SyMind (approved):

hardfist (approved):

PR #14650, opened 2026-07-01
perf(ci): speed up rust benchmark simulation build

stormslowly:
## Why

The `Rust Bench (simulation)` CI job spends **~6m48s** in the build step. Root cause: the `codspeed` profile `inherits = "release"`, which sets `codegen-units = 1`. LLVM backend codegen then runs single-threaded, so critical-path crates serialize (rspack_core alone ~76s; a ~2m17s stretch compiles one crate at a time) while the 4-core runner sits idle. `-Zthreads=8` only parallelizes the frontend.

## What

`[profile.codspeed]`:
- `codegen-units = 16` — parallelize backend codegen across cores.
- `debug = "full"` → `false` — DWARF is unnecessary for callgrind instruction counting; `strip = "none"` still keeps symbols, so function-level flamegraphs remain. Generated machine code and instruction counts are unchanged by debug level.

## Note

`codegen-units` changes generated code, so the first CodSpeed run shows a one-time baseline recalibration across all benchmarks (not a regression).

Before/after build step: **6m48s → TBD** (measuring in this run).

> Draft — collecting timing data.


stormslowly:
Closing — both explored levers regress the simulation build vs the 6m48s baseline:

- `codegen-units=16` (github 4-core): **7m14s**. The wide phase already saturates 4 cores; extra codegen units are pure overhead (rspack_core 76s → 102s).
- self-hosted large runner + sccache (warm): **8m11s**. Weak single-core (rspack_core 76s → 216s, 2.8×); the critical path is `cgu=1` single-threaded large crates, which more cores cannot help.

Root cause: the build critical path is single-threaded (cgu=1) codegen of large crates (rspack_core etc.), bottlenecked on single-core performance. github-hosted 4-core + rust-cache (deps cached) at **6m48s is already near-optimal** for this configuration.

PR #14653, opened 2026-07-01
perf: defer persistent cache snapshot updates until save time

ahabhgk:
## Summary
- Collect snapshot and build-dependency updates as deferred tasks instead of applying them immediately
- Merge deferred updates into `FileSystemStorage::save` so persistence work stays batched
- Update persistent cache call sites and tests to validate the new update flow

## Testing
- Added/updated Rust tests for snapshot update generation and build-dependency update application
- Not run (not requested)


ahabhgk:
@codex review

PR #14658, opened 2026-07-01
feat(core): add NormalModuleFactory resolveError hook

### LINKED ISSUE #14640: [Feature]: expose a JS hook to recover from NormalModuleFactory resolution failures
### What problem does this solve?

We are porting [`vercel/ncc`](https://github.com/vercel/ncc) to Rspack in this PR:

- PR: https://github.com/vercel/ncc/pull/1304
- Scenario/comment: https://github.com/vercel/ncc/pull/1304#issuecomment-4846997078
- Working fork used for the port: https://github.com/ScriptedAlchemy/rsncc

ncc preserves a runtime missing-module behavior: for unresolved non-external requests, it rewrites the request to a generated `@@notfound.js?...` module so the bundled app throws at runtime when that missing import path is actually executed.

With the current Rspack JS plugin API, the only behavior-preserving place we found for this is `normalModuleFactory.hooks.beforeResolve`. The plugin calls `normalModuleFactory.getResolver("normal").resolve(...)` to check whether the request exists; if it fails with NotFound, it rewrites the request to ncc's `@@notfound.js?...` module.

That works, but it means successful requests are resolved twice: once by the plugin preflight and once again by Rspack's normal module factory. A reviewer on the ncc PR correctly pointed out that the ideal shape would be to hook the failed-resolution path instead of pre-resolving everything.

The blocker is that Rspack does not appear to expose that recovery point to JS plugins today:

- `beforeResolve` runs before the normal resolver and can only mutate/ignore before knowing whether Rspack resolution fails.
- `afterResolve` and `createModule` only run after successful resolution, so they cannot recover from NotFound.
- `normalModuleFactory.hooks.resolve` is typed/exposed to JS as a mutation hook returning `void`/`JsResolveData`. In the Rust bridge, the JS `NormalModuleFactoryResolveTap` updates `request`/`context` and then returns `Ok(None)`, so JS cannot bail out with a recovered module/result.
- In core, the native `NormalModuleFactoryResolve` hook already has a `NormalModuleFactoryResolveResult`, and the ideal recovery point seems to be the `Err(err)` branch in `resolve_normal_module`, after resolver file/missing dependencies have been collected and before the resolver error is returned.

### Suggested solution

Expose a JS API that runs when normal module resolution fails, before Rspack reports the module-not-found error. For example:

```js
normalModuleFactory.hooks.resolveError.tapAsync(
  "plugin",
  (resolveData, error, callback) => {
    if (isNotFound(error)) {
      resolveData.request = pathToRuntimeNotFoundModule;
      return callback(null, { retry: true });
    }

    callback(error);
  }
);
```

The exact API shape could be different, but it would need to let a JS plugin do one of these:

- see the resolver error and decide whether it is recoverable;
- mutate `resolveData.request` and ask Rspack to retry resolution once;
- or return a recovered module/ignored result from JS, similar in spirit to the native `NormalModuleFactoryResolveResult`.

This would let ncc preserve runtime missing-module semantics without double-resolving every successful module.

### Possible implementation points

From local exploration of Rspack's codebase, the relevant areas seem to be:

- `crates/rspack_core/src/normal_module_factory.rs`
  - `NormalModuleFactoryResolveResult`
  - `NormalModuleFactory::resolve_normal_module`
  - the `Err(err)` branch after `resolve(resolve_args, plugin_driver).await`
- `crates/rspack_binding_api/src/plugins/interceptor.rs`
  - `NormalModuleFactoryResolveTap` currently maps a successful JS tap to `Ok(None)`, so JS cannot currently bail/recover there.
- `packages/rspack/src/NormalModuleFactory.ts`
  - current public hook types for `beforeResolve`, `resolve`, `afterResolve`, and `createModule`

### Alternatives considered

Keeping the current `beforeResolve` preflight preserves ncc behavior, but duplicates resolver work for successful modules.

Using `afterResolve` or `createModule` is too late because those hooks only run after success.

Extending JS `hooks.resolve` to return a real recoverable result may also work, but it still does not directly address the "real resolver failed with NotFound" use case unless it can run after the native resolver error or receive that error.

ScriptedAlchemy:
## Summary

Adds a `resolveError` hook to `NormalModuleFactory` that is called when resolving a module request fails (e.g. the requested module cannot be found). A tap can recover from the failure by mutating the resolve data (such as rewriting `resolveData.request`) and returning `{ retry: true }`, which retries the resolution once with the modified data.

Behavior:

- If no tap requests a retry, the original resolution error is reported as usual.
- The resolution is retried at most once; if the retry also fails, the new error is reported and the hook is not consulted again.
- The hook only fires for ordinary path-resolution failures — not for scheme (URI) specifiers, loader resolution failures, or other error kinds.
- File/missing dependencies recorded by the failed attempt are preserved (extended rather than overwritten), so watch mode still invalidates and rebuilds once the originally-missing file appears.

Implementation notes:

- Rust core: resolution failure is surfaced as a `ResolveNormalModuleOutcome::ResolveFailed` outcome and the retry policy lives in `resolve_normal_module`, keeping the once-function policy-free.
- The hook's `error` argument reuses the canonical `RspackError` binding type.
- Docs added for the new hook (en + zh), marked as Rspack-specific.

## Related links

- Closes https://github.com/web-infra-dev/rspack/issues/14640

## Checklist


ScriptedAlchemy:
Addressed the P2 review finding in 20fed21: module-not-found resolution failures are now tagged with a `ModuleNotFoundError` code (matching webpack's error name) and the `resolveError` retry path is gated to exactly those. Other resolver errors (invalid `package.json`, exports violations, IO errors) surface directly without consulting the hook — covered by a new `resolve-error-not-recoverable` config test whose tap throws if invoked for a malformed-package-json resolution failure. Docs updated to state the module-not-found scope, and stats/diagnostics snapshots updated for the new error code.

ScriptedAlchemy:
Good catch — fixed in e2b658d: `ResolveError::Recursion` no longer gets the `ModuleNotFoundError` code, so cyclic-alias failures never satisfy the retry gate and always surface their diagnostic directly. Verified the existing `resolve-overflow-error` stats case and all resolve-error/diagnostics/error suites still pass.

ScriptedAlchemy:
Addressed in e9461001f6: `MatchedAliasNotFound` is now tagged with `MODULE_NOT_FOUND_ERROR_CODE` (keeping its alias-specific message), so alias-target misses can be recovered via `resolveError` while recursion and non-missing resolver errors remain excluded. Added a `resolve-error-alias` config test that rewrites a broken `resolve.alias` target to a fallback and retries.

ScriptedAlchemy:
Addressed in 60eede4 by taking the suggested approach: module-not-found classification now happens on the original `rspack_resolver` variant (`NotFound` | `MatchedAliasNotFound` | `ExtensionAlias`) in `map_rspack_resolver_error`, before the mapping loses that distinction — the per-arm/per-branch tagging is gone. Extension-alias misses are now recoverable via `resolveError` (new `resolve-error-extension-alias` config test), and recursion plus config/metadata errors remain excluded.

ScriptedAlchemy:
Disagreeing with this one: `matches!(error, ResolveError::NotFound(_) | MatchedAliasNotFound(..) | ExtensionAlias(..))` does not move `error`. The expanded `match` only moves the scrutinee when an arm pattern binds by value; `_`/`..` patterns bind nothing, so `error` remains usable and is passed to `map_rspack_resolver_error_untagged` afterwards. Evidence: `cargo check -p rspack_core` passes locally and the Rust check / all builds are green on this exact head (88cc4d5). Resolving as not-an-issue.

PR #14667, opened 2026-07-03
fix(mf): emit manifest assets with forward-slash names on Windows

### LINKED ISSUE #14309: [Bug]: mf-stats.json and mf-manifest.json emitted twice causing corrupt output with `manifest.filePath` on Windows
### System Info

System:
    OS: Windows 11 10.0.26100
    CPU: (22) x64 Intel(R) Core(TM) Ultra 7 165H
    Memory: 33.41 GB / 63.64 GB
  Binaries:
    Node: 22.22.3 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.8 - C:\Program Files\nodejs\npm.CMD

**Versions:**
- `@rspack/core@2.0.6`
- `@rsbuild/core@2.0.9`
- `@module-federation/enhanced@2.5.1`

## Description

When `manifest.filePath` is set in the `ModuleFederationPlugin` options, `mf-stats.json` and `mf-manifest.json` are emitted twice per compilation on Windows. The second emission can corrupt the files by appending content to the already-written file, resulting in invalid JSON.

## Root Cause

Rspack's built-in Rust-side `MfManifestPlugin` emits the initial manifest assets using `path.join(filePath, statsFileName)`, which on Windows produces backslash-separated paths (e.g. `client\\mf-stats.json`). All other Rspack assets consistently use forward slashes.

`@module-federation/manifest`'s `StatsPlugin` then tries to find the pre-emitted asset via `compilation.getAsset(statsFileName)`, where `statsFileName` is computed by `getManifestFileName()` from `@module-federation/sdk` using forward slashes (e.g. `client/mf-stats.json`):

https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/container/ModuleFederationManifestPlugin.ts#L182
```typescript
return {
  statsFileName: join(filePath, statsFileName),    // → "client\\mf-stats.json" on Windows
  manifestFileName: join(filePath, manifestFileName),
};
```

https://github.com/module-federation/core/blob/main/packages/manifest/src/StatsPlugin.ts#L59
```typescript
const existedStats = compilation.getAsset(this._statsManager.fileName);
// → looks for "client/mf-stats.json", not found → falls through to legacy emitAsset branch
```

Since client\\mf-stats.json ≠ client/mf-stats.json, getAsset() returns undefined and StatsPlugin incorrectly falls into the legacy webpack/old-rspack emitAsset branch – even though modern Rspack is being used. This emits the files a second time onto already-existing assets, causing corruption.

## Expected Behavior

StatsPlugin finds the pre-emitted asset via getAsset() and uses updateAsset() (the modern Rspack path). Each file is emitted exactly once with valid JSON content.

## Proposed Fix

In @rspack/core: Use forward slashes for asset names consistently:
```typescript
// ModuleFederationManifestPlugin.ts
import { posix } from 'node:path';

return {
  statsFileName: posix.join(filePath, statsFileName),
  manifestFileName: posix.join(filePath, manifestFileName),
};
```

## Note
The issue does not occur without filePath since the default asset names contain no path separators.

Image

### Reproduce link

https://github.com/hpoepping/mf-manifest-test

### Reproduce Steps

1. Use @module-federation/enhanced with manifest: { filePath: "client/" } in any Rsbuild setup
2. Build on Windows
3. Observe mf-stats.json and mf-manifest.json appearing twice in the build report with different sizes, and the resulting files containing invalid JSON (content appended after the closing })

The file corruption (invalid JSON with appended content) does not occur consistently – the double emission is reproducible, but whether it results in a corrupt file appears to be non-deterministic. mf-stats.json appears to be affected more often than mf-manifest.json.

johncarmack1984:
## Summary

Fixes #14309.

When `manifest.filePath` is set, `getFileName()` builds the emitted asset names with `path.join`, so on Windows they come out backslash-separated: `client\mf-stats.json`. Compilation asset names always use forward slashes. `@module-federation/manifest`'s `StatsPlugin` looks the assets up by the forward-slash names, finds nothing, and emits both files a second time, as @hpoepping diagnosed in the issue.

This normalizes both names to forward slashes in `getFileName()`, so every consumer of the names (the Rust manifest plugin's `emit_asset`, `IndependentSharedPlugin`, `SharedUsedExportsOptimizerPlugin`) stays consistent. It also handles a user-supplied `filePath` that already contains backslashes.

Verified on windows-latest with a minimal rsbuild + `@module-federation/enhanced` host and `manifest.filePath: 'client/'` (`@rspack/core` 2.1.2). Before the fix, both slash variants exist as separate compilation assets and each file is written twice:

```
mf asset names reported by the compiler:
  ["client/mf-stats.json", "client/mf-manifest.json",
   "client\\mf-stats.json", "client\\mf-manifest.json"]

build output:
  dist\client\mf-manifest.json    0.77 kB      dist\client\mf-manifest.json    0.91 kB
  dist\client\mf-stats.json       0.82 kB      dist\client\mf-stats.json       0.96 kB
```

The double emission is deterministic. The two writes then race for the same path on disk, so the on-disk result ranges from stale manifest content to the appended invalid JSON in the issue report. With the fix, the same app emits each file once, at the forward-slash path, matching the macOS output byte for byte.

New test case `configCases/container-1-5/manifest-file-path-separator` feeds a Windows-shaped `filePath` (`custom\path`) through the real emission pipeline and asserts that no asset name contains a backslash and that the manifest parses at the forward-slash location. It fails before the fix and passes after, on every platform.

## Related links

- #14309

## Checklist

### DIFF
--- a/packages/rspack/src/container/ModuleFederationManifestPlugin.ts
+++ b/packages/rspack/src/container/ModuleFederationManifestPlugin.ts
@@ -173,14 +173,17 @@ export function getFileName(
   const insertSuffix = (name: string, suffix: string): string => {
     return name.replace(JSON_EXT, `${suffix}${JSON_EXT}`);
   };
+  // Asset names always use forward slashes, even on Windows where
+  // `path.join` produces backslash separators.
+  const toAssetName = (name: string): string => name.replace(/\\/g, '/');
   const manifestFileName = fileName ? addExt(fileName) : MANIFEST_FILE_NAME;
   const statsFileName = fileName
     ? insertSuffix(manifestFileName, '-stats')
     : STATS_FILE_NAME;
 
   return {
-    statsFileName: join(filePath, statsFileName),
-    manifestFileName: join(filePath, manifestFileName),
+    statsFileName: toAssetName(join(filePath, statsFileName)),
+    manifestFileName: toAssetName(join(filePath, manifestFileName)),
   };
 }
 
--- a/tests/rspack-test/configCases/container-1-5/manifest-file-path-separator/expose-a.js
+++ b/tests/rspack-test/configCases/container-1-5/manifest-file-path-separator/expose-a.js
@@ -1,0 +1,1 @@
+export default 'expose-a';
--- a/tests/rspack-test/configCases/container-1-5/manifest-file-path-separator/index.js
+++ b/tests/rspack-test/configCases/container-1-5/manifest-file-path-separator/index.js
@@ -1,0 +1,27 @@
+const fs = require("fs");
+const path = require("path");
+
+it("should emit manifest assets with forward-slash names for a Windows-style filePath", () => {
+	const assetNames = __STATS__.assets.map(asset => asset.name);
+	expect(assetNames).toContain("custom/path/mf-manifest.json");
+	expect(assetNames).toContain("custom/path/mf-stats.json");
+	for (const name of assetNames) {
+		expect(name).not.toContain("\\");
+	}
+});
+
+it("should write valid manifest files to the forward-slash location", () => {
+	const manifestPath = path.join(
+		__dirname,
+		"custom",
+		"path",
+		"mf-manifest.json"
+	);
+	const statsPath = path.join(__dirname, "custom", "path", "mf-stats.json");
+	expect(fs.existsSync(manifestPath)).toBe(true);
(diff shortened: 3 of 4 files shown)


ahabhgk (approved):
Thanks!

PR #14673, opened 2026-07-03
fix(esm_library): unwrap dynamic source-phase imports in modern-module output

elecmonkey:
## Summary

In modern-module (ESM library) output, a dynamic source-phase import like `import.source('./add.wasm')` was rendered without unwrapping the module namespace, so callers got `{ default: WebAssembly.Module }` instead of the `WebAssembly.Module` value the source-phase contract promises:

```js
// before
const m = await import("./chunk.js").then(__webpack_require__.bind(...));
new WebAssembly.Instance(m); // ❌ Argument 0 must be a WebAssembly.Module
```

The standard dynamic-import template (`ImportDependencyTemplate`) already appends `.then(m => m["default"])` for source-phase imports, but the ESM library template (`DynamicImportDependencyTemplate`) never checked `ImportDependency::get_phase()`, so the unwrap was missing on this path.

## Fix

`crates/rspack_plugin_esm_library/src/dependency/dyn_import.rs`:

- Route all normal branches through a single exit point that returns the promise expression, instead of each doing its own `source.replace`.
- At that exit, when `get_phase().is_source()`, append `.then((m) => (m["default"]))` (via `returning_function`), mirroring the standard template.

```js
// after
const m = await import("./chunk.js")
  .then(__webpack_require__.bind(...))
  .then((m) => m["default"]); // ✅ WebAssembly.Module
```

## Checklist


JSerFeng:
@codex review

JSerFeng (approved):

PR #14682, opened 2026-07-04
feat(hmr): emit css.c / css.r in the hot-update manifest for precise CSS updates

### LINKED ISSUE #14747: HMR: adding the very first native CSS import never loads the stylesheet (css hmr runtime not registered)
### System Info

rspack `main` (also reproduces on the `feat/hmr-precise-css-updates` branch, PR #14682)

### Details

When a compilation contains **no native CSS module at all**, `CssPlugin` does not inject `CssLoadingRuntimeModule` (and with it the `__webpack_require__.hmrC.css` download handler): the injection is gated on `HAS_CSS_MODULES | CSS_INJECT_STYLE | CSS_STYLE_SHEET` (`crates/rspack_plugin_css/src/plugin/impl_plugin_for_css_plugin.rs`).

If a hot update then adds the **first** native CSS import, the new compilation emits the stylesheet asset (and, with #14682, lists the chunk in the manifest's `css.c`), but `hotCheck` enumerates the `hmrC` handlers of the **old** runtime — updated runtime modules only take effect in the apply phase. Since the old runtime never registered `hmrC.css`, nobody requests the stylesheet: the JS part of the update applies fine while the styles silently never show up until a full reload.

Note the asymmetry: `CssExtractRspackPlugin` injects its css loading runtime whenever `HMR_DOWNLOAD_UPDATE_HANDLERS` (or `ENSURE_CHUNK_HANDLERS`) is required (`crates/rspack_plugin_extract_css/src/plugin.rs`), i.e. always under HMR, so the same "first stylesheet ever" scenario works for extracted CSS. The native CSS runtime should follow the same rule: when HMR is enabled, register the css HMR handler even if the current compilation has no CSS modules yet.

### Reproduce

`rspack.config.js`:

```js
const { rspack } = require('@rspack/core');

module.exports = {
  context: __dirname,
  mode: 'development',
  entry: { main: './src/index.js' },
  plugins: [new rspack.HtmlRspackPlugin({ template: './src/index.html' })],
  module: {
    rules: [{ test: /\.css$/, type: 'css' }],
  },
};
```

`src/index.js` (no css import yet):

```js
document.getElementById('root').textContent = 'no css yet';

import.meta.webpackHot.accept();
```

`src/blue.css`:

```css
body {
  background-color: rgb(163, 255, 255);
}
```

1. `rspack dev`, open the page — no stylesheet, as expected.
2. Add the first css import to `src/index.js`:

```diff
+import './blue.css';
 document.getElementById('root').textContent = 'no css yet';
```

**Expected**: the hot update loads `main.css` and the background turns blue (this is what happens with `CssExtractRspackPlugin` in the same scenario).

**Actual**: the JS update applies, no CSS request is made, the page stays unstyled until a manual reload.

A playwright assertion against this setup fails with the background stuck at `rgba(0, 0, 0, 0)` while the same test shape passes for `CssExtractRspackPlugin` (see `tests/e2e/cases/css/hmr-css-added` on #14682).

stormslowly:
### Why

The CSS HMR handlers re-apply the stylesheet of every chunk in the manifest `c` list. The runtime chunk is always in `c`, so a pure JS edit re-fetches and re-applies CSS that never changed (#11377, the style flash behind rsbuild#7951). There is also no way to remove a stylesheet when a chunk loses its CSS, or to load one when a chunk gains its first CSS.

The browser cannot know a stylesheet is unchanged without downloading it first, so this is only fixable in the compiler. The HMR plugin now collects each chunk's css / mini-css module hashes from the content-hash hook (computed anyway for chunk hashing — no extra work) and diffs them against the previous compilation, so the manifest tells the runtime exactly which stylesheets changed.

Before / after on a js-only edit (Chrome, native CSS):

| stylesheet | before | after |
|---|---|---|
| 0.5 MB / 5k rules | ~50 ms + 1 css request | ~5 ms, 0 requests |
| 2.1 MB / 20k rules | ~170 ms | ~5 ms |
| 4.1 MB / 40k rules | ~385 ms | ~6 ms |

### What

- The manifest gains one field per CSS runtime — `css` (built-in `type: 'css'`, matches webpack's shape) and `miniCss` (`CssExtractRspackPlugin`) — listing the chunks whose CSS changed (`.c`) or went away (`.r`), so each runtime fetches only the stylesheets it needs:

  ```json
  {"c":["runtime","main"],"r":[],"m":[],"miniCss":{"c":["main"]}}
  ```

- The two fields are separate so mixed native+extract builds never fetch each other's stylesheets (a shared list caused `ChunkLoadError`).
- The native css hmr runtime is registered whenever HMR and a css module rule are configured, even before the first css module exists (migrated from #14754). Fixes #14747.
- Runtimes act on their field: absent → no CSS work; `.r` → remove the `<link>`; `.c` with no existing `<link>` → load the new stylesheet.
- The runtime-side `cssTextKey` compare is removed — the manifest already guarantees the fetched CSS differs.
- `fix(runtime)`: the injected context-write IIFE could be swallowed by a trailing `//` comment of the preceding runtime module, emitting a broken hot-update chunk.

### Tests

- `hotCases/css/*` (manifest snapshots, both runtimes): unchanged / reorder / add / remove / chunk-removal / first-css-ever.
- e2e `cases/css/hmr-*` (17 Chromium cases): js-only zero requests, `<link>` add/remove lifecycle, async and never-loaded chunks, splitChunks, mixed-runtime isolation, css modules, esm output.
- e2e `lazy-compilation/css-update`: the rsbuild#7951 flash scenario.

Note: loading newly added CSS requires the filename be derivable from the chunk id (e.g. `[name].css`); hash-carrying names keep the previous behavior — same constraint as webpack.


hardfist:
@hardfist PTAL

hardfist:
@codex review

crates/rspack_core/src/compilation/mod.rs
```diff
@@ -461,6 +461,41 @@ impl Compilation {
...
+  /// appeared or disappeared across a rebuild.
+  pub fn chunk_css_content_signature(&self, chunk_ukey: &ChunkUkey) -> Option<RspackHashDigest> {
+    use rspack_hash::RspackHash;
+    let module_graph = self.get_module_graph();
+    let chunk = self
+      .build_chunk_graph_artifact
+      .chunk_by_ukey
+      .expect_get(chunk_ukey);
+    let chunk_graph = &self.build_chunk_graph_artifact.chunk_graph;
+    let mut css_modules =
```
hardfist: **[P1]** `correctness` — **CSS change signature can miss real stylesheet updates**

`chunk_css_content_signature()` no longer matches the inputs used to render/hash the actual CSS asset. It only looks at `SourceType::Css` plus `css/mini-extract` and then re-sorts by module identifier, but the CSS plugins build the stylesheet from ordered `CssImport` + `Css` modules. That means an `@import`-only change or an import-order change can alter the emitted stylesheet while this signature stays stable, so the hot-update manifest omits `css.c` and the browser keeps stale CSS. Please derive this from the same ordered module list/content-hash path the CSS asset uses instead of the ad-hoc hash here.

hardfist (commented):
## Code Review: 🛑 REQUEST_CHANGES

This PR improves CSS HMR by adding a dedicated `css` namespace to the hot-update manifest and teaching both CSS runtimes to skip unchanged styles, remove vanished stylesheets, and load newly added ones. The added coverage around unchanged/add/remove cases and the runtime-mode comment fix both look directionally good.

The main issue I found is in the new CSS change detection logic: the signature used to decide whether to emit `css.c` is weaker than the actual stylesheet render/hash inputs, so some real CSS changes can be missed and leave stale styles in place. I’d fix that before merging.

**Verdict:** REQUEST_CHANGES — The new manifest plumbing looks good overall, but the CSS signature computation can miss real stylesheet changes and break HMR correctness.

**1 finding(s)** posted as inline comments below.

| Severity | Category | File | Title |
|----------|----------|------|-------|
| P1 | correctness | `crates/rspack_core/src/compilation/mod.rs` | CSS change signature can miss real stylesheet updates |

---
*Automated review by [RomeOS](https://romeos.io) Code Review · commit d717e84*

crates/rspack_core/src/compilation/mod.rs
stormslowly: Fixed in 48c34576ef — the signature now re-invokes the `content_hash` hook and digests the css-related entries **before** the chunk-hash salt is applied in `process_chunk_hash`, so it is derived from exactly the ordered module list (including `CssImport`) both css pipelines render and hash from. A source-type pre-check skips the hook for css-less chunks.

Added a `css-extract-reorder-css` hot case that was red before the fix, and the existing `module-graph/import-order-change` snapshot now correctly reports `css.c` for the reordered native css cascade.

stormslowly:

crates/rspack_core/src/compilation/create_hash/mod.rs
```diff
@@ -563,5 +574,44 @@ async fn process_chunk_hash(
...
+  let output = &compilation.options.output;
+  let digest_of = |source_type: SourceType| {
+    content_hashes
+      .get(&source_type)
+      .map(|hasher| hasher.clone().digest(&output.hash_digest))
+  };
+  let css = digest_of(SourceType::Css).filter(|digest| {
+    let empty_digest = RspackHasher::from(output).digest(&output.hash_digest);
+    *digest != empty_digest
+  });
```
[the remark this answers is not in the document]
stormslowly: Good catch — hoisted: the empty digest is now computed once in `create_hash` and passed down to `process_chunk_hash` / `chunk_css_content_hash`, so the per-chunk check no longer constructs a hasher.

JSerFeng:
@codex review

tests/e2e/cases/css/hmr-async-chunk-css/rspack.config.js
```diff
@@ -0,0 +1,35 @@
...
+  module: {
+    rules: [
+      {
+        test: /\.css$/,
+        type: 'javascript/auto',
+        use: [rspack.CssExtractRspackPlugin.loader, 'css-loader'],
+      },
+    ],
+  },
+  experiments: {
```
chenjiahan: Rspack v2 does not have a `experiments.css` config

tests/e2e/cases/css/hmr-native-css-lifecycle/rspack.config.js
```diff
@@ -0,0 +1,34 @@
...
+  mode: 'development',
+  entry: {
+    main: './src/index.js',
+  },
+  output: {
+    cssFilename: '[name].css',
+    cssChunkFilename: '[name].css',
+  },
+  devServer: {
+    hot: true,
```
chenjiahan: Why do we need `hot: true`?

chenjiahan: It should be the default behavior

tests/e2e/cases/css/hmr-async-chunk-css/src/index.js
```diff
@@ -0,0 +1,5 @@
+import(/* webpackChunkName: "feature" */ './feature.js').then(() => {
+  document.getElementById('root').textContent = 'feature loaded';
+});
+
+import.meta.webpackHot.accept();
```
JSerFeng, line 5:
```suggestion
import.meta.hot.accept();
```

tests/e2e/cases/css/hmr-precise-css-updates/rspack.config.js
```diff
@@ -0,0 +1,34 @@
...
+    rules: [
+      {
+        test: /\.css$/,
+        type: 'javascript/auto',
+        use: [rspack.CssExtractRspackPlugin.loader, 'css-loader'],
+      },
+    ],
+  },
+  experiments: {
+    css: false,
```
JSerFeng: already removed

tests/e2e/cases/css/hmr-native-css-lifecycle/rspack.config.js
```diff
@@ -0,0 +1,34 @@
...
+    rules: [
+      {
+        test: /\.css$/,
+        type: 'css',
+      },
+    ],
+  },
+  experiments: {
+    css: true,
+  },
```
JSerFeng: already removed

tests/e2e/cases/css/hmr-css-added/rspack.config.js
```diff
@@ -0,0 +1,34 @@
...
+      {
+        test: /\.css$/,
+        type: 'javascript/auto',
+        use: [rspack.CssExtractRspackPlugin.loader, 'css-loader'],
+      },
+    ],
+  },
+  experiments: {
+    css: false,
+  },
```
JSerFeng: already removed

tests/e2e/cases/css/hmr-css-added/src/index.js
```diff
@@ -0,0 +1,3 @@
+document.getElementById('root').textContent = 'no css yet';
+
+import.meta.webpackHot.accept();
```
JSerFeng, line 3:
```suggestion
import.meta.hot.accept();
```

tests/e2e/cases/css/hmr-async-chunk-css/rspack.config.js
```diff
@@ -0,0 +1,35 @@
...
+    css: false,
+  },
```
JSerFeng:
```suggestion
```

JSerFeng (dismissed):

tests/e2e/cases/css/hmr-async-chunk-css/src/index.js
stormslowly: Tried it: `import.meta.hot` is not recognized by the parser (`expr_name::IMPORT_META_HOT` only matches `import.meta.webpackHot`), so the accept is silently dropped and every js edit falls back to a full page reload — the js-only test in this PR catches exactly that (link identity marker lost + unexpected css re-request). Keeping `import.meta.webpackHot.accept()`.

crates/rspack_plugin_css/src/runtime/css_loading_with_hmr.ejs
```diff
@@ -33,7 +43,6 @@ var cssTextKeyOfLink = <%- basicFunction("link") %> {
 		}
 		seen[url] = true;
 		var oldTag = cssLoadStylesheet(chunkId, url);
-		if (!oldTag) return;
 		promises.push(
```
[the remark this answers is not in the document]
stormslowly: Confirmed and fixed. A new e2e case (`hmr-mixed-css-runtimes`: native css in an async chunk + extracted css in the entry) reproduced exactly this — an extract-only edit made the native handler fetch a synthesized `main.css?hmr=…`, and the resulting ChunkLoadError rejected the whole hot update. The manifest now hashes the two css kinds separately and ships them in separate namespaces: `css` (native runtime) / `miniCss` (CssExtractRspackPlugin), each consumed only by its own handler.

tests/e2e/cases/css/hmr-css-added/src/index.js
stormslowly: Maybe after `import.meta.hot` is merged.

crates/rspack_core/src/compilation/create_hash/mod.rs
```diff
@@ -563,5 +594,34 @@ async fn process_chunk_hash(
...
+  let digest_of = |source_type: SourceType| {
+    content_hashes
+      .get(&source_type)
+      .map(|hasher| hasher.clone().digest(&output.hash_digest))
+  };
+  ChunkCssHashes {
+    css: digest_of(SourceType::Css).filter(|digest| digest != empty_css_digest),
+    // The content-hash key CssExtractRspackPlugin emits under.
+    mini_css: digest_of(SourceType::Custom("css/mini-extract".into())),
+  }
```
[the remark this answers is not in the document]
stormslowly: Done — the source type is now interned once in a function-local `LazyLock` and copied per chunk.

matthewdavis-oai:
Tested this branch against a large React Router/Rsbuild application where a small TSX edit invalidates a watched Tailwind stylesheet. This exposed one correctness gap in precise CSS HMR and one source of redundant refetches.

The extract loader currently injects `Date.now()` into every hot CSS proxy. A prose-only TSX edit consequently changes the proxy JS and takes the legacy `cssReload`/`reloadAll` path, refetching every stylesheet (20-27 requests per edit in the application) even when CSS is byte-identical.

After removing that timestamp, a real Tailwind-utility edit emitted changed CSS in the `main` stylesheet chunk but did not apply it. The chunk had no changed JS/runtime modules, and `CssUpdate::Changed` is currently added to `miniCss.c` only inside the `!new_modules.is_empty() || !new_runtime_modules.is_empty()` branch. Moving the CSS-update insertion alongside `CssUpdate::Removed` fixes this without incorrectly marking the chunk as a JS update.

I prepared a small patch and an async-TSX/Tailwind browser regression. The regression verifies both:

- prose edit: no `miniCss`, zero CSS requests, same stylesheet link, no navigation;
- new utility: the separate stylesheet chunk appears in `miniCss.c`, exactly one CSS request, computed style updates, and the DOM marker is preserved.

Validation on this PR head with the patch: 18/18 Chromium CSS-HMR tests passed across normal and incremental compilation (including extracted/native CSS, direct edits, added/removed links, async chunks, mixed runtimes, and unchanged JS updates); the focused rebuilds were 78-81 ms. `cargo check --locked --package rspack_plugin_hmr`, rustfmt, Prettier, and `git diff --check` also pass.

Patch/regression: https://gist.github.com/matthewdavis-oai/64c8b0dfef65a326ccbdfc1772099409

Happy to split this into a follow-up PR if that is easier to merge.

stormslowly:
@matthewdavis-oai  thanks a lot for your testing.
I am trying to reproduce this case locally. 

The gist link your provides is 404. Could you plz check the link?

stormslowly:
@matthewdavis-oai Thanks for the detailed review — both problems reproduced and are fixed in 950547b20f, following your analysis:

1. **Loader timestamp removed**: `hotLoader()` no longer injects `// ${Date.now()}` into the css proxy module, so a loader re-run over byte-identical css no longer produces a hot update (and no longer triggers the legacy `cssReload`/`reloadAll` refetch storm).
2. **css-only chunk updates**: the `miniCss.c` insertion was gated behind the js-update branch, so a chunk holding only extracted css (e.g. a splitChunks style chunk) missed its manifest entry when its stylesheet changed — previously masked by the timestamp churn. The insertion now runs independently of js module updates.

Added a regression e2e case `tests/e2e/cases/css/hmr-split-css-chunk` covering both: a real css edit on a css-only chunk applies with exactly one stylesheet fetch, and a byte-identical recompile (a `dep.txt` dependency touch, mimicking Tailwind) produces zero stylesheet requests.

JSerFeng:
@codex review

packages/rspack/src/builtin-plugin/css-extract/loader.ts
```diff
@@ -43,7 +43,6 @@ export function hotLoader(
     if(module.hot) {
       (function() {
         var localsJsonString = ${localsJsonString};
-        // ${Date.now()}
         var cssReload = require(${stringifyRequest(
           context.loaderContext,
           path.join(import.meta.dirname, 'cssExtractHmr.js'),
```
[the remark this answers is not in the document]
stormslowly: Fixed in 4cc67ab82b. With `runtime: false` the plugin now marks the compiler (via a `Symbol.for` key, since the plugin and the loader are bundled separately), and `hotLoader` keeps injecting the `Date.now()` change signal for that configuration only, so css-only edits still reach `cssReload()`. The default path stays signal-free. Covered by a new e2e case `tests/e2e/cases/css/hmr-extract-no-runtime` that fails without the fix.

stormslowly:
@codex code review

crates/rspack_plugin_extract_css/src/runtime/css_loading_with_hmr.ejs
```diff
@@ -17,36 +14,50 @@ var extractCssApplyHandler = function (options) {
...
+	});
+	cssChunkIds.forEach(function (chunkId) {
 		var href = <%- REQUIRE_SCOPE %>.miniCssF(chunkId);
 		if (!href) return;
 		var fullhref = <%- PUBLIC_PATH %> + href;
 		if (seen[fullhref]) return;
 		seen[fullhref] = true;
 		var oldTag = extractCssFindStylesheet(href, fullhref);
-		if (!oldTag) return;
 		promises.push(new Promise(function (resolve, reject) {
```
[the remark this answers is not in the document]
stormslowly: Fixed in 5053df413d — and the native `css` runtime had the same hole, so both handlers are gated now. The no-old-tag path only proceeds when the chunk is actually on the page: it is an initial chunk of this runtime (rendered into the template, covering an initial chunk gaining its first stylesheet), or the jsonp runtime's persistent `hmrS` state records it as fully loaded (covering a loaded async chunk gaining css or re-adding a removed stylesheet). The state expression is rendered via `render_hmr_runtime_state_expression`, so both the classic and the lexical runtime-globals modes resolve it. Regression e2e cases: `hmr-extract-unloaded-async-css` and `hmr-native-unloaded-async-css` (both red before the fix).

stormslowly:
@codex code review

crates/rspack_plugin_css/src/runtime/mod.rs
```diff
@@ -414,11 +416,13 @@ impl RuntimeModule for CssLoadingRuntimeModule {
...
 
-        if with_css_hmr {
+        if with_hmr {
           let source_with_hmr = context.runtime_template.render(
             &self.template_id(TemplateId::WithHmr),
             Some(serde_json::json!({
-              "_is_neutral_platform": is_neutral_platform
+              "_is_neutral_platform": is_neutral_platform,
+              "_initial_chunk_ids": stringify_chunks(&all_initial_chunk_ids, 1),
+              "_js_state_expression": render_hmr_runtime_state_expression(runtime_template, "jsonp"),
```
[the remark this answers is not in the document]
stormslowly: Fixed in d6dcf1fac9. Both css runtimes now derive the hmr state expression from the chunk loading active for the chunk (`render_chunk_loading_hmr_state_expression`: import → `hmrS_module`, importScripts → `hmrS_importScripts`, require/async-node → their keys, custom loaders fall back to jsonp behind the existing `typeof` guard). Regression e2e: `tests/e2e/cases/css/hmr-esm-async-css-added` — `chunkLoading: 'import'`, an async chunk loaded at startup gains its first stylesheet and self-accepts, so the stylesheet can only arrive through the hmr handler; red before the fix (js half applied, stylesheet never requested).

LingyuCoder:

### PR Overview

**Primary classification: Feature development**, with secondary bug-fix and performance aspects. The PR changes the internal CSS HMR contract so native CSS and `CssExtractRspackPlugin` can update, add, or remove only the stylesheets whose rendered content changed. It also fixes hot-update runtime code generation when a preceding runtime module ends in a line comment.

### Previous Design

Both CSS handlers consumed the JavaScript `c` list from the hot-update manifest. Consequently, a JS-only edit re-requested every existing stylesheet associated with an updated chunk and compared CSSOM text in the browser; a missing `<link>` was skipped, so a chunk could neither gain its first stylesheet nor restore one after removal, and CSS removal was not represented precisely. The extract loader's unconditional timestamp also made byte-identical CSS recompilations take the legacy reload path.

### New or Changed Design

The HMR plugin now captures the unsalted native/extracted CSS digests produced by the CSS plugins' `content_hash` hooks, keeps the previous successful compilation's per-chunk baseline, and emits separate `css.{c,r}` and `miniCss.{c,r}` lists. Each runtime consumes only its own namespace, removes vanished links, loads changed/new links, and gates the no-existing-link path to initial chunks or async chunks recorded as loaded by the active JS chunk-loading HMR state. `CssExtractRspackPlugin({ runtime: false })` retains its loader change signal because it has no `hmrC.miniCss` handler.

The regression coverage is broad: unchanged JS edits, CSS-only chunks, add/remove/re-add lifecycle, removed chunks, import-order changes, CSS Modules, mixed native/extract runtimes, multiple updated chunks, unloaded async chunks, `chunkLoading: "import"`, first-CSS-ever, lazy compilation, and `runtime: false` are all represented in hot cases and/or browser E2E cases.

### webpack and Other Bundlers

webpack has an analogous native CSS HMR path and passes `update.css` to download handlers, but current webpack emits only `css.r`; its CSS runtime still iterates the general JavaScript chunk list and serializes `cssRules` to suppress byte-identical replacements. Rspack's `css.c` therefore goes beyond the current webpack implementation by moving change detection to compilation time, while `miniCss` is an additional namespace needed to isolate Rspack's extracted-CSS runtime from native CSS. See webpack's [manifest assembly](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/HotModuleReplacementPlugin.js#L887-L928), [CSS HMR runtime](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/css/CssLoadingRuntimeModule.js#L477-L603), and [handler dispatch](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/hmr/HotModuleReplacement.runtime.js#L397-L413).

### API and Documentation

No public configuration or TypeScript API changes were found. The observable changes are to the internal hot-update manifest and the appended download-handler arguments; existing handlers that ignore extra arguments remain compatible. Adding a stylesheet when the stale runtime cannot derive its filename from the chunk id (for example, a newly introduced `[contenthash]` filename) remains unsupported; the tests intentionally use hash-independent names for this path. This limitation is explained in the PR description, but no permanent documentation change is included.

### Correctness Risks

No material correctness defect was identified in the final CSS HMR implementation. Earlier findings about incomplete CSS signatures, mixed-runtime cross-loading, CSS-only chunks, unloaded async chunks, `runtime: false`, and ESM chunk-loading state are addressed in the final head with focused regressions; all 18 review threads are resolved.

One concrete scope risk remains outside the HMR design: `pnpm-lock.yaml` changes `html-rspack-plugin`/`rspack-vue-loader` from `@rspack/lite-tapable@1.1.4` to `1.1.2` and `ts-checker-rspack-plugin` from `1.1.5` to `1.1.2`, despite no related package-manifest change. This silently changes workspace tooling and is unrelated to precise CSS HMR, so it warrants confirmation or regeneration against the current base.

Local validation on `7da50e2` passed `cargo check --locked --package rspack_plugin_hmr` and `cargo fmt --all --check`. GitHub reports 32 successful checks and 7 conditionally skipped checks; Linux/WASM E2E and the 11-suite Ecosystem CI report are green. Browser E2E was not rerun locally.

The runtime direction is favorable: JS-only edits avoid stylesheet requests and browser-side CSSOM serialization. The PR description reports approximately 50/170/385 ms before versus 5/5/6 ms after for 0.5/2.1/4.1 MB stylesheets; these numbers are author-provided and were not independently reproduced here.

Compilation-side cost is low but non-zero: on HMR builds the plugin clones/digests up to two CSS hashers per hashed chunk and currently constructs the empty native-CSS digest inside the per-chunk hook; HMR runtimes may also include the CSS handler before the first CSS module exists. CodSpeed reports 46 unchanged and 47 skipped benchmarks, Rsdoctor reports no bundle changes in its five projects, and the binding-size report shows +32.13 KB (+0.05%). No CSS-heavy HMR compilation benchmark was found, so that narrow overhead is not directly measured.

<details>
<summary>中文版本</summary>

### PR 概览

**主要分类：功能开发**，同时包含次要的缺陷修复和性能优化。本 PR 修改了内部 CSS HMR 协议，使原生 CSS 与 `CssExtractRspackPlugin` 只更新、添加或移除渲染内容确实发生变化的 stylesheet；同时修复了前一个 runtime module 以行注释结尾时 hot-update runtime 代码生成可能损坏的问题。

### 旧设计

两套 CSS handler 都消费 hot-update manifest 中 JavaScript 的 `c` 列表。因此仅修改 JS 也会重新请求更新 chunk 关联的所有现有 stylesheet，并在浏览器中比较 CSSOM 文本；缺少 `<link>` 时会直接跳过，所以 chunk 无法获得第一份 stylesheet，也无法在移除后重新加载，CSS 移除也没有被精确表达。extract loader 无条件注入的时间戳还会让 CSS 内容未变化的重新编译进入旧的 reload 路径。

### 新设计或设计变更

HMR plugin 现在从 CSS plugins 的 `content_hash` hooks 中捕获未加盐的原生/抽取 CSS digest，保留上一次成功编译的逐 chunk 基线，并分别生成 `css.{c,r}` 与 `miniCss.{c,r}`。每套 runtime 只消费自己的 namespace，移除消失的 link、加载变化或新增的 link，并把“没有现有 link”的路径限制在 initial chunks，或被当前 JavaScript chunk-loading HMR state 记录为已加载的 async chunks。`CssExtractRspackPlugin({ runtime: false })` 没有 `hmrC.miniCss` handler，因此仍保留 loader change signal。

回归覆盖较完整：hot cases 和/或浏览器 E2E 覆盖了 JS 未引起 CSS 变化、CSS-only chunks、添加/移除/重新添加生命周期、chunk 移除、import 顺序变化、CSS Modules、原生/抽取混合 runtime、多个 chunk 同时更新、未加载的 async chunks、`chunkLoading: "import"`、首次添加 CSS、lazy compilation 与 `runtime: false`。

### webpack 与其他打包器

webpack 有类似的原生 CSS HMR 路径，也会把 `update.css` 传给 download handlers，但当前 webpack 只生成 `css.r`；它的 CSS runtime 仍遍历通用的 JavaScript chunk 列表，并序列化 `cssRules` 来抑制内容相同的替换。因此，Rspack 的 `css.c` 通过把变更检测移到编译阶段，超出了当前 webpack 的实现；`miniCss` 则是为了隔离 Rspack 的抽取 CSS runtime 与原生 CSS 而增加的 namespace。参见 webpack 的 [manifest 组装](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/HotModuleReplacementPlugin.js#L887-L928)、[CSS HMR runtime](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/css/CssLoadingRuntimeModule.js#L477-L603) 和 [handler 分发](https://github.com/webpack/webpack/blob/4c2279dbd9663645a64da9598410ac742258839a/lib/hmr/HotModuleReplacement.runtime.js#L397-L413)。

### API 与文档

未发现公共配置或 TypeScript API 变更。可观察到的变化位于内部 hot-update manifest 以及追加的 download-handler 参数；忽略额外参数的现有 handler 保持兼容。当旧 runtime 无法仅根据 chunk id 推导新增 stylesheet 的文件名时（例如首次引入带 `[contenthash]` 的文件名），新增 stylesheet 仍不受支持；相关测试为该路径特意使用了不依赖 hash 的名称。PR 描述解释了这一限制，但没有包含长期文档更新。

### 正确性风险

在最终 CSS HMR 实现中未发现实质性正确性缺陷。此前关于 CSS signature 不完整、混合 runtime 交叉加载、CSS-only chunks、未加载的 async chunks、`runtime: false` 与 ESM chunk-loading state 的问题，都已在最终 head 中修复并加入聚焦回归；全部 18 个 review threads 均已解决。

HMR 设计之外仍有一个明确的范围风险：在没有相关 package manifest 变更的情况下，`pnpm-lock.yaml` 将 `html-rspack-plugin`/`rspack-vue-loader` 使用的 `@rspack/lite-tapable` 从 `1.1.4` 改为 `1.1.2`，并将 `ts-checker-rspack-plugin` 从 `1.1.5` 改为 `1.1.2`。这会静默改变 workspace tooling，且与精确 CSS HMR 无关，因此值得基于当前 base 确认或重新生成。

在 `7da50e2` 上，本地 `cargo check --locked --package rspack_plugin_hmr` 与 `cargo fmt --all --check` 均通过。GitHub 显示 32 个 checks 成功、7 个按条件跳过；Linux/WASM E2E 与包含 11 个 suite 的 Ecosystem CI 报告均为绿色。浏览器 E2E 未在本地重新运行。

### 性能回归风险

runtime 方向是正向的：仅修改 JS 时不再发起 stylesheet 请求，也不再在浏览器中序列化 CSSOM。PR 描述给出的数据是：0.5/2.1/4.1 MB stylesheet 分别从约 50/170/385 ms 降至 5/5/6 ms；这些是作者提供的数据，本次未独立复现。

编译侧成本较低但不为零：HMR 构建中，plugin 会对每个参与 hash 的 chunk clone/digest 最多两个 CSS hasher，并且当前在逐 chunk hook 内构造原生 CSS 的空 digest；HMR runtime 也可能在第一份 CSS module 出现之前就包含 CSS handler。CodSpeed 报告 46 项不变、47 项跳过；Rsdoctor 的五个项目没有 bundle 变化；binding size 增加 32.13 KB（+0.05%）。未找到 CSS-heavy HMR 编译 benchmark，因此这部分窄范围开销没有被直接测量。

</details>

pnpm-lock.yaml
```diff
@@ -3164,8 +3164,8 @@ packages:
       selfsigned:
         optional: true
 
-  '@rspack/lite-tapable@1.1.4':
-    resolution: {integrity: sha512-ZXl+NUVMkLCmi5aLXgM9I8EQ4/UH4tfhQHuZhwKSR8MUOVT1q/LuFs8No4B0UZ512WGruEUhq4FjW3yheKvIhg==}
+  '@rspack/lite-tapable@1.1.2':
```
LingyuCoder: Why we need to modify this

JSerFeng:
@codex review

crates/rspack_plugin_css/src/plugin/impl_plugin_for_css_plugin.rs
```diff
@@ -409,6 +410,20 @@ async fn compilation(
   Ok(())
 }
 
+// Native CSS is configured through module rules (`type: "css*"`), so the runtime
+// chunk must carry the css hmr handler whenever such a rule exists, even before any
+// stylesheet is imported. Otherwise the first stylesheet added via HMR is emitted but
+// never requested, because the runtime baked at the initial build lacks `hmrC.css`.
+fn has_css_module_rule(rules: &[ModuleRule]) -> bool {
```
JSerFeng, line 417: Check the module type through rule config, Is this expected ?

crates/rspack_plugin_hmr/src/lib.rs
```diff
@@ -54,8 +97,85 @@ async fn compilation(
...
+// the digests change exactly when the chunk's css content does.
+#[plugin_hook(CompilationContentHash for HotModuleReplacementPlugin, stage = 100)]
+async fn content_hash(
+  &self,
+  compilation: &Compilation,
+  chunk_ukey: &ChunkUkey,
+  hashes: &mut HashMap<SourceType, RspackHasher>,
+) -> Result<()> {
+  // The content-hash key CssExtractRspackPlugin emits under.
+  static MINI_EXTRACT_CSS: LazyLock<SourceType> =
```
JSerFeng, line 111: Should we use &str here directly.

crates/rspack_plugin_runtime/src/runtime_module/utils.rs
```diff
@@ -68,6 +69,33 @@ pub fn render_hmr_runtime_state_expression(
...
+      chunk.get_entry_options(&compilation.build_chunk_graph_artifact.chunk_group_by_ukey)
+    })
+    .and_then(|options| options.chunk_loading.as_ref())
+    .unwrap_or(&compilation.options.output.chunk_loading);
+  let key = match chunk_loading {
+    ChunkLoading::Enable(ChunkLoadingType::Import) => "module",
+    ChunkLoading::Enable(ChunkLoadingType::ImportScripts) => "importScripts",
+    ChunkLoading::Enable(ChunkLoadingType::Require) => "require",
+    ChunkLoading::Enable(ChunkLoadingType::AsyncNode) => "readFileVm",
+    _ => "jsonp",
```
LingyuCoder: Should `CunkLoadingType::Custom(_)` also use `jsonp`?

crates/rspack_plugin_css/src/plugin/impl_plugin_for_css_plugin.rs
stormslowly: It should be determined ahead of time even if there are no CSS modules in compilation.
This flag is used to handle the case: 
in the first compilation, there are no CSS modules; in the second compilation, a static-imported CSS module. So we need the CSS module HMR loading runtime code deployed before so the second compilation's css chunk can be loaded to the browser.

crates/rspack_plugin_hmr/src/lib.rs
stormslowly: In my opinion, the SourceType key is more expressive than primitive str.

LingyuCoder (approved):

PR #14711, opened 2026-07-07
chore(ci): surface Windows Defender status after disable attempt

stormslowly:
The "Disable CPU hogs (Windows)" step runs with `continue-on-error: true` and `$ErrorActionPreference = 'SilentlyContinue'`, so a green check does **not** prove Defender realtime monitoring was actually disabled. `Set-MpPreference -DisableRealtimeMonitoring` is commonly rejected silently when Tamper Protection is enabled, and the swallowed error leaves us blind to it.

This makes the actual Defender state observable in the logs.

## Before
```
✓ Disable CPU hogs (Windows)   # green even if nothing was disabled
```

## After
```
== Windows Defender status (post disable attempt) ==
RealTimeProtectionEnabled : False
IsTamperProtected         : False
AntivirusEnabled          : True
```

Motivation: a recent `main` Windows build failed at `cargo codegen` with `serde_derive.dll: LoadLibraryExW failed: Invalid access to memory location. (os error 998)`. That loader flake is often attributed to AV scanning freshly-linked proc-macro DLLs, but we currently cannot confirm whether Defender is truly off on these runners. This unblocks that diagnosis without changing the disable behavior itself.


stormslowly:
Addressed the review: extracted the whole "Disable CPU hogs (Windows)" step (disable + Defender-status logging) into a shared composite action \`.github/actions/windows-disable-cpu-hogs\`, and replaced all three inline copies with a \`uses:\` reference. This removes the pre-existing triplication so the snippet now lives in one place (35658ef).

hardfist (approved):

PR #14735, opened 2026-07-09
fix: clean up runtime global rendering

LingyuCoder:
## Summary

- render Rspack ESM module registration through `__rspack_context.m.add` while keeping webpack output on `__webpack_require__.add`
- treat control-only runtime globals as non-renderable require-scope fields
- update runtime-mode ESM snapshots after removing unused control bindings


LingyuCoder:

ahabhgk (approved):

PR #14748, opened 2026-07-10
perf: deduplicate repeated dynamic reexport code

JSerFeng:
## Summary

- Add a shared runtime helper for repeated dynamic `export *` code.
- Keep the existing inline code path when a module has fewer than 16 dynamic star reexports.
- Preserve the existing checked reexport implementation; sharing that path was measured but did not improve compressed size.
- Compare the shared runtime against the inline path and cover legacy getter capture semantics.

Dynamic star reexports currently emit the same getter collection loop for every dependency. This change moves that loop into `__webpack_require__.re` only when the helper cost is amortized, while keeping the low-cardinality output unchanged.

For 16 dynamic reexports in a production build with the SWC minimizer, compared with a mechanical old-inline baseline:

The helper preserves source enumeration order, single and multiple exclusion handling, environment-specific getter capture, and the existing define-after-collection behavior.

The integration coverage compares 15 inline reexports with 16 runtime reexports in the same build. It checks every exported value, duplicate export precedence, default and named exclusions, and live bindings. A separate legacy-output case covers the `var + bind` getter path when `const` and arrow functions are disabled.

## Related links

N/A

## Checklist


JSerFeng:
@codex review

LingyuCoder (approved):

PR #14754, opened 2026-07-11
fix(css): register css hmr runtime when native css has no module yet

stormslowly:
## Why

When a native-css project's initial build has **no css module**, `CssPlugin` never injects `CssLoadingRuntimeModule`, so the `hmrC.css` download handler is missing from the initial runtime chunk. The runtime chunk is not re-emitted on HMR, so once the first stylesheet is added by a hot update there is no handler to request it — the JS update applies while styles silently never load.

`CssExtractRspackPlugin` already registers its css loading runtime whenever HMR is enabled, so the same "first stylesheet ever" scenario works for extracted css. Native css should follow the same rule.

## What

Register the css hmr runtime (and provide its `GET_CHUNK_CSS_FILENAME` / `PUBLIC_PATH` dependencies) whenever HMR is enabled **and** a native css module rule (`type: "css*"`) is configured — even if the compilation has no css module yet. Detection scans module rules (native css is enabled via rules, not the deprecated `experiments.css`), and the scan only runs under HMR, so non-watch builds are unaffected.

Added a hot case asserting `__webpack_require__.hmrC.css` is registered when native css is configured but not yet imported.

## Scope

This is the **registration** half of #14747. The handler logic that actually attaches the newly-added stylesheet (removing the `if (!oldTag) return` bail and driving from the `css.c` / `css.r` hot-update manifest) lands in #14682; together they make the first css load end-to-end. Registration must happen here because the runtime chunk is baked once at the initial build.

ref #14747, #14682

### DIFF
--- a/crates/rspack_plugin_css/src/plugin/impl_plugin_for_css_plugin.rs
+++ b/crates/rspack_plugin_css/src/plugin/impl_plugin_for_css_plugin.rs
@@ -8,10 +8,11 @@ use rspack_core::{
   Compilation, CompilationContentHash, CompilationId, CompilationParams, CompilationRenderManifest,
   CompilationRuntimeRequirementInTree, CompilerCompilation, CssBuildInfo, CssModuleRenderCondition,
   DependencyType, ManifestAssetType, Module, ModuleFactoryCreateData, ModuleGraph,
-  ModuleIdentifier, ModuleType, NormalModuleCreateData, NormalModuleFactoryAfterResolve,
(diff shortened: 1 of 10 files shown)


stormslowly:
Migrated into #14682 (commit c4d3c0566d) so the registration half and the handler half of #14747 land together — the branch now also carries an end-to-end browser test (`tests/e2e/cases/css/hmr-native-first-css`) proving the very first native css import hot-loads.

PR #14757, opened 2026-07-12
fix(watcher): serialize watch and close on the native watcher

### LINKED ISSUE #14790: Native watcher watch/close overlap permits undefined behavior and can crash
## Summary

## Minimal reproduction shape

```js

  );
  watch();
  watch();
  watcher.pause();
}
```

## Proposed boundary

stormslowly:
## Why

On macOS, tearing down and restarting the native watcher (`experiments.nativeWatcher`) intermittently aborts the process:

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
    reason: '-[OS_os_log addObject:]: unrecognized selector sent to instance ...'
libc++abi: terminating due to uncaught exception of type NSException  →  SIGABRT
```

Symbolicated crash stack (release-debug build), on a tokio worker thread:

```
notify::fsevent::FsEventWatcher::append_path        // CFArrayAppendValue(self.paths, ...)
notify::fsevent::FsEventWatcher::watch
rspack_watcher::disk_watcher::DiskWatcher::watch
rspack_watcher::FsWatcher::wait_for_event
rspack_binding_api::native_watcher::NativeWatcher::watch   // spawned task
```

A use-after-free between two entry points that were never kept apart:

- `watch()` spawns a **detached** task and returns to JS immediately. The task registers paths with FSEvents via `append_path`, which mutates the watcher's `CFMutableArray` (`self.paths`).
- `close()` drops the watcher — `Drop` → `CFRelease(self.paths)`.
- They land on the *same* instance: `NativeWatchFileSystem` memoizes one `NativeWatcher` per compiler and clears `#inner` only after `close()`'s promise resolves. The `# Safety` note on `close` asserted callers must not overlap the two, but nothing enforced it.

So `append_path` writes into the CFArray while `close` frees it. The freed slot gets reused by an `os_log` object, and the next `CFArrayAppendValue` dispatches `addObject:` to a non-array → NSException → SIGABRT.

**Where the window is.** `DiskWatcher::watch` calls into notify only for patterns it isn't already watching, and on macOS `WatcherRootAnalyzer` collapses every dependency into a single recursive common root that is stable across rebuilds. `append_path` therefore runs on a watcher's **first** `watch()` — one recursive FSEvents subscription over the whole project root, which is far from instantaneous — while steady-state rebuilds dedupe to a no-op. The race is a `close()` (or a second `watch()`, which would also see an empty pattern set) landing inside that first registration.

Instrumenting a real consumer confirms it is reachable: Rstest rebuilds its dev server whenever a config file changes, and its watch-mode restart test edits the config while startup is still in progress. `close()` arrives **3 ms** after the first `watch()` on the same memoized `NativeWatcher`, and the process aborts in `append_path`.

This is distinct from the `analyzer/root.rs` path-tree assert — it still reproduces with that fix applied.

## What

Serialize the two entry points that reach the disk watcher with an async mutex:

- a watch task acquires it **before** touching the watcher, and holds it across the whole `watcher.watch()`;
- `close` aborts the in-flight watch task, then acquires the same mutex before tearing the watcher down.

`watch` and `close` are the only paths that reach `DiskWatcher` (`FsWatcher::watch` → `wait_for_event` → `disk_watcher.watch()`, and `FsWatcher::close` → `disk_watcher.close()`). `pause` and `trigger_event` take `&self` and only touch the executor / trigger, never the FSEvents state, so they stay outside the lock — the `# Safety` note is updated to say exactly that.

Cancellation-safe by construction: a future can only be dropped at an `.await`, never mid-FFI, so an aborted task always releases the guard with the watcher in a consistent state. Holding the mutex in `close` is proof that no task is inside `append_path`. It cannot deadlock either — `abort()` always precedes the lock wait, so the current holder is guaranteed to reach an await and release.

Also adds `rspack_napi::runtime::spawn_handle`, which returns the `JoinHandle` needed to abort (the existing `spawn` drops it).

> **Rejected alternative** (raised in review): have each new watch task `abort()` **and `await`** its predecessor. Not cancellation-safe — `close` aborting the head of the chain cancels it at its `prev.await`, which drops the predecessor's `JoinHandle` and **detaches** it; the abandoned tasks keep running into `append_path` while `close` drops the watcher. The test below reproduced both the resulting SIGSEGV and a hang (notify's `stop()` spinning forever on `CFRunLoopIsWaiting` against corrupted state), which is why this uses a lock.

## Test

`tests/rspack-test/NativeWatcherLifecycle.test.js` drives `binding.NativeWatcher` directly: overlapping watch tasks on a fresh watcher followed by an immediate close, repeated. The file set is large enough that the work preceding `append_path` (path-manager update, mtime baselines, path-tree rebuild) keeps the registration in flight long enough for the close to land on it. It runs in the `nativeWatcher` project (`retry: 0`, `maxConcurrency: 1`) so a native abort cannot be masked by a retry.

## Verification

macOS arm64, binding built from this branch:

| | pre-fix | abort+await chain | this PR (mutex) |
| --- | --- | --- | --- |
| `NativeWatcherLifecycle` | **6/6 crash** | 4/6 fail (3 hang, 1 SIGSEGV) | **6/6 pass** |

close: https://github.com/web-infra-dev/rspack/issues/14790

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -470,8 +470,8 @@ export declare class NativeWatcher {
    * # Safety
    *
    * This function is unsafe because it uses `&mut self` to call the watcher asynchronously.
-   * It's important to ensure that the watcher is not used in any other places before this function is finished.
-   * You must ensure that the watcher not call watch, close or pause in the same time, otherwise it may lead to undefined behavior.
+   * `watch` and `close` are serialized internally, but calling `pause` while this function is
+   * in flight is still undefined behavior.
    */
   close(): Promise<void>
   pause(): void
--- a/crates/rspack_binding_api/src/native_watcher.rs
+++ b/crates/rspack_binding_api/src/native_watcher.rs
@@ -2,6 +2,7 @@ use std::{
   boxed::Box,
   panic::AssertUnwindSafe,
   path::{Path, PathBuf},
+  sync::Arc,
   time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
@@ -59,6 +60,8 @@ pub struct NativeWatchUndelayedEvent {
 pub struct NativeWatcher {
   watcher: FsWatcher,
   closed: bool,
+  watch_task: Option<tokio::task::JoinHandle<()>>,
+  watcher_lock: Arc<tokio::sync::Mutex<()>>,
 }
 
 fn timestamp_to_system_time(millis: u64) -> SystemTime {
@@ -81,6 +84,8 @@ impl NativeWatcher {
     Self {
       watcher,
       closed: false,
+      watch_task: None,
+      watcher_lock: Arc::new(tokio::sync::Mutex::new(())),
     }
   }
 
@@ -110,8 +115,15 @@ impl NativeWatcher {
 
     let start_time = start_time.get_u64().1;
 
+    if let Some(prev) = self.watch_task.take() {
+      prev.abort();
+    }
+
+    let watcher_lock = Arc::clone(&self.watcher_lock);
+    let mut watch_task = None;
     reference.share_with(env, |native_watcher| {
-      rspack_napi::runtime::spawn(async move {
+      watch_task = Some(rspack_napi::runtime::spawn_handle(async move {
+        let _guard = watcher_lock.lock_owned().await;
         native_watcher
           .watcher
           .watch(
@@ -123,9 +135,10 @@ impl NativeWatcher {
             Box::new(js_event_handler_undelayed),
           )
           .await
-      });
+      }));
       Ok(())
     })?;
+    self.watch_task = watch_task;
 
     Ok(())
   }
@@ -160,6 +173,10 @@ impl NativeWatcher {
     let shared_reference = reference.share_with(Env::from_raw(env.raw()), |native_watcher| {
       rspack_napi::runtime::spawn(async move {
         let result = AssertUnwindSafe(async {
+          if let Some(task) = native_watcher.watch_task.take() {
+            task.abort();
+          }
+          let _guard = Arc::clone(&native_watcher.watcher_lock).lock_owned().await;
           native_watcher
             .watcher
             .close()
--- a/crates/rspack_binding_api/src/native_watcher.rs
+++ b/crates/rspack_binding_api/src/native_watcher.rs
@@ -161,8 +161,8 @@ impl NativeWatcher {
   /// # Safety
   ///
   /// This function is unsafe because it uses `&mut self` to call the watcher asynchronously.
-  /// It's important to ensure that the watcher is not used in any other places before this function is finished.
-  /// You must ensure that the watcher not call watch, close or pause in the same time, otherwise it may lead to undefined behavior.
+  /// `watch` and `close` are serialized internally, but calling `pause` while this function is
+  /// in flight is still undefined behavior.
   pub unsafe fn close<'env>(
     &mut self,
     env: &'env Env,
--- a/crates/rspack_napi/src/runtime.rs
+++ b/crates/rspack_napi/src/runtime.rs
@@ -38,6 +38,14 @@ where
   std::mem::drop(spawn_inner(future));
 }
 
+pub fn spawn_handle<F>(future: F) -> tokio::task::JoinHandle<F::Output>
+where
+  F: Future + Send + 'static,
+  F::Output: Send + 'static,
+{
+  spawn_inner(future)
+}
+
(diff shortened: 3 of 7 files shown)


crates/rspack_binding_api/src/native_watcher.rs
```diff
@@ -110,8 +112,13 @@ impl NativeWatcher {
 
     let start_time = start_time.get_u64().1;
 
+    if let Some(prev) = self.watch_task.take() {
+      prev.abort();
+    }
```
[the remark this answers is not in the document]
stormslowly: Fixed in c4db59c.

Confirmed the reasoning holds: `abort()` only takes effect at an `.await`, so the previous task can still be executing the synchronous `FsWatcher::watch` → `wait_for_event` → `disk_watcher.watch` → FSEvents `append_path` section when the new task starts. That section is not short either (`path_manager.update`, `record_initial_file_mtimes` stat syscalls, `analyzer.analyze`, `scanner.scan`), so two tasks could end up in `CFArrayAppendValue` on the same `CFMutableArray` concurrently — the same corruption this PR fixes. The asymmetry was the giveaway: `close()` already did `abort()` **+ `await`**, `watch()` only aborted.

Moved the abort+await into the newly spawned task, before it touches the watcher:

```rust
let prev_watch_task = self.watch_task.take();

reference.share_with(env, |native_watcher| {
  watch_task = Some(rspack_napi::runtime::spawn_handle(async move {
    if let Some(prev) = prev_watch_task {
      prev.abort();
      let _ = prev.await;
    }
    native_watcher.watcher.watch(...).await
  }));
  Ok(())
})?;
```

Re-verified on macOS arm64 with the native watcher enabled (config-restart reproduction, binding built from this branch):

| | result |
| --- | --- |
| restart crash rate | 0/10 (was ~50% before this PR) |
| normal watch (create / update / delete + CLI shortcuts) | 5/5 pass |

The second row matters here: `prev.await` now runs on every watch cycle, so it confirms the added wait does not stall the watch loop.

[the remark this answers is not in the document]
stormslowly: Following up: your diagnosis was right, but the remedy was not — and the regression test I added for it caught that.

Implementing exactly what was suggested (abort **and await** the previous `JoinHandle` inside the newly spawned task) produced a binding that still failed 4/6 runs of the new test: 3 hangs and 1 SIGSEGV.

The chain is not cancellation-safe. Five `watch()` calls in one tick chain as A→B→C→D→E, each awaiting its predecessor. `close()` takes only E, aborts it and awaits it — but E is parked at `prev.await`, so cancelling it there **drops D's `JoinHandle`, detaching D**. D/C/B/A keep running while `close()` drops the watcher, so `append_path` still hits freed memory (SIGSEGV), or notify's `stop()` spins forever on `CFRunLoopIsWaiting` against the corrupted state (the 30s hangs). Aborting the head of a `JoinHandle` chain abandons the tail.

So this revision serializes with an async mutex instead: a watch task acquires it before touching the watcher and holds it across the whole `watch()`, and `close` aborts the in-flight task and then acquires the same mutex before dropping. A future can only be dropped at an `.await`, never mid-FFI, so an aborted task always releases the guard with the watcher consistent — and holding the mutex in `close` is proof that no task is inside `append_path`. No task holds another's handle, so nothing can be detached.

| | pre-fix | abort+await chain | mutex |
| --- | --- | --- | --- |
| `NativeWatcherLifecycle` | 6/6 crash | 4/6 fail (3 hang, 1 SIGSEGV) | **6/6 pass** |
| `NativeWatcher.part1/2/3` | — | — | 122 / 55 / 60 pass |

crates/rspack_napi/src/runtime.rs
hardfist, line 1: why need spawn_handle here and what does spawn_handle mean here?

crates/node_binding/napi-binding.d.ts
```diff
@@ -470,8 +470,8 @@ export declare class NativeWatcher {
    * # Safety
    *
    * This function is unsafe because it uses `&mut self` to call the watcher asynchronously.
-   * It's important to ensure that the watcher is not used in any other places before this function is finished.
-   * You must ensure that the watcher not call watch, close or pause in the same time, otherwise it may lead to undefined behavior.
+   * `watch` and `close` are serialized internally, but calling `pause` while this function is
```
hardfist, line 473: so calling pause while watching still can cause panic | abort issue?

hardfist, line 473: this api's semantic is not same as js version

stormslowly:
We have a better one https://github.com/web-infra-dev/rspack/pull/14829

PR #14758, opened 2026-07-12
perf: optimize ESM linking passes

JSerFeng:
## Summary

- remove an unused per-module output-path resolution pass before ESM module analysis
- process namespace-object propagation as an insertion-ordered worklist without cloning the full set each round
- borrow hoisted module sets during link traversal instead of cloning them

## Why

ESM linking performed async output-path resolution once per module and discarded every result. This added serial work on a high-cardinality structure before the parallel module analysis phase. Namespace propagation and hoisted-module traversal also cloned collections that can grow with the module graph.

The new flow preserves linking order and emitted output while removing those repeated computations and allocations.

## Checks

- `pnpm run build:cli:dev`
- `pnpm exec rstest EsmOutput.test.js` (456 passed)
- `cargo clippy -p rspack_plugin_esm_library --all-targets --all-features`
- `cargo fmt --all --check`
- `git diff --check`
- `pnpm run test:unit` (9 host-contention timeout flakes; every timed-out case passed on isolated rerun)


JSerFeng:
@codex review

SyMind (approved):

PR #14759, opened 2026-07-12
perf: parallelize delayed chunk-group walks in code splitting

JSerFeng:
## Summary

- keep the existing `build_chunk_graph` queue, connect/merge loop, named chunk, and cycle behavior as the source of truth
- parallelize only independent async `ProcessBlock` roots already emitted into the legacy delayed queue
- snapshot each root's `min_available_modules`, runtime, and chunk mask, then perform the read-only graph walk with Rayon
- validate speculative results immediately before ordered commit and rerun stale roots through the original serial queue
- merge all pending available-module masks once per unique chunk group, then replay roots in the original order
- preserve the serial walk identity as `(dependencies block, module context)`, including the module context forwarded to async blocks
- batch existing non-entrypoint async-block connections during commit, while flushing before legacy chunk-group creation paths
- avoid runtime clones, empty leaf-module extraction, and duplicate global-order checks on the commit path
- enable the parallel path by default; set `RSPACK_EXPERIMENTAL_PARALLEL_CODE_SPLITTING=0` for an emergency serial rollback

This does not collect pending roots ahead of the existing merge loop. The old loop still decides when work becomes available, which preserves its handling of cycles and `webpackChunkName` merges.

### Aeolus `apps/abi`

Production full build, cold Rspack cache for every sample, release binding, and the `Compilation:code_splitting` span:

| mode | samples | median | mean |
| --- | --- | ---: | ---: |
| serial delayed-root path | 389 / 343 / 345 / 343 ms | 344 ms | 355 ms |
| parallel delayed-root path | 190 / 191 / 190 / 186 ms | 190 ms | 189.3 ms |

The isolated code-splitting phase improved by **44.8% at the median** and **46.7% at the mean**. Full-build wall time improved by 2.1% at the median (26.88 s to 26.33 s); code splitting is a small share of the full build.

### In-repo walltime benchmark

The benchmark builds 64 independent async chunk groups, each with 8 branches and 16 leaves. It is registered in the walltime target so Rayon uses the real multi-threaded pool; the simulation target still covers single-thread fallback overhead.

| mode | median |
| --- | ---: |
| serial rollback | 4.396 ms |
| parallel default | 2.479 ms |

Criterion reported **-43.9%** (`p < 0.05`) for the code-splitting benchmark.

The local stripped Darwin ARM64 CI-profile binding also decreased by 16,552 bytes versus the previously failing PR head after removing temporary phase/cardinality instrumentation. Linux CI remains the authoritative binary-size check.

## Correctness

The regression case covers a shared named async chunk, changing available-module inputs, shared modules across roots, and a cyclic dependency.

Every speculative result is validated immediately before its ordered commit. A stale result is rerun synchronously with the original `process_queue`; later independent results remain eligible and are validated against any state changed by that rerun. Default-on and serial-rollback builds emit byte-identical files for both config and runtime-mode variants of the regression case.

### DIFF
--- a/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
+++ b/crates/rspack_core/src/chunk_graph/chunk_graph_chunk.rs
@@ -367,13 +367,20 @@ impl ChunkGraph {
   }
 
   pub fn connect_chunk_and_modules(&mut self, chunk: ChunkUkey, modules: &[ModuleIdentifier]) {
-    for module in modules.iter() {
-      let cgm = self.expect_chunk_graph_module_mut(*module);
-      cgm.chunks.insert(chunk);
-    }
-    let cgc = self.expect_chunk_graph_chunk_mut(chunk);
+    let cgc = self
+      .chunk_graph_chunk_by_chunk_ukey
+      .entry(chunk)
+      .or_default();
+    cgc.modules.reserve(modules.len());
     for module in modules.iter() {
       cgc.modules.insert(*module);
+    }
+    for module in modules.iter() {
+      let cgm = self
+        .chunk_graph_module_by_module_identifier
+        .entry(*module)
+        .or_default();
+      cgm.chunks.insert(chunk);
     }
   }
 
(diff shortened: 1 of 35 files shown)


JSerFeng:
@codex review

Issue #14770, opened 2026-07-13
[Perf]: identical SplitChunks work and ChunkSplit mutations repeat on every hot rebuild


matthewdavis-oai:
### Summary

After fixing a false-negative in Rspack's code-splitting cache, a visible, constant-length `leaf.tsx` edit in a large Rsbuild/React Router app consistently reuses the chunk graph but still executes the full SplitChunks pass. The client graph has roughly **408 route entries / 5,452 chunks**; the edited module builds in **4-5 ms**, while SplitChunks costs **0.81-1.40 s** on each hot rebuild.

This is distinct from the existing leaf/chunk-graph report https://github.com/web-infra-dev/rspack/issues/14768 and from HMR `process_assets` https://github.com/web-infra-dev/rspack/issues/14767: those address earlier/later stages. There is currently no open SplitChunks-incremental issue or PR.

### Attribution

Four consecutive hot edits in the post-cache-fix native trace:

```text
                                  edit 1   edit 2   edit 3   edit 4
code_splitting (client, reuse)       26ms     29ms     28ms     21ms
SplitChunks                         1.15s    0.85s    0.81s    1.40s
root.css loader chain               1.13s    1.12s    0.93s    1.00s
leaf.tsx                               ~5ms     ~5ms     ~5ms     ~5ms
```

Each cache decision sees seven mutations/five affected modules and unchanged topology. SplitChunks' two `prepare_module_group_map` spans account for only about **80-190 ms combined**; most of the pass is repeated full-graph grouping/application work. The browser run completes true HMRs with no CSS requests or reloads. The CSS row is the complete loader chain and is a separate opportunity, not time attributed to SplitChunks.

Artifacts:

```text
trace:       hmr-codesplit-focused-20260713.jsonl
sha256:      2f2ceb002d18fc6d8dbb6084a3aae22eee1c836f94dd541de70a90ae5e51b7dc
HMR result:  rspack-codesplit-candidate-focused-result.json
sha256:      edbbf933c8270778849a35238505f3720199bee4a3e0050e9fc22df8a8ec874f
cache patch: rspack-codesplit-outgoing-cache-fix.patch
sha256:      c0ccae88e9588754089a71d4aa7a8136f0592005f89f6234302694af847f01b1
native:      rspack.linux-x64-gnu.node
sha256:      43aeb4bb914277fba080227d85de59fd45c73f329b393ef05be87d81e3ff2418
```

### Why it repeats

[...]

Issue #14765, opened 2026-07-13
Persistent-cache snapshot work scales with unchanged Tailwind file dependencies during HMR


matthewdavis-oai:
### Problem

While profiling a one-file TSX HMR edit in a large Rsbuild/Rspack application, the TSX transform itself takes about 28 ms, but a warm rebuild still spends roughly 0.4--0.8 s in persistent-cache snapshot/metadata work. The compilation has thousands of Tailwind-watched source files. Making the persistent cache read-only consistently removes about 0.7--0.8 s from the web rebuild, so this is synchronous rebuild work rather than background cache writes.

### Findings

1. `Compilation::{context,missing,build}_dependencies()` appended `self.file_dependencies` to their respective added iterators. This made every compilation-level file dependency also look like an added context, missing, and build dependency. The chains must append `self.context_dependencies`, `self.missing_dependencies`, and `self.build_dependencies` respectively.
2. Loaders such as Tailwind can remove and re-add thousands of unchanged watched files during a rebuild. Forwarding every re-added file to `save_snapshot` is expensive, but filtering that set only by `compilation.modified_files` is not generally safe: `Watching.invalidate()`, directory-only watcher provenance, or a loader/plugin mutating an already-tracked file during compilation can omit the exact file and allow a later cold-cache start to accept stale output.

### Current status

The compilation dependency-delta correctness fix and focused disjoint-iterator regression are stacked in #14772 (`601ac45`): context, missing, and build dependencies now append only to their corresponding added iterators. The regression is `compilation_level_dependencies_stay_in_their_own_added_iterators`.

The more aggressive re-added-file snapshot filter is deliberately **not** included. The remaining work for this issue is to establish a safe snapshot-provenance contract and measure a correctness-preserving filter. Directory, missing, build, added, and removed dependency semantics must remain intact.

The earlier standalone patch/digest and its snapshot regression are historical experiments, not a supported patch to apply. This snapshot investigation is independent of the precise CSS-HMR work in #14682: reducing compiler-side metadata work still matters when an edited TSX produces identical CSS.

matthewdavis-oai:
The three compilation dependency-delta chains and a focused disjoint-iterator
regression are now included in #14772: compilation-level context, missing, and
build dependencies append only to their corresponding added iterators rather
than being appended to every file-backed delta.

The more aggressive exact-file snapshot filter is deliberately deferred.
`Watching.invalidate()` or directory-only watcher provenance can omit an
[...]

Issue #6633, opened 2024-05-28
[Bug]: hmr failed when adding or modifying new initial chunk


SoonIter:
### System Info

  System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Max
    Memory: 296.03 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.13.1 - ~/Library/Caches/fnm_multishells/996_1716862511218/bin/node
    npm: 10.5.2 - ~/Library/Caches/fnm_multishells/996_1716862511218/bin/npm
    pnpm: 8.15.4 - ~/Library/Caches/fnm_multishells/996_1716862511218/bin/pnpm
  Browsers:
    Chrome: 125.0.6422.112
    Chrome Canary: 127.0.6504.0
    Safari: 17.5

### Details

Cannot find module '@icon-park/react'

It seems to be an issue of esm hmr.

Add the import in the devlopment environment, only updated `index.js`, but not updated `vendors-node_modules_pnpm_icon-park_react_1_4_2_react-dom_18_3_1_react_18_3_1_nod E_modules.js` (the bundle of the iconPark package)

![faKUSl028L](https://github.com/web-infra-dev/rspack/assets/79413249/c3f12323-8305-4813-92f2-b7bb988b8c2f)

### Reproduce link

https://github.com/SoonIter/rsbuild-provided-exports-issue

### Reproduce Steps

1. pnpm install
2. npm run dev
3. modify index.js, and add import from iconPark
```
import {
  SunOne,
  // ACane,      <--- add import
} from '@icon-park/react';
```

SoonIter:
There is still a problem with this place in 0.7.0
<img width="400" alt="image" src="https://github.com/web-infra-dev/rspack/assets/79413249/4b03a5ad-c58f-4a0e-8aeb-0a7d01009ebc">

@JSerFeng said he might take a look at this issue

SoonIter:
related to https://github.com/web-infra-dev/rspack/issues/6612

JSerFeng:
FYI: https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/847

SoonIter:
related to
https://github.com/webpack/webpack/issues/18548
https://github.com/JSerFeng/webpack-lazy-compilation-issue/tree/hmr

SoonIter:
webpack devServer should watch the index.html and initial chunk changes and trigger the page reload. 
[...]

PR #14772, opened 2026-07-13
perf(hmr): consolidate fast incremental updates and lazy-runtime correctness

### LINKED ISSUE #14766: [Perf]: native watcher drops concrete created child paths from raw change events
### Summary

Loaders that register a source context need the concrete file path from the
watcher's undelayed `change`/`remove` events so they can scan only the edited
file. Rspack's native watcher collapsed a newly created child to its registered
parent and its one-slot N-API raw callback queue silently dropped sibling
events. A context-only loader was consequently forced to rescan the entire
source tree during HMR.

This is separate from compilation aggregation: like Watchpack, native
`modifiedFiles`/`removedFiles` should contain registered dependencies and
affected contexts, while the undelayed event stream reports the concrete
filesystem path.

### Minimal reproduction

Register `src` as a context dependency and `src/existing.js` as a file
dependency, then edit `existing.js` and create `created.js` in the same watch
tick. Before the fix, the native aggregate is correct but its raw callback
contains only the parent or first queued event:

```text
modifiedFiles = [src, src/existing.js]
raw changes   = [src]                 # created.js is missing
```

A deterministic synchronous burst makes the queue-loss problem clear: with
2,048 registered edits the recorded old one-slot callback delivers **0/2,048** concrete
paths while the aggregate remains complete.

### Fix stacked in #14772

The consolidated HMR PR separates the two watcher projections instead of
overloading one event vector:

- registered paths and affected contexts feed compilation aggregation;
- exactly one concrete `Create`/`Change`/`Remove` path feeds the undelayed
  callback when an associated dependency exists;
- ignored and unrelated paths stay suppressed;
- a delayed `Create` for an already-disappeared, unregistered context child
  is normalized to `Remove`, covering rapid create/delete and rename races;
- the raw callback queue is lossless for bursts, detects a closed/failed enqueue,
  aborts an obsolete queue on rewatch, and the JS bridge rejects callbacks
  from stale or closed watch generations.

The shared/scanner representation remains single-allocation and the normal
trigger path stores its one raw event inline, so the parity fix adds no
per-event vector or scanner-batch clone.

### Measurements and validation

The final release-equivalent binding delivers every concrete context child and
keeps the aggregate at exactly the watched context for **2,048 creates,
removes, and delayed-create-after-delete events**. For registered changes,
the captured final run (collected while a Rust check was active) reports:

| synchronous edits | concrete paths | raw callbacks | delivery | process RSS delta |
| ---: | ---: | ---: | ---: | ---: |
| 2,048 | 2,048/2,048 | 2,747 | 35.0 ms | 3.0 MiB |
| 8,192 | 8,192/8,192 | 13,985 | 179.0 ms | 9.5 MiB |
| 16,384 | 16,384/16,384 | 22,976 | 447.6 ms | 19.3 MiB |

Extra callbacks above the edit count are genuine OS notifications; all
expected paths and kinds are present. A separate idle run records
33.6/85.1/146.4 ms delivery for the same registered sizes. The regression
suite covers create-only,
simultaneous edit/create, ignored descendants, a **128-file raw burst**,
remove, delayed create/delete, aggregate parity, and stale/post-close raw
callbacks. Focused Rust validation passes **35/35 unit + 2/2 pending-event +
4/4 real-watcher tests** with Clippy warnings denied; the native JS matrix is
stable across repeated runs.

Related incremental-loader report:
https://github.com/tailwindlabs/tailwindcss/issues/20326

### LINKED ISSUE #14767: [Perf]: HMR process_assets performs graph-wide old-module work for unchanged runtimes
### Summary

A traced single-file React Refresh edit in a large route graph spends roughly
0.94--1.12 s of CPU in `HotModuleReplacementPlugin:process_assets`, despite
the edited module taking only a few milliseconds. The compilation has about
5,500 chunks. The pass linearly searches the current chunk collection for
every recorded chunk and repeatedly walks/clones old memberships even when an
ordinary edit changes only a small number of chunks.

### Findings and corrected fast path

There are three separable costs/correctness boundaries:

1. Resolving every recorded chunk ID by scanning `chunk_by_ukey` is
   `O(chunks^2)`. Build a `ChunkIdMap<ChunkUkey>` once and use indexed lookup.
2. Completely removed module IDs can be derived once from
   `CompilationRecords.modules`, excluding orphan graph modules with no old
   chunk membership, instead of cloning a per-chunk `remaining_modules` set.
3. When incremental chunk assets are enabled, `Mutation::ChunkSetHashes`
   identifies the surviving chunks whose rendered content changed. Unchanged
   surviving chunks can be skipped, **except** when their runtime membership
   changed. If the pass is disabled, fall back to the existing full scan.

The initial proposal to gate the entire old-membership loop solely on
`removed_from_runtime` was too aggressive. A module can move out of an
installed chunk, remain in the same runtime, and change code generation during
SplitChunks repartitioning; the installed old chunk still needs its hot update.
The corrected loop skips an old membership only when the runtime is stable
and the module is still present in that current chunk.

Runtime membership also requires an explicit guard on the changed-chunk fast
path: a shared async chunk can change from runtimes `[a,b]` to `[b]` while both
its hash and content hash remain identical. Runtime `a` must still receive the
`r`/`m` disposal update.

The implementation and committed regressions are stacked in #14772.

### Trace evidence

Representative native `OVERVIEW` samples from consecutive one-file edits:

```text
HotModuleReplacementPlugin:process_assets busy=985ms
HotModuleReplacementPlugin:process_assets busy=982ms
HotModuleReplacementPlugin:process_assets busy=944ms
HotModuleReplacementPlugin:process_assets busy=1.03s
HotModuleReplacementPlugin:process_assets busy=1.08s
HotModuleReplacementPlugin:process_assets busy=1.02s
HotModuleReplacementPlugin:process_assets busy=973ms
HotModuleReplacementPlugin:process_assets busy=1.00s
HotModuleReplacementPlugin:process_assets busy=1.12s
HotModuleReplacementPlugin:process_assets busy=970ms
```

At 5,500 chunks, the old linear lookup alone performs about 15.1 million
successful comparisons (up to 30.25 million for missing chunks). The repeated
membership/runtime work is more expensive in a shared route graph.

### Generic release-binding measurement

A grouped fixture with 1,057 modules, 194 emitted cold chunks, and 195 chunks
tracked on rebuild (three groups of four leaf edits) shows:

```text
                         baseline median   final consolidated
leaf rebuild                 403.33 ms             30.17 ms
code splitting                28.26 ms              0.55 ms
chunk hashing                 51.55 ms              4.42 ms
chunk assets                  70.00 ms              5.84 ms
HMR process_assets            10.84 ms              0.51 ms
changed chunks                  195/195                2/195
```

The optimized path completes **40/40** measured leaf updates, emits exactly two
hot-JS assets and one manifest with the edited marker, and retains an
approximately **13x** leaf-rebuild improvement. Focused regressions cover
stable runtimes, partial/final runtime removal with an unchanged hash,
moved-and-edited modules, and the
disabled-`CHUNK_ASSET` fallback.

An independent pre-lazy-consolidation release-binding comparison against the
earlier optimized binding (five groups of four edits, with emitted-value and
removal assertions) reduces leaf median **36.37 -> 29.77 ms (-18.1%)** and HMR
`process_assets` **1.13 -> 0.52 ms (-54.0%)**; removal improves
**114.51 -> 103.18 ms (-9.9%)**. Cold compilation also improves
**381.80 -> 365.85 ms (-4.2%)**, so the small cold-stage CodSpeed signal does
not reproduce in the grouped local A/B. An idle, bracketed 2.1.4 comparison
isolating the final correctness commits measures leaf/cold/removal
**29.41/382.79/101.64 ms -> 30.17/367.59/102.14 ms**
(**+2.6%/-4.0%/+0.5%**) with all runtime-loss/removal assertions passing,
showing no material regression from the final guards.

Related trace-backed reports: #14765 (persistent snapshot scope/update work),
#14766 (native watcher created-child provenance), #14768/#14769 (chunk-graph
reuse), and #14770 (repeated SplitChunks work).

### LINKED ISSUE #14768: [Perf]: editing an existing leaf falsely rebuilds the chunk graph
### Summary

Editing an **existing leaf module** falsely logs `new module detected, rebuilding chunk graph` and forces a full chunk-graph rebuild on every watch update. This affects large route graphs disproportionately and defeats the otherwise-correct skip path for unchanged outgoings.

The behavior is reproducible on Rspack 2.1.3 and the investigated main revision using many independent entries that share an existing leaf:

```text
initial: rebuild chunk graph; entrypoints change detected
leaf edit: rebuild chunk graph; new module detected, rebuilding chunk graph
```

A 250-entry reproduction reports roughly 43--51 ms for the update and emits the false-positive diagnostic. The application trace has about 5,500 chunks and spends roughly 0.5--1.6 s in chunk-graph/code-splitting work for a single TSX edit.

### Cause

The code-splitter leaf optimization intentionally omits leaves from `block_modules_runtime_map`. `BuildChunkGraphArtifact::can_skip_rebuilding_legacy` treated a missing previous runtime-map entry as a genuinely new module, even when the module already existed in the recovered chunk graph and still had no outgoings. Consequently, every existing-leaf edit invalidated the skip heuristic.

There are also upstream watch fixtures filtered with the reason `incremental.buildChunkGraph heuristic update is disabled`, including the pre-order-index leaf-edit case.

### Fix and regression

When a previous runtime-map entry is absent, distinguish an existing leaf from a genuinely new module:

```rust
if outgoings.is_empty() && self.chunk_graph.try_get_module_chunks(&module).is_some() {
  continue;
}
```

This preserves the rebuild for new modules or changed outgoings while allowing a stable existing leaf to use the cached chunk graph. The regression stacked in #14772 re-enables the previously filtered pre-order-index watch case and asserts that its leaf edit does not rebuild the graph.

### Validation and current status

With a rebuilt native debug binding, the 250-entry leaf-edit regression emits 250 valid hot updates and no longer logs either `rebuild chunk graph` or `new module detected`. Rustfmt, source checks, patch apply checks, and JS syntax/formatting are clean.

The fix and re-enabled regression are stacked in #14772 (`ce7ca20`, with shared outgoing-preparation work in `d138a89..e8740f6`). The final watch fixture covers stable leaf reuse and real topology-change fallbacks. The earlier gist/digest is superseded by the reviewable PR stack.

### LINKED ISSUE #14769: [Perf]: unchanged async outgoings falsely invalidate the chunk-graph cache during HMR
# Incremental chunk-graph cache falsely misses for an unchanged module with async outgoings

We traced slow React Router HMR in a large Rsbuild application to `BuildChunkGraphArtifact::can_skip_rebuilding_legacy`. A text-only `leaf.tsx` edit is visible through `Page.tsx`; the client compiler repeatedly reports `module outgoings change detected: src/root.tsx?react-router-route` and fully rebuilds the chunk graph even though route topology is unchanged.

## Attribution

The focused native trace attributes the client rebuild primarily to code splitting, followed by SplitChunks:

```text
Compilation:code_splitting        2.10s   2.22s   2.58s
Compilation:SplitChunks            0.60s   0.91s   0.79s
Compilation:create_hash           ~0.11s
Compilation::create_chunk_assets ~0.06s
Compilation:process_assets        ~0.22s
```

With verbose `codeSplittingCache` logging enabled, all 13 observed baseline hot rebuilds miss on the root route. `rebuild chunk graph` is 1.51--2.84 s (median 2.13 s). The browser harness completes 10/10 true HMRs but reports a 6.17 s median web compile and 7.37 s median edit-to-visible-DOM time; there are no CSS requests or reloads.

The corrected compiler then completes the same 10/10 browser run with **zero hot chunk-graph misses**: median web compile drops to 5.05 s (-1.13 s, 18%), edit-to-visible-DOM to 6.18 s (-1.20 s, 16%), hot-manifest start from 6.52 s to 5.45 s, and hot-JS start from 7.18 s to 6.03 s. Node remains comparable (3.21 s baseline, 3.33 s candidate); no CSS requests or reloads occur. A final uncontended run with one uniform native binding, constant-size visible tokens, and verbose diagnostic stats removed again completes 10/10: web **5.03 s**, visible DOM **5.97 s**, manifest **0.44 s**, node **3.21 s**, zero CSS/reloads. Additional post-run cleanup rebuilds also reuse the graph.

The root becomes affected because the Tailwind loader watches the edited source and extracted CSS uses `CssDependency::could_affect_referencing_module() == AffectType::Transitive`. This explains the reach into the root route; it does **not** require a CSS-content change. The React Router route transform wraps component exports and inserts a static `virtual/react-router/with-props` import; it does not inject an async import for this route.

## Cause

`ModuleGraphModule::all_dependencies()` explicitly includes both root and async-block dependencies and is ordered by dependency-creation order. The old cache predicate flattened that list, grouped it by target module, and compared the result only with the cached root block:

```rust
modules.get(&DependenciesBlockIdentifier::Module(module))
```

`CodeSplitter::prepare` uses different semantics: it filters non-module/non-context and weak dependencies, partitions each connection with `get_parent_block`, sorts ordered dependencies by `source_order`, appends unordered dependencies, and stores root and async blocks independently. Consequently, an unchanged async outgoing was compared against the root block and guaranteed a false miss; creation-order/source-order differences could also falsely miss for sync dependencies.

## Native reproduction and measurements

The regression/fix stacked in #14772 uses the fixture:

```text
tests/rspack-test/watchCases/build-chunk-graph/async-block-outgoings/
```

Its affected route has two sync imports, a re-export of the edited non-leaf module, and one stable async import:

```js
import "./side-effect";
import { first } from "./sync-first";
import { second } from "./sync-second";
export { value } from "./leaf";

export const sync = `${first}:${second}`;
export const load = () => import(/* webpackChunkName: "lazy" */ "./lazy");
```

The 250-entry watch/HMR matrix edits the body, renames the async chunk, retargets it, and adds a second async block. Baseline rebuilds on every edit:

```text
round  cause                 rebuild  time
1      leaf-only             yes      253ms  module outgoings change detected: route.js
2      rename-chunk          yes      212ms
3      leaf-after-rename     yes      234ms  module outgoings change detected: route.js
4      retarget-async        yes      231ms
5      leaf-after-retarget   yes      200ms  module outgoings change detected: route.js
6      add-async             yes      238ms
7      leaf-after-add        yes      263ms  module outgoings change detected: route.js
```

The corrected optimized binding reuses stable topology and still rebuilds for each real topology change:

```text
round  cause                 rebuild  time
1      leaf-only             no       173ms
2      rename-chunk          yes      166ms  module async blocks change detected: route.js
3      leaf-after-rename     no       151ms
4      retarget-async        yes      148ms  module removal detected
5      leaf-after-retarget   no       159ms
6      add-async             yes      163ms  module async blocks change detected: route.js
7      leaf-after-add        no       165ms
```

Unchanged-edit median drops from 244 ms to 162 ms in the small fixture; more importantly, the client can avoid a 1.5--2.8 s chunk-graph rebuild. Both matrices emit valid hot updates: the final edited token is present in `src_route_js.*.hot-update.js`, and all 250 runtime updates are emitted. A second matrix with a non-leaf edit, mixed ESM/CJS/re-export, persistent cache, SplitChunks, and extracted/context CSS imported by the affected module also reuses correctly (189 ms, no errors).

## Fix and remaining work

The corrected predicate mirrors `CodeSplitter::prepare`:

- checks that the affected module's async-block IDs/order and existing chunk-group options are unchanged;
- conservatively rebuilds for nested blocks;
- applies the same module/context/weak filtering and `source_order` ordering;
- groups current connections separately for the root and each async block, then compares each with its corresponding cached block across runtimes;
- preserves existing active/false-connection and module-removal behavior.

The fix is intentionally scoped to legacy chunk-graph reuse. SplitChunks still runs after a successful cache hit because the saved graph snapshot precedes that optimization. A post-fix focused trace confirms that the client consistently reuses code splitting (21--29 ms, 7 mutations/5 affected modules), while SplitChunks still costs **0.81--1.40 s**. The same trace attributes another **0.92--1.11 s** to the complete loader chain rebuilding `styles/root.css`; `leaf.tsx` itself builds in roughly 4--5 ms. Tailwind instrumentation is only 54--82 ms on these edits (3--4 ms scan, 51--78 ms dependency registration, essentially zero utility generation); most of the chain delay is downstream processing of its unchanged approximately 1.9 MB CSS output. These are separate incremental opportunities, not residual chunk-graph-cache misses.

A correctness-preserving SplitChunks fast path must capture the reusable graph immediately after `optimize chunks` and before runtime modules/assets are attached. Reusing the previous final `BuildChunkGraphArtifact` is unsafe. It should require a code-splitting cache hit, unchanged per-source-type module sizes and selection metadata, unchanged export usage when enabled, and no function-valued SplitChunks filters/name/test/layer callbacks; real topology or threshold changes must fall back to the existing path.

### Current fix and validation

The shared connection-preparation helper, root/async outgoing comparison, stable async identity, terminal-leaf reuse, and eight-step incremental-watch regression are stacked in #14772 (`d138a89..e8740f6`, `ce7ca20`). The final fixture covers stable body/source-position edits, rename/retarget/add/remove topology misses, synchronous ordering/side effects, terminal-leaf reuse, and emitted async values. Validation includes optimized native builds, topology and extracted-CSS/non-leaf matrices, emitted-hot-update content checks, focused Rust/core/watcher suites, syntax/formatting checks, and the 10/10 full-browser HMR run.

The earlier standalone patch, local repro paths, and binding digests are historical and superseded by the consolidated source/test diff in #14772.

### LINKED ISSUE #14790: Native watcher watch/close overlap permits undefined behavior and can crash

### LINKED ISSUE #14791: CopyRspackPlugin loses configured priority order at scale and emits Windows separators
## Summary

`CopyRspackPlugin` currently sorts copied results only by priority using an
unstable sort. At realistic pattern counts, equally prioritized forced copies
can be reordered, so the configured last pattern does not reliably win.
Separately, Copy asset names are normalized with platform separators; on
Windows both ordinary destinations and rendered `[name]`/`[path]` templates
can emit backslash-separated asset keys even though asset names are expected
to use `/`.

## Discriminating reproduction

Configure 64 existing files to the same destination:

```js
const patterns = Array.from({ length: 64 }, (_, index) => ({
  from: `assets/${index}.txt`,
  to: "copied/value.txt",
  toType: "file",
  force: true,
  priority: index % 5
}));
```

The last configured priority-4 pattern is 59. With the priority-only unstable
sort, a native optimized binding consistently emits pattern 44 instead, both
initially and after mixed cache-hit/miss rebuilds. The same scale harness
observes 94 instead of 124 at 128 patterns and 309 instead of 509 at 512.

There is a second, independent tie: one forced glob can contribute many
results with the same priority and pattern index. With 64 files in that glob
and 32 interleaved single-file patterns, the Copy debug log identifies file 39
as the last enumerated result, but result-level unstable sorting emits file 27
(and larger mixes can emit file 0). Changing file 0 then incorrectly changes
the destination even though file 39 should still win.

For Windows, a glob such as `assets/*/*.txt` copied to `copied` should expose
`compilation.getAsset("copied/assets/a/one.txt")`; the native test matrix
instead reports the expected key missing because the emitted key contains
backslashes.

The same boundary affects templates. With
`{ from: "assets/template", to: "template/[path][name][ext]" }` and
`assets/template/deep/one.txt`, the portable key is
`template/deep/one.txt`. Windows normalizes the configured template to
backslashes before interpolation, and the previous template branch returned
that rendered key unchanged. The native Windows matrix consistently reports
`compilation.getAsset("template/one.txt")` missing in all four attempts of
the cache-policy regression while all 8,930 other tests pass.

## Proposed boundary and measurements

PR #14772 sorts the small pattern list by `(priority, pattern_index)` and
emits each pattern's results in their enumerated order. This preserves both
last-configured precedence and within-glob overwrite order while avoiding a
sort over every copied asset. It normalizes the final asset key for both
ordinary and rendered template destinations to POSIX and adds
64-pattern/mixed-cache, within-glob, nested-template initial/rebuild, and
Windows-key regressions. It
also prepartitions static cache hits, uses one structured cache hit/total log,
and avoids source, glob-filter, and permission-path allocations that do not
apply to static hits.

A bracketed release-binding comparison with 512 static patterns improves
all-hit/one-miss/52-miss pattern work from `1.163/1.380/20.76 ms` to
`0.626/0.863/19.94 ms`; compilation medians improve `48/41/66 ms` to
`40/38/58 ms` with every asset and cache-hit assertion checked. Dynamic
transform controls retain the exact call count.

The change is conservative: callbacks, templates, permission copying, and
provenance-free builds do not reuse static results.

### LINKED ISSUE #14792: HMR preserve finalization leaves re-instantiated modules active and concurrent runners share warnings
## Summary

Two HMR correctness boundaries surfaced in concurrent hot-case coverage:

1. Web/worker hot cases currently share the process console. Cases that
   temporarily replace `console.warn` can capture warnings from a neighboring
   concurrent case, producing duplicate disposed-module warnings and flaky
   assertions.
2. A removed module factory can be re-required during a
   `preserveDisposedModuleFactories` apply transaction. That creates a fresh
   cached module which is unlinked/deleted at finalization but remains
   `hot.active`. A retained closure can then require a child after idle and
   silently reattach the removed parent instead of producing the expected
   disposed-module warning.

The worker matrix reproduced the first problem consistently as an extra
`from disposed` warning on different worker targets depending on scheduling.
The second problem is discriminated by retaining a closure created during the
apply callback, calling it after `NEXT_HMR`, and checking that the removed
parent is not reattached.

## Proposed boundary

PR #14772 gives each `WebRunner` an inheriting scoped console, restores warning
hooks in `finally`, and adds a deterministic two-runner isolation regression.
The runtime finalizer marks a transient re-instantiated module inactive before
unlink/cache deletion; the hot fixture checks its post-idle disposed warning
and parent/child graph integrity across node, web, and worker runtimes.

The runtime change is limited to the existing preserve transaction and adds
no additional option or change to normal HMR disposal.

### LINKED ISSUE #14793: [Bug]: extracted CSS HMR reloads each stylesheet twice and can retain a duplicate link
### Problem

With `CssExtractRspackPlugin` HMR enabled, one stylesheet edit can trigger **two independent browser reload paths** for the same extracted asset. The delayed second reload briefly leaves two stylesheet links and makes the existing `shared-stylesheet-dedup` incremental E2E regression intermittent.

This is observable even though the compiler performs exactly one rebuild per edit. It is separate from the within-one-update href de-duplication fixed in #14580.

### Reproduction and measurements

Using the existing fixture:

```text
tests/e2e/cases/css/shared-stylesheet-dedup/
```

run Chromium incremental HMR with four workers and repeat a request-count probe ten times. Each probe performs 20 sequential `src/index.css` edits, waits for the computed color and one stylesheet link after every edit, and records cache-busted `/static/style.css?...` responses and the link count immediately afterwards.

On #14772 head `df63754` with its exact CI-profile native binding:

```text
200 edits
200 compiler rebuilds
392 unique cache-busted stylesheet responses
38 post-assertion two-link observations
maximum stylesheet-link count: 2
10/10 probe runs completed
```

The ordinary two-edit regression passes 40/40 repetitions locally (four workers); one Linux CI run reports `Expected: 1, Received: 2` after the second edit while the computed color has already updated. Windows, macOS, WASM, Linux unit matrices, and the Rust/watcher checks on that head pass.

### Cause

There are two owners for extracted-CSS HMR:

1. `packages/rspack/src/builtin-plugin/css-extract/loader.ts` injects a changing proxy signal and registers `module.hot.dispose(... cssReload())`. `cssExtractHmr.ts` debounces that clone/reload by 50 ms.
2. `crates/rspack_plugin_extract_css/src/runtime/css_loading_with_hmr.ejs` immediately downloads and swaps the extracted stylesheet through `hmrC.miniCss`.

The native handler correctly de-duplicates chunk IDs within its own invocation, but it cannot de-duplicate the loader's later, independent reload. The later clone explains both approximately two requests per edit and a link-count assertion that can briefly pass before another link appears.

### Fix direction and validation

Extracted-CSS HMR should have a single owner when the native `miniCss` handler is present, while retaining the loader fallback for `CssExtractRspackPlugin({ runtime: false })`. A minimal compatible guard is:

```js
module.hot.dispose(function (data) {
  data.value = localsJsonString;
  if (!__webpack_require__.hmrC || !__webpack_require__.hmrC.miniCss) {
    cssReload();
  }
});
```

Suppressing only the redundant loader reload and leaving the native runtime unchanged produces the discriminating result:

```text
200 edits
200 compiler rebuilds
200 unique cache-busted stylesheet responses
0 two-link observations
maximum stylesheet-link count: 1
10/10 probe runs completed
```

The regression should extend the existing E2E case with repeated edits and a request-count assertion, and separately exercise `runtime: false` so the fallback and CSS-module/local invalidation semantics remain covered. Avoid adding sleeps/retries or weakening the link assertion.

#14682 is the broader, maintainable performance fix: its precise `miniCss.c/r` manifest makes the extracted runtime react only to actual CSS changes, removes the changing loader proxy signal when the runtime is active, and retains the signal for `runtime: false`. This issue supplies a focused, repeatable correctness/performance discriminator for that work and for #14772.

matthewdavis-oai:
## Summary

Fixes #14768. Fixes #14769. Fixes #14767. Fixes #14766. Fixes #14790.
Fixes #14791. Fixes #14792. Fixes #14793. Refs #14401, #14411, #14682, #14757, #14765, #14770,
#14779, #14789, #12904, and #6633.

This consolidates the trace-backed HMR, chunk-graph, Copy/watch, lifecycle,
and lazy-compilation correctness work into one reviewable stack. The
performance paths add no configuration switches and fall back whenever the
provenance required for a safe fast path is unavailable. The lazy-HMR port
retains its narrowly scoped, documented `preserveDisposedModuleFactories`
apply option and exposes typed `Compilation.watchInvalidationKind` so
integrations do not need process-global Symbols. The lazy/runtime source is
ported from #14753 with the original co-author attribution retained.

### What each part fixes

| Area | Cause | Fix and correctness boundary |
| --- | --- | --- |
| Chunk-graph reuse | Async outgoings were compared with the root block; source order/filtering differed from `CodeSplitter::prepare`; existing zero-outgoing leaves looked like new modules. | Share connection preparation/filter/order logic, compare root and async blocks independently, retain group-option/nested-block bailouts, and reuse only an existing dependency-less root already present in the recovered chunk graph. |
| Dynamic/global entry correctness | Reuse checked only entry names, so changed dynamic entry options, ordered roots, includes, or request attribution could leave an installed chunk graph stale while another entry kept the modules alive. Function-valued filename/public-path options were rewrapped every generation, causing false misses or stale closures under blind reuse. | Compare all topology-affecting entry fields, ordered/deduplicated global and named roots, original requests, and compilation-scoped includes. Refresh current filename/public-path functions on a safe hit and conservatively disable incremental chunk assets for full-hash/function filenames. Duplicate shared roots still reuse. |
| JS entry/include bridge correctness | Dependency reuse ignored the raw resolution context; equal relative requests from different contexts could silently omit a module. Unnamed entries and includes were marked as ordinary entries and could be replaced by lazy proxies that never execute. | Key dependency identity by raw context/request/name/layer, preserve dependency IDs across output-option changes, mark unnamed entries and all includes eager, and retain ordinary named-entry laziness. |
| Stable async identity | Moving an unchanged `import()` changed its source-location-based block ID and forced a full graph rebuild. | Keep semantic dependency/modifier identity, remove source-location identity, and assign collision-safe occurrence suffixes for repeated blocks. |
| HMR `process_assets` | Every recorded chunk was linearly searched, all memberships were cloned/scanned, and an ordinary edit revisited unchanged chunks. | Index chunk IDs once, calculate actually removed modules once, and use `ChunkSetHashes` to skip unchanged surviving chunks. Runtime-membership changes and disabled incremental chunk assets always take the full path; an edited module that moves chunks is still emitted to its installed old chunk. |
| Copy/watch correctness and reuse | Static patterns rescanned unchanged assets and cache hits allocated/logged once per pattern; an empty but valid lazy-watch rebuild was indistinguishable from a provenance-free rebuild and recopied every asset; glob contexts could omit an initially empty sibling; ancestor events could leave cached files stale; equal-priority forced copies used an unstable priority-only result sort, so later configured patterns or later results within one glob could lose nondeterministically; Windows could emit backslash-separated asset keys. Linux stale-watch cleanup was quadratic and overlapping registrations could silently remove child watches. | Prepartition cache hits/misses under a dense indexed cache, carry the existing lazy-watch provenance to the native compilation so an empty lazy generation can safely reuse its static result, report one structured hit/total counter, run only misses, avoid redundant source/permission allocations, register the stable glob base and invalidate symmetric path overlap, sort the smaller pattern list by `(priority, pattern index)` while preserving each result vector's enumeration order, normalize emitted asset names to POSIX, and index stale paths/recursive ancestors. Dynamic/callback/template/permission and empty normal/programmatic paths remain conservative. |
| Lazy compilation/watch provenance | A lazy activation could invalidate unrelated compilers, paused watcher batches could be replayed or lost, mixed lazy/file generations could be published stale or misclassified, and prefix matching routed compiler 10 to compiler 1. | Target the activated child with exact path/query matching, atomically drain/acknowledge native batches and clear paused Watchpack aggregation, keep the native aggregate callback queue lossless while the executor limits it to one in-flight batch, snapshot typed invalidation provenance per compilation, classify artifact-dependent children as normal, and let normal/file-backed work dominate a single coalesced generation. |
| Lazy HMR/runtime filename correctness | A chunk can become initial during an update, extracted async CSS using `[fullhash]` could call an undefined hash runtime, and filename-requirement checks repeatedly traversed all reachable chunks. Removed factories may still be needed until an apply transaction settles. | Keep referenced async-entry filenames aligned with code generation, derive/cache the JS/CSS full-hash requirement once per compilation (including per-chunk and extracted-CSS templates), preserve disposed factories only for the transaction, and clean parent/child links on every settle/error path. |
| Native watcher lifecycle | The N-API close path asynchronously shared/mutated an `&mut FsWatcher`; overlapping watch/pause/close was explicitly documented as undefined behavior, the native matrix reproduced an access violation (`0xC0000005`), immediate rewatch/double-close can hang a local binding stress loop, and a retained prior JS `Watcher` handle could pause, drain, or close the live reused native watcher. | Own the watcher behind an async mutex plus atomic closed state, serialize watch/close, reject post-close use, keep short synchronous operations locked, permit only the owning JS watch generation to pause/drain/close or inject callbacks, acknowledge obsolete aggregates without affecting the live batch, and stress overlapping watch/pause/double-close, stale handles/shims, a 400-file no-pause close race, and immediate recreation. |
| Native watcher raw context events | A new/changed/removed child below a context dependency was collapsed to its parent, the one-slot N-API raw callback queue silently dropped sibling events, and conflating raw paths with compilation aggregates duplicated directory callbacks and inflated invalidation sets. | Keep Watchpack-compatible registered/context aggregation and a separate concrete raw projection, deliver concrete child paths to long-lived `watchFileSystem.on('change'|'remove')` listeners across watch cycles, preserve ignored/unrelated suppression, normalize a delayed create for an already-removed unregistered child, use a lossless raw queue with stale-generation/close guards, and avoid scanner-batch cloning and the extra per-event raw-vector allocation. |
| Extracted-CSS HMR ownership | The native `hmrC.miniCss` handler and the loader-level, debounced `cssReload()` both replaced the same stylesheet, causing nearly two browser requests per edit and a transient duplicate link even with one compiler generation. | Let the native handler own replacement when present and retain the loader fallback only for `CssExtractRspackPlugin({ runtime: false })`; CSS-module locals invalidation and dispose data remain unchanged. |
| HMR test isolation/runtime disposal | Concurrent web/worker hot cases shared process-global `console.warn`, so a neighboring disposed-module case could capture extra warnings. A removed factory re-required during a preserve transaction created a fresh cached module that stayed `hot.active` after finalization, allowing retained closures to silently reattach stale parent edges. | Give each `WebRunner` an inheriting scoped console and always restore warning hooks in `finally`; deactivate the transient module before unlink/cache deletion, and assert its post-idle closure warns and cannot reattach the removed parent. |
| Compilation dependency deltas | Context, missing, and build added iterators incorrectly appended compilation-level file dependencies. | Append the corresponding dependency sets and add a focused disjoint-iterator regression. The more aggressive persistent-snapshot filtering proposal in #14765 is deliberately **not** included because provenance-free invalidation can make it unsound. |
| Long-session lifecycle | `ModuleGraphConnectionWrapper` and CSS/SRI/Rsdoctor compilation maps retained prior generations. | Call the existing connection cleanup and remove all compilation-keyed plugin entries in `clear_cache(old_id)`. |

### Commit/issue map

- `d138a89..e8740f6`, `ce7ca20`: async/root outgoing reuse, stable async
  identity, terminal-leaf reuse, and connection cleanup (#14768/#14769).
- `4710507`, `8887a86`, `0aabc05`, `f521bcc`: conservative Copy reuse,
  recursive/context correctness, compilation-map cleanup, and scalable
  watcher/Copy recovery (#14765/#14766).
- `f10f433`: changed-chunk HMR fast path with runtime-loss/moved-module
  guards (#14767).
- `601ac45`, `182ab3c`: disjoint compilation-dependency iterators and the
  connection-wrapper GC regression (#14765).
- `32d9804`, `2ee3fd6`, `cd55890`, `6f0d4e3`: consolidated lazy/runtime
  source, dependent-generation/watch-delta correctness, lossless native
  delivery/wrap, and exact compiler-prefix routing (#14753/#14779/#12904/#6633).
- `5936572`: dynamic/global entry, include, binding-context, and
  function-template correctness/reuse.
- `f861edd`: full-hash runtime filename correctness and scale.
- `17d545c`: native/Watchpack in-flight file coalescing, active-aggregate and
  invalid-hook correctness, and the constant-time empty-pending fast path.
- `ee631e7`: retain Watchpack coverage on WASI while skipping only the
  unsupported native watcher variants exposed by the Node-24 WASM matrix.
- `c73ba1f`: compact Copy cache-hit dispatch and structured attribution,
  deterministic inter-pattern priority and intra-glob enumeration ordering,
  POSIX asset names, and allocation-light graph reuse (#14791).
- `af562a1`: sound native-watcher serialization/recreation and an explicit
  direct binding dependency for the lifecycle regression (#14790/#14757).
- `43a2264`: isolated hot-test consoles, deterministic paused-Watchpack
  coverage, and disposed-closure cleanup (#14792). The independently
  developed native-watcher lifecycle fix and its macOS crash reproduction are
  also tracked in #14757. The consolidated variant removes the unsafe
  shared-reference boundary and serializes synchronous controls as well.
- `5a6b4f2`: use the CI-required typed `collect::<FuturesOrdered<_>>()` form
  at both Copy dispatch sites; it invokes the same ordered-future
  implementation and preserves concurrency, input order, and codegen shape.
- `99af575`: refresh the six generated hot snapshots affected solely by the
  new `module.hot.active = false` disposal line.
- `f49bfbc`: preserve concrete native context-child events without
  changing compilation aggregates, make raw bursts lossless, and reject stale
  watcher callbacks (#14766).
- `df63754`: normalize the final Copy asset key after either ordinary-path or
  template rendering, fixing the remaining deterministic Windows
  `[name]`/nested-`[path]` lookup failure without adding a hot-path call or
  changing cache eligibility (#14791).
- `ef90a0f`: keep extracted-CSS HMR single-owned, preserve the loader
  fallback for `runtime: false`, and add repeated-edit/request-count and
  affected-snapshot regressions (#14793).
- `eb10e07`: carry explicit lazy-watch provenance into the native
  compilation, safely reuse an unchanged Copy pattern on an empty lazy
  rebuild, preserve the existing Rust/three-argument binding APIs, expose the
  Perfetto field, and cover the 2,000-file and empty-normal freshness
  boundaries (#14791).
- `e92e185`: isolate retained native-watch handles, callbacks, and shims from
  the live generation; acknowledge stale aggregates without pausing or
  purging the active batch; cover long-lived raw listeners and stale
  lifecycle operations (#14766/#14790).

The final commits are deliberately separated so the entry bridge, runtime,
watcher/Copy, lazy middleware, and in-flight coalescing changes can be lifted
or reviewed independently.

### Latest correctness pass

The final adversarial pass adds two narrow boundaries. First, a legitimate
lazy watch rebuild can have empty modified/removed sets; Copy now reuses its
unchanged static pattern only when explicit lazy provenance reaches the native
compilation. Empty normal invalidation and repeated `compiler.run()` still
recompute, including a source mutation with its original mtime restored. The
original Rust rebuild API and three-argument binding call remain compatible;
the provenance marker is an optional trailing binding argument and appears on
the `Compiler:rebuild` Perfetto span. Second, retained watcher handles,
callbacks, and shims from an earlier generation cannot pause, drain, close,
or inject into the current native watcher; obsolete aggregates are still
acknowledged so delivery cannot stall. Long-lived raw listeners continue to
receive concrete created/removed children across cycles.

The focused final matrix is **24/24** across Copy, native generation,
lifecycle/raw-event, and lazy artifact-chain cases. It includes a real
2,000-file activation, empty-normal and repeated-run freshness, native and
Watchpack coalescing, raw bursts, stale handles/shims, and 1,000 lifecycle
races.

### Measured evidence

A grouped release-binding HMR fixture with **1,057 modules / 194 emitted cold
chunks / 195 chunks tracked on rebuild** (three groups of four leaf edits)
shows:

| Phase | Baseline median | Initial optimized | Final consolidated |
| --- | ---: | ---: | ---: |
| leaf rebuild | 403.33 ms | 40.34 ms | 30.17 ms |
| code splitting | 28.26 ms | 0.66 ms | 0.55 ms |
| chunk hashing | 51.55 ms | 5.22 ms | 4.42 ms |
| chunk assets | 70.00 ms | 8.69 ms | 5.84 ms |
| HMR `process_assets` | 10.84 ms | 1.14 ms | 0.51 ms |

The baseline changes **195/195** tracked chunks per edit; the optimized path
changes **2/195**, emits exactly two hot-JS assets plus one manifest with the
edited marker, and preserves runtime/module removal behavior. In a separate
stability run, the final binding completes **40/40** measured leaf edits with the exact two-chunk
boundary, approximately a **13x** median rebuild reduction for the sparse
fixture.

An independent pre-lazy-consolidation release-binding comparison against the
earlier optimized binding (five groups of four edits, with
emitted-value/removal assertions)
finds additional headroom: leaf median **36.37 -> 29.77 ms (-18.1%)**, HMR
`process_assets` **1.13 -> 0.52 ms (-54.0%)**, rebuild chunk assets
**6.80 -> 5.63 ms (-17.2%)**, hashing **5.14 -> 4.17 ms (-18.9%)**, and
chunk runtime requirements **2.61 -> 2.00 ms (-23.4%)**. Cold median is
**381.80 -> 365.85 ms (-4.2%)** with cold chunk assets
**43.22 -> 40.45 ms (-6.4%)** and runtime requirements
**13.14 -> 11.10 ms (-15.5%)**; removal improves
**114.51 -> 103.18 ms (-9.9%)**. The small CodSpeed cold-stage signals do
not reproduce in this grouped local comparison.

An idle, bracketed 2.1.4 comparison isolates the earlier four correctness
commits from their preceding combined binding. Leaf/cold/removal medians are
**29.41/382.79/101.64 ms -> 30.17/367.59/102.14 ms**
(**+2.6%/-4.0%/+0.5%**), with every rebuild/runtime-loss/removal assertion
passing and phase deltas within roughly **0.2 ms**. This does not show a
material regression from the final entry/runtime/watch guards.

For Copy, a matched debug-native watcher A/B with **374 files/29.92 MiB** and
three repetitions per group gives static-pattern medians of
**65.78 -> 1.69 ms (-97.4%)**, compilation **135 -> 72 ms (-46.7%)**, and
edit-to-done **150.34 -> 80.13 ms (-46.7%)**. The mixed case retains exactly
80 dynamic transform calls and two static-pattern hits, with pattern work
**67.23 -> 21.98 ms (-67.3%)**, compilation **131 -> 95 ms (-27.5%)**, and
edit-to-done **151.90 -> 105.68 ms (-30.4%)**.

The empty-lazy Copy boundary has a separate bracketed CI-profile A/B using
**2,000 x 16 KiB static assets (about 32 MiB)**. Every run asserts exactly
2,000 emitted assets, `watchInvalidationKind: lazy`, and empty modified and
removed sets. The preceding binding records **0/1** Copy hits on **10/10**
activations with **320.5 ms** median edit-to-done (**292--365 ms**) and
**310.5 ms** compiler time (**282--354 ms**). The final binding records
**1/1** on **10/10** and improves those medians to **215 ms**
(**202--248 ms**) and **206 ms** (**192--235 ms**): roughly **105 ms / 33%**
saved. Empty normal/programmatic invalidation still recomputes, including a
source mutation whose mtime is restored. The compatible optional provenance
argument adds **0 B** to the stripped CI-profile binding; total size remains
**69,831,832 B**, **+49,152 B** over main and **2,048 B** below the gate.

A grouped release-binding Copy-pattern scale comparison (baseline ->
prepartition candidate -> baseline; three warmups and twelve candidate/twenty-
four pooled-baseline timed rebuilds per scenario) covers
**16 x 24 x 80,000 B (~30.72 MB)**, **128 x 2 x 512 B**, and
**512 x 1 x 256 B**. Every rebuild checks every asset, exact cache-hit count,
and dynamic-transform count. At 512 static patterns, run-pattern medians for
all-hit/one-miss/52-miss improve **1.163/1.380/20.76 ms ->
0.626/0.863/19.94 ms**; compiler medians improve **48/41/66 -> 40/38/58 ms**
and p95 improves **63/61/100 -> 55/50/75 ms**. Dynamic controls retain the
exact 24/2/1 transform calls per rebuild; 512-pattern run-pattern medians
improve **1.494/1.670/21.91 -> 1.005/1.083/21.26 ms**. The structured cache
counter was added after this binding and replaces per-hit debug-string
allocation with one standard hit/total record; these timings are deliberately
not attributed to that later logger change.

The two concurrent Copy joins intentionally collect directly into
`FuturesOrdered`.
With the pinned futures 0.3.32 implementation, `join_all` switches strategy
at 30 futures and generates both `MaybeDone` and ordered poll/drop paths;
symbol attribution showed several 7--10 KB Copy-specific instances. Direct
ordered collection preserves concurrency, input order, and successful-sibling
error semantics while removing that duplicate small-path monomorphization.
The synchronous pattern-order/emission step is a small dedicated helper, so it
is not duplicated in the instrumented async state machine. The exact
CI-equivalent, path-trimmed binding is **69,831,832 B** stripped versus
**69,782,680 B** for main (**+49,152 B**), leaving **2,048 B** under the
**51,200 B** binary-size gate; it contains zero local source/registry paths.

A final bracketed guard run (24 pooled baseline/12 candidate rebuilds)
confirms the extraction does not regress Copy scale: at 512 static patterns,
all-hit/one-miss/52-miss pattern medians are
**0.831/0.949/23.72 -> 0.625/0.841/23.63 ms** and compiler p95 is
**105/72/90 -> 61/63/87 ms**; the dynamic control is
**1.048/1.187/23.71 -> 0.934/1.095/23.75 ms** with exactly one transform.
All asset, cache-hit, and overwrite-order assertions pass; remaining
compiler-median differences are within run noise.

The scale harness also discriminates a pre-existing correctness bug. With 64
interleaved forced copies and `priority = index % 5`, the configured final
priority-4 pattern is 59 but the priority-only unstable sort emits 44; at 128
and 512 patterns the observed winners are 94 instead of 124 and 309 instead
of 509, across initial and every mixed-cache rebuild. Result-level sorting
still leaves all files within one glob tied: with 64 glob results and 32
interleaved patterns the last enumerated file is 39, but the result sort emits
27 (and larger mixes can emit 0). Sorting the much smaller pattern list by
`(priority, pattern_index)` and preserving each result vector's order restores
both configured and within-glob precedence. The regression covers initial,
early miss, winner miss, all-hit, and intra-pattern rebuilds.

The stale-watch cleanup path previously compared every current registration
against every stale path. A standalone optimized microbenchmark that retains
half of the registrations and removes the other half shows the indexed,
component-aware path scaling with tree depth instead of the Cartesian product:

| stale/current watches | previous median | indexed median |
| ---: | ---: | ---: |
| 2,048/4,096 | 569.08 ms | 0.92 ms |
| 4,096/8,192 | 2,428.60 ms | 1.93 ms |
| 8,192/16,384 | 9,669.48 ms | 3.94 ms |

This retains recursive-child and near-prefix-sibling semantics; it is not an
inotify-registration benchmark. Native recursive registration still traverses
ignored descendants before filtering events (5,000 ignored top-level
directories produced 10,023 native watches and about 1.25 s ready time,
versus 23 watches/about 0.32 s with Watchpack). That separate scalability
limitation and an ignored-aware registration direction are documented in
#14789.

The native raw-event path has a separate correctness and scale boundary. A
deterministic synchronous burst of **2,048** registered edits delivers
**0/2,048** unique concrete paths at the JS raw callback in the recorded
previous one-slot/non-blocking N-API run even though compilation aggregation is
complete; this is callback-queue backpressure, not evidence that the OS watcher
generated zero events. The final watcher delivers every
expected path/kind for 2,048 context-only creates, removes, and delayed
create-after-delete events while keeping the aggregate at exactly the watched
context. In the captured final run at **2,048/8,192/16,384** registered
edits, all paths arrive with **2,747/13,985/22,976** raw callbacks,
**35.0/179.0/447.6 ms** delivery, and approximately **3.0/9.5/19.3 MiB**
whole-process RSS deltas while a Rust check was active. An earlier idle run
records **33.6/85.1/146.4 ms** delivery. Extra callbacks above the edit count
include explicit triggers and OS notifications; no expected event is lost. The
normal trigger
stores its single raw event inline and scanner batches retain one allocation.

A dense runtime fixture with **192 entries, 192 shared async chunks, and about
37,000 import edges** separates filename correctness from the remaining cold
compile cost. In clean, bracketed output-full-hash runs, the preceding/final
binding ranges are **3,746--3,879 / 3,826--3,861 ms** wall time; module graph
is **2,715--2,810 / 2,741--2,825 ms**, code splitting
**351--378 / 357--368 ms**, runtime requirements **210--227 / 202--226 ms**,
hashing **248--249 / 246--260 ms**, and chunk assets
**72 / 71--75 ms**. Async-full-hash and cache-group overrides show the same
throughput-neutral pattern, with every filename/runtime-helper assertion
passing. Across two dense runs, the four CSS factory and HMR/URL parser hooks
fire **74,112** times in total; their combined close-span time is about
**293 ms**. The useful next upstream target is repeated factory/parser-hook
dispatch and resource scheduling in dense graphs, not another
runtime-template traversal.

An earlier optimized-stack run in a large route-graph browser harness
completed **20/20** true HMR/SSR updates with one web/node generation per
edit, no steady CSS or lazy requests, and no navigation. Visible p50 was
**4.20--4.29 s** for the first 15 measured edits and **5.17 s** for the final
five as the process grew; corresponding browser compilation medians were
**2.89/3.26/3.98 s**. This predates the final `ChunkSetHashes`, terminal-leaf,
and Copy/watcher correctness commits, so it is phase attribution and a useful
target, not a claim about the exact final SHA. Remaining trace targets are
module-graph scheduling, asset materialization/source maps, repeated
SplitChunks work (#14770), and per-generation memory growth.

The large multi-entry leaf trace makes that boundary concrete: the edited TSX
transform itself is only **4--6 ms**. In the historical pre-alias trace, the
development-server root-stylesheet loader costs **~0.92--1.11 s**; removing
that redundant server graph eliminates the cost. Remaining warm compiler
costs are SplitChunks (**~0.75--1.40 s**), runtime requirements
(**~0.63--0.95 s**), and process assets (**~0.2 s**).
In the dense cold fixture, module graph is about **2.76 s/~70%**, code
splitting **~0.36 s**, runtime requirements **~0.22 s**, and hashing
**~0.26 s**; four parser/factory hooks fire **74,112** times across two runs
(about **146 ms/run** combined). The useful next targets are hook
dispatch/resource scheduling and asset materialization, not another
filename-template traversal.

Related open upstream work is intentionally **not** included in this stack:

- #14411 is a draft for native-watcher time-info parity. It populates
  `fileTimeInfoEntries`/`contextTimeInfoEntries` on the native callback and
  leaves `Watcher.getInfo()` and the undelayed change time out of scope; it is
  separate from the lifecycle/coalescing fixes here.
- #14682 tracks precise native/extracted CSS HMR (`css.c`/`css.r` and
  `miniCss` namespaces), eliminating stylesheet re-fetch/reparse on a
  JavaScript-only update and handling CSS add/remove. That browser-runtime
  work is separate from this stack's compiler-phase costs; the large-route
  run above already observes zero steady CSS requests. A clean trial merge
  with this stack has no source conflicts, but its CI-profile binding is
  **69,852,312 B** stripped (**+69,632 B** over main), exceeding the
  **51,200 B** binary-size gate by **18,432 B**; keeping it independently
  reviewable avoids duplicating its 29 attributed commits. The loader guard
  here remains compatible and prevents a second owner whenever a proxy module
  is disposed.
- #14401 is a draft, production-oriented source-map generation optimization
  that reduces mappings allocation/copying. It does not address the
  incremental source-map cache/re-emission pass identified in this trace,
  but is useful adjacent profiling context.

### Committed regressions and validation

The Linux `chromium-incremental` E2E run on `df63754` exposed the extracted-
CSS double-owner race: the runtime `hmrC.miniCss` handler and loader-level
`cssReload()` could both swap the same stylesheet, leaving two links after an
edit. A four-worker, ten-repeat browser matrix now exercises **20 edits x 10
repetitions** for each runtime mode (**400 edits total**), asserts one link
and exactly one unique/cache-busted HMR stylesheet response per edit, and
passes both modes. The unguarded baseline performs one compiler rebuild per
edit but produces **392** unique stylesheet responses and **38** post-
assertion two-link observations for **200** edits; the guarded run produces
exactly **200/200** with zero duplicate links. All 400 guarded updates compile
in **11--53 ms** (p50 **20 ms**, p95 **32 ms**). This fixes the duplicate-
request/link boundary without claiming the precise JavaScript-only CSS work
elimination tracked in #14682.

- async-outgoing incremental-watch fixture: eight steps cover leaf reuse,
  async chunk rename/retarget/add/remove misses, a same-module
  source-position move, a dependency-less terminal-leaf edit, synchronous
  order/side effects, and emitted async values; the existing pre-order-index
  leaf fixture is re-enabled;
- four Rust async-identity cases cover location moves, duplicate blocks,
  unrelated insertion, and occurrence-suffix collisions;
- four HMR-process-assets cases cover stable runtimes, partial/final runtime
  removal with an unchanged chunk hash, moved-and-edited modules, and the
  disabled `CHUNK_ASSET` fallback;
- Copy matrix covers native watcher and Watchpack sibling creation, ancestor
  file/directory events, add/edit/remove-last/re-add, ignores,
  callback/template/permission bypass, and repeated `compiler.run()` without
  watcher provenance; it also covers 64 interleaved equal-priority forced
  destinations, mixed cache hit/miss ordering, POSIX asset keys on Windows,
  simple/nested `[name]`/`[path]` templates on both initial and unrelated
  rebuilds with an explicit no-backslash-key assertion, and successful-sibling
  caching across an errored glob recomputation. The
  watcher-backed assertion waits for its specific copied asset so queued
  startup events cannot create a false failure. A missing required glob
  reports exactly one diagnostic and recovers cleanly;
- watcher unit/integration regressions cover an initially empty child, new
  grandchildren, recursive/non-recursive mode transitions, recursive-parent
  removal with a retained child, pruning overlapping descendants, and
  generation-aware pending-event delivery/drain/acknowledgement, callback
  teardown, uint32 generation wrap, and thousands of stale/retained sibling
  registrations with recursive children and near-prefix siblings;
- native raw-event regressions cover create-only, simultaneous edit/create,
  ignored descendants, a 128-file lossless burst, remove, delayed
  create/delete, exact Watchpack-compatible aggregate parity, and rejection of
  stale/post-close callback generations;
- dynamic/global-entry regressions cover option/filename changes, ordered
  import roots, unnamed global roots, named/global includes, original request
  attribution, unchanged duplicate-root reuse, and both shared and freshly
  captured filename/public-path functions across repeated rebuilds;
- binding-entry regressions cover eager unnamed entries/named and global
  includes, ordinary named-entry laziness, and equal relative requests from
  distinct raw contexts for both `addEntry` and `addInclude`;
- lazy/MultiCompiler and runtime regressions cover hook order, artifact
  dependency ordering, native and Watchpack coalesced file provenance without
  a duplicate dependent generation, canonicalization-safe native invalidation,
  the non-consuming paused-Watchpack signal and guaranteed in-flight release
  before close, initial JS/CSS chunks,
  worker chunks, full-hash filename templates (including extracted async CSS),
  compiler-prefix collision, and disposed factories across success, queued,
  and error paths; the preserved-factory case retains a closure created during
  apply and proves its post-idle disposed warning/no parent reattachment;
- WebRunner constructs two isolated runners and proves `console.warn`
  overrides cannot cross-contaminate or mutate the process console;
- the native lifecycle regression overlaps rewatch/pause/double-close, runs
  the 400-file/20-directory/five-watch/no-pause immediate-close stress shape,
  and proves that an immediately recreated watcher receives its full
  dependency set and first event. The test package explicitly declares its
  direct `@rspack/binding` dependency and lock entry instead of relying on
  incidental workspace resolution;
- connection-wrapper GC regression discriminates the leak with forced GC;
  CSS/SRI/Rsdoctor cleanup is additionally covered by the compilation-map
  source regression and native plugin/HMR smoke;
- focused validation is **54/54** base tests across ten files, **2/2** native
  watcher-generation tests, and **3/3 x5** repeated native/Watchpack
  coalescing/getInfo cases; the standalone correctness scenarios and focused
  Rust/core/watcher suites also pass. JS syntax and formatting are clean. The
  exact CI-produced WASI binding passes **10/10** focused Copy/coalescing/
  getInfo cases with Watchpack enabled; the native matrix passes **12/12 x3**.
  The final CI-profile binding additionally passes **20/20** base
  Copy/lazy/watch cases across six files, **7/7** native-generation/raw/
  lifecycle cases across three files on each of Node 22/24 (**54/54**), and
  the six regenerated hot snapshots,
  including the 128-event raw burst, 1,000 overlapping lifecycle iterations,
  the 30-iteration
  no-pause/400-file close race, and immediate-recreate delivery. The
  earlier CI-profile worker/console matrix passes **349/349** on both
  Node versions with `--maxConcurrency 4 --retry 0`, and the final six preserved-
  factory runtime variants pass (**18/18** selected assertions). With the
  repository-supported `-t preserve-disposed-module-factories` filter, the
  primary HMR cases take **121--207 ms** in the final Node-22 run; the long-form
  `--testNamePattern` initializes unrelated hot cases and is not a valid
  focused timing. Focused watcher Clippy passes with warnings denied. The
  final commits pass `git diff --check`; the consolidated stack retains
  whitespace in one generated `initial-chunks-hmr` runtime snapshot.
  All CI passes on the exact `ef90a0f` head: Linux Node 20/24, WASM,
  macOS/Windows Node 22, all watcher platforms, Rust/Clippy, Dylint, bench,
  and binary size are green. Linux E2E passes **132/132** in **41.8s**,
  resolving the preceding head's sole shared-stylesheet duplicate-link
  failure; Windows passes **8,931/8,931 enabled tests with 4 skipped**,
  including all eleven `CopyPluginCache` tests and the simple/nested
  template-key regression. The maintained local CSS matrix and two
  regenerated CSS-recovery snapshots also pass. The native source is
  unchanged by the loader fix, so stripped size remains **69,831,832 B**.
  The prior Rust-check CI failures (`clippy::needless_borrow` and
  `clippy::from_iter_instead_of_collect` in Copy) are fixed.

The small CodSpeed signals on the predecessor PRs were about 3% for
cold-stage microbenchmarks (`create_chunk_assets` and `runtime_requirements`;
the latter explicitly compared different runtime environments). The grouped
release A/B above does not reproduce them, and the HMR fast path returns
before doing work on a cold compilation.

The PR is intentionally kept as one handoff stack, with each topic small
enough to review or lift independently.


ScriptedAlchemy:
WASM Node 24 is deterministically failing both new native-watcher matrix rows (all 4 attempts timed out):

- `CopyPluginCache.test.js`: `watches the stable glob base and copies a newly populated sibling with native watcher`
- `LazyCompilationArtifactChain.test.js`: `keeps dependent and coalesced file-backed generations normal with native watching`

The corresponding Watchpack rows pass. `rstest.config.mts` already excludes `NativeWatcher*.test.js` under `process.env.WASM`; these mixed-mode files should omit only their `nativeWatcher === true` row when `process.env.WASM`, preserving their Watchpack coverage.

Failing job: https://github.com/web-infra-dev/rspack/actions/runs/29369339433/job/87210554619

ScriptedAlchemy:
The binding-size failure is also real, not a runner flake: `17d545c` is 69,864,600 bytes versus 69,782,680 on `03080ba`, an 81,920-byte increase. The gate allows 51,200 bytes, so this is 30,720 bytes over the limit. The workflow's own report comment failed with a fork-token `403`, which hid the delta from the PR timeline.

Failing job: https://github.com/web-infra-dev/rspack/actions/runs/29369339433/job/87212276014

crates/rspack_core/src/dependencies_block.rs
```diff
@@ -143,6 +140,17 @@ impl AsyncDependenciesBlock {
...
+    &mut self,
+    identifier: AsyncDependenciesBlockIdentifier,
+    occurrence: usize,
+  ) {
+    let mut id = String::with_capacity(identifier.0.as_str().len() + "|occurrence=".len() + 20);
+    id.push_str(identifier.0.as_str());
+    write!(id, "|occurrence={occurrence}").expect("write to String should not fail");
+    self.id = id.into();
+  }
+
```
JSerFeng, line 153: seems that loc is indeed  necessary for identifier here, as the loc may change at rebuild, and it should causing the AsyncDependenciesBlock change as well, because the error message and `origin` field depend on `loc` to display, but current implementation won't trigger change, as AsyncDependenciesBlockIdentifier during rebuild remains the same.

ScriptedAlchemy: Agreed — `loc` needs to remain part of the identifier unless the reuse path explicitly refreshes location-derived state. `can_skip_rebuilding_legacy` compares the current block identifiers/options and can reuse the cached chunk groups; during a rebuild, `CodeSplitter` copies `block.loc()` into `ChunkGroup::origins` (`code_splitter.rs` around 1708/1719 and 1814/1832), and stats later emit `origin.loc`. With the proposed location-independent ID, a line-only move can pass the reuse check and leave stale origin/diagnostic locations. The `async_block_identifier_is_stable_when_source_location_moves` expectation is therefore too strong; preserving that optimization would require refreshing cached origins separately.

PR #14781, opened 2026-07-14
fix: invalidate the target compiler for lazy compilation

### LINKED ISSUE #14778: [Bug]: Lazy compilation invalidates every compiler in a MultiCompiler
### System Info

System:
  OS: macOS 26.5.1
  CPU: (10) arm64 Apple M5
  Memory: 8.67 GB / 32.00 GB
  Shell: 5.9 - /bin/zsh
Binaries:
  Node: 26.0.0 - /opt/homebrew/bin/node
  Yarn: 1.22.22 - /opt/homebrew/bin/yarn
  npm: 11.12.1 - /opt/homebrew/bin/npm
  pnpm: 11.8.0 - /opt/homebrew/bin/pnpm
  bun: 1.3.14 - /opt/homebrew/bin/bun
Browsers:
  Safari: 26.5

### Details

When lazy compilation is enabled in a multi-config build, activating a module for one compiler triggers rebuilds for every compiler.

The middleware creates separate state for each compiler, but passes the parent MultiCompiler to `lazyCompilationMiddlewareInternal`. The activation request then calls `MultiWatching.invalidate()`, which invalidates every child watcher.

Only the compiler associated with the request should be invalidated.

### Reproduce link

_No response_

### Reproduce Steps

1. Create a multi-config build with `lazyCompilation` enabled for at least two configs.
2. Start the compiler in watch mode.
3. Trigger a dynamic import belonging to one compiler.
4. Observe that the other compiler is also rebuilt.

Kanary-channelio:
## Summary

When using lazy compilation with a MultiCompiler, each compiler-specific middleware received the parent MultiCompiler. Activating a module therefore invalidated every watcher.

Pass the corresponding compiler to the middleware so only its watcher is invalidated.

## Related links

Closes #14778

## Validation

- `pnpm --filter @rspack/core build`
- `git diff --check`

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
+++ b/packages/rspack/src/builtin-plugin/lazy-compilation/middleware.ts
@@ -72,11 +72,7 @@ export const lazyCompilationMiddleware = (
 
       middlewareByCompiler.set(
         options.prefix,
-        lazyCompilationMiddlewareInternal(
-          compiler,
-          activeModules,
-          options.prefix,
-        ),
+        lazyCompilationMiddlewareInternal(c, activeModules, options.prefix),
       );
 
       applyPlugin(c, options, activeModules);


JSerFeng (approved):

stormslowly:
@Kanary-channelio 
 Thank you!

PR #14785, opened 2026-07-14
feat: warn on unknown import.meta properties

intellild:
## Summary

- Add a traceable warning when a statically known, unknown `import.meta` property is replaced with `undefined` while `parser.javascript.importMeta` is `true`.
- Keep the existing transformation result unchanged.
- Cover identifier access, static computed properties, member chains, optional chains, `typeof`, and destructuring.
- Do not warn for dynamic computed-property access because the property name cannot be determined statically; preserve evaluation of the key expression.
- Do not warn when unknown properties are preserved, including the `output.module: true` default and granular `importMeta` options.
- Keep known properties and exact `DefinePlugin` replacements unchanged.
- Reuse parser `Atom` values when rendering unknown property access to avoid intermediate string allocations.

## Why

Rspack already replaces unsupported, unknown `import.meta` properties with `undefined`. This PR adds diagnostics for statically analyzable accesses so configuration mistakes are visible, without changing the generated runtime behavior.

For example:

```js
console.log(import.meta.environment);
```

With `parser.javascript.importMeta: true`, the expression continues to be replaced with `undefined` and now produces an actionable warning:

```text
Accessing unknown `import.meta` property 'environment' is replaced with undefined.
```

Dynamic access does not warn because its property name cannot be determined statically:

```js
import.meta[getPropertyName()];
```

The key expression is still evaluated exactly once.

Preserved unknown properties also do not warn. Output modules default to `"preserve-unknown"`:

```js
module.exports = {
  experiments: {
    outputModule: true,
  },
  output: {
    module: true,
  },
};
```

Granular parser options can also preserve an unknown property without warning, including with non-module output:

```js
module.exports = {
  output: {
    module: false,
  },
  module: {
    parser: {
      javascript: {
        importMeta: {
          environment: false,
        },
      },
    },
  },
};
```

An exact `DefinePlugin` key is handled before `ImportMetaPlugin` and does not produce an unknown-property warning:

```js
new DefinePlugin({
  'import.meta.MY_ENV': JSON.stringify('canary'),
});

console.log(import.meta.MY_ENV); // "canary"
```

## Related links

N/A

## Validation

- `cargo fmt --all --check`
- `pnpm run build:js`
- `pnpm run test -t "normalCases/esm/import-meta"`
- `pnpm run test -t "configCases/module/import-meta-parser-options"`
- `pnpm run test -t "configCases/module/import-meta-disabled-unknown-property"`
- `pnpm run test -t "configCases/plugins/define-plugin-import-meta"`
- `pnpm run test -t "statsOutputCases/track-returned"`
- Commit hooks: Rust formatting, Prettier, and JavaScript lint

## Checklist

- [x] Tests updated.
- [x] Documentation not required.

---
_by OpenAI Codex_

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/define_plugin/walk_data.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/define_plugin/walk_data.rs
@@ -262,9 +262,11 @@ impl WalkData {
         define_record = define_record
           .with_on_evaluate_identifier(Box::new(move |record, parser, _ident, start, end| {
             parser
-              .evaluate(
+              .evaluate_with_range(
                 code_to_string(&record.code, None, None).into_owned(),
                 "DefinePlugin",
+                start,
+                end,
               )
               .map(|mut evaluated| {
                 evaluated.set_range(start, end);
@@ -296,7 +298,7 @@ impl WalkData {
             Cow::Owned(format!("typeof ({code})"))
           };
           parser
-            .evaluate(typeof_code.into_owned(), "DefinePlugin")
+            .evaluate_with_range(typeof_code.into_owned(), "DefinePlugin", start, end)
             .map(|mut evaluated| {
               evaluated.set_range(start, end);
               evaluated
@@ -310,7 +312,7 @@ impl WalkData {
             Cow::Owned(format!("typeof ({code})"))
           };
           parser
-            .evaluate(typeof_code.to_string(), "DefinePlugin")
+            .evaluate_with_range(typeof_code.to_string(), "DefinePlugin", start, end)
             .and_then(|evaluated| {
               if !evaluated.is_string() {
                 return None;
--- a/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
@@ -6,7 +6,7 @@ use rspack_core::{
   DependencyRange, ImportMeta, ImportMetaKnownProperties, ResolvedModuleOptions, RscMeta,
   RscModuleType, RuntimeGlobals, RuntimeRequirementsDependency, property_access,
 };
-use rspack_error::{Error, Severity};
+use rspack_error::{Error, Label, Severity};
 use rspack_util::{SpanExt, json_stringify_str};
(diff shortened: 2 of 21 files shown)


intellild:
@codex review

Timeless0911:
@elecmonkey Pls help verify this on Rslib PR.

elecmonkey (approved):
This PR works well in Rslib. Transforming `import.meta.env` into `void 0` may cause a `TypeError` at runtime when accessing properties such as `import.meta.env.MODE`.

However, this behavior matches rollup, esbuild, and Rolldown. Therefore, I think aligning with the ecosystem is reasonable.

cc @Timeless0911 

ref:
- [rollup repl](https://rollupjs.org/repl/?version=4.62.4&shareable=eyJleGFtcGxlIjpudWxsLCJtb2R1bGVzIjpbeyJjb2RlIjoiY29uc29sZS5sb2coaW1wb3J0Lm1ldGEuZW52Lk1PREUpIiwiaXNFbnRyeSI6dHJ1ZSwibmFtZSI6Im1haW4uanMifV0sIm9wdGlvbnMiOnsib3V0cHV0Ijp7ImVzTW9kdWxlIjoiaWYtZGVmYXVsdC1wcm9wIiwiZXhwb3J0cyI6ImF1dG8iLCJmb3JtYXQiOiJjanMifX19)
- [esbuild repl](https://esbuild.github.io/try/#dAAwLjI4LjEALS1mb3JtYXQ9Y2pzAGNvbnNvbGUubG9nKGltcG9ydC5tZXRhLmVudi5NT0RFKQ)
- [rolldown repl](https://repl.rolldown.rs/#eNptkM9qwzAMxl9F6NIWiqGwU67bjmMv4Itx5ODOkYLjdIXgd5/isH/Qk/1Jn3760IoBuxUj93Q3Zd7+jN2vPqNX6YVnSWSSDMc4TpKLGak4Q3wzb+8vryf1EXYlL1TPmCWlXj7Z6FiIwx/sg86+YIfCCj2FyPTc2lAhZBnh8D12sGyZ7s2qRrek9v4MHFfLALKUaSkdNAEQJI9OpUV/nS1uxWq5niyjZp2c/3ADmessGnBP+a+252ssiz1NpJdhH0lRumIjbSAl3dR2MRfzhPULfYhy5g==)

PR #14787, opened 2026-07-14
perf(copy-plugin): reuse unchanged static copy patterns during watch

matthewdavis-oai:
## Summary

Cache static `CopyRspackPlugin` pattern results across watcher-backed rebuilds and rerun only patterns whose registered file or context dependencies changed.

The plugin currently re-globs and re-reads every matched input on every rebuild. The fast path is automatic and deliberately conservative:

- reuses only when `modified_files` or `removed_files` provides watcher provenance;
- re-registers cached file/context dependencies on every hit and invalidates on a concrete file, removed file, reported directory, or any context descendant;
- never caches transforms, function/template destinations, permission copies, or a pattern that produced diagnostics;
- keeps cache ownership per native plugin/compiler instance and preserves the existing result/priority/asset path.

No public API or configuration option is added.

## Measured result

A release-binding watcher fixture with three patterns, **374 files / 29.9 MiB**, and an unrelated entry edit produced these medians:

```text
all static patterns       baseline       cached
run-pattern work            38.64 ms       0.45 ms
compiler                    99.00 ms      63.00 ms
edit-to-done               121.17 ms      71.32 ms
patterns reused                 0 / 3          3 / 3
copied assets                     374            374
```

With one dynamic transform pattern, its **80** transform calls remain intact while the two static patterns reuse; edit-to-done improves **164.60 ms -> 98.29 ms**.

## Validation

The optimized release binding and focused watcher matrix cover:

- unrelated-entry hit and copied-asset equality;
- directory-source edit, glob add, glob removal, ignored input, and empty-glob remove-last/re-add;
- transform edit and exact transform-call count;
- repeated `compiler.run()` with no watcher provenance (must miss, preventing stale output);
- shared JavaScript plugin across a dependent MultiCompiler;
- **48** consecutive persistent-cache watch rebuilds with all copied assets and changed markers preserved.

`rustfmt --edition 2024 --check crates/rspack_plugin_copy/src/lib.rs` and `git diff --check` pass.

### DIFF
--- a/crates/rspack_plugin_copy/src/lib.rs
+++ b/crates/rspack_plugin_copy/src/lib.rs
@@ -22,7 +22,7 @@ use rspack_error::{Diagnostic, Error, Result};
 use rspack_hash::{HashDigest, HashFunction, HashSalt, RspackHashDigest, RspackHasher};
 use rspack_hook::{plugin, plugin_hook};
 use rspack_paths::{Utf8Path, Utf8PathBuf};
-use rspack_util::fx_hash::FxDashSet;
+use rspack_util::fx_hash::{FxDashMap, FxDashSet};
 use sugar_path::SugarPath;
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


matthewdavis-oai:
Superseded by #14772, where the validated changes are now stacked for a single review surface.

PR #14808, opened 2026-07-16
fix(stats): avoid live chunk access when serializing stale stats

SyMind:
## Summary

- avoid traversing live `compilation.chunks` when generating entrypoint stats
- use the `assets` and `auxiliaryAssets` data from the stats snapshot instead
- prevent stale stats from accessing chunk wrappers removed by a subsequent watch compilation
- add a regression test ensuring entrypoint stats do not read live compilation chunks

## Checklist


SyMind:
@codex review

LingyuCoder (approved):

PR #14813, opened 2026-07-16
fix: parse preserved createRequire arguments

JSerFeng:
## Summary

- Defer walking `createRequire(import.meta.url)` arguments until the created require keep/strip state is known.
- Let regular parser plugins process the argument when the call is preserved, so options such as `parser.importMeta` remain authoritative.
- Escape generated `import.meta.url` string literals without changing the existing single-quoted output style.
- Extend ESM output coverage for default import-meta handling, `parser.importMeta: false`, `requireResolve: false`, and resource paths containing quotes.

## Related links

- #14798.


JSerFeng:
@codex review

Timeless0911 (approved):

PR #14814, opened 2026-07-16
fix(css-extract): fix stale CSS HMR when extracted filename is hashed

### LINKED ISSUE #6869: [Bug]: live reload of styles lazy components is broken when using hashes in names
### System Info

  System:
    OS: macOS 14.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 6.46 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.12.1 - /usr/local/bin/node
    Yarn: 1.22.18 - /usr/local/bin/yarn
    npm: 10.5.0 - /usr/local/bin/npm
    pnpm: 8.4.0 - ~/Library/pnpm/pnpm
    bun: 1.0.28 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 126.1.67.116
    Chrome: 126.0.6478.62
    Chrome Canary: 128.0.6544.0
    Edge: 111.0.1661.54
    Safari: 17.5
    Safari Technology Preview: 17.4
  npmPackages:
    @rspack/cli: ^0.7.4 => 0.7.4 
    @rspack/core: ^0.7.4 => 0.7.4 
    @rspack/plugin-react-refresh: ^0.7.4 => 0.7.4 

### Details

When using dev server and HMR I noticed the following:

1. updating styles for the main components work correctly.

2. updating styles for components lazily loaded via `import()` is left for one iteration.

Example:

* change the background color from `blue` to `red`, save the file.
* HMR is triggered, judging by the console.

Expected result: background is `red`
Actual result: the background still `blue`.

 * now change the background color from `red` to `yellow`, save the file.
* HMR is triggered, judging by the console.

Expected result: background is `yellow`
Actual result: background turned `red`

I tried to reproduce the minimal repl and the problem seems to appear when using `[fullhash]` or `[contenthash]` in the `output.filename` / `output.chunkFilename`.

(although at some point the styles of the main components stopped updating completely)

### Reproduce link

https://github.com/dartess/rspack-hmr-bug

### Reproduce Steps

1. run `npm run dev`
2. change background-color in `src/Page/Page.module.css` with some another color `color#1`

Expected result: background is `color#1`
Actual result: the background still `indianred`.

3. change background-color in `src/Page/Page.module.css` with some another color `color#2`

Expected result: background is `color#2`
Actual result: the background turned `color#1`.

gultyayev:
## Motivation

With `CssExtractRspackPlugin({ filename: '[name].[contenthash:8].css' })` + HMR, editing a CSS file never updates the page. Not the off-by-one described in #6869 — permanent, for every edit. The console reports success, a network request fires each cycle, and a fresh stylesheet link element is created and parsed, but its content never changes.

**Root cause:** the extract-css update handler (`hmrC.miniCss`) resolves the stylesheet URL through `__webpack_require__.miniCssF`, the chunk filename runtime module. Like every other runtime module, `miniCssF` is re-shipped and re-run when an update is *applied* — but the handler runs while the update is being *downloaded*, so for the very update that changed the `[contenthash]` it still returns the pre-edit filename. Re-fetching that stale URL (with a cache-busting query) never picks up the new content, and every subsequent edit hits the same wall.

#14682 reshaped the CSS hot-update manifest into `miniCss: { c: [...], r: [...] }` (changed/removed chunk ids) but still resolves the stylesheet URL through `miniCssF` in the handler, so the contenthash bug remained. This PR builds on that shape.

Fixes #6869.

## Changes

The hot-update manifest carries the fresh filename, and the handler requests it.

- **Manifest (`rspack_plugin_hmr`):** the `miniCss` entry gains an optional **`f`** map (`{chunkId: freshCssFilename}`), computed for each chunk whose extracted CSS changed from the chunk's own emitted extract-css asset. Matched by asset ownership (`ManifestAssetType::Custom("extract-css")`, the same tag `rspack_plugin_sri` relies on), not by filename suffix, so a chunk carrying both native CSS and extract-css output can't have the wrong one picked. Only the extract-css (`miniCss`) consumer populates it; the native css runtime passes an empty map.
- **Extract-css HMR handler (`rspack_plugin_extract_css`):** fetches `miniCss.f[chunkId]`, falling back to `miniCssF(chunkId)` for unhashed names. The installed `<link>` is still located through `miniCssF`, which at download time names the stylesheet currently on the page — the initial one on the first update, or the one the previous update installed (its runtime module was re-applied then).

Nothing else needs to change: once an update is applied, `miniCssF` is fresh again, so lazy chunk loading, prefetch and preload already resolve the current name. `packages/rspack/src/runtime/cssExtractHmr.ts` (the loader's `module.hot.dispose` fallback used when `runtime: false`) is unchanged.

### Tests

- e2e (real chromium): `tests/e2e/cases/css/contenthash-hmr` asserts the computed color updates across **two** consecutive edits with exactly one stylesheet element remaining — this one fails on `main`. `contenthash-hmr-lazy-chunk` edits a lazy chunk's CSS *before* its first `import()` and asserts the loaded stylesheet reflects the edit, guarding the post-apply path.
- Hot-update snapshots regenerated; the diffs are exactly the new `miniCss.f` field plus the handler change in the cases that re-ship the css loading runtime module. Native-CSS (`type: 'css'`) cases correctly gain no `f`.

### DIFF
--- a/crates/rspack_core/src/runtime_globals.rs
+++ b/crates/rspack_core/src/runtime_globals.rs
@@ -291,6 +291,13 @@ define_runtime_globals! {
 
   const HAS_FETCH_PRIORITY;
 
+  /**
+   * object mapping updated chunk ids to the CSS filenames emitted for them by the
+   * in-flight hot update, so CSS HMR handlers can resolve the fresh stylesheet URL
+   * instead of relying on the (possibly stale) chunk filename runtime function
+   */
+  const HMR_CSS_FILENAMES;
+
   // amd module support
   const AMD_DEFINE;
   const AMD_OPTIONS;
@@ -449,6 +456,7 @@ pub fn runtime_globals_property_name(runtime_globals: &RuntimeGlobals) -> Option
     RuntimeGlobals::CSS_STYLE_SHEET => "css",
     RuntimeGlobals::ASYNC_STARTUP => "asyncStartup",
     RuntimeGlobals::HAS_FETCH_PRIORITY => "has fetch priority",
+    RuntimeGlobals::HMR_CSS_FILENAMES => "hmrCF",
 
     RuntimeGlobals::RSC_MANIFEST => "rscM",
     RuntimeGlobals::TO_BINARY => "tb",
--- a/crates/rspack_core/src/runtime_globals.rs
+++ b/crates/rspack_core/src/runtime_globals.rs
@@ -291,6 +291,16 @@ define_runtime_globals! {
 
   const HAS_FETCH_PRIORITY;
 
+  /**
+   * object mapping chunk ids to the extract-css filenames emitted for them by hot
+   * updates seen so far, so the extract-css HMR handler and its chunk loading can
+   * resolve the fresh stylesheet URL instead of relying on the (possibly stale)
+   * chunk filename runtime function. Namespaced as "mini css" (matching the
+   * `miniCssF` chunk filename function) to stay distinct from any equivalent
+   * mechanism for the native (non-extracted) CSS runtime.
+   */
+  const HMR_MINI_CSS_FILENAMES;
+
   // amd module support
   const AMD_DEFINE;
   const AMD_OPTIONS;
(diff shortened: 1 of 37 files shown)


stormslowly:
@gultyayev, thank you a lot.
I am planning to merge this fix after https://github.com/web-infra-dev/rspack/pull/14682 which will make the shape of update json for css chunks

gultyayev:
@stormslowly I've rebased & adapted the fix to the new changes

PR #14818, opened 2026-07-17
fix: reduce checked reexport output size

LingyuCoder:
## Summary

- group checked `export *` reexports into a single conditional init fragment so the runtime condition is emitted once per dependency
- render checked reexports as a loop from 8 keys onward, while preserving the static fallback for smaller sets and legacy output environments
- add the `__webpack_require__.cr` runtime helper only when repeated call sites amortize its cost across all runtime copies
- include runtime-group size in the code-generation cache key and add output-shape, semantics, and cache regression coverage

## Motivation

Checked normal reexports previously emitted a complete runtime condition and property-definition statement for every possible export. Large CommonJS reexport surfaces therefore repeated the same `__webpack_require__.j` condition many times and increased minified and gzip output.

The runtime condition intentionally remains at the call site. This change only deduplicates it within each checked reexport dependency and extracts the property loop when doing so is size-positive.

## Size impact

Measured with production minimization and gzip level 9:

- representative shared asset: 1674 B → 1046 B minified (-37.51%), 482 B → 450 B gzip (-6.64%)
- seven-asset fixture total: 9769 B → 9141 B minified, 4678 B → 4646 B gzip
- dedicated helper case (8 calls, one runtime): 224 B less minified and 20 B less gzip than equivalent inline loops
- the other six assets in the representative fixture remain byte-identical

## Test plan

- `pnpm run test:unit` — 46 test files passed; 8811 passed, 4 skipped
- `cargo test -p rspack_core cache_key_accounts_for_runtime_group_size`
- `cargo test -p rspack_plugin_javascript checked_reexport_`
- Prettier checks for the changed JavaScript fixtures
- commit hooks: rustfmt, Prettier, and JavaScript lint

## Related links

N/A

## Checklist

- [x] Tests updated.


LingyuCoder:
After trying it in a real-world project, we found that although the minified size can be reduced somewhat, there is almost no improvement in the gzipped size. This is because the reduction comes from duplicated code, which gzip already compresses very effectively. Therefore, we won’t move forward with this for now.

PR #14828, opened 2026-07-17
chore: update simd-json and value-trait to fix stringify issue  for CPUs without AVX/SSE4 capabilities

### LINKED ISSUE #14701: [Bug]: CSS extract build fails on Linux runners with duplicated build-time module ids
### System Info

  System:
    OS: Linux 6.11 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (4) x64 Intel(R) Core(TM) i7-5930K CPU @ 3.50GHz
    Memory: 3.15 GB / 3.82 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 24.18.0 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 11.16.0 - /usr/local/bin/npm

@rspack/core >= 2.1.0

## Problem

After upgrading from Rspack 2.0.8 to 2.1.0, CSS builds started failing on some Linux GitLab runners when using Rsbuild 2.

The issue can be reproduced with a minimal project created via `npm create rsbuild@latest`. Even importing an empty `.scss`/CSS file can trigger the failure.

## Error

error   Build error: 
./src/components/app.scss:1:1
  × Module build failed (from ./node_modules/@rspack/core/dist/cssExtractLoader.js):
  ╰─▶   × Cannot read properties of undefined (reading 'build time')

Image

## Investigation

The crash happens during executeModule for CSS extraction. The JS layer tries to access:
codegenResults.map[id]["build time"]
but id does not exist in the codegen results map.
Native diagnostics showed that the generated CSS module code contained duplicated build-time module ids, for example:
__webpack_require__("javascript/dynamic|.../noSourceMaps.jsjavascript/dynamic|.../noSourceMaps.js")
while the actual codegen result key only existed once:
javascript/dynamic|.../noSourceMaps.js
This means the issue is not caused by the JS execute-layer lookup itself, but by native code generation rendering an invalid module id expression.
## Environment
-Rspack/Rsbuild regression between 2.0.8 and 2.1.0
-Reproduced on Linux GitLab runners
-Affected CSS extraction / cssExtractLoader
-Could not reproduce locally on Windows/macOS or Windows Docker
-Runner cache was cleared and did not affect the issue
## Expected behavior
CSS extraction should render valid build-time module ids and the build should complete successfully.

### Reproduce link

https://github.com/hpoepping/rsbuild-css-test

### Reproduce Steps

1. npm create rsbuild@latest
2. npm run build (via GitLab CI/CD)

I have not been able to reproduce the issue outside of GitLab CI/CD yet.

intellild:
## Summary

- Update the workspace `simd-json` constraint from 0.17.0 to 0.17.3.
- Update the locked `simd-json` package from 0.17.0 to 0.17.3.
- Update the locked transitive `value-trait` package from 0.12.1 to 0.12.2.

## Example

A compilation that parses JSON through `simd-json` now both declares and resolves version 0.17.3, so a fresh dependency resolution receives the same patch release as the checked-in lockfile.

## Validation

- `cargo check --workspace --locked`

---
_by OpenAI Codex_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5683,9 +5683,9 @@ checksum = "e320a6c5ad31d271ad523dcf3ad13e2767ad8b1cb8f047f75a8aeaf8da139da2"
 
 [[package]]
 name = "simd-json"
-version = "0.17.0"
+version = "0.17.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4255126f310d2ba20048db6321c81ab376f6a6735608bf11f0785c41f01f64e3"
+checksum = "e32d7ab2678282d21e53374fbead7119b7eacbede73685dcaac472870a29a11c"
 dependencies = [
  "halfbrown",
  "ref-cast",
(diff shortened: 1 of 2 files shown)


intellild:

SyMind (approved):

PR #14829, opened 2026-07-17
fix(watcher): run FsWatcher on a single owner task to fix concurrent &mut races

stormslowly:
## Why

`NativeWatcher.watch()` / `close()` each spawned a detached task holding a `&'static mut NativeWatcher` obtained from `Reference::share_with`. On the multi-thread runtime, in-flight tasks mutate the same `FsWatcher` concurrently, and this `&mut` race shows up as two observed crashes:

- **Release builds (macOS)**: one task appends to notify's FSEvents watched-paths `CFMutableArray` while a concurrent `close` releases it — use-after-free, killing the process with `NSInvalidArgumentException: -[OS_os_log addObject:]` / SIGSEGV (the `NativeWatcher.part3` CI flake).
- **Debug builds**: two aliased `&mut Executor` race on the same `JoinHandle`, tripping tokio's `assertion failed: !curr.is_join_waker_set()`.

Repro (5× `watch()` + `close()` per iteration on one watcher): https://gist.github.com/stormslowly/b44c6b1540bc70a98a2aeaa8695e13a9 — crashes within ~15 iterations on `@rspack/binding@2.1.4`, 300 iterations clean after this PR.

## What

- **`rspack_watcher`**: `FsWatcher` becomes a handle; the watching state (`FsWatcherInner`) is owned by a single task spawned in `FsWatcher::new` (requires an ambient tokio runtime). `watch()`/`close()` enqueue synchronously and return ack futures, so operations execute strictly in call order — `watch` receives incremental added/removed sets, so ordering is part of the contract and now holds for every consumer of the crate. After `close`, later `Watch` ops are rejected through `on_error` and later `Close` ops resolve; the owner task exits when the handle is dropped, so a leaked/GC'd watcher still cleans up.
- `pause()` / `trigger_event()` stay synchronous and lock-free (`paused` flag hoisted out of `Executor`, trigger behind a mutex).
- **Binding**: `Reference::share_with` and the `unsafe fn close` are removed; an eager `closed` flag makes watch-after-close throw deterministically.
- **`NativeWatchFileSystem`**: `close()` detaches `#inner` immediately and resets `#isFirstWatch`, so close → `compiler.watch()` gets a fresh native watcher with a full (non-incremental) registration.
- New test: watching after close is rejected through `on_error` instead of silently dropping the callbacks.

Verified: crash repro 3×100 iterations clean (previously died within ~15); `NativeWatcher.part1/2/3` 237 passed; same-tick add/remove ordering stress passed; `cargo test -p rspack_watcher` passed.

### DIFF
--- a/crates/rspack_binding_api/src/native_watcher.rs
+++ b/crates/rspack_binding_api/src/native_watcher.rs
@@ -1,11 +1,10 @@
 use std::{
   boxed::Box,
-  panic::AssertUnwindSafe,
   path::{Path, PathBuf},
+  sync::Arc,
   time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
-use futures::FutureExt;
 use napi::bindgen_prelude::*;
 use napi_derive::*;
 use rspack_paths::ArcPath;
@@ -69,14 +68,17 @@ fn timestamp_to_system_time(millis: u64) -> SystemTime {
 impl NativeWatcher {
   #[napi(constructor)]
   pub fn new(options: NativeWatcherOptions) -> Self {
-    let watcher = FsWatcher::new(
-      FsWatcherOptions {
-        follow_symlinks: options.follow_symlinks.unwrap_or(false),
-        poll_interval: options.poll_interval,
-        aggregate_timeout: options.aggregate_timeout,
-      },
-      to_fs_watcher_ignored(options.ignored),
-    );
+    // FsWatcher::new spawns its owner task, so it needs the runtime context.
+    let watcher = rspack_napi::runtime::block_on(async {
+      FsWatcher::new(
+        FsWatcherOptions {
+          follow_symlinks: options.follow_symlinks.unwrap_or(false),
+          poll_interval: options.poll_interval,
+          aggregate_timeout: options.aggregate_timeout,
+        },
+        to_fs_watcher_ignored(options.ignored),
+      )
+    });
 
     Self {
       watcher,
@@ -85,10 +87,8 @@ impl NativeWatcher {
   }
 
   #[napi]
-  #[allow(clippy::too_many_arguments)]
   pub fn watch(
     &mut self,
-    reference: Reference<NativeWatcher>,
     files: (Vec<String>, Vec<String>),
     directories: (Vec<String>, Vec<String>),
     missing: (Vec<String>, Vec<String>),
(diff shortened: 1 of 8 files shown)


stormslowly:
@codex code review

crates/rspack_binding_api/src/native_watcher.rs
```diff
@@ -69,14 +68,17 @@ fn timestamp_to_system_time(millis: u64) -> SystemTime {
...
-    let watcher = FsWatcher::new(
-      FsWatcherOptions {
-        follow_symlinks: options.follow_symlinks.unwrap_or(false),
-        poll_interval: options.poll_interval,
-        aggregate_timeout: options.aggregate_timeout,
-      },
-      to_fs_watcher_ignored(options.ignored),
-    );
+    // FsWatcher::new spawns its owner task, so it needs the runtime context.
+    let watcher = rspack_napi::runtime::block_on(async {
```
hardfist: why add async here?

hardfist: if you want to add tokio runtime context for jobs in FsWatcher::new, maybe you should move this logic into FsWatcher::new, and there's another way to enter into tokio context https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter

hardfist: I'm not sure whether it's necessary to use tokio::spawn in current callstack

hardfist: it seems not necssary to introduce tokio runtime in FsWatcher::new(), something like this seems better
```rust
fn spawn_owner_thread(mut inner: FsWatcherInner) -> Sender<WatcherOp> {
  let (tx, mut rx) = tokio::sync::mpsc::unbounded_channel();

  std::thread::spawn(move || {
    let runtime = tokio::runtime::Builder::new_current_thread()
      .enable_all()
      .build()
      .expect("create watcher runtime");

    runtime.block_on(async move {
      let mut closed = false;

      while let Some(op) = rx.recv().await {
        match op {
          WatcherOp::Watch { .. } => {
            inner.watch(/* ... */).await;
          }
          WatcherOp::Close { .. } => {
            inner.close().await;
            closed = true;
          }
        }
      }
    });
  });

  tx
}
```

hardfist (approved):

PR #14815, opened 2026-07-16
perf: optimize module concatenation root search

ScriptedAlchemy:
## Summary

`ModuleConcatenationPlugin` repeatedly rebuilt the same incoming-module analysis for every candidate root. This PR:

- caches incoming analysis per `(module, runtime)` for one optimization pass
- keeps root-specific chunk checks and graph mutations outside the shared cache
- reuses root-search allocations and defers bailout formatting
- skips deferred-module lookup when the `deferImport` experiment is disabled
- caches module-depth sort keys and clones the root chunk set once per concatenation
- shares unsupported-syntax bailout payloads instead of deep-cloning them
- uses the shared-root and bailout benchmarks from #14817 and adds runtime-isolation and unsupported-syntax coverage

## Performance

Large synthetic support application:

| Measurement | Before | After | Improvement |
| --- | ---: | ---: | ---: |
| Server plugin | 10.47 s | 1.60–1.64 s | ~84.5% |
| Server root search | 9.56 s | 0.86–0.89 s | ~90.9% |
| Profiled server build | 26.1 s | 17.4–17.6 s | ~33% |
| Cold release build | 58.08 s | 29.07 s | 49.95% |

Cold release figures are five-run medians with caches cleared. They compare published `@rspack/core` 2.1.0 with a local 2.1.4 distribution and therefore include intervening `main` changes. The plugin and root-search profiles isolate this PR's target path.

CodSpeed compares this branch with the stacked benchmark base from #14817: shared roots improved from 30.2 ms to 17.6 ms (+71.61%), bailouts from 107.4 ms to 97.7 ms (+9.95%), and the existing case from 14.4 ms to 13.4 ms (+7.77%). The aggregate report improved 20.16%. CodSpeed flags differing runner environments, so local profiles and support-app measurements remain the primary magnitude check. The ecosystem benchmark improved every reported wall-time case by 0.87–6.80%.

The latest follow-up changes were measured against the previous PR head with deterministic Callgrind instruction counts:

| Case | Previous head | Current head | Improvement |
| --- | ---: | ---: | ---: |
| Shared roots | 63,055,803 | 61,785,291 | 2.02% |
| Unsupported syntax | 62,044,869 | 61,740,194 | 0.49% |

## Correctness

- cache entries are runtime-keyed and live for one plugin invocation
- root chunk compatibility, recursive rollback, ESM checks, and `used_as_inner` ownership remain root-local and sequential
- parallel work only formats immutable bailout data; graph writes retain their original order
[...]

PR #14834, opened 2026-07-17
perf: reduce module concatenation hot-path overhead

LingyuCoder:
## Summary

- Share unsupported-syntax bailout payloads instead of cloning their module lists.
- Use cached-key sorting for repeated readable identifiers.
- Clone root chunk state once per root search instead of once per candidate.

## Related links

- Split from #14815.
- Stack 4/4; based on #14833.

## Checklist

- [x] Tests updated in the preceding stack layer.
- [x] Documentation not required.


LingyuCoder:

JSerFeng (dismissed):

stormslowly (approved):

PR #14839, opened 2026-07-17
refactor(core): reuse stable async chunk graphs

### LINKED ISSUE #14769: [Perf]: unchanged async outgoings falsely invalidate the chunk-graph cache during HMR

LingyuCoder:
## Summary

- Reuse incremental chunk graphs when entry topology and root or async outgoing connections remain stable.
- Keep async block identity location-independent while refreshing recovered origin locations and generation-time entry options.
- Add watch regressions for async ordering, entry and include topology, and location-only edits.

## Related links

Source PR: #14772

Closes #14769

## Checklist


crates/rspack_core/src/dependencies_block.rs
```diff
@@ -248,3 +279,74 @@ impl From<AsyncDependenciesToInitialChunkError> for rspack_error::Error {
     error
   }
 }
+
+#[cfg(test)]
```
LingyuCoder: No need to add these simple rust test cases

LingyuCoder:

PR #14842, opened 2026-07-17
fix(watcher): serialize native watcher lifecycle

### LINKED ISSUE #14790: Native watcher watch/close overlap permits undefined behavior and can crash

LingyuCoder:
## Summary

- Serialize native watch, pause, trigger, and close operations through a shared lifecycle state.
- Release closed JavaScript handles immediately so stale operations cannot reach replacement watchers.
- Add stress coverage for immediate close, repeated close, and overlapping watch and pause operations.

## Related links

Source PR: #14772

Closes #14790

## Checklist


LingyuCoder:
The changes in this PR have already been merged into `main` via #14829, so this PR will be closed.

PR #14846, opened 2026-07-17
fix(hmr): keep extracted CSS reload single-owned

### LINKED ISSUE #14793: [Bug]: extracted CSS HMR reloads each stylesheet twice and can retain a duplicate link

LingyuCoder:
## Summary

- Let the native miniCss HMR handler own extracted stylesheet replacement when it is present.
- Retain the loader reload fallback when the extracted runtime is disabled.
- Cover repeated edits, request counts, single-link ownership, and the runtime-disabled fallback.

## Related links

Source PR: #14772

Closes #14793

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/css-extract/loader.ts
+++ b/packages/rspack/src/builtin-plugin/css-extract/loader.ts
@@ -39,6 +39,8 @@ export function hotLoader(
   },
 ): string {
   const localsJsonString = JSON.stringify(JSON.stringify(context.locals));
+  // The extracted-CSS runtime owns stylesheet replacement when present; keep
+  // the loader reload only as the fallback for `CssExtractRspackPlugin({ runtime: false })`.
   return `${content}
     if(module.hot) {
       (function() {
@@ -60,7 +62,9 @@ export function hotLoader(
         }
         module.hot.dispose(function(data) {
           data.value = localsJsonString;
-          cssReload();
+          if (!__webpack_require__.hmrC || !__webpack_require__.hmrC.miniCss) {
+            cssReload();
+          }
         });
       })();
     }
--- a/tests/e2e/cases/css/shared-stylesheet-dedup-runtime-false/index.test.ts
+++ b/tests/e2e/cases/css/shared-stylesheet-dedup-runtime-false/index.test.ts
@@ -1,9 +1,6 @@
 import { test, expect } from '@/fixtures';
 
-// The hot-update lists both the `style` and `main` chunks while the fixed
-// `filename` maps them to one stylesheet. Without de-duplication the handler
-// re-fetched it once per chunk and leaked one <link> per update.
-test('should keep a single stylesheet link when several updated chunks share it', async ({
+test('keeps the loader CSS reload fallback when the extracted runtime is disabled', async ({
   page,
   fileAction,
 }) => {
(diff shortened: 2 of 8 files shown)


LingyuCoder:
This PR has been contained in https://github.com/web-infra-dev/rspack/pull/14682，so this PR can be close

PR #14850, opened 2026-07-18
fix(esm-library): support worker entries in modern module output

elecmonkey:
## Summary

Rslib uses modern-module output to generate clean ESM worker chunks. This PR treats `NewWorker` as an entry dependency, ensuring the worker module is executed when its chunk loads instead of only being registered as a module factory.


elecmonkey:
@codex review

PR #14854, opened 2026-07-19
fix: warn on invalid chunk names

### LINKED ISSUE #14383: [Feature]: Error (or warn) when a chunk name produces an emitted filename that differs from the URL the runtime requests

When a chunk name contains a `?` (e.g. because a `splitChunks.cacheGroups.*.name(module)` callback derives the name from
a module path that includes a resource query like `index.aff?raw`), Rspack silently produces **two different names for
the same chunk**:

- **On disk**, the emitted filename is truncated at the `?`: `library-index.aff` - losing the rest of the template,
  including `[contenthash]` and the `.js` extension.
- **In the runtime code**, the chunk-loading URL is built from the full, untruncated name:
  `library-index.aff?raw.<contenthash>.js`.

The build succeeds with no error and no warning. The divergence is only discoverable at runtime - and in browsers it is
usually _masked_, which makes it far worse (see "Real-world impact" below).

I would like Rspack to **fail the build** (or at minimum emit a warning) when a chunk name contains characters that make
the emitted filename differ from the URL the runtime will request - practically: when the chunk name contains `?` (also
worth considering `#`).

This is technically inherited behavior - webpack 5 and Rspack 1.x do exactly the same - so I'm filing it as a feature
request rather than a regression. But it's a silent output-correctness violation, and a one-line validation would have
surfaced it years earlier in CI instead of as a production-grade incident.

## Reproduction

6 files, no dependencies beyond `@rspack/core` / `@rspack/cli` 2.0.8.

**rspack.config.js**

```js
const path = require('path')

module.exports = {
	mode: 'production',
	context: __dirname,
	entry: './src/index.js',
	target: 'node',
	output: {
		path: path.join(__dirname, 'dist'),
		clean: true,
		chunkFilename: '[name].[contenthash].js',
	},
	module: {
		rules: [{resourceQuery: /raw/, type: 'asset/source'}],
	},
	optimization: {
		minimize: false,
		splitChunks: {
			cacheGroups: {
				libraries: {
					chunks: 'all',
					test: /[\\/]vendor[\\/]/,
					enforce: true,
					// Derives the chunk name from the module path - for an asset
					// module imported as `index.aff?raw` the resource query leaks
					// into the chunk name.
					name(module) {
						return `library-${module.identifier().split('/').pop()}`
					},
				},
			},
		},
	},
}
```

**src/index.js**

```js
import('./dicts/en_US.js')
	.then(m => console.log('DICT OK:', JSON.stringify(m.getDictionary())))
	.catch(err => {
		console.error('DICT FAILED:', err.message)
		process.exitCode = 1
	})
```

**src/dicts/en_US.js**

```js
import aff from '../../vendor/fake-dict/index.aff?raw'
import dic from '../../vendor/fake-dict/index.dic?raw'
export const getDictionary = () => ({aff, dic})
```

**vendor/fake-dict/package.json**

```json
{"name": "fake-dict", "version": "1.0.0"}
```

**vendor/fake-dict/index.aff** / **vendor/fake-dict/index.dic** - any text content:

```
AFF-DICTIONARY-CONTENT
```

### Steps

```bash
npm install
npx rspack build   # compiles successfully, zero warnings
ls dist
node dist/main.js
```

### Actual result

```
$ ls dist
895.e90f52d09ea3b8e0.js   library-index.aff   library-index.dic   main.js
#                         ^^^^^^^^^^^^^^^^^ truncated at '?': no contenthash, no .js

$ node dist/main.js
Error: Cannot find module './library-index.dic?raw.33258b7eda25f3f2.js'
#      runtime requests the FULL name that was never emitted
```

The chunk name became `library-index.aff?raw` / `library-index.dic?raw`; the emitted file kept only the part before the
`?`, while the runtime URL kept everything.

## Real-world impact 

With `target: 'node'` the mismatch at least fails fast. In a **browser** setup it is silently masked, e2e tests caught the degradation:

1. A `splitChunks` group's `name(module)` derived chunk names from module paths of hunspell dictionaries imported with
   `?raw` (`dictionary-en/index.aff?raw`) → chunk name `library-index.aff?raw`.
2. The emitted file `library-index.aff` (hashless!) was uploaded; the runtime requested
   `library-index.aff?raw.<contenthash>.js`. CDN treats `?raw.<hash>.js` as a query string and serves
   `library-index.aff` with HTTP 200. Everything _appeared_ to work for years.
3. Because the filename carries **no contenthash, the URL never changes between deploys**, while its content can. Any
   change to chunk ids or the runtime layout turns the cached artifact stale-but-still-served.
4. Upgrading Rspack 1.x → 2.x changed the chunk-loading global (`self.webpackChunk` → `self.rspackChunk`) and chunk id
   format. The CDN kept serving the old Rspack-1-built file for the frozen URL; it pushed into a global the new runtime
   never reads → `Loading chunk 18780 failed. (missing: …/library-index.aff?raw.<contenthash>.js)` - with the file
   returning HTTP 200. This is a genuinely confusing failure to debug from the browser side.

A build-time validation would be nice so we can catch it reliably.

## Environment

- Rspack 2.0.8 (also reproduces on Rspack 1.7.4 and webpack 5.107.2)
- Node 24.12.0, macOS (darwin arm64)

1. **Error** during chunk naming / asset emission when the rendered chunk filename differs from the name used to build
   the runtime URL (practically: chunk name contains `?` or `#`).
2. **Warning** with the same detection, if erroring is considered too breaking.
3. Alternatively, **sanitize consistently on both sides** (strip/encode the query in both the emitted filename _and_ the
   runtime URL) - this keeps builds working but still silently produces hashless names when `[contenthash]` follows the
   `?`, so 1 or 2 seem strictly better.

### Expected output

A build-time error (preferred) or warning, e.g.:

```
ERROR in chunk library-index.aff?raw
Chunk name contains '?' - the emitted filename ("library-index.aff") would not
match the URL requested by the runtime ("library-index.aff?raw.<contenthash>.js").
Chunk names must be valid filename fragments. Check splitChunks cacheGroup
`name` callbacks for module paths containing resource queries.
```

HashemKhalifa:
## Summary

- Emit a compilation warning when an emitted chunk name contains `?` or `#`.
- Explain that these characters can make the emitted filename differ from the URL requested by the runtime.
- Validate names before JavaScript eligibility exits so CSS-only chunks are covered without duplicate warnings.
- Add diagnostic regression coverage for JavaScript and CSS-only chunks.

## Related links

Fixes #14383

## Validation

- `cargo clippy -p rspack_plugin_javascript --all-targets --all-features`
- `cargo test -p rspack_plugin_javascript`
- Development native binding build
- `rstest Diagnostics.test.js -t config/invalid-chunk-name --project base`
- Repository pre-commit Rust formatting, JavaScript formatting, and lint hooks
- Independent semantic and changed-file safety review

## Checklist


HashemKhalifa:
@claude review

HashemKhalifa:
@chenjiahan could you please have a look :)

LingyuCoder:
Thanks! If the chunkFilename does not contain `[name]` then the invalid chunk names will not affect the runtime chunk loading and the warnings will be confusing. So I just add a little filter.

LingyuCoder (dismissed):

LingyuCoder (approved):

PR #14874, opened 2026-07-20
fix(worker): add module type when options are omitted in module output

elecmonkey:
## Summary

When `output.module` is enabled, Worker chunks are emitted as ESM.

Previously, when `new Worker()` had a second argument, Rspack appended `{ type: "module" }`:

```js
new Worker(url, options);

// ↓

new Worker(url, Object.assign({}, options, { type: "module" }));
```

However, when the second argument was omitted, Rspack left it unchanged:

```js
new Worker(url);

// ↓

new Worker(url);
```

This caused the browser to create a classic Worker for an ESM chunk. This PR adds `{ type: "module" }` when the second argument is omitted as well.


elecmonkey:
@codex review

JSerFeng (approved):

Timeless0911 (approved):

PR #14885, opened 2026-07-21
fix(asset): avoid JavaScript output for URL-only assets

JSerFeng:
## Summary

When an emitted asset resource is referenced exclusively by `new URL(..., import.meta.url)` dependencies in `new-url-relative` mode, emit only the asset source type instead of also generating an unused JavaScript module export.

The source-type decision now follows the CSS module pattern: inspect every incoming connection and derive the result from the dependency itself. `URLDependency` exposes the parser mode captured when the dependency was created, so asset generation no longer infers that mode indirectly from the issuer module's current parser options.

The predicate remains conservative: mixed imports or requires, default and `relative` URL modes, inline/source/bytes assets, and `emit: false` continue to generate JavaScript when needed.

Regression coverage includes URL-only, mixed import plus URL, and non-static URL mode cases, with module concatenation enabled and disabled. A dedicated `esmOutputCases/new-url/url-only-asset` snapshot verifies that modern-module output contains the static URL and emitted resource but no asset JavaScript module factory or runtime.

## Related links

N/A

## Validation

- `git diff --check`
- `cargo check -p rspack_plugin_asset -p rspack_plugin_javascript`
- `corepack pnpm --dir tests/rspack-test run test -t "configCases/asset-url/new-url-relative"`
- `corepack pnpm --dir tests/rspack-test run test -t "configCases/asset-url"` (20 tests passed)
- `corepack pnpm --dir tests/rspack-test exec rstest EsmOutput.test.js` (460 tests passed)

## Checklist


elecmonkey (approved):
output is correct

PR #14890, opened 2026-07-21
fix: disable incremental chunk assets for entry fullhash filenames

### LINKED ISSUE #14869: [Bug]: Incremental rebuild keeps stale per-entry [fullhash] filename
## System Info

```text
System:
  OS: macOS 26.5.2
  CPU: (14) arm64 Apple M3 Max
Binaries:
  Node: 24.16.0
  npm: 10.9.8
  pnpm: 11.8.0
npmPackages:
  @rspack/core: 2.1.4
```

Also reproduced against current `main` at `cfa680e76b623c42d13088a6d46bfcd1f75da2f7`.

## Details

An entry-specific filename containing `[fullhash]` is not enough to disable incremental `CHUNK_ASSET` reuse when the build chunk graph is recovered from cache.

The repro has two entries. `first` uses `[name].[fullhash].js` and contains a stable async import so that the build-chunk-graph cache is reusable. Only `second` changes. The second compilation has a new full hash, but `first` remains on disk only under the first compilation's hash.

Expected: `first.<second-compilation-hash>.js` exists.

Actual: only `first.<first-compilation-hash>.js` exists.

A fresh compilation produces the expected filename.

### Invalidation analysis

The generic guard in [`create_chunk_assets`](https://github.com/web-infra-dev/rspack/blob/cfa680e76b623c42d13088a6d46bfcd1f75da2f7/crates/rspack_core/src/compilation/create_chunk_assets/mod.rs#L35-L55) checks only the four global output filename templates. The entry-specific guard lives in [`code_splitter`](https://github.com/web-infra-dev/rspack/blob/cfa680e76b623c42d13088a6d46bfcd1f75da2f7/crates/rspack_core/src/compilation/build_chunk_graph/code_splitter.rs#L461-L473), so it is skipped when `BUILD_CHUNK_GRAPH` is reused.

`CHUNK_ASSET` must be invalidated because the compilation full hash is an input to this chunk's asset name even when that chunk's content hash and chunk graph are unchanged. The module graph, code generation, and chunk content hashes do not need global invalidation for this case: their semantic inputs are unchanged; only asset rendering/naming depends on the new global hash.

## Reproduce link

https://github.com/hardfist/rspack-repro-incremental-fullhash-entry

## Reproduce Steps

1. Clone the reproduction repository.
2. Run `npm install`.
3. Run `npm test`.
4. Observe that the assertion reports the new `first.<fullhash>.js` as missing and lists only the old hash filename.

ahabhgk:
## Summary

Fix https://github.com/web-infra-dev/rspack/issues/14869

- Detect fullhash placeholders in entry-specific filenames.
- Disable incremental chunk asset passes when those filenames depend on the full hash.
- Add a watch test covering fullhash changes across rebuilds.

## Testing

- Added a watch integration test for entry filenames using `[fullhash]`.
- Not run (not requested).

### DIFF
--- a/crates/rspack_core/src/compilation/create_chunk_assets/mod.rs
+++ b/crates/rspack_core/src/compilation/create_chunk_assets/mod.rs
@@ -32,7 +32,7 @@ pub async fn create_chunk_assets(
   compilation: &mut Compilation,
   plugin_driver: SharedPluginDriver,
 ) -> Result<()> {
-  if (compilation.options.output.filename.has_hash_placeholder()
+  let filename_has_hash_placeholder = compilation.options.output.filename.has_hash_placeholder()
     || compilation
       .options
       .output
@@ -47,7 +47,16 @@ pub async fn create_chunk_assets(
       .options
       .output
       .css_chunk_filename
-      .has_hash_placeholder())
+      .has_hash_placeholder()
+    || compilation.entries.values().any(|entry| {
+      entry
+        .options
+        .filename
+        .as_ref()
+        .is_some_and(|filename| filename.has_hash_placeholder())
+    });
+
+  if filename_has_hash_placeholder
     && let Some(diagnostic) = compilation.incremental.disable_passes(
       IncrementalPasses::CHUNK_ASSET,
       "Chunk filename that dependent on full hash",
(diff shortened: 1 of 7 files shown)


ahabhgk:
@codex review

hardfist (approved):

Issue #14889, opened 2026-07-21
[Bug]: `rspack build` can exit 0 while some declared chunk assets are never written to output.path


dewalchaturvedi:
### System Info

Rspack:    @rspack/core 1.7.1, @rspack/cli 1.7.1
Node:      20.11.1
OS:        Linux (glibc, x64) — CI container (Debian bullseye-slim)
PM:        yarn 3.5.0
Mode:      production, target: web

The build runs the client and a separate server compilation concurrently
(two rspack processes via concurrently), in a memory-constrained CI
container (--max-old-space-size=16384, ~900–1800 output files). Our current
hypothesis is that the disk-write phase hit resource pressure (EMFILE / ENOSPC
/ a killed worker) and some writes were dropped, but rspack did not surface
this as a fatal error — which is the core problem regardless of trigger.

Expected behavior

If rspack cannot write every asset it declared to output.path, the build
should fail with a non-zero exit code (or at minimum emit a hard error).
A build that exits 0 must guarantee that every emitted asset is on disk.

Actual behavior

rspack exits 0 with a subset of the declared chunk assets missing from
output.path, no error, no warning.

What would help / suggested fix

- A post-emit assertion in rspack that every asset in the compilation was
actually written to output.path, failing the build otherwise.
- Treat individual asset write failures (EMFILE/ENOSPC/worker crash) during
emit as fatal instead of being swallowed.
- Guidance on whether concurrent rspack compilations writing to different
output.paths can interfere under load.

Happy to add diagnostics (a strace/--stats verbose, FD limits, container
memory) — please tell me what output would be most useful to pin the trigger.

### Details

On a large production build (a micro-frontend that emits ~930 JS/CSS chunk
files), `rspack` **intermittently exits 0 while only a subset of the chunk
files it declared actually get written to `output.path`**. There is no error
and no warning — the process reports success.

We caught this by comparing two CI builds of the **exact same commit**:

| | Build A (broken) | Build B (good) |
|---|---|---|
| Assets listed by the build (stats / manifest) | 933 | 933 |
| Chunk files actually present in `output.path` | **770** | **933** |
| Referenced-but-missing files | **163** | 0 |
| Process exit code | **0** | 0 |
| Content hashes of the shared chunks | **identical** | identical |

Because the source (and therefore the deterministic content hashes) are
identical across the two builds, this is not a hashing/naming problem — Build A
simply **failed to write 163 files that it had already accounted for**, and
[...]

PR #14891, opened 2026-07-21
fix: propagate asset emit errors

LingyuCoder:
## Summary

Propagate failures from parallel asset emit tasks back to the compiler.

The emit scope returns both task join results and the result of each asset write, but those results were previously discarded. A filesystem failure such as `EMFILE` could therefore leave an asset missing while the compiler still reported success.

This change:
- returns asset write failures and task failures from `emitAssets`
- updates emitted asset versions and runs `afterEmit` only after every emit task succeeds
- adds a regression case that injects `EMFILE` from the output file system

## Related links

Related to #14889 , not sure this PR can solve it or not

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/mod.rs
+++ b/crates/rspack_core/src/compiler/mod.rs
@@ -382,7 +382,7 @@ impl Compiler {
       .incremental
       .passes_enabled(IncrementalPasses::EMIT_ASSETS);
 
-    rspack_parallel::scope(|token| {
+    let emit_results = rspack_parallel::scope(|token| {
       self
         .compilation
         .assets()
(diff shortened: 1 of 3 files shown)


crates/rspack_core/src/compiler/mod.rs
```diff
@@ -411,6 +411,10 @@ impl Compiler {
     })
     .await;
 
+    for result in emit_results {
+      result.map_err(|error| rspack_error::error!("Emit asset failed: {error}"))??;
```
hardfist, line 415: it seems it only report the first error?

LingyuCoder: Yes, this is a fatal error not a compilation error, so there is no need to report all.

LingyuCoder:

hardfist (approved):

PR #14892, opened 2026-07-21
fix(sharing): support ESM containers in shared tree-shaking

jcampalo:
## Summary

Fixes two bugs that prevented Module Federation shared tree-shaking from working with `library: { type: 'module' }` (ESM) containers:

1. **Unbounded fallback recursion.** `filterPlugin` in `IndependentSharedPlugin` excludes the MF plugin from the per-dependency fallback child compilers by name, but only listed `'ModuleFederationPlugin'`. The instance created by `@module-federation/rspack` reports `name === 'RspackModuleFederationPlugin'`, so it wasn't excluded, each child compiler re-applied MF and recursed without converging. Added `'RspackModuleFederationPlugin'` to the default exclusion list.

2. **`× Library name must be unset` for module containers.** `SharedContainerPlugin` set a library `name` unconditionally and passed it to the native binding, which rejects a `name` on a `type: 'module'` library. The fallback compile failed and shared tree-shaking silently no-op'd. Now the `name` is omitted when the library type is `module`.

## Test

Adds `configCases/sharing/reshake-share-esm-container`, a regression test that builds a shared tree-shaking fallback with `library: { type: 'module' }` + `experiments.outputModule`. It verifies the container is emitted as a `.mjs` ESM module exposing the federation `get`/`init` interface and tree-shaken down to the configured `usedExports`. Without the fix the fallback compile fails with `× Library name must be unset`; with it, the case passes. All existing `configCases/sharing/*` continue to pass.

## Checklist

- [x] Documentation updated (or not required) — not required.


jcampalo:
@hardfist @ScriptedAlchemy @2heal1 Can someone review this?

tests/rspack-test/configCases/sharing/reshake-share-esm-container/index.js
2heal1, line 1: The current test only checks the emitted files and their source code, but does not verify that the generated fallback container can actually be loaded and used.

Could we add an execution test that imports the emitted .mjs container, calls init and get, and verifies the returned shared module exports? I don't think this PR is ready to merge without this fallback availability coverage.

2heal1:
I verified the fallback locally and found that the ESM container fails under webpack runtime mode with `Cannot read properties of undefined (reading 'consumes')`.

`ConsumeSharedRuntimeModule` accesses `__webpack_require__.f.consumes`, but `__webpack_require__.f` has not been initialized. Since `IndependentSharedPlugin` always installs `ConsumeSharedPlugin` for these child compilations, I suggest initializing the ensure-chunk handlers object there before the consume runtime executes.

With this initialization, the complete `import → init → installInitialConsumes → get` flow works. I think this should be fixed as part of this PR.

jcampalo:
@2heal1 Thanks for the help, I've tried to follow your comments, let me know if it looks better. Whatever suggestion you have I'll do my best to address them until you feel confident about it, specially with rust which is far from being my area of expertise and I'm more guided by LLMs.

PR #14896, opened 2026-07-22
[POC] feat(hmr): support opt-in import.meta.hot compatibility facade

intellild:
## Summary

- add an opt-in `experiments.importMetaHot` flag, disabled by default
- expose `import.meta.hot` as a Vite-compatible interface instead of aliasing `import.meta.webpackHot`
- implement it as a shallow facade over `module.hot`: webpack still owns dependency boundaries, disposal, data handoff, evaluation, and error handling
- add only the adaptation state needed for namespace callbacks, per-dependency array callbacks, repeated self-accept callbacks, and compatibility data cleanup
- preserve existing `module.hot` and `import.meta.webpackHot` behavior; default-off builds retain the existing HMR runtime
- provide opt-in TypeScript declarations through `@rspack/core/import-meta-hot`
- keep `import.meta.glob` unchanged and out of scope

## Example

```js
const rspack = require("@rspack/core");

module.exports = {
  experiments: {
    importMetaHot: true
  },
  plugins: [new rspack.HotModuleReplacementPlugin()]
};
```

```js
if (import.meta.hot) {
  import.meta.hot.accept(["./a.js", "./b.js"], modules => {
    // The callback runs once per changed dependency. Unchanged slots are undefined.
    console.log(modules);
  });

  import.meta.hot.dispose(data => {
    data.updateCount = (data.updateCount || 0) + 1;
  });
}
```

For TypeScript, consumers can opt in without changing the existing bundler-resolution declarations:

```ts
/// <reference types="@rspack/core/import-meta-hot" />
```

## Compatibility

The facade calls native `module.hot.accept(dep)` to establish each HMR boundary. The extra runtime is enabled only by `experiments.importMetaHot` and handles semantics that differ from webpack's callback contract. It also covers multiple dependency callbacks, multiple self-accept callbacks, duplicate resolved dependency IDs, partial update failures, and module remove/re-add updates without switching the HMR runtime shape.

## Validation

- focused Rust workspace checks for the affected crates
- `pnpm run test:unit` — 46 files, 8836 passed, 4 skipped
- `pnpm run test:type`
- focused config cases for opt-in, default-off, production, and parser overrides
- focused HMR cases for dependency, self-accept, mixed handlers, failures, and remove/re-add
- `pnpm run lint:js`
- `pnpm run lint:rs`
- `cargo lint`
- `pnpm run check-dependency-version`
- `cargo fmt --all --check`

## Related links

N/A

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

---
_by OpenAI Codex_


intellild:
Close for wrong direction

PR #14897, opened 2026-07-22
fix: handle context changes in import.meta.glob hooks

intellild:
## Summary

- clarify the responsibilities of context-module request, parser context, compiler context, and resolved resource
- store `compiler_context` as the shared `Context` type instead of cloning it into a new `String` for every context dependency
- compile the initial `import.meta.glob` scan request once before `beforeResolve`
- treat the `context`, `request`, `recursive`, and pattern returned by `beforeResolve` as authoritative resolver/traversal input
- preserve the parser-derived glob request context for matching, following webpack's `requestContext` behavior
- keep `afterResolve.context` consistent with the value returned by `beforeResolve` without storing an extra `glob_match_context` in `ContextModule`
- extract glob request compilation and matching into `context_module_factory/glob.rs`
- add coverage for `import.meta.webpackContext` and glob context-module hooks

## Problem

Glob context modules have two different path coordinates:

1. the mutable `context + request` pair used by `ContextModuleFactory` to resolve the physical scan directory;
2. the parser-derived context used to interpret source glob patterns.

For example:

```js
// /project/src/index.js
const modules = import.meta.glob("./local/*.js", { eager: true });
```

A plugin can relocate resolution:

```js
compiler.hooks.contextModuleFactory.tap("Plugin", factory => {
  factory.hooks.beforeResolve.tap("Plugin", data => {
    data.context = "/project/fixtures";
  });
});
```

webpack resolves the hook's final `context + request`, but it does not rebase `./local/*.js` onto the hook-modified context. The pattern remains relative to `/project/src` (or to the explicit `base` option).

Rspack previously stored the generated glob request, compared the hook result with that snapshot, and conditionally recompiled the request after the hook. This equality-based compensation could overwrite an explicit same-value `request` or `recursive` assignment, produce inconsistent values between multiple taps, and make the request observed by hooks differ from the request used by the resolver.

## Fix

The data flow now follows webpack more closely:

```text
Parser
  stores raw glob patterns
  context          = importer directory or explicit import.meta.glob base
  compiler_context = compiler options.context
       ↓
Compile initial glob scan request once
  derive request + recursive before exposing them to hooks
       ↓
beforeResolve hook
  context/request/recursive/pattern may be modified
  returned values are authoritative for resolution and traversal
       ↓
Resolver
  resolve(hook context, hook request) → resource
  no glob request recompilation or hook-result patching
       ↓
afterResolve hook
  receives the hook-modified resolve context through transient factory state
  resource/recursive/pattern remain mutable
       ↓
ContextModule build
  walk final resource using final recursive
       ↓
Glob matcher
  relative pattern → file relative to parser-derived context
  root pattern     → file relative to compiler_context
```

`ContextOptions.context` therefore remains the glob equivalent of webpack's `requestContext`. The hook-modified resolution context is passed only through the factory's `resolve → afterResolve` call and is not persisted as another glob matching field.

## Example

When a plugin intentionally relocates scanning while preserving the original source glob coordinate, it can update both fields:

```js
factory.hooks.beforeResolve.tap("Plugin", data => {
  data.context = "/project/fixtures";
  data.request = "../src/local";
  data.recursive = false;
});
```

The resolver scans `/project/src/local`, while the source pattern is still interpreted relative to `/project/src`, exposing `./local/value.js`. Changing only `context` may scan a different directory, but it does not silently reinterpret the source glob.

## Test plan

- `cargo clippy -p rspack_core --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `configCases/hooks/context-module-resolve-context` in base and runtime-mode configurations
- all four `normalCases/context/import-meta-glob` variants
- all four `normalCases/context/import-meta-webpack-context` variants
- Prettier and `git diff --check`

This PR targets `main` and is independent of the former PR stack.

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
@codex review

intellild:
@codex review

SyMind (approved):

PR #14904, opened 2026-07-22
fix: compare async block values in code splitting cache

JSerFeng:
## Summary

- generate `AsyncDependenciesBlockIdentifier` from the parent module and the block's module-local index
- validate code-splitting cache reuse from each block's actual value: group options, ordered target modules, and per-runtime connection states
- rebuild the chunk graph when an affected module has no previous runtime snapshot to compare
- refresh cached chunk-group origin metadata when code splitting is reused
- add a watch regression covering loc-only changes and async-block target changes

## Why

Async block identity previously included dependency resource data and source location. A loc-only edit could therefore change the block key even when its code-splitting topology was unchanged.

Identity is now a stable module-local slot used for graph lookup. Cache reuse is decided separately from the block's code-splitting value. When the value is unchanged, the cached chunk graph is reused and its origin `module`, `loc`, and `request` are updated. When a block points to different modules, changes its group/runtime behavior, or has no previous runtime snapshot that can prove equivalence, the chunk graph is rebuilt.

This keeps loc-only rebuilds on the fast path without allowing stale chunk topology or origin metadata.

### DIFF
--- a/crates/rspack_core/src/artifacts/build_chunk_graph_artifact.rs
+++ b/crates/rspack_core/src/artifacts/build_chunk_graph_artifact.rs
@@ -1,19 +1,31 @@
 use std::mem;
 
 use futures::Future;
-use rspack_collections::{IdentifierIndexMap, IdentifierMap};
+use rspack_collections::IdentifierMap;
 use rspack_error::Result;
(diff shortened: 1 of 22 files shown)


JSerFeng:
@codex review

PR #14905, opened 2026-07-22
[POC] feat: add in-memory loader cache option

intellild:
## Summary

Adds a compiler-lifetime, in-memory cache boundary to loader entries through `module.rules[].use[].cache`.

- A cache hit skips the cache-enabled loader and loaders to its right in the `use` array, then continues with loaders to its left.
- The cache works for both JavaScript/parallel loaders and builtin Rust loaders.
- Cache entries preserve content, source maps, additional data, and loader dependencies.
- A resource listed in `modified_files` misses the cache. Entries for `removed_files` are deleted after a successful rebuild.
- The cache is intentionally memory-only and does not serialize loader options or persist across compiler instances.

Example with global compilation cache disabled:

```js
const path = require("node:path");

module.exports = {
  cache: false,
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          path.resolve(__dirname, "post-loader.js"),
          {
            loader: path.resolve(__dirname, "expensive-loader.js"),
            cache: true,
          },
          path.resolve(__dirname, "preprocess-loader.js"),
        ],
      },
    ],
  },
};
```

After the first build, an unrelated watch update reuses the output at the cache boundary, skipping `expensive-loader.js` and `preprocess-loader.js`. Updating the matched resource invalidates the entry and runs them again.

Validation:

- `pnpm run build:cli:dev`
- `pnpm run build:js`
- `pnpm run test:unit`
- `pnpm run test:base -t "loaders/cache-loader"`
- `cargo test -p rspack_core loader::cache::tests`
- `cargo clippy -p rspack_loader_runner -p rspack_core -p rspack_binding_api --all-targets -- -D warnings`

## Related links

- Inspired by [webpack-contrib/cache-loader](https://github.com/webpack-contrib/cache-loader), without API or persistence compatibility.

## Checklist

- [x] Tests updated.
- [x] Documentation updated (inline configuration API documentation).

---
_by OpenAI Codex_


intellild:
This POC has finished its duty.

PR #14917, opened 2026-07-23
refactor(runtime): collect runtime module variables from templates

LingyuCoder:
## Summary

- collect runtime-module variable names from top-level EJS `var` and `fn` declarations
- require every runtime module to declare its variable list explicitly, using an empty list for modules that do not declare surrounding-scope variables
- make parameterless `#[impl_runtime_module]` register each non-generic runtime module provider automatically; EJS-backed modules return their extracted variables
- carry the collected metadata through runtime module generation while preserving the existing EJS `var` and `fn` helpers

## Stack

1. #14917 (this PR)
2. #14741

## Checks

- `pnpm run build:js`
- `pnpm run test:unit`
- `cargo clippy --workspace --all-targets --tests --locked -- -D warnings`
- `pnpm run format-ci:js`
- `pnpm run format-ci:toml`

## Checklist

### DIFF
--- a/crates/rspack_core/src/runtime_module.rs
+++ b/crates/rspack_core/src/runtime_module.rs
@@ -247,6 +247,14 @@ pub trait RuntimeModule:
   fn template(&self) -> Vec<(String, String)> {
     vec![]
   }
+  /// Names that this runtime module may declare in the surrounding chunk scope.
+  ///
+  /// Runtime modules backed by EJS templates should derive these names from top-level `var()` and
+  /// `fn()` declarations. `#[impl_runtime_module]` registers the provider so the names can be
+  /// reserved before runtime module instances are created.
+  fn runtime_module_variables() -> &'static [&'static str]
+  where
+    Self: Sized;
   async fn generate(
     &self,
     context: &RuntimeModuleGenerateContext<'_>,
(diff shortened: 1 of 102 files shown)


crates/rspack_core/src/runtime_module.rs
```diff
@@ -247,6 +247,17 @@ pub trait RuntimeModule:
...
+  ///
+  /// Runtime modules backed by EJS templates should derive these names from top-level `var()` and
+  /// `fn()` declarations. `#[impl_runtime_module]` registers the provider so the names can be
+  /// reserved before runtime module instances are created.
+  fn runtime_module_variables() -> &'static [&'static str]
+  where
+    Self: Sized,
+  {
+    &[]
+  }
```
JSerFeng: We can drop the default impl for this, If we don't provide default impl, it'll force the developer who implement this to check if the really need runtime_module_variables

LingyuCoder: Fixed in c7ed816c02. `RuntimeModule::runtime_module_variables` no longer has a default implementation. All 67 `RuntimeModule` implementations now declare it explicitly; modules without surrounding-scope variables return `&[]`, while template-backed modules keep their extracted variable lists.

JSerFeng (approved):

PR #14920, opened 2026-07-23
[POC] feat: prototype first-class entries for CSS and URL targets

intellild:
## Summary

This draft PR prototypes a unified first-class entry rendering path for module types that do not inherently require a JavaScript facade.

- avoid emitting JavaScript bootstrap output for configured CSS-only entries
- represent `new URL()` targets as async entries instead of hard-coding asset-module behavior
- render JavaScript, CSS, WebAssembly, and asset URL targets from their actual source types
- preserve `parser.javascript.url` behavior for `true`, `relative`, and `new-url-relative`
- keep `executeModule` compatible by using the normal module export during module execution, then resolving the async entry URL during final chunk rendering

## Why

The previous implementation treated URL dependencies as asset-module exports and assumed a JavaScript `require()` facade. The same assumption also caused CSS-only configured entries to produce an unnecessary JavaScript runtime file.

That model does not extend cleanly to `new URL()` targets that are JavaScript, CSS, or WebAssembly modules. This POC makes the target module and its emitted chunk first-class, while generating JavaScript glue only when the referencing environment actually needs it.

## Examples

A configured CSS entry such as:

```js
module.exports = {
  entry: "./style.css"
};
```

now emits the bundled CSS without an otherwise empty JavaScript bootstrap file.

A URL dependency can target different module types through the same path:

```js
new URL("./worker.js", import.meta.url);
new URL("./theme.css", import.meta.url);
new URL("./module.wasm", import.meta.url);
new URL("./image.png", import.meta.url);
```

Each target becomes an async entry and emits according to its source type. The final URL expression still follows `parser.javascript.url`; for example, `new-url-relative` remains `new URL("./output-file", import.meta.url)` and does not introduce a runtime chunk.

## Validation

- `cargo clippy -p rspack_plugin_javascript -p rspack_plugin_esm_library --all-targets -- -D warnings`
- `pnpm run test:rs`
- targeted URL mode, async entry module type, ESM output, tree-shaking, and `importModule` cases
- full `pnpm run test:unit`: 8,805 passed and 4 skipped; three known watcher file-timing cases failed, while the three URL-related stale assertions found by the run were updated and passed when rerun in isolation

## Related links

None.

## Checklist

- [x] Tests updated.
- [x] Documentation not required for this POC.

---
_by OpenAI Codex_


intellild:
close in favor of #15469

PR #14923, opened 2026-07-23
fix: respect DllPlugin context for concatenated modules

### LINKED ISSUE #10799: [Bug]: DllPlugin manifest does not respect context if module is concatenated
### System Info

  System:
    OS: macOS 15.5
    CPU: (10) arm64 Apple M1 Pro
    Memory: 149.44 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.0 - ~/.nvm/versions/node/v18.18.0/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.18.0/bin/yarn
    npm: 9.8.1 - ~/.nvm/versions/node/v18.18.0/bin/npm
    pnpm: 9.1.2 - ~/.nvm/versions/node/v18.18.0/bin/pnpm
  Browsers:
    Chrome: 137.0.7151.120
    Edge: 137.0.3296.93
    Safari: 18.5

### Details

Concatenated modules have a static `lib_ident` that does not properly respect the `context` (or any other options) passed into it: https://github.com/web-infra-dev/rspack/blob/006d33e1935b356c1be4d37874eb663773db404f/crates/rspack_core/src/concatenated_module.rs#L1498-L1500

This diverges from Webpack's behavior, where the root module's `libIdent` method is called: https://github.com/webpack/webpack/blob/bba9874f0d896da00465b3947493f8e1948374c0/lib/optimize/ConcatenatedModule.js#L727-L729

This ends up causing the manifest produced by the `LibManifestPlugin` (via the `DllPlugin`) to be incorrect, which can make it difficult to have consumers consume DLLs properly.

### Reproduce link

https://github.com/pgoldberg/rspack-concatenated-lib-ident-bug

### Reproduce Steps

1. Run `pnpm install && pnpm build`
2. Notice that in the DLL manifest (`packages/dlls/lib-dll/build/rspack-output/lib.dll.json`) has a path that does not properly respect the `context` specified in the DLL's Rspack configuration (it is `"../../lib/index.mjs"`, but it should be `./packages/lib/index.mjs` because the [`context` given to the `DllPlugin` is `path.resolve('../../..')`](https://github.com/pgoldberg/rspack-concatenated-lib-ident-bug/blob/91713d0ec7d0761b4fe9c7e076877858aa180c46/packages/dlls/lib-dll/rspack.config.js#L21))
3. Notice that in the app's bundle output (`packages/app/build/rspack-output/app.js`), the library is not properly externalized
4. If you set `optimization.concatenateModules` to `false` in the DLL's [`rspack.config.js`](https://github.com/pgoldberg/rspack-concatenated-lib-ident-bug/blob/91713d0ec7d0761b4fe9c7e076877858aa180c46/packages/dlls/lib-dll/rspack.config.js), then the `context` is properly respected and the library is properly externalized

hardfist:
## Summary

Fix DLL manifest generation for concatenated modules by resolving the concatenation root before computing its library identifier. This ensures `DllPlugin.context` is honored, allowing `DllReferencePlugin` consumers to match and externalize the DLL correctly.

Add a regression case that forces module concatenation, verifies the concatenated module is present, and asserts that the manifest key is relative to the configured DLL context.

## Related links

Closes #10799

## Checklist

### DIFF
--- a/crates/rspack_plugin_dll/src/lib_manifest_plugin.rs
+++ b/crates/rspack_plugin_dll/src/lib_manifest_plugin.rs
@@ -128,7 +128,9 @@ async fn emit(&self, compilation: &mut Compilation) -> Result<()> {
       // Match webpack by deriving the library identifier from the concatenation root.
       let module_for_ident = module
         .as_concatenated_module()
-        .and_then(|module| module_graph.module_by_identifier(&module.get_root()))
+        .and_then(|concatenated_module| {
+          module_graph.module_by_identifier(&concatenated_module.get_root())
+        })
         .unwrap_or(module);
       let ident = module_for_ident.lib_ident(LibIdentOptions { context });
 
--- a/crates/rspack_plugin_dll/src/lib_manifest_plugin.rs
+++ b/crates/rspack_plugin_dll/src/lib_manifest_plugin.rs
@@ -125,7 +125,12 @@ async fn emit(&self, compilation: &mut Compilation) -> Result<()> {
         None => &compilation.options.context,
       };
 
-      let ident = module.lib_ident(LibIdentOptions { context });
+      // Match webpack by deriving the library identifier from the concatenation root.
+      let module_for_ident = module
+        .as_concatenated_module()
+        .and_then(|module| module_graph.module_by_identifier(&module.get_root()))
+        .unwrap_or(module);
+      let ident = module_for_ident.lib_ident(LibIdentOptions { context });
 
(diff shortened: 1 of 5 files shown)


hardfist:
@hardfist PTAL

hardfist:
@hardfist review

hardfist:
@hardfist PTAL

stormslowly (approved):

PR #14944, opened 2026-07-25
fix(cache): invalidate persistent cache when build dependencies snapshot is missing

### LINKED ISSUE #14864: [Bug]: missing snapshot_build recovery keeps stale build-dependency output
### Reproduction

Repository: https://github.com/hardfist/rspack-persistent-cache-missing-build-deps-recovery-repro

This is a JavaScript repro using the public compiler API and the published `@rspack/core@2.1.4` package:

```sh
pnpm install
pnpm repro
```

The repro uses a loader that reads an external `flag.txt` file. That file is explicitly listed in `cache.buildDependencies`, so changing it should invalidate the persistent cache.

Steps:

1. Build with `flag.txt = A`.
2. Remove all records from the `snapshot_build` scope and recompute its pack integrity hash.
3. Change `flag.txt` to `B` and start a fresh process against the cache.
4. Compare with a cold build.

Observed on `@rspack/core@2.1.4`:

```
recovered exit: 0
recovered output contains A: true
recovered output contains B: false
cold build has errors: false
cold output contains B: true

BUG REPRODUCED: missing build-dependency recovery keeps stale output.
```

The recovered build succeeds but silently emits the old result even though the explicitly configured build dependency changed.

### Expected behavior

An absent/incomplete `snapshot_build` scope should invalidate the cache or conservatively rerun build-dependency validation. A changed file listed in `cache.buildDependencies` must not be ignored during persistent-cache recovery.

### Environment

- `@rspack/core@2.1.4`
- Node.js 24
- macOS arm64

lazerg:
## Summary

Persistent-cache recovery trusted the `snapshot_build` scope even when it was empty. `BuildDeps::validate` only returned `Invalid` when a tracked file was modified or removed, so an absent build-dependency snapshot fell through to `Valid { tracked_files: 0 }`. Recovery then reused the cached artifacts and kept the old output, ignoring a file listed in `cache.buildDependencies` that had changed.

`validate` now also invalidates when the BUILD scope has no records while build dependencies were configured and the rest of the cache is still present, which only happens when that snapshot was lost. A genuine first build (no cache at all) stays valid.

## Related links

Fixes #14864

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
+++ b/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs
@@ -27,6 +27,8 @@ pub enum BuildDepsValidationResult {
     modified_files: ArcPathSet,
     removed_files: ArcPathSet,
   },
+  /// The snapshot is missing while the rest of the cache is present.
+  Missing,
 }
 
 /// Build dependencies manager
@@ -102,7 +104,7 @@ impl BuildDeps {
   ///
   /// If any build dependencies have changed, this method will return an invalid result.
   pub async fn validate(&mut self, storage: &dyn Storage) -> Result<BuildDepsValidationResult> {
-    let (_, modified_files, removed_files, no_changed_files) = self
+    let (is_hot_start, modified_files, removed_files, no_changed_files) = self
       .snapshot
       .calc_modified_paths(storage, SnapshotScope::BUILD)
       .await?;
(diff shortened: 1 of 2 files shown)


lazerg:
@chenjiahan could you take a look at this cache invalidation fix when you have a moment?

lazerg:

PR #14945, opened 2026-07-25
fix(cache): recover from corrupt persistent cache module graph

### LINKED ISSUE #14858: [Bug]: corrupt persistent-cache module graph recovery panics
### Reproduction

Repository: https://github.com/hardfist/rspack-persistent-cache-make-recovery-repro

This is a Node-level repro using the public `@rspack/core` compiler API. It pins the latest published package at the time of filing, `@rspack/core@2.1.4`; it does not build or call Rspack's Rust crates.

Run:

```sh
pnpm install
pnpm repro
```

The script creates a real filesystem persistent cache, corrupts the first value in `occasion_make_module_graph/0.pack`, recomputes that pack's integrity hash, and starts a fresh Rspack process against the cache.

On `@rspack/core@2.1.4`, the child aborts with:

```text
unexpected module graph deserialize failed
```

The parent reports `BUG REPRODUCED` and exits 0.

### Expected behavior

A malformed persistent-cache entry should return a recovery error so the cache context can invalidate/reset that scope. It should not abort the process.

### Environment

- `@rspack/core@2.1.4`
- Node.js 24.16.0
- macOS arm64

lazerg:
## Summary

A corrupt module graph entry in the persistent cache made recovery call `.expect()` on the decode result, which aborted the whole process. It now returns the decode error instead, so `load_occasion` resets the make scope and the next build recomputes it, the same way snapshot and build-dependency load failures are already handled.

The devtool and minimize occasions can skip a bad entry, but this one can't: dropping a single node would leave the module graph inconsistent, so the whole scope is reset.

## Related links

Closes #14858

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/occasion/make/module_graph.rs
+++ b/crates/rspack_core/src/cache/persistent/occasion/make/module_graph.rs
@@ -129,17 +129,21 @@ pub async fn recovery_module_graph(
   let mut need_check_dep = vec![];
   let mut mg = ModuleGraph::default();
   let mut module_to_lazy_make = ModuleToLazyMake::default();
+  let mut decode_error = None;
   storage
     .load(SCOPE)
     .await?
     .into_par_iter()
-    .map(|(_, v)| {
-      codec
-        .decode::<Node>(&v)
-        .expect("unexpected module graph deserialize failed")
-    })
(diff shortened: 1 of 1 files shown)


lazerg:
@chenjiahan mind reviewing this one too, on recovering from a corrupt module graph entry?

lazerg:

PR #14946, opened 2026-07-25
fix(cache): recover from corrupt persistent cache meta instead of panicking

### LINKED ISSUE #14859: [Bug]: corrupt persistent-cache meta recovery panics
### Reproduction

Repository: https://github.com/hardfist/rspack-persistent-cache-meta-recovery-repro

This is a Node-level repro using the public `@rspack/core` compiler API. It pins the latest published package at the time of filing, `@rspack/core@2.1.4`; it does not build or call Rspack's Rust crates.

Run:

```sh
pnpm install
pnpm repro
```

The script creates a real filesystem persistent cache, truncates the first value in `meta/0.pack` to one byte, recomputes that pack's integrity hash, and starts a fresh Rspack process against the cache.

On `@rspack/core@2.1.4`, the child aborts with:

```text
should decode success
```

The parent reports `BUG REPRODUCED` and exits 0.

### Expected behavior

A malformed persistent-cache entry should return a recovery error so the cache context can invalidate/reset the meta scope. It should not abort the process.

### Environment

- `@rspack/core@2.1.4`
- Node.js 24.16.0
- macOS arm64

lazerg:
## Summary

Recovering the persistent cache `meta` scope decoded its stored value with `.expect("should decode success")`, so a corrupted entry aborted the whole process instead of being handled as a recoverable failure. Propagate the decode error instead. `CacheContext::load_occasion` already treats a failed recovery by resetting the occasion's scope, so a bad `meta` entry now falls back to a clean rebuild rather than a panic.

## Related links

Fixes #14859

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/occasion/meta/mod.rs
+++ b/crates/rspack_core/src/cache/persistent/occasion/meta/mod.rs
@@ -60,11 +60,29 @@ impl Occasion for MetaOccasion {
       return Ok(());
     };
 
-    let meta: Meta = self.codec.decode(&value).expect("should decode success");
+    let meta: Meta = self.codec.decode(&value)?;
     if get_current_dependency_id() != 0 {
       panic!("The global dependency id generator is not 0 when the persistent cache is restored.");
     }
     set_current_dependency_id(meta.max_dependencies_id);
     Ok(())
+  }
+}
+
+#[cfg(test)]
+mod test {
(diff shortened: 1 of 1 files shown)


lazerg:
@chenjiahan this is the companion fix for the meta scope, similar idea to the module graph one, would appreciate a look when you're free.

lazerg:

PR #14947, opened 2026-07-25
fix: keep incremental missing-module diagnostic location stable

### LINKED ISSUE #14857: [Bug]: incremental missing-module diagnostics can point at a later use instead of the import
### Reproduction

- Rspack: @rspack/core@2.1.4
- Repository: https://github.com/hardfist/rspack-repro-incremental-error-location

The entry imports the same missing request twice:

~~~js
import { first } from "./dep.js";
import { second } from "./dep.js";
console.log(first, second);
~~~

The script builds once, deletes dep.js, and compares an incremental rebuild with a cold build.

~~~text
incremental rebuild after deleting dep.js: 3:13-18
cold build with dep.js missing:             1:1-34
BUG REPRODUCED: incremental location 3:13-18 differs from cold location 1:1-34.
~~~

### Expected behavior

The same missing request should report a stable location regardless of whether the module graph was repaired incrementally or rebuilt cold.

### Suspected area

The repair path groups affected dependencies in crates/rspack_core/src/compilation/build_module_graph/graph_updater/repair/process_dependencies.rs, while the diagnostic is assigned from create_data.dependencies[0].loc() in crates/rspack_core/src/compilation/build_module_graph/graph_updater/repair/factorize.rs. The incremental path can select a different dependency from the cold path when a group is reconstructed from affected dependency collections.

lazerg:
## Summary

When a module imports the same missing request more than once, an incremental rebuild could report the "Module not found" error at a different import than a cold build. The incremental repair path groups the affected dependencies straight from a `HashSet`, so the dependency that supplies the diagnostic location was picked in arbitrary order. Sort each group back into the parent module's source order before factorizing, so the reported location stays the same whether the graph was repaired incrementally or built cold.

## Related links

Closes #14857

## Checklist


lazerg:
@chenjiahan whenever you have time, could you check this diagnostic location fix?

LingyuCoder:
@codex review

LingyuCoder:
Thanks for tracking this down and providing the regression test!

The repair path is only one of several paths that feed dependency groups into factorization. Restoring dependency order there fixes this particular path, but it does not make diagnostic location selection intrinsically stable.

I opened #15456 to address this where the diagnostic location is assigned: it selects the earliest source location among all dependencies in the same factorization group. This keeps the diagnostic output stable regardless of the ordering of the upstream dependency collection.

#15456 reuses your test case verbatim and credits you as a co-author. Thanks again for the investigation and contribution!

Issue #14961, opened 2026-07-27
Diagnostics: memory/time scale as O(error count × line length) with no cap — a 267KB input can drive the build process past 10GB RSS (78GB observed on a real-world file)


cezaraugusto:
### System Info

- rspack: reproduced on `@rspack/core` **2.1.1** and **2.1.2** (latest at time of writing), JS API, `mode: 'production'`
- OS: macOS 15 (darwin arm64), Node 23.8.0, 32GB RAM
- No loaders, no plugins — minimal config below

### Details

When a module produces many recoverable parse errors, diagnostic construction/rendering appears to be unbounded in both **count** and **per-diagnostic snippet size**:

- Each diagnostic's rendered snippet embeds the full offending source line, so cost is roughly **O(error count × line length)**. For generated/minified single-line files, both factors grow with file size — memory blows up quadratically.
- There is no cap on the number of diagnostics materialized (webpack renders all errors too, but does not embed multi-MB source lines per error).

Measurements (repro below; `/usr/bin/time -l`, `stats.toJson({errors:true})`):

| escapes → diagnostics | input file | shape | wall time | max RSS |
| --- | --- | --- | --- | --- |
| 1,002 | 5 KB | single line | ~1s | 42 MB |
| 10,002 | 50 KB | single line | 13.1s | **4.08 GB** |
| 10,002 | 53 KB | one string per line | 0.83s | 1.33 GB |
| 50,001 | 267 KB | one string per line | 67.4s | **10.98 GB** |

The single-line vs multi-line rows isolate the mechanism: same error count, ~same byte size — 5× slower and 3× the memory when the offending line is long, because every diagnostic renders a snippet of the same 50KB line. Even with short lines, 50k diagnostics cost 11GB.

[...]

PR #14967, opened 2026-07-28
perf: share source code across diagnostics

hardfist:
## Summary

- store diagnostic source code in shared immutable storage
- reuse source storage across parser errors and SWC diagnostics
- avoid intermediate source allocations when converting SWC sources
- add a regression test for repeated parse errors

## Performance impact

Diagnostic creation previously cloned the full module source for every parse error, making memory usage grow with `error count × source size`. Sharing one immutable source allocation per file makes each additional diagnostic clone only a small reference.

Measured with [rspack-issue-14961](https://github.com/cezaraugusto/rspack-issue-14961) on macOS arm64 and Node.js 24.16 using `@rspack/core` 2.1.5 dev bindings. Values are medians of three warm runs comparing `main@019c3aa232` with this PR.

| Scenario | Metric | main | PR | Change |
| --- | ---: | ---: | ---: | ---: |
| 10k errors, compile only | Peak RSS | 1,305.7 MB | 143.5 MB | **-89.0%** |
| 10k errors, compile only | Build time | 151 ms | 72 ms | **-52.3%** |
| 1k errors, single-line full workflow | Peak RSS | 292.5 MB | 282.0 MB | -3.6% |
| 1k errors, multi-line full workflow | Peak RSS | 162.4 MB | 149.1 MB | -8.2% |

The rendered output remains unchanged at 7,819,499 characters for the 1k single-line case. This PR addresses source clone amplification; bounding diagnostic rendering remains separate.

Addresses one source of memory amplification reported in #14961.

### DIFF
--- a/crates/rspack_javascript_compiler/src/compiler/parse.rs
+++ b/crates/rspack_javascript_compiler/src/compiler/parse.rs
@@ -61,7 +61,7 @@ impl JavaScriptCompiler {
           .with_context(ast::Context::new(self.cm, Some(self.globals)))
       })
       .map_err(|errs| {
-        let source: Arc<str> = fm.src.to_string().into();
+        let source: Arc<str> = Arc::from(fm.src.as_ref());
         BatchErrors(
           errs
             .dedup_ecma_errors()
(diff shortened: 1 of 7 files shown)


crates/rspack_javascript_compiler/src/compiler/parse.rs
```diff
@@ -61,11 +61,12 @@ impl JavaScriptCompiler {
           .with_context(ast::Context::new(self.cm, Some(self.globals)))
       })
       .map_err(|errs| {
+        let source: Arc<str> = fm.src.to_string().into();
```
SyMind: change to `Arc::from(fm.src.as_ref())` or `Arc<String>`

SyMind (approved):

PR #14993, opened 2026-07-29
fix(runtime): remove redundant runtime declarations

LingyuCoder:
## Summary

- Treat runtime globals defined by native runtime modules as owned even when they are also exposed through the runtime context.
- Remove redundant lexical declarations for globals such as `publicPath`, `scriptNonce`, and `initializeSharing`.
- Update the affected modern-module runtime snapshots.

## Problem

`define` means a runtime module creates the binding, while `force_context` only means that the binding must also be exposed through the runtime context. The previous ownership calculation excluded `force_context` fields from `define`, so the chunk renderer could emit a redundant declaration before the runtime module definition:

```js
var publicPath;
var publicPath = "/public/path/";
```

The redeclaration is valid JavaScript, but it is unnecessary output.

## Why this works

Native runtime modules already emit the lexical definition, for example `var publicPath = ...`. Context accessors or ESM exports then expose that same binding; they do not require a separate outer declaration. Therefore, fields in both `define` and `force_context` should remain owned by the runtime module.

`STARTUP` remains excluded because it follows separate bootstrap ownership rules.

## Correctness considerations

This changes declaration ownership only for native runtime modules whose fields are present in both `define` and `force_context`. Custom-source runtime modules and `RuntimeModuleFromJs` remain excluded.

The PR-specific risk is a missing binding if such a runtime module reports a field in `define` but does not emit its definition in the final output scope. The updated modern-module snapshots cover the affected output shapes.

## Related links

- https://github.com/web-infra-dev/rslib/pull/1806#issuecomment-5103444616

## Checklist

## Testing

- `cargo fmt --all --check`
- `pnpm --parallel --filter "@rspack/*" test` (46 Rspack test files, 8883 passed, 4 skipped; CLI 89 passed, 1 skipped)


LingyuCoder:

JSerFeng (approved):

PR #15000, opened 2026-07-30
fix(cache): align cache version semantics with webpack

hardfist:
## Summary

- keep one persistent cache per compiler path and remove multi-version storage support
- validate the Rspack package version, `cache.version`, and `buildDependencies` through one compatibility gate
- retain `cache.maxVersions` as a deprecated JS-only no-op that warns when explicitly configured

## Behavior

`compilerPath` selects the cache subdirectory and is hashed into a filesystem-safe directory name. The effective compatibility version, `<rspack-version>|<cache.version>`, is stored in that directory metadata rather than used as part of the path.

Before any artifact is reused, one validator checks the effective version first and then `buildDependencies`, matching webpack's PackContainer flow. If either validation fails, the complete cache for that compiler path is discarded and rebuilt. This prevents an unused scope, such as the source-map cache, from surviving an invalidating build.

A shared storage directory with root and child compilers looks conceptually like:

```text
<cache.storage.directory>/
  rspack_v_<hash(root compilerPath)>/
  rspack_v_<hash(child-a compilerPath)>/
  rspack_v_<hash(child-b compilerPath)>/
```

Each compiler path owns exactly one subdirectory. Changing `cache.version` overwrites that compiler cache in place; switching back does not restore an older version.

`cache.version` remains the caller-controlled compatibility key for application configuration. Changes such as `mode` or compiler `name` should be included explicitly when they need to invalidate cache:

```js
cache: {
  type: "persistent",
  version: `app-v3:${mode}`,
}
```

## Compatibility

`cache.maxVersions` remains accepted by the JavaScript configuration API for backward compatibility, but it is deprecated and has no effect. Rspack emits a warning when the option is explicitly configured.


crates/rspack_core/src/cache/persistent/occasion/meta/mod.rs
```diff
@@ -56,11 +59,21 @@ impl Occasion for MetaOccasion {
 
   #[tracing::instrument("Cache::Occasion::Meta::recovery", skip_all)]
   async fn recovery(&self, storage: &dyn Storage) -> Result<()> {
+    let scopes = storage.scopes().await?;
     let Some((_, value)) = storage.load(SCOPE).await?.pop() else {
+      if scopes
```
hardfist: what's the usage of this

crates/rspack_storage/src/filesystem/mod.rs
```diff
@@ -1,62 +1,80 @@
...
+async fn cleanup_stale_directories(fs: ScopeFileSystem, stale_fs: ScopeFileSystem) -> Result<()> {
+  stale_fs.ensure_exist().await?;
 
-    for stale_version in stale_versions {
-      let _ = stale_version.remove().await;
-    }
+  // The previous multi-version layout used
+  // `rspack_v_<compiler-hash>_<version-hash>`. It is incompatible with the
+  // compiler-path-only layout and can be removed during migration.
+  let legacy_directories = fs
```
hardfist: no need to back compatible with old version

hardfist:
@codex review

crates/rspack_storage/src/filesystem/meta.rs
```diff
@@ -57,9 +57,9 @@ impl Meta {
         ))
       })?;
 
-      // Ignore malformed version ids before they can become cleanup targets.
-      if let Some(version) = Version::parse(version) {
-        meta.access_times.insert(version, timestamp);
+      // Ignore malformed directory names before they can become cleanup targets.
+      if let Some(cache_directory) = CacheDirectory::parse(cache_directory) {
+        meta.access_times.insert(cache_directory, timestamp);
```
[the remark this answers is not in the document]
hardfist: no need to back compat

LingyuCoder:

### PR Overview

Primary classification: **Bug fix**. This PR replaces version-keyed multi-generation persistent-cache directories with one directory per `compilerPath`, stores the effective `<Rspack package version>|<cache.version>` compatibility key in metadata, validates that key together with `buildDependencies`, and invalidates the complete compiler cache when compatibility fails. Writable caches are reset; readonly caches keep their data but block subsequent artifact reads. `cache.maxVersions` remains accepted by the JavaScript configuration surface, but is deprecated and becomes a no-op.

### Problem and Trigger

With the previous layout, changing `cache.version` from `v1` to `v2` and then back to `v1` could select and reuse the retained `v1` directory instead of overwriting it. That can replay stale artifacts when the affected inputs are intentionally excluded from ordinary snapshot checking. Separately, a changed build dependency reset the BUILD snapshot and gated later reads, but an occasion not visited by that invalidating compilation—for example, the source-map cache while source-map caching is disabled—could remain on disk and be restored by a later build.

### Previous Design and Root Cause

The directory version hash included `compilerPath`, serialized persistent-cache options, the Rspack package version, `compiler_options.name`, and `compiler_options.mode`; `maxVersions` retained several such directories. Compatibility was therefore represented by directory selection rather than by an authoritative marker inside the selected cache. Build-dependency failure set the per-build load gate, while the remaining snapshot and occasion scopes were reset only when their individual load paths were reached. Those two choices caused the old-generation reuse and unused-scope survival paths.

### Fix Design

The compiler-path hash now selects a stable directory, while [`CacheValidation`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/validation.rs#L72-L151) checks the stored effective version first and then validates the BUILD snapshot, including newly configured dependency roots. Any mismatch or validation error enters one invalidation path: writable storage calls [`reset_all`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/context.rs#L145-L154), with the task queue sequencing that reset before the replacement save, while readonly storage keeps the invalid read gate set. This directly addresses the root causes: switching versions overwrites the same compiler cache, and an invalidation cannot leave an unvisited artifact scope reusable.

The validation order is analogous to webpack's `PackContainer` restore flow, which rejects a version mismatch before checking build-dependency snapshots and only exposes pack data after both checks succeed ([webpack source](https://github.com/webpack/webpack/blob/9d3b707d9edc7ca537bb3ce5943af2ce3c19e7e5/lib/cache/PackFileCacheStrategy.js#L1322-L1423)). Rspack additionally performs an explicit check for newly configured dependency roots.

### Correctness Risks

No material output-correctness defect was identified in the normal version/build-dependency invalidation flow. The added tests cover version switching without generation resurrection, whole-cache reset, newly added build dependencies, readonly invalidation state, compiler-path separation, and the `maxVersions` deprecation. Current required CI aggregation is green across lint, Rust checks/tests, and the Linux, Windows, macOS, and WASM test paths that ran.

There is an intentional behavioral compatibility boundary: `compiler_options.name`, `compiler_options.mode`, and persistent-cache options are no longer automatically part of the compatibility key. Default cache naming still separates the common name/mode cases, but configurations that explicitly share `cache.storage.location`, mutate options after defaults, or change snapshot/serialization-affecting options must also change `cache.version`. Likewise, existing valid `maxVersions` configurations still parse but no longer retain reusable generations. The updated cache documentation explains this caller responsibility and the deprecation, but the change remains observable for users who relied on the previous automatic isolation.

- **Verified edge case:** a configured build dependency that matches `snapshot.immutablePaths` will invalidate the entire cache on every subsequent process start. [`Snapshot::add`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/snapshot/mod.rs#L44-L48) deliberately stores no strategy for immutable paths, so the root can never appear in `no_changed_files`; the new [`pending` check](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs#L99-L127) therefore classifies it as newly added whenever metadata exists. Output remains conservative, but persistent-cache hits are lost. The changed tests do not cover this combination.
- The immediately preceding `rspack_v_<compiler-hash>_<version-hash>` layout is intentionally not migrated. Its names are rejected by the new parser, ignored by metadata loading, and never moved into `_stale`, so `maxAge` will not reclaim those directories; upgrades may require one-time manual cleanup to avoid retained disk usage.
- No general hot-path regression is apparent: the extra work is concentrated at startup and on invalidation. The latest CodSpeed report found no change in 46 executed benchmarks, but 47 were skipped, and ecosystem-per-commit CI was skipped; neither result directly exercises the immutable-build-dependency or legacy-layout cases.

<details>

### PR 概览

主要分类：**缺陷修复**。该 PR 将按版本区分、可保留多代的持久化缓存目录改为每个 `compilerPath` 只使用一个目录；将实际的 `<Rspack package version>|<cache.version>` 兼容性键存入元数据；统一校验该键与 `buildDependencies`；并在兼容性校验失败时使该 compiler 的完整缓存失效。可写缓存会被重置；readonly 缓存保留数据，但后续 artifact 读取会被阻止。JavaScript 配置层仍接受 `cache.maxVersions`，但该选项已弃用且不再产生作用。

### 问题与触发条件

在旧目录布局下，将 `cache.version` 从 `v1` 改为 `v2` 后再切回 `v1`，可能重新选择并复用之前保留的 `v1` 目录，而不是覆盖它。当相关输入被有意排除在普通 snapshot 校验之外时，这会恢复过期 artifact。另一方面，构建依赖发生变化时只会重置 BUILD snapshot 并阻止后续读取；如果本次失效构建没有访问某个 occasion，例如关闭 source-map 缓存时未访问 source-map cache，该 scope 仍可能留在磁盘上，并在之后的构建中被恢复。

### 旧设计与根本原因

旧目录版本哈希包含 `compilerPath`、序列化后的持久化缓存选项、Rspack package version、`compiler_options.name` 与 `compiler_options.mode`，同时由 `maxVersions` 保留多个目录。因此，兼容性由目录选择表达，而不是由选中缓存内部的权威标记表达。构建依赖校验失败时会设置单次构建的加载门禁，但其余 snapshot 与 occasion scope 只有在各自的加载路径实际执行时才会重置。这两个设计分别导致了旧版本缓存被重新复用，以及未访问 scope 在失效后继续存活。

### 修复设计

现在由 compiler-path 哈希选择稳定目录，而 [`CacheValidation`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/validation.rs#L72-L151) 会先检查已存储的实际版本，再校验 BUILD snapshot，并检查新配置的依赖根路径。任何版本不匹配或校验错误都会进入同一条失效路径：可写 storage 调用 [`reset_all`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/context.rs#L145-L154)，task queue 会保证该重置先于替换缓存的保存执行；readonly storage 则持续保持无效读取门禁。该方案直接解决根因：切换版本会覆盖同一个 compiler cache，且一次失效不会让未访问的 artifact scope 继续可复用。

该校验顺序与 webpack 的 `PackContainer` 恢复流程一致：先拒绝版本不匹配，再检查构建依赖 snapshot，只有两项都通过后才暴露 pack 数据（[webpack 源码](https://github.com/webpack/webpack/blob/9d3b707d9edc7ca537bb3ce5943af2ce3c19e7e5/lib/cache/PackFileCacheStrategy.js#L1322-L1423)）。Rspack 还额外显式检查了新配置的依赖根路径。

### 正确性风险

在常规版本与构建依赖失效流程中，没有发现影响输出正确性的实质缺陷。新增测试覆盖了版本切换时不恢复旧代缓存、完整缓存重置、新增构建依赖、readonly 失效状态、compiler-path 隔离以及 `maxVersions` 弃用。当前 required CI 汇总为绿色，包括 lint、Rust check/test，以及实际运行的 Linux、Windows、macOS 和 WASM 测试路径。

这里存在一个有意引入的行为兼容边界：`compiler_options.name`、`compiler_options.mode` 和持久化缓存选项不再自动进入兼容性键。默认缓存命名仍会隔离常见的 name/mode 场景，但显式共享 `cache.storage.location`、在 defaults 之后修改选项，或变更会影响 snapshot/序列化的配置时，也必须同步更新 `cache.version`。同样，现有合法的 `maxVersions` 配置仍可解析，但不再保留可复用的多代缓存。更新后的缓存文档说明了调用方责任和弃用情况，但对于依赖旧版自动隔离行为的用户，这仍是可观察的变化。

### 性能回归风险

- **已确认的边界情况：** 如果配置的 build dependency 匹配 `snapshot.immutablePaths`，后续每次启动新进程都会使完整缓存失效。[`Snapshot::add`](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/snapshot/mod.rs#L44-L48) 会有意跳过 immutable path，不为其保存 strategy，因此该根路径永远不会出现在 `no_changed_files` 中；只要元数据存在，新增的 [`pending` 检查](https://github.com/web-infra-dev/rspack/blob/41d1ff59a7a85ba3f4467d4f8f4fdb3f3c9d37c1/crates/rspack_core/src/cache/persistent/build_dependencies/mod.rs#L99-L127) 就会把它判定为新加入的依赖。输出仍然是保守且正确的，但持久化缓存命中会全部丢失。变更后的测试没有覆盖这一组合。
- 紧邻本 PR 之前的 `rspack_v_<compiler-hash>_<version-hash>` 布局被有意排除在迁移范围之外。新 parser 会拒绝这些名称，元数据加载会忽略它们，也不会将其移动到 `_stale`，因此 `maxAge` 无法回收这些目录；升级后可能需要一次手动清理，以避免磁盘占用长期保留。
- 未发现常规热路径上的普遍回归：新增工作主要集中在启动与失效阶段。最新 CodSpeed 报告中，实际运行的 46 个 benchmark 没有变化，但另有 47 个被跳过，ecosystem-per-commit CI 也被跳过；两者都没有直接覆盖 immutable build dependency 或旧目录布局场景。

</details>

crates/rspack_core/src/cache/persistent/context.rs
```diff
@@ -366,3 +427,157 @@ fn append_paths_group(output: &mut String, label: &str, paths: &ArcPathSet) {
     output.push_str("\n  - ...");
   }
 }
+
+#[cfg(test)]
```
LingyuCoder: These rust test cases are not needed

crates/rspack_core/src/cache/persistent/validation.rs
```diff
@@ -0,0 +1,311 @@
...
+
+  fn restore_meta(meta: Meta) {
+    if get_current_dependency_id() != 0 {
+      panic!("The global dependency id generator is not 0 when the persistent cache is restored.");
+    }
+    set_current_dependency_id(meta.max_dependencies_id);
+  }
+}
+
+#[cfg(test)]
```
LingyuCoder: Should also been removed

crates/rspack_storage/src/filesystem/cache_directory.rs
```diff
@@ -0,0 +1,77 @@
...
+  }
+}
+
+impl AsRef<str> for CacheDirectory {
+  fn as_ref(&self) -> &str {
+    self.as_str()
+  }
+}
+
+#[cfg(test)]
```
LingyuCoder, line 57: These too

crates/rspack_storage/src/filesystem/mod.rs
```diff
@@ -203,3 +210,63 @@ impl Storage for FileSystemStorage {
     Ok(names)
   }
 }
+
+#[cfg(test)]
```
LingyuCoder, line 214: These too

LingyuCoder (approved):

PR #15001, opened 2026-07-30
fix: preserve context exports in rspack runtime mode

LingyuCoder:
## Summary

- derive the CompatibilityPlugin require identifier by rendering `RuntimeGlobals::REQUIRE_SCOPE` with the non-export runtime representation
- recognize `__webpack_require__` in webpack mode and `__rspack_context` in Rspack modes without hardcoded identifiers
- avoid misclassifying a user-defined `context` binding as a nested runtime require binding
- leave top-level collision handling to the ESM linker and add execution coverage in both runtime modes

## Testing

- `cargo check -p rspack_core -p rspack_plugin_javascript`
- `corepack pnpm --filter @rspack/binding run build:dev`
- targeted EsmOutput and RuntimeModeEsmOutput cases

### DIFF
--- a/crates/rspack_core/src/runtime_template.rs
+++ b/crates/rspack_core/src/runtime_template.rs
@@ -867,6 +867,14 @@ impl ModuleCodeTemplate {
     self.runtime_globals.render(runtime_globals)
   }
 
+  pub fn render_runtime_scope(&self) -> String {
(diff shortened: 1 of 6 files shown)


LingyuCoder:

JSerFeng:
@codex review

JSerFeng (approved):

PR #15002, opened 2026-07-30
fix: avoid runtime scope conflicts in module factories

LingyuCoder:
## Summary

- render module-factory runtime scopes through the module code template, keeping `rspackRequire` as the outer scope for `RspackExport` factories while avoiding redundant or conflicting factory parameters
- collect runtime requirements added during module-content rendering before choosing factory arguments, including requirements introduced by devtool wrappers and concatenated modules
- decouple chunk code templates from runtime-module Dojang templates so chunk rendering does not keep runtime template registration alive
- render SRI hash storage with `RuntimeCodeTemplate` and retain only the rendered require name in the compilation context
- update affected runtime-mode snapshots and add regression coverage for CommonJS `rspackRequire` binding conflicts and modern-module SRI async chunks

## Related links

- Follow-up to #15001
- Downstream case: https://github.com/web-infra-dev/rslib/pull/1806

## Testing

- `cargo clippy --workspace --all-targets --tests --locked -- -D warnings`
- `corepack pnpm --filter @rspack/binding run build:dev`
- targeted `EsmOutput` and `RuntimeModeEsmOutput` cases for `cjs-rspack-require` and `modern-module-sri`

## Checklist

- [x] Tests updated.
- [x] Documentation updated (not required; internal rendering behavior only).

### DIFF
--- a/crates/rspack_core/src/runtime_template.rs
+++ b/crates/rspack_core/src/runtime_template.rs
@@ -362,7 +362,7 @@ impl RuntimeTemplate {
     RuntimeCodeTemplate::new(
       self.compiler_options.clone(),
       self.render_mode.runtime_module_render_mode(),
-      self.dojang.clone(),
+      Some(self.dojang.clone()),
     )
   }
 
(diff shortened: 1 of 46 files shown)


LingyuCoder:

JSerFeng:
@codex review

JSerFeng (approved):

PR #15012, opened 2026-07-30
fix(worker): handle SharedWorker string name option

elecmonkey:
## Summary

Fix the string name overload of `SharedWorker`.

Previously, Rspack transformed:

```js
new SharedWorker(url, "chat");
```

into:

```js
new SharedWorker(url, Object.assign({}, "chat", { type: "module" }));
```

This converted the string into an object without a `name` property, causing the SharedWorker name to be lost.

This PR handles the second argument based on its syntax:

- Object literals keep the existing `Object.assign` behavior.
- String literals are preserved for classic output and converted to `{ name: "chat", type: "module" }` for module output.
- Expressions whose type is unknown at compile time use a runtime check to distinguish strings from objects and are evaluated only once.

> Since the SharedWorker flag and option kind jointly describe one rendering semantic, they are encoded together to keep `add_dependencies` within Clippy's argument limit.

The new behavior is limited to `SharedWorker`. `Worker`, Service Worker, Node.js Worker, and custom worker aliases are unchanged.

## Related links

- Follow-up to #14874
- https://html.spec.whatwg.org/multipage/workers.html#shared-workers-and-the-sharedworker-interface

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/worker_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/worker_plugin.rs
@@ -44,8 +44,8 @@ struct ParsedNewWorkerOptions {
 #[derive(Debug, Clone, Copy)]
 enum ParsedNewWorkerOptionsKind {
   Object,
-  String,
-  Unknown,
+  SharedWorkerString,
+  SharedWorkerUnknown,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


elecmonkey:
@codex review

crates/rspack_plugin_javascript/src/parser_plugin/worker_plugin.rs
```diff
@@ -46,15 +54,28 @@ struct ParsedNewWorkerImportOptions {
   pub ignored: Option<bool>,
 }
 
-fn parse_new_worker_options(arg: &ExprOrSpread) -> ParsedNewWorkerOptions {
+fn parse_new_worker_options(arg: &ExprOrSpread, is_shared_worker: bool) -> ParsedNewWorkerOptions {
   let obj = arg.expr.as_object();
-  let name = obj
-    .and_then(|obj| get_literal_str_by_obj_prop(obj, "name"))
-    .map(|str| str.value.to_string_lossy().into());
+  let string = arg.expr.as_lit().and_then(|lit| lit.as_str());
```
elecmonkey: When a valid call uses a spread second argument such as `new SharedWorker(url, ..."chat")` with `output.module` enabled, `arg.expr` is still a string literal even though `arg.spread` is set, so this classifies it as `SharedWorkerString`. Because `arg.span()` includes the ellipsis, the rewrite emits `{ name: ..."chat", type: "module" }`, which is invalid JavaScript. Check the spread marker before classifying the argument as a literal string, or handle it through the runtime wrapper.

elecmonkey: Fixed in 4aff4e0866. Spread arguments are no longer classified as string literals and now use the runtime wrapper. Added a regression test for `new SharedWorker(url, ..."spread-string")` covering both classic and module output.

JSerFeng (approved):

PR #15014, opened 2026-07-30
fix(json): match webpack source map behavior for generated modules

matthewdavis-oai:
## Summary

- Match webpack's JSON generator by using an unmapped raw source for generated JavaScript in every source-map mode.
- Avoid attributing generated JSON module code to the original JSON file while preserving JavaScript mappings, `sourcesContent`, JSON exports, and `__proto__` behavior.
- Extend the existing source-map fixture across `source-map` and `cheap-module-source-map`; the fixture fails before the fix and passes afterward.

## Related links

- [webpack JSON generator](https://github.com/webpack/webpack/blob/main/lib/json/JsonGenerator.js)

## Checklist

### DIFF
--- a/crates/rspack_plugin_json/src/lib.rs
+++ b/crates/rspack_plugin_json/src/lib.rs
@@ -17,7 +17,7 @@ use rspack_core::{
   ParseOption, ParserAndGenerator, ParserOptions, Plugin, RuntimeSpec, SourceType, UsageState,
   UsedNameItem,
   diagnostics::ModuleParseError,
-  rspack_sources::{BoxSource, OriginalSource, RawStringSource, Source, SourceExt},
+  rspack_sources::{BoxSource, RawStringSource, Source, SourceExt},
 };
 use rspack_error::{Error, IntoTWithDiagnosticArray, Result, TWithDiagnosticArray, error};
 use rspack_util::{itoa, location::byte_line_column_to_offset};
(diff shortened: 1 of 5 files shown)


yifancong:
@matthewdavis-oai  This change was made at that time so that rsdoctor could obtain the actual size of the JSON bundled into the final assets (https://github.com/web-infra-dev/rspack/pull/12950/changes).

After #15014, JSON modules generally no longer suffer from "missing parsedSize". However, this parsedSize is no longer guaranteed to equal the actual byte size of the JSON code within the final bundle.

jsonModuleSizes only passes: { identifier, size }
without delivering the final generated JSON JavaScript content. Moreover, this size is estimated before codegen based on the following Rust code:
```rust
"module.exports = ".len() + json::stringify(final_json).len()
```
It reads build_info.json_data and export usage for tree-shaking, instead of extracting code from the final asset.

Therefore, if we attribute the output of JSON modules to JSON files, what specific errors or issues are you encountering on your end?
————————————————————————————————————————————————————————————

- New:

After checking, collect_json_module_sizes calculates the used exports of JSON files (https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_rsdoctor/src/module_graph.rs#L34-L79). I think this PR can be merged. @SyMind

SyMind (approved):

PR #15047, opened 2026-08-03
perf(devtool): reduce source map persistent cache size

SyMind:
## Summary

- stores compact source map metadata and source-content ordinals instead of `sourcesContent` or rendered source map assets;
- restores `sourcesContent` by borrowing strings from the current asset's `Source` tree;
- stores only the debug ID and `sourceMappingURL` comments required to reconstruct the final asset;
- stores the final rendered comment for external source maps, while retaining `[url]` for inline source maps and regenerating the complete base64 data URL on cache hits;
- uses the source map filename to distinguish inline source maps from separately emitted source map assets;
- parallelizes cache construction and restoration across assets;
- falls back to recomputation and removes invalid cache entries when source content cannot be restored.

As a result, persistent cache size primarily scales with mappings, source map metadata, compact `u32` indices, and small comment strings, while cache hits can still skip the expensive source map computation for both inline and external source maps.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4456,6 +4456,7 @@ dependencies = [
  "futures",
  "itertools 0.14.0",
  "memchr",
+ "rayon",
  "regex",
  "rspack_collections",
  "rspack_core",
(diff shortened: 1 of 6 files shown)


SyMind:
@codex review

PR #15058, opened 2026-08-04
chore(build): enable createRequire parser in Rslib builds

Timeless0911:
## Summary

Rslib's `createRequire` parser resolves static requests from source, which breaks `@rspack/core`'s output-relative watchpack request and changes the generated binding require shape. This PR enables the parser in both Node and browser builds, routes watchpack through the existing compiled dependency external, and keeps the `RSPACK_BINDING` codmod compatible with the generated output. The explicit parser overrides can be removed after the next Rslib upgrade.

## Related links

- https://github.com/web-infra-dev/rslib/pull/1814
- https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/30813488415

## Checklist

### DIFF
--- a/packages/rspack/rstack.config.ts
+++ b/packages/rspack/rstack.config.ts
@@ -134,7 +134,7 @@ const codmodPlugin: RsbuildPlugin = {
 
       return [
         binding.replace(
-          `__rspack_createRequire_require(process.env.RSPACK_BINDING ? process.env.RSPACK_BINDING : "@rspack/binding")`,
+          `__rspack_createRequire_require(process.env.RSPACK_BINDING || "@rspack/binding")`,
         ),
       ];
     }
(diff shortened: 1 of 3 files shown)


packages/rspack/rstack.config.ts
```diff
@@ -123,37 +123,20 @@ const codmodPlugin: RsbuildPlugin = {
...
+      if (!binding) {
         throw new Error(
           'Cannot find any binding require statements to replace',
         );
       }
 
-      return edits;
+      return [
+        binding.replace(
+          `__rspack_createRequire_require(process.env.RSPACK_BINDING ? process.env.RSPACK_BINDING : "@rspack/binding")`,
```
chenjiahan: Should we use `||` here?

```suggestion
          `__rspack_createRequire_require(process.env.RSPACK_BINDING || "@rspack/binding")`,
```

Timeless0911: nice catch, updated.

chenjiahan (approved):

Issue #5596, opened 2024-02-04
support `import.meta.webpackHot.data`


h-a-n-a:
`webpack-test/hotCases/chunks/update-chunk-loading-runtime`
`webpack-test/hotCases/recover/recover-after-self-error`

PR #15076, opened 2026-08-05
fix: handle optional chains in member hook traversal

zhsama:
## Summary

- Traverse member expressions wrapped in optional chains when resolving parser hooks.
- Reuse extracted static member names so computed properties behave like dot access.
- Add regression coverage for nested and computed `import.meta.webpackHot` chains.

Rspack previously treated `import.meta.webpackHot.data?.components` as an unsupported `import.meta` chain and emitted `undefined.data...`. This fixes the parser traversal generically instead of special-casing HMR or Octane.

## Related links

- Exposed by https://github.com/octanejs/octane/issues/485
- Octane workaround: https://github.com/octanejs/octane/pull/562
- Follow-up to https://github.com/web-infra-dev/rspack/issues/5596

## Validation

- New config case, including runtime-mode config
- Existing optional-chaining and `import.meta.webpackHot` config cases
- Relevant HMR update, recovery, and optional-chaining cases
- `cargo fmt --all --check`
- `cargo check -p rspack_plugin_javascript --all-targets --locked`
- JS formatting and pre-commit lint checks

## Checklist

- [x] Documentation updated (not required; internal parser fix).

## Provenance

- [x] An agent produced this diff (`agent-authored`).


LingyuCoder:
LGTM, thanks

LingyuCoder (approved):

PR #14962, opened 2026-07-27
feat: add builds for binding-linux-(ppc64,s390x)-gnu

logonoff:
## Summary

In binding.js, there are references to the packages `@rspack/binding-linux-s390x-gnu` and `@rspack/binding-linux-ppc64-gnu` which are not built or published. This results in runtime errors on `s390x` and `ppc64le` with "Cannot find native binding".

https://github.com/web-infra-dev/rspack/blob/fe6890c7e11829a6459f1e9641b28aace4c0c268/crates/node_binding/binding.js#L295-L316

This PR adds the full build and release pipeline for both targets:

- Added `powerpc64le-unknown-linux-gnu` and `s390x-unknown-linux-gnu` to napi targets
- Created `npm/linux-ppc64-gnu` and `npm/linux-s390x-gnu` packages
- Added CI build steps using `@napi-rs/cross-toolchain` (which supports both architectures)
- Added matrix entries to all release workflows (canary, stable, preview-commit)
- Added `powerpc64le` → `ppc64` and `s390x` → `s390x` mappings to `build-npm.cjs`
- Bumped `json-escape-simd` from 3.1.0 to 3.1.1 — v3.1.0 has a big-endian bug that silently produces wrong escapes on s390x in release builds (e.g. `"node:module"` → `"node:modle\0…"`), corrupting ESM paths, sourcemaps, and data URLs: https://github.com/napi-rs/json-escape-simd/pull/88

No musl variants are included — `s390x-unknown-linux-musl` is only Rust Tier 3.

## Related links

Closes https://github.com/web-infra-dev/rspack/issues/14958

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #15088, opened 2026-08-06
docs(release): group release targets by Node.js support tier

stormslowly:
## Why

The release target list in `releasing.md` was a flat list, which gave no hint that the targets are supported very unevenly. Two of them (`powerpc64le-unknown-linux-gnu`, `s390x-unknown-linux-gnu`, added in #14962) are Node.js Tier 2 platforms and, like riscv64 and the musl/32-bit Windows targets, are only build-verified in CI — no test suite ever runs on them.

This groups the list by the support tier from the [Node.js platform list](https://github.com/nodejs/node/blob/main/BUILDING.md#platform-list) and states explicitly what Rspack guarantees for everything below Tier 1: the binding builds, nothing more.

Before:

```
- x86_64-unknown-linux-gnu
- aarch64-unknown-linux-gnu
- riscv64gc-unknown-linux-gnu
- powerpc64le-unknown-linux-gnu
...
```

After:

```
**Tier 1**
- x86_64-unknown-linux-gnu
...
**Tier 2**
- powerpc64le-unknown-linux-gnu
...
**Experimental**
...
**Others**
...
```

Docs only, both `en` and `zh`. The set of targets is unchanged and still matches the release matrix in `reusable-release-npm.yml`.

### DIFF
--- a/website/docs/en/contribute/development/releasing.md
+++ b/website/docs/en/contribute/development/releasing.md
@@ -14,21 +14,38 @@ The latest stable release follows the Semantic Versioning specification (x.y.z).
 
 The [full release workflow](https://github.com/web-infra-dev/rspack/actions/workflows/release.yml?query=is%3Asuccess) is triggered manually by Rspack maintainers on Tuesday with the complete release notes.
 
-During the release, the following binary artifacts for the target platforms are built:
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
👍

PR #15093, opened 2026-08-07
feat: expose ExternalModule chunkCondition hook

### LINKED ISSUE #15066: [Feature]: allow plugins to override where `ExternalModule`s are placed
### Background

`ExternalModule::chunk_condition` currently hard-codes webpack's historical rule - an external module may only live in a chunk that contains an entry module: https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/external_module.rs#L1092

```rust
fn chunk_condition(&self, chunk_key: &ChunkUkey, compilation: &Compilation) -> Option<bool> {
  match self.resolve_external_type() {
    "css-import" | "module" if !self.place_in_initial => Some(true),
    _ => Some(compilation.build_chunk_graph_artifact.chunk_graph.get_number_of_entry_modules(chunk_key) > 0),
  }
}
```

webpack made exactly this decision overridable in 5.106.0 via [webpack#20632](https://github.com/webpack/webpack/pull/20632), by turning it into a tappable `SyncBailHook` reachable through `ExternalModule.getCompilationHooks(compilation).chunkCondition`. webpack's own `ContainerReferencePlugin` taps it to keep module federation remotes in entry chunks. rspack has no equivalent, so plugins that need the pre-hook behaviour cannot get it.

### Why this matters outside of bundling

In the Atlassian Data Center ecosystem, `@atlassian/webresource-webpack-plugin` maps "provided dependencies" (modules supplied by the host product at runtime) to externals, and generates an XML descriptor that tells the product which resource bundle to fetch for each generated chunk. The dependency is attached to the bundle of whichever chunk the `ExternalModule` ends up in, so chunk placement decides whether the host product downloads that resource **eagerly on every page** or **lazily when the dynamic import runs**.

### Reproduction

rspack 1.7.6, config below:

```js
module.exports = {
  mode: 'development',
  entry: {
    onlyExternal: './src/only-external.js', // import(/* webpackChunkName: 'lazy-external' */ 'ext-dep')
    withModule: './src/with-module.js',     // import(/* webpackChunkName: 'lazy-module' */ './lazy.js')
  },
  externals: { 'ext-dep': "require('ext-dep')" },
  externalsType: 'var',
  optimization: {
    minimize: false, runtimeChunk: false, splitChunks: false,
    chunkIds: 'named', moduleIds: 'named',
  },
};
```

`./src/lazy.js` is `import dep from 'ext-dep'; export default () => dep;`

Observed:

```
emitted: lazy-module.js, onlyExternal.js, withModule.js
chunk onlyExternal initial=true  modules=["./src/only-external.js", "external var \"require('ext-dep')\""]
chunk withModule  initial=true  modules=["./src/with-module.js",   "external var \"require('ext-dep')\""]
chunk lazy-module initial=false modules=["./src/lazy.js"]
```

Two consequences:

1. `lazy-external` is never emitted. The external is inlined into the initial chunk and the dynamic
   import compiles to `Promise.resolve().then(__webpack_require__.t.bind(...))`, so there is no
   loadable chunk left to associate a lazily fetched resource with.
2. The external is hoisted into **every** initial chunk that can reach it, including `withModule`,
   whose only importer of `ext-dep` is the async `lazy.js`.

webpack solution:

```js
// webpack/lib/ExternalModule.js
chunkCondition(chunk, compilation) {
  const { chunkCondition } = ExternalModule.getCompilationHooks(compilation);
  const condition = chunkCondition.call(chunk, compilation);
  if (condition !== undefined) return condition;
  // ...existing default...
}
```
webpack's own `ContainerReferencePlugin` taps it to keep module federation remotes in entry chunks: https://github.com/webpack/webpack/blob/main/lib/container/ContainerReferencePlugin.js The API we would like in rspack, for source-compatibility with webpack plugins:

```js
import { ExternalModule } from '@rspack/core';
class MyPlugin {
  apply(compiler) {
    compiler.hooks.compilation.tap('MyPlugin', (compilation) => {
      ExternalModule.getCompilationHooks(compilation).chunkCondition.tap('MyPlugin', () => true);
    });
  }
}
```

intellild:
## Summary

- for `import()` externals, keep them in the initial chunk to avoid loading them asynchronously twice and to improve runtime performance.
- expose `ExternalModule.getCompilationHooks(compilation).chunkCondition` to JavaScript
- bridge the JavaScript `SyncBailHook` through the binding to the native Rust `ExternalModule` chunk-condition path
- preserve the existing Rust fallback behavior when JavaScript taps return `undefined`
- use the hook in `ContainerReferencePlugin` to keep remote externals in entry chunks
- port webpack's `module-externals-in-async-chunk` coverage and add a Rspack integration test that verifies a JavaScript tap changes chunk placement
- behavior changes https://github.com/web-infra-dev/rspack/pull/15093#issuecomment-5215402938

## Minimal example

An external used only by a conditionally loaded async module can stay in that async chunk:

```js
// index.js
if (globalThis.USE_NODE_FS) {
  import("./async.js");
}
```

```js
// async.js
export { readFile } from "external";
```

```js
// rspack.config.js
module.exports = {
  externals: { external: "fs" },
  externalsType: "module-import",
  experiments: { outputModule: true },
  output: { module: true }
};
```

For the static import in `async.js`, `module-import` resolves to `module`. Keeping that external in the async chunk means the native `fs` import is evaluated only when `globalThis.USE_NODE_FS` is true and the async chunk is loaded:

```js
// main.mjs
if (globalThis.USE_NODE_FS) {
  __webpack_require__.e("async");
}

// async.mjs
import * as external from "fs";
```

If the external were forced into the entry chunk, the generated entry would eagerly contain the native import:

```js
// main.mjs
import * as external from "fs";

if (globalThis.USE_NODE_FS) {
  __webpack_require__.e("async");
}
```

In an environment where `fs` is unavailable, that eager import can fail during startup even when the condition is false and the application never needs `async.js`.

A plugin that needs the external initialized before async code can override the placement through the new hook:

```js
const { ExternalModule } = require("@rspack/core");

compiler.hooks.compilation.tap("KeepExternalInEntry", compilation => {
  ExternalModule.getCompilationHooks(compilation).chunkCondition.tap(
    "KeepExternalInEntry",
    chunk => compilation.chunkGraph.getNumberOfEntryModules(chunk) > 0
  );
});
```

### Generated output

For `externals: { external: "import fs" }` and `await import("external")`, the generated structure changes as follows (simplified).

Before, the call site loaded a separate chunk before evaluating the external:

```js
const fs = await __webpack_require__
  .e(/* import() */ "external")
  .then(__webpack_require__.bind(__webpack_require__, "external"));
```

Now, the call site directly references an independent ExternalModule in the current chunk:

```js
const fs = await Promise.resolve(/* import() */).then(
  __webpack_require__.bind(__webpack_require__, "external")
);

// Independent ExternalModule
"external"(module) {
  module.exports = import("fs");
}
```

## Related links

- https://github.com/webpack/webpack/pull/20632

## Validation

- `pnpm run build:js`
- `pnpm run test -t "configCases/externals/module-externals-chunk-condition-hook"`
- `pnpm run test -t "configCases/externals/module-externals-in-async-chunk"`
- `pnpm run test -t "configCases/container-1-5/container-reference"`
- `pnpm run test -t "configCases/split-chunks/max-size-split"`
- `cargo fmt --all --check`
- `pnpm exec rs fmt --check`
- `git diff --check`

## Checklist

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
## Summary

Assuming `place_in_initial == false`:

| Raw type | Resolved type | `main` | This branch |
|---|---|---|---|
| `css-import` | `css-import` | Allowed in any chunk | Allowed in any chunk |
| `module` | `module` | Allowed in any chunk | Allowed in any chunk |
| `import` | `import` | Allowed in async chunks | Entry chunks only |
| Static `module-import` reference | `module` | Allowed in async chunks | Allowed in async chunks |
| Dynamic `module-import` reference | `import` | Allowed in async chunks | Entry chunks only |

The core behavior change is:

```text
main
  checks the raw external type
        ↓
"import" externals may remain in async chunks

this branch
  also checks the resolved external type
        ↓
ExternalModules that ultimately use native import() must be placed in entry chunks
```

As a result:

- A static `module-import` reference resolves to `module` and can remain in an async chunk. This prevents an external import behind an unused conditional branch from being evaluated eagerly.
- An `import` external or a dynamic `module-import` reference resolves to `import` and is restricted to entry chunks.
- This is also why the `import-attributes` call site changes from `__webpack_require__.e(chunkId)` to directly binding an independent ExternalModule in the entry chunk.

---
_by OpenAI Codex_

intellild:
@codex review

JSerFeng (dismissed):

JSerFeng:
@codspeedbot is this regression real

JSerFeng (approved):

SoonIter:
<agent: daily-job rstack ecosystem-ci>

### Ecosystem CI regression: `rspack/rstest`

**Attribution:** Confirmed source

**Impact:** Rstest's define-config e2e child process now fails consistently in `ExternalModule.getCompilationHooks`, leaving the Rspack ecosystem row red.

#### Why this PR

This PR added the ExternalModule hook registration path and made `getCompilationHooks` call `checkCompilation`, which performs a strict `instanceof Compilation` check. In Rstest's Rsbuild-based define fixture, the compiler's compilation can cross an `@rspack/core` package-object boundary; the new internal registration path therefore rejects the otherwise valid compilation before the test can run. The same Rstest commit passed immediately before this merge and fails with the same stack on the merge commit and later Rspack commits, so this is a Rspack compatibility regression rather than a downstream change or runner flake.

#### Suggested next step

Rspack should bind the ExternalModule hook registry to the compiler's own core instance, or avoid the cross-package `instanceof` check on this internal registration path while keeping the public mixed-version diagnostic. Verify with a focused dual-core-instance ExternalModule hook fixture and `pnpm test --stack rspack rstest`.

<details>
<summary>Evidence and attribution checks</summary>

- Current run: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/31569316797
- Tested upstream: `rspack` at `4e8371ee4f1ec56129c86bcde84060d1577053ea`
- First bad: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/31566085981/job/94018132479 at `cce3b45bf103a8ad8b34dc86a95c615d0d3a3b33`
- Previous good: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/31564970866/job/94014874650 at `778f6f5c11fb1d09a73919c9b16c00c610384a7a`
- Downstream control: both jobs checked out Rstest `6047a7f5dcf22ac33ce9c10c841ef342ddcbdfdb`
- Failure signature: `build/index.test.ts` / define config fails after three attempts with `The 'compilation' argument must be an instance of Compilation`, at `ExternalModule.getCompilationHooks`; 917 other tests pass
- Flaky/pre-existing check: not observed in the immediately previous run; deterministic on the merge commit and three later Rspack commits
- Config check: not applicable; the failing stack is the newly registered core hook path itself
- Alternatives ruled out: #15155 is a later unrelated ESM-library change; the Rstest SHA is unchanged; frozen lockfile install and the exact stack are stable across the red runs

</details>

pdrosatlassian:
@intellild could this be backported to version 1? 

Thank you very much for quick feature introduction 🙇

intellild:
> @intellild could this be backported to version 1?
> 
> Thank you very much for quick feature introduction 🙇

new features are not backported to v1, please upgrade to v2

chenjiahan (commented):
Can we document this new hook?

intellild:
> Can we document this new hook?

docs updated in #15163

PR #15101, opened 2026-08-07
ci: verify native vs npm wasm-opt on the self-hosted runner

stormslowly:
Measurement only — **not for merge**. Supports #15100.

## Why

#15100 replaces the `binaryen` npm package with the upstream native release binary for `wasm-opt`. The 19x speedup behind that change was measured on a 16-core devbox. The threading factor depends on core count, so the benefit has to be confirmed on the runner that actually builds the release.

The regular PR CI never builds wasm with `profile: release`, so nothing exercises this path outside a canary run.

## What

A single job on the linux self-hosted runner that builds `wasm32-wasip1-threads` with `release-wasi`, then runs both `wasm-opt` builds — same version 131, same input, same machine — and reports timings plus a sha256 comparison of the two outputs to the step summary.

The sha256 assertion is the point as much as the timing: it shows the native binary is not producing a different artifact.

### DIFF
--- a/.github/workflows/verify-wasm-opt.yml
+++ b/.github/workflows/verify-wasm-opt.yml
@@ -1,0 +1,99 @@
+name: Verify wasm-opt
+
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/verify-wasm-opt.yml'
+
+permissions:
+  contents: read
+
+env:
+  CARGO_INCREMENTAL: 0
+  BINARYEN_VERSION: '131'
+
+jobs:
+  verify:
+    name: Verify wasm-opt (wasm32-wasip1-threads, release)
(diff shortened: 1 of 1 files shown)


stormslowly:
Result — [run 31151423743](https://github.com/web-infra-dev/rspack/actions/runs/31151423743):

| | value |
|---|---|
| runner | `rspack-ubuntu-22.04-large` |
| nproc | 192 |
| input | 32,419,871 B |
| npm `binaryen` (JS) | 363,611 ms |
| upstream native | 15,118 ms |
| speedup | **24.1x** |
| output size | 29,809,648 B |
| sha256 match | **yes** |

Confirms #15100 on the runner that actually builds the release: 5m48s saved per wasm job, byte-identical artifact.

stormslowly:
Measurement done, result carried into #15100.

PR #15142, opened 2026-08-10
[POC] feat: add loader chain caching

intellild:
## Summary

Introduce loader chains as the loader runner's execution unit and add opt-in per-loader caching through a new `cache` option alongside `parallel`.

Consecutive cacheable loaders are merged into one loader chain so they share a single input hash calculation and cache lookup. Consecutive JavaScript loaders are also merged, keeping JS yielding and worker scheduling scoped to the chain. The chain planning strategy is isolated from execution so the optimization can be benchmarked or disabled independently.

The cache key includes the loader input, loader requests, loader types, and stable serialized loader options. Cached results replay dependencies and supported build side effects, while loader implementation or dependency changes invalidate the entry. Both memory and persistent compiler caches are supported.

### Example

```js
module.exports = {
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          {
            loader: "first-loader",
            options: { target: "modern" },
            cache: true
          },
          {
            loader: "second-loader",
            cache: true,
            parallel: true
          }
        ]
      }
    ]
  }
};
```

In this example, the two consecutive cached loaders execute as one loader chain. Rspack hashes the chain input once and reuses the complete chain result on subsequent compilations when the input, loader implementations, options, and tracked dependencies are unchanged.

## Validation

- `pnpm run build:js`
- `pnpm run test:rs`
- `cargo lint`
- Loader-chain planner Rust tests: 10 passed
- New watch, `cacheable(false)`, persistent-cache, and loader-parallel cases passed
- Main unit suite: 9061 passed; three browserslist cases hit an existing shared temporary-file cleanup race under parallel execution, and the affected 11 cases passed when rerun in isolation

## Related links

None.

## Checklist

---
_by OpenAI Codex_

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -959,6 +959,12 @@ export interface JsLoaderContext {
   buildDependencies: Array<string>
   loaderItems: Array<JsLoaderItem>
   loaderIndex: number
+  /**
+   * Inclusive start and exclusive end of the current JavaScript execution
+   * span inside the loader chain.
+   */
+  loaderChainStart: number
+  loaderChainEnd: number
   loaderState: Readonly<JsLoaderState>
   __internal__error?: RspackError
   /**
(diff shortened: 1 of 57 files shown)


intellild:
This POC has finished its duty.

PR #15147, opened 2026-08-11
perf: batch splitChunks JavaScript callbacks

JSerFeng:
## Summary

This PR batches only function-form `splitChunks.name` callbacks across the N-API boundary while preserving the public per-item callback API and the original callback suspension points.

It now:

- batches `name` requests in bounded groups
- sends a compact payload containing modules, deduplicated chunks, offsets, and chunk indices


JSerFeng:
@codex review

SyMind (approved):

PR #15152, opened 2026-08-11
perf: parallelize ensure chunk condition checks

intellild:
## Summary

- parallelize the initial module chunk-condition scan with `rspack_parallel::scope`
- parallelize target chunk discovery for modules that violate their current chunk condition
- keep per-module traversal state and complexity counters local, then aggregate the results
- skip source chunks whose condition was already evaluated as `false`

## Why

This explores whether the two module-oriented passes in `EnsureChunkConditionsPlugin` can overlap their work.

For example, when multiple modules need chunk-condition evaluation, each module can scan its currently connected chunks independently. When several modules must be moved, each module can also search its relevant chunk groups and parent groups independently before the final chunk graph mutation is applied.

## Validation

- `cargo check -p rspack_plugin_ensure_chunk_conditions`
- `cargo clippy -p rspack_plugin_ensure_chunk_conditions -- -D warnings`
- `configCases/externals/module-externals-chunk-condition-hook`
- `configCases/externals/module-externals-in-async-chunk`
- `configCases/split-chunks/max-size-split`

## Performance status

The local `rust@split_chunks` Criterion run measured about `6.347 ms`, roughly `10.9%` slower than the existing local baseline. This draft keeps the direct parallel implementation available for CodSpeed analysis and follow-up iteration.

---
_by OpenAI Codex_


intellild:
close for no benefit

PR #15153, opened 2026-08-11
refactor: use trait objects for scoped tasks

intellild:
## Summary

- define boxed trait-object aliases for scoped tasks and their futures
- make `Spawner::spawn` accept the erased task type directly
- update every existing `rspack_parallel::scope` call site on `main`

## Why

`Spawner::spawn` previously used generic closure and future parameters, so each call site produced another monomorphized spawn implementation. Erasing both types at the API boundary lets the compiler reuse one implementation. The tradeoff is explicit boxing and dynamic dispatch for each scoped task; task scheduling and lifetime guarantees remain unchanged.

For example:

```rust
// Before
spawner.spawn(move |used| async move {
  process(used).await
});

// After
spawner.spawn(Box::new(move |used| {
  Box::pin(async move {
    process(used).await
  })
}));
```

This PR is intentionally based on `main` and does not include the parallel ensure-chunk-conditions changes from #15093.

## Validation

- `cargo fmt --all --check`
- `git diff --check`
- `cargo test -p rspack_parallel --target-dir /tmp/rspack-parallel-target`
- `cargo clippy -p rspack -p rspack_plugin_copy -p rspack_plugin_size_limits --target-dir /tmp/rspack-dyn-clippy-target -- -D warnings`

---
_by OpenAI Codex_

### DIFF
--- a/crates/rspack_core/src/compilation/code_generation/mod.rs
+++ b/crates/rspack_core/src/compilation/code_generation/mod.rs
@@ -170,61 +170,63 @@ pub(crate) async fn code_generation_modules(
       // SAFETY: await immediately and trust caller to poll future entirely
       let s = unsafe { token.used((compilation_ref, &module_graph, cache_counter, job)) };
 
-      s.spawn(|(this, module_graph, cache_counter, job)| async move {
-        let options = &this.options;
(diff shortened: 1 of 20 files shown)


intellild:
close for no benefit

PR #15154, opened 2026-08-11
perf: skip empty JavaScript hook registers

intellild:
## Summary

- add a small non-generic `JsTapRegister` trait that only exposes Rust-local empty-state checks
- implement that trait once for `RegisterJsTapsInner`; the same inner also implements each required `Interceptor<Hook>`
- load `Arc<RegisterJsTapsInner>` through a generic hook boundary, clone the Arc into the existing interceptor path, and retain a non-generic trait object only for `has_js_taps()/is_empty()`
- return from generated hook calls before tracing/register dispatch when Rust taps, ordinary interceptors, and real JS taps are all empty
- share each binding register through one `Arc<RegisterJsTapsInner>`, with the cache and `AtomicUsize` embedded directly in the inner allocation
- report the final `JsTap[]` length synchronously from JavaScript to Rust and remove the global `NonSkippableRegisters` lock

Example: a frequently called `buildModule` hook with no JavaScript taps previously still appeared non-empty because its JS register was installed as an ordinary interceptor. Every call entered tracing and the additional-tap path before Rust could discover that there was nothing to run. With this change, JavaScript records a count of `0` in the shared register inner, so the hook returns using only a Rust-local atomic read and performs no Rust-to-JavaScript dispatch.

When a plugin later installs a JavaScript tap, the existing bootstrap/dynamic refresh updates the same atomic state to non-zero. The next hook call executes the same `RegisterJsTapsInner` through its existing typed `Interceptor<BuildModuleHook>` implementation. There is no `into_js_tap_register`, closure, `Any`, downcast, erased tap payload, or second register executor.

## Related links

N/A

## Validation

- `cargo test -p rspack_macros_test --test hook` (9 passed)
- `cargo clippy -p rspack_binding_api -p rspack_macros_test --tests -- -D warnings`
- `pnpm run format-ci:js`
- merge-base CI-profile stripped native binding: `+20,472 bytes` (below the 50 KiB limit by `30,728 bytes`)

## Checklist

---
_by OpenAI Codex_

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -337,6 +337,7 @@ export declare class JsCompilation {
 export declare class JsCompiler {
   constructor(compilerPath: string, options: RawOptions, builtinPlugins: BuiltinPlugin[], registerJsTaps: RegisterJsTaps, outputFilesystem: ThreadsafeNodeFS, intermediateFilesystem: ThreadsafeNodeFS | undefined | null, inputFilesystem: ThreadsafeNodeFS | undefined | null, resolverFactoryReference: JsResolverFactory, unsafeFastDrop: boolean, platform: RawCompilerPlatform)
   setNonSkippableRegisters(kinds: Array<RegisterJsTapKind>): void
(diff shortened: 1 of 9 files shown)


intellild:
close for no benefit

PR #15161, opened 2026-08-12
chore: migrate rspack_resolver crate into the workspace

stormslowly:
## Why

`rspack_resolver` currently lives in [rstackjs/rspack-resolver](https://github.com/rstackjs/rspack-resolver) and is consumed from crates.io (`=0.10.0`). Moving the crate into this workspace lets resolver changes be developed, tested, and released together with rspack: the crate now follows the unified workspace version and ships with every rspack crates.io release, and resolver changes are tested by CI in the same PR that uses them.

The npm package `@rspack/resolver` stays in the original repository and will switch its napi binding to the crates.io `rspack_resolver` once the first rspack release containing it is published.

## What

- Import the crate into `crates/rspack_resolver`: sources, all tests (181 unit/integration tests + 12 doc tests, counts verified identical to the original repository), fixtures (including the pnpm lockfile so fixture dependency versions asserted by tests stay pinned), benches sources, and examples. `fuzz/`, the napi binding, and the npm packages are not migrated.
- Replace the crates.io dependency with a path dependency; the crate uses `version.workspace = true`, so `cargo-workspaces` picks it up in the regular crates release with no workflow changes.
- Add `ci-resolver.yml`: a paths-gated workflow (3-OS test matrix with fixture installs, lib clippy, wasm32-unknown-unknown check) that runs only when the resolver, workspace manifests, or the toolchain change. The shared rust test/clippy exclude the crate: its imported test sources are kept verbatim and predate the workspace lint set.
- Slim the workspace `tokio` entry from `["rt", "rt-multi-thread"]` to `["rt"]`: tokio hard-errors on `rt-multi-thread` for wasm32-unknown-unknown, and inherited workspace features cannot be disabled per-target. Crates needing the multi-thread runtime in isolated builds (`rspack_resolver` non-wasm target, `xtask/benchmark`) declare it explicitly; other CI builds get it through feature unification, unchanged from before.

Files intentionally differing from the original repository (everything else is byte-identical): workspace rustfmt (`style_edition = "2024"`) reformatted 14 files (formatting only), and 7 call sites were mechanically adjusted for workspace lints injected via rustflags (6× `.unwrap()` → `.expect(...)`, 1× `.map` → `.inspect`). A crate-local `clippy.toml` keeps the imported sources shielded from the workspace `disallowed-methods` list.

## Release note

Before the next crates release, the crates.io trusted-publishing / ownership settings of `rspack_resolver` must be updated to allow publishing from this repository (currently bound to rstackjs/rspack-resolver).


hardfist (approved):

crates/rspack_resolver/pnpm-workspace.yaml
chenjiahan, line 1: Do we need another pnpm workspace in our monorepo?

crates/rspack_resolver/examples/tsconfig_references/app/package.json
chenjiahan, line 1: Why do we need this

crates/rspack_resolver/pnpm-workspace.yaml
stormslowly: This is for Resolver to set up deps for testing.
It's not included in the root monorepo workspace.

PR #15162, opened 2026-08-12
fix(cli): normalize string `context` from loaded config

### LINKED ISSUE #15145: [Bug]: Windows: forward-slash context produces duplicate module records for the same file
### System Info

OS:      Windows 11 Home 10.0.26200
Node:    22.13.0
pnpm:    11.21.0
rspack:  @rspack/core 2.0.4
         @rspack/cli 2.0.4
Other:   @nx/angular-rspack 23.1.1, Angular 22.0.4, @ngrx/signals 22.0.0-rc.0

### Details

On Windows, when `context` is supplied with forward slashes (`D:/proj/app`), rspack can create **two module records for the same physical file** — one identified with a native path, one with a mixed-separator path. The two records differ by exactly one character.

The resulting bundle contains two copies of the module, which breaks `instanceof` and DI-token identity for any library that relies on module-level singletons. In our case an NgRx SignalStore fails to construct:

```
ERROR TypeError: Cannot set properties of undefined (setting 'entityMap')
    at new SignalStore (ngrx-signals.mjs:401)
```

#### The two records

```
D:\Projects\ng-rspack-app\node_modules\.pnpm\@ngrx+signals@22.0.0-rc.0_@_edb829a6efa93a4118da456da13d58e4\node_modules\@ngrx\signals\fesm2022\ngrx-signals.mjs
D:/Projects\ng-rspack-app\node_modules\.pnpm\@ngrx+signals@22.0.0-rc.0_@_edb829a6efa93a4118da456da13d58e4\node_modules\@ngrx\signals\fesm2022\ngrx-signals.mjs
```

They differ **only in the first separator after the drive letter** (`\` vs `/`); everything after is identical. Both are 158 characters. Notably:

- no `\\?\` extended-length prefix is involved (it never appears on any module)
- drive-letter case is identical
- neither path exceeds `MAX_PATH`

#### Which importer gets which record

The two records partition cleanly, with no overlap:

| record | imported by |
|---|---|
| `D:\…` (native) | project source files (`.ts`), via bare `@ngrx/signals` |
| `D:/…` (mixed) | `ngrx-signals-entities.mjs`, via bare `@ngrx/signals` |

The second importer is a **package self-reference**: `@ngrx/signals/entities` imports `@ngrx/signals`, resolved through its own `exports` map. So a file reachable both as a package entry point from outside *and* via a self-reference from a secondary entry point ends up with two records. Only `ngrx-signals.mjs` collides; `ngrx-signals-entities.mjs` itself has a single record.

Across the whole compilation, 123 of 135 module resources carried the mixed `D:/…` form and 12 carried the native form.

#### The resolver is not the source

Driving the resolver standalone via `experiments.resolver.ResolverFactory`, seeded with the exact `compiler.options.resolve` dumped from the live build (`modules`, `conditionNames`, `exportsFields`, `mainFields`, `importsFields`, `symlinks: true`, and Nx's `tsConfig.configFile`), it returned the **native** path in every case — including from inside the package directory, and with the context supplied in native, forward-slash, and mixed form. The mixed form is introduced above the resolver.

#### Cause

The compiler's `context` is forward-slashed:

```
compiler.context : "D:/Projects/ng-rspack-app/demo"
process.cwd()    : "D:\Projects\ng-rspack-app\demo"
```

Our integration sets `context` from `__dirname` in an `rspack.config.ts`. Since **`@rspack/cli` 2.0.0** the TS config is loaded through the prebundled **jiti** (`jiti.import(configPath)`), and jiti derives `__dirname` from a file URL — so on Windows it is forward-slashed. `@rspack/cli` 1.x defaulted to `--config-loader register`, which loads the config as a real CJS module and yields a native `__dirname`.

I checked the published tarballs: jiti is absent from every `1.7.x` and first appears in `@rspack/cli` 2.0.0.

Important: **the core behaviour is not new.** With a forward-slash context, `@rspack/core` 1.3.5, 1.6.8, 1.7.0, 2.0.4 and 2.1.8 all produce the duplicate. The 2.0.0 CLI change did not introduce the divergence — it made configs start passing a forward-slash context, which exposes it.

### Reproduce link

https://github.com/gultyayev/ng-rspack-app-repro

### Reproduce Steps

To reveal the bug - clone the repo. `pnpm install`, `pnpm start`, try opening "Checkout (deep lazy)" route. You'll get an error in the console. Below, are some extra details captured by claude debugging the cause.

Ingredients:

1. Windows, pnpm (any package reached through `.pnpm` symlinks).
2. A `context` containing forward slashes — e.g. `context: __dirname` in an `rspack.config.ts` loaded by `@rspack/cli` >= 2.0.0.
3. A dependency with a secondary entry point that **self-references its own package name** (`@ngrx/signals/entities` → `@ngrx/signals`).
4. Application code that also imports that package root directly.

Then inspect the module list — the package root appears twice.

Minimal diagnostic plugin used to detect this (normalising away separator direction and case before grouping):

```ts
const DupeCheck = {
  apply(compiler) {
    compiler.hooks.done.tap('DupeCheck', (stats) => {
      const byBase = new Map();
      for (const m of stats.compilation.modules) {
        const r = m.resource;
        if (!r) continue;
        const base = r.replace(/^\\\\\?\\/, '').toLowerCase().replace(/\\/g, '/');
        byBase.set(base, [...(byBase.get(base) ?? []), r]);
      }
      for (const [base, paths] of byBase) {
        if (paths.length > 1) console.log('[DupeCheck]', base, paths);
      }
    });
  },
};
```

Toggling only the separator convention of `context`, with nothing else changed:

| `context` | duplicate groups | module path forms | runtime |
|---|---|---|---|
| `D:/Projects/ng-rspack-app/demo` | 3 (incl. the ngrx one) | 123 mixed, 12 native | throws |
| `D:\Projects\ng-rspack-app\demo` | **0** | 132 native, 0 mixed | works |

Consistently, a full `@rspack/cli` 1.6.8 + `@rspack/core` 1.6.8 toolchain yields a native `context` and 0 duplicates, while `@rspack/core` 1.6.8 with `@rspack/cli` 2.0.4 yields a forward-slash context and reproduces the bug — i.e. the CLI, not core, decides whether it triggers.

All measurements were taken after `nx reset` and deleting `dist/` and every build cache, with `--skip-nx-cache`, and with the active `node_modules` symlinks verified to point at the intended versions.

### Expected Behavior

One physical file yields exactly one module record, regardless of the separator convention of `context`. rspack accepts a config-supplied `context` in any form and its resolver already normalises to native paths, so module identifiers should be normalised the same way.

### Actual Behavior

A forward-slash `context` on Windows causes some module identifiers to be emitted as `D:/…` with the remainder in native separators, while others are fully native. When one file is reached through both paths it gets two records and is emitted twice, breaking module-level identity.

### Suggested Fix

Two places, either of which resolves it:

1. **`@rspack/core`** — normalise `context` (or module identifiers derived from it) to native path separators on Windows. This is the more complete fix and protects every integration.
2. **`@rspack/cli`** — have the jiti config loader expose a native `__dirname` / `__filename` on Windows, restoring 1.x behaviour.

As a local workaround, normalising the context at the config level (`context: path.resolve(__dirname)`) removes the duplication entirely.

stormslowly:
## Why

Since `@rspack/cli` 2.0.0 the TS config is loaded through the prebundled jiti, which derives `__dirname` from a file URL. On Windows that yields a POSIX-style path, so a config doing `context: __dirname` hands the compiler `D:/proj/app` instead of `D:\proj\app`.

The compiler compares that `context` against native paths, so the same physical file can end up with two module records differing only by the first separator — the bundle then contains two copies of the module and module-level singletons (DI tokens, `instanceof`) break.

Before / after on Windows, for `context: __dirname`:

```
before: compiler.context === "D:/Projects/app/demo"
after:  compiler.context === "D:\\Projects\\app\\demo"
```

The CLI now runs `path.normalize` over a string `context` right after the config is loaded, for both single and multi-compiler configs, so what reaches the compiler matches `process.cwd()` again. No-op on POSIX. Can be dropped once jiti stops rewriting `__dirname`.

Closes #15145

### DIFF
--- a/packages/rspack-cli/src/utils/loadConfig.ts
+++ b/packages/rspack-cli/src/utils/loadConfig.ts
@@ -39,7 +39,9 @@ const loadConfigByPath = async (
     );
   }
 
-  return normalizeContext(content);
+  return options.configLoader === 'native'
+    ? content
+    : normalizeContext(content);
 };
 
 const isConfigObject = (value: unknown): value is Record<string, unknown> =>
--- a/packages/rspack-cli/src/utils/loadConfig.ts
+++ b/packages/rspack-cli/src/utils/loadConfig.ts
@@ -39,7 +39,7 @@ const loadConfigByPath = async (
     );
   }
 
-  return content;
+  return normalizeContext(content);
 };
 
 const isConfigObject = (value: unknown): value is Record<string, unknown> =>
@@ -53,6 +53,21 @@ const isRspackConfig = (
 const checkIsMultiRspackOptions = (
   config: RspackOptions | MultiRspackOptions,
 ): config is MultiRspackOptions => Array.isArray(config);
+
+/**
+ * jiti exposes a POSIX-style `__dirname` on Windows, so a `context` built from
+ * it keeps forward slashes and no longer matches the native paths the compiler
+ * compares it against. Drop this once jiti stops rewriting `__dirname`.
+ */
+const normalizeContext = (config: LoadedRspackConfig): LoadedRspackConfig => {
+  for (const item of checkIsMultiRspackOptions(config) ? config : [config]) {
+    if (typeof item.context === 'string') {
+      item.context = path.normalize(item.context);
+    }
+  }
+
+  return config;
+};
 
 /**
  * Loads and merges configurations from the 'extends' property


packages/rspack-cli/src/utils/loadConfig.ts
```diff
@@ -39,7 +39,7 @@ const loadConfigByPath = async (
     );
   }
 
-  return content;
+  return normalizeContext(content);
```
chenjiahan: Only replace context when using `jiti` or `auto`

```suggestion
  return options.configLoader === 'native'
    ? content
    : normalizeContext(content);
```

chenjiahan (approved):
👍

PR #15165, opened 2026-08-12
fix(asset): preserve ESM imports in module output for "preserve" importMode

JSerFeng:
## Summary

- record non-concatenated `asset/resource` imports using `generator.importMode: "preserve"` as structured code-generation data when `output.module` is enabled
- register those imports in the modern-module linker's `raw_import_stmts`, allocate their bindings with the chunk-level deconfliction pass, and reference the final binding from the module factory
- avoid asset-specific init fragments in modern-module output; the normal JavaScript renderer uses `ExternalModuleInitFragment` only as a compatibility fallback when no chunk linker is available
- add `esmOutputCases` and watch coverage for non-concatenated assets, application-symbol renaming, multiple similar asset names, runtime modes, and incremental stability
- verify asset import bindings do not collide with module-external raw imports, `ExternalModuleInitFragment` helpers, or webpack/rspack runtime imports

The root cause was that the non-concatenated preserve branch always rendered `require(...)`. That is invalid in module output. Emitting a raw import from an asset-specific init fragment also bypasses the modern-module linker's normal import ownership and symbol allocation.


JSerFeng:
@codex review

elecmonkey (approved):
Verified locally: built this branch against an rslib project with CJS source and ESM output. Before, the ESM chunk emitted `module.exports = require("./static/image/a.png")`, which throws in an ES module. Now it hoists `import a from "./static/image/a.png"` and the body reads `module.exports = a`. No bare `require` remains.

PR #9467, opened 2025-02-26
fix: handle css with BOM

inottn:
## Summary
fix #9461 

## Checklist

- [x] Tests updated (or not required).
- [ ] Documentation updated (or not required).

Issue #9461, opened 2025-02-25
[Bug]: File encoded as UTF-8 with BOM makes output file invalid


jparkrr:
### System Info

  System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M1
    Memory: 334.39 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.13.1 - ~/.asdf/installs/nodejs/22.13.1/bin/node
    npm: 10.9.2 - ~/.asdf/installs/nodejs/22.13.1/bin/npm
    pnpm: 9.15.2 - /opt/homebrew/bin/pnpm
    bun: 1.2.1 - /opt/homebrew/bin/bun
    Watchman: 2024.12.02.00 - /Users/justin/code/custom-watchman/Cellar/watchman/2024.12.02.00_1/bin/watchman
  Browsers:
    Chrome: 133.0.6943.127
    Safari: 18.3
  npmPackages:
    @rspack/cli: ^1.2.6 => 1.2.6 
    @rspack/core: ^1.2.6 => 1.2.6 

### Details

I guess this is/was an issue with webpack as well, but it didn't happen for me when I was using webpack https://github.com/webpack/webpack/issues/13642

One of my dependencies has a css file encoded as UTF-8 with BOM. This causes Chrome to not be able to fully parse the CSS that is output in development mode.

![Image](https://github.com/user-attachments/assets/42a8c37c-3711-4cf7-9c7f-b96d7aaae719)

Note that in my actual project, the file with the incorrect UTF-8 BOM encoding is in one of my dependencies so it's not easy for me to re-encode it.

### Reproduce link

https://github.com/jparkrr/bom-bug-rspack-repo

### Reproduce Steps

1. Clone repo https://github.com/jparkrr/bom-bug-rspack-repo
2. `pnpm run dev:rspack`
3. See that the css rule isn't being applied to make the text blue.

PR #15169, opened 2026-08-12
fix(css-extract): strip BOM from module content before concatenation

doxxx93:
## Summary

`CssExtractRspackPlugin` concatenates CSS module contents verbatim. When a loader prepends a UTF-8 BOM (U+FEFF) to a module's output, that BOM ends up at a **non-zero** offset in the concatenated bundle, where it is no longer a BOM but a stray character in the middle of a stylesheet. `LightningCssMinimizerRspackPlugin` then fails to parse at that point and **silently deletes the rule that follows it**. No error, no warning, exit code 0.

This is not hypothetical: it shipped broken CSS on a production site for two weeks. The upright `@font-face` of a self-hosted code font was deleted, only the italic face survived, and every code block rendered in italic because CSS font matching falls back to an italic face when a family has no normal one.

Minimal repro, with `a.css` (`.a{color:red}`) and `b.css` (an `@font-face` rule followed by `.b{color:blue}`) both going through `[CssExtractRspackPlugin.loader, 'css-loader']`, and `b.css` additionally through a loader that prepends a BOM (which is what dart-sass does for compressed output containing non-ASCII):

- `minimize: false` → 113 bytes, `.a{color:red}\n\n<BOM>@font-face{...}` (bytes `7d 0a 0a ef bb bf 40`), rule intact
- with `LightningCssMinimizerRspackPlugin` → 27 bytes, `.a{color:red}.b{color:#00f}` — the `@font-face` rule is gone, no diagnostic

#9467 already strips a BOM, but only at the head of a source file. It does not reach this path: with `css-loader` the module Rspack parses is JavaScript, and the BOM lives inside a JS string literal in the middle of that module rather than at its head, so `remove_bom` is never called for it. Putting the BOM in the source file instead of injecting it from a loader does not reproduce, which is exactly the gap.

The fix strips a leading BOM per CSS module in `rspack_plugin_extract_css` before the contents are concatenated. It is applied before the `@import url` check so that a BOM-prefixed `@import` module is still routed correctly.

The existing `configCases/css/css-with-bom` case only asserts `css[0] !== "﻿"`, so it passes even with this bug present. The new `configCases/css/css-extract-with-bom` case asserts the BOM is absent **anywhere** in the bundle and that the rule following it survives minification. Both assertions fail before this change and pass after.

## Related links

- Downstream report and full analysis: facebook/docusaurus#12358
- Prior BOM fix that covers source files only: #9467 (issue #9461)

## Checklist

---

Follow-up worth considering, deliberately **not** in scope here: the silent data loss itself. `LightningCssMinimizerRspackPlugin` drops an unparseable rule with no diagnostic at all. The JS `lightningcss` package throws on the same input and cssnano tolerates it; only Rspack's minimizer loses data silently. Even with this fix, any other unparseable construct would still vanish without a warning.


intellild:
@codex review

crates/rspack_plugin_extract_css/src/plugin.rs
```diff
@@ -411,7 +411,9 @@ despite it was not able to fulfill desired ordering with these modules:
     let mut external_source = ConcatSource::default();
 
     for module in used_modules {
-      let content = Cow::Borrowed(module.content.as_str());
+      // Loaders such as dart-sass prepend a BOM, which is only meaningful at the
+      // start of a file. Concatenated as-is it lands mid-chunk and breaks parsers.
+      let content = Cow::Borrowed(remove_bom_str(&module.content));
```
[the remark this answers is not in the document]
doxxx93: Good catch, confirmed and fixed in 1c3252242c.

I measured it with `devtool: 'source-map'` and `css-loader` `sourceMap: true`, comparing identical CSS with and without a loader-injected BOM:

```
without BOM:            AAAA,GAAG,UAAU;AACb,GAAG,WAAW
with BOM (old commit):  AAAA,IAAG,UAAU;AACb,GAAG,WAAW
```

The second segment moved from `G` (+3) to `I` (+4), so generated column 4 pointed at the character that lives at column 3. Exactly the off-by-one you described, and it only shows up when the first line carries more than one mapping segment.

The BOM is now stripped through `remove_bom` where a source map is attached, so `ReplaceSource` moves the mappings with it, and the mappings are byte-identical to the BOM-free build. `remove_bom_str` is kept for the `@import url` branch, which is rendered separately and has no source map, and for the `starts_with_at_import` check so a BOM-prefixed `@import` is still routed correctly.

Added `configCases/css/css-extract-with-bom-source-map` to guard it: it builds the same CSS twice, once through a BOM-injecting loader, and asserts both maps have identical `mappings`. Verified it fails on the previous commit and passes on this one.

doxxx93:
@codex review

doxxx93:
ah not possible

intellild:
@codex review

doxxx93:
@intellild update or rebase branch needed?

intellild (dismissed):

crates/rspack_plugin_extract_css/src/plugin.rs
```diff
@@ -430,8 +434,9 @@ despite it was not able to fulfill desired ordering with these modules:
         if let Some(header) = header {
           external_source.add(header);
         }
+        let content = remove_bom_str(&content);
```
intellild: Is this redundant? BOM is removed in line 491

intellild (changes requested):

crates/rspack_plugin_extract_css/src/plugin.rs
doxxx93: Half redundant, half not — thanks, digging into it found a real bug.

The rendering side was redundant. Both arms now build one `BoxSource` that goes through a single `remove_bom`, so `remove_bom_str` is gone and `rspack_core` is untouched by this PR again.

The `@import` branch is not, and I had it backwards at first. `starts_with_at_import` runs on the raw content, so a leading BOM hides the `@import url` prefix and the module falls into the `else` branch — it gets emitted into `source`, which is appended after `external_source`. The `@import` then sits behind style rules, where browsers ignore it. Reproduced on current `main`, with `css-loader` `import: false` so the `@import` stays in the module content (dart-sass passes plain `@import url(...)` through verbatim, and prepends a BOM to compressed output containing non-ASCII):

```
main:  .first{color:black}\n\n<BOM>@import url("https://example.com/x.css");\n.z{color:green}
fixed: @import url("https://example.com/x.css")\n.z{color:green}\n.first{color:black}
```

So the check is normalised, and once it is, that branch genuinely can carry a BOM. `configCases/css/css-extract-with-bom-at-import` covers it and fails on `main`.

That is a pre-existing bug rather than one this PR introduced, so say the word if you would rather I split it into its own PR and keep this one to the minifier data loss.

Also switched the fixtures from a literal U+FEFF to the `\uFEFF` escape, matching `configCases/css/css-with-bom`, so an editor cannot silently eat it.

intellild (approved):

PR #15193, opened 2026-08-14
feat: reuse unchanged NormalModule builds in new cache

### LINKED ISSUE #15190: Reuse unchanged NormalModule builds in memory-backed new_cache
## What to build

Add the first end-to-end NormalModule build-cache path for the experimental new cache when using the memory backend.

An unchanged NormalModule should be built once, restored into a freshly factorized module on the next compilation, and integrated into the rebuilt module graph without running the build again. The cache entry must preserve all build-owned state needed for behavior equivalent to a real build while retaining the fresh module's factorization-time configuration.

Cache lookup, snapshot validation, restore, and store should follow the existing new-cache facade style. This ticket covers ordinary build-module-graph work only; persistent restore, explicit force rebuild propagation, and ModuleExecutor caching are separate follow-up work.

## Acceptance criteria

- [ ] Only NormalModule participates; unsupported module types bypass the cache.
- [ ] The cache uses a dedicated modules namespace, the module identifier as its item key, and no etag.
- [ ] A first compilation builds the module and stores a complete, cloneable build-state entry.
- [ ] A second compilation on the same Compiler restores an unchanged module without executing its build.
- [ ] A changed file dependency invalidates the entry and executes a real build.
- [ ] Memory-backed cache entries use the same snapshot validation abstraction intended for persistent entries, with default snapshot options.
- [ ] Cache hits skip buildModule and succeedModule and invoke stillValidModule after restoring the module.
- [ ] Cache misses invoke buildModule, build the module, store before succeedModule, and then invoke succeedModule.
- [ ] Cache read, validation, snapshot, or write failures fall back to a real build without failing an otherwise successful compilation.
- [ ] newCache disabled and cache disabled behavior remains unchanged.
- [ ] Behavior is covered through an existing JavaScript test harness; no Rust unit test is added.
- [ ] No public or internal module-build hit/miss metric is introduced.

## Blocked by

None — can start immediately.

hardfist:
## Summary

- cache completed `NormalModule` builds in the experimental new cache
- restore unchanged builds from memory or persistent storage after snapshot validation
- rebuild when source or value dependencies change

Closes #15190

### DIFF
--- a/crates/rspack_core/src/compilation/build_module_graph/graph_updater/repair/add.rs
+++ b/crates/rspack_core/src/compilation/build_module_graph/graph_updater/repair/add.rs
@@ -123,6 +123,8 @@ impl Task<TaskContext> for AddTask {
       plugin_driver: context.plugin_driver.clone(),
       runtime_template: context.runtime_template.create_module_code_template(),
       fs: context.fs.clone(),
+      cache: context.cache.clone(),
+      value_cache_versions: context.value_cache_versions.clone(),
       forwarded_ids,
     })])
   }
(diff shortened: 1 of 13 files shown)


hardfist:
@codex review

PR #15195, opened 2026-08-14
perf(core): precompile filename templates cache

intellild:
## Summary

- store filename string templates as `Ustr` and reuse compiled metadata through a global identity-hashed `DashMap`
- compile templates into plain byte ranges and placeholder IDs, with hash length/encoding parameters kept in a side table
- render compiled segments into one output buffer while preserving the existing staged replacement behavior and `AssetInfo` updates
- reject static templates and filename-function results longer than 65,535 UTF-8 bytes
- extend the existing asset filename case with Unicode literals, repeated and differently parameterized hashes, base64 parameters, and invalid-placeholder preservation
- render async WebAssembly filenames during module code generation so emitted assets and runtime requests use the same encoded hash

## Why

Filename templates are cloned broadly across output and chunk options, then rendered repeatedly. The previous renderer scanned and rebuilt the whole string for each placeholder phase. Keeping only the interned raw template in `FilenameKind` makes copies cheap, while the global compiled cache lets identical templates share their parsed representation without changing persistent-cache semantics.

## Example

For `assets/[name].[hash:1234].js`, compilation now records:

- plain segments as byte ranges into the original `Ustr`
- `[name]` and `[hash:1234]` as placeholder segment IDs
- `1234` in the placeholder side table, so rendering does not parse the token again

If a replacement introduces a later-stage placeholder, such as `[file]` producing `foo-[hash].js`, the later hash replacement still runs as before.

## WASM runtime filename fix

Before this fix, the emitted WebAssembly asset was named through the general filename renderer, but the WASM loading runtime rebuilt the path from `module.id` and a rendered module hash. That runtime path only truncated `wasmModuleHash` and did not preserve the placeholder's requested encoding.

For example:

```js
output: {
  webassemblyModuleFilename: "[id].[hash:base64:8].wasm"
}
```

The manifest emitted a base64-encoded filename, while the runtime requested the first eight characters of the ordinary hash digest, causing the asynchronous WASM request to return 404.

The fix now:

- resolves the complete WASM filename once during module code generation
- stores the filename and `AssetInfo` in the module's code-generation data
- embeds the real filename in the generated WASM JavaScript module factory
- reuses the same filename and asset metadata from `renderManifest`
- reduces the WASM loading and compile runtime helpers to accepting a filename instead of separately accepting `module.id` and a hash

This follows the asset-module data flow and guarantees that runtime loading uses the exact filename that was emitted.

## Validation

- `pnpm run build:js`
- `pnpm run test -t "configCases/asset/filename"` from `tests/rspack-test`
- `pnpm run test -t "configCases/wasm/runtime-hash-encoding"` from `tests/rspack-test`
- `pnpm run lint:rs`
- `cargo lint`
- workspace Rust tests, excluding `rspack_storage` and `rspack_watcher` as required for sandbox execution
- verified that 65,535-byte templates succeed and longer templates panic before interning

---
_by OpenAI Codex_


intellild:
@codex review

intellild:

intellild:
@codex review

intellild:
@codex review

intellild:
## Background: why webpack passes `module.id` and `wasmModuleHash`

I traced this design through webpack's history. The runtime reconstruction is primarily a historical consequence of webpack 4's WASM chunk-loading architecture, rather than a requirement of hash encoding.

### 2017: WASM loaded before its JavaScript module factory

The original WebAssembly proof of concept loaded WASM as part of `ensureChunk(chunkId)`. The chunk runtime:

1. looked up the WASM module IDs belonging to the chunk
2. iterated those IDs
3. rebuilt each binary filename from `wasmModuleId` plus a runtime `id -> hash` map
4. fetched and compiled the binary
5. stored the compiled result in `__webpack_require__.w[wasmModuleId]`
6. executed the JavaScript module factory afterwards

At that point the loader genuinely only had the module ID available, so resolving the filename in the chunk runtime was a natural design.

Source: [webpack's 2017 WebAssembly proof of concept](https://github.com/webpack/webpack/commit/41a1d602e1e406a3132bd176d5fb3876ea0e6ec3).

### 2019: async modules inherited the runtime filename renderer

When webpack introduced side-by-side async WASM modules, the JavaScript module factory began calling a shared instantiate helper. The implementation moved loading closer to the module factory, but retained the existing `module.id -> runtime filename` design.

Source: [side-by-side WASM support based on async modules](https://github.com/webpack/webpack/commit/de610236161d8ea8d8c0bc9823a85a58b7cd4cb3).

### 2020: `wasmModuleHash` removed an incorrect runtime hash map

The explicit `wasmModuleHash` argument was added later. The commit motivation was:

> avoid using getChunkModuleMaps as it doesn't support different modules per runtime

The change moved `getRenderedModuleHash(module, runtime)` into the generated module factory and passed that rendered string to the shared runtime helper. It did not introduce the parameter to support digest conversion; it was a way to remove the runtime `module id -> hash` map while preserving the existing filename renderer.

Source: [webpack commit introducing the explicit hash argument](https://github.com/webpack/webpack/commit/8948263db866655f85b39235695e8937411c3fd2).

### Asset modules were designed around a different data flow

Asset modules compute their complete filename during code generation, write the real filename/URL into their generated JavaScript, and store the filename plus asset metadata in code-generation data for `renderManifest` to consume.

Source: [initial asset-module improvements](https://github.com/webpack/webpack/commit/491cb57cb48e2552d58f575b9cd5b0aac4cd7612).

That model is simpler because the runtime and asset emission consume the same resolved filename.

### Inline digest support exposed the legacy boundary

Per-placeholder digest syntax such as `[hash:base64:8]` was added much later than the WASM runtime design. A rendered hash string is enough for truncation, but it is not a good representation for requesting a different encoding later in runtime code.

Source: [inline hash digest support](https://github.com/webpack/webpack/commit/ecc00c0959cdae56e8ab77d80ef48082a17168f3).

Webpack's recent hashed-WASM work reaches the same conclusion for re-encoded compilation hashes: a runtime hash expression cannot safely be re-encoded as though it were the original digest, so the settled value is inlined instead.

Sources: [analyzable async WASM URLs](https://github.com/webpack/webpack/commit/7a19dd8014b738a5c83158545f2819fcdde2e3fe), [hashed WASM filename handling](https://github.com/webpack/webpack/commit/bc65f6b5114cca5a028bc1499d2d1c6bfac08624).

### Direction used in this PR

For module IDs and module hashes, both inputs are already available during WASM code generation. This PR therefore resolves the complete filename once and shares that result between:

- the generated WASM JavaScript module factory
- the emitted WASM manifest entry
- the loading/compile runtime helper

The runtime no longer needs to know how filename placeholders or hash encodings work, and cannot request a path different from the emitted asset.

---
_by OpenAI Codex_

LingyuCoder:

### PR Overview

This is primarily a performance optimization of filename-template rendering. It also fixes two correctness gaps exposed by parameterized hashes: runtime chunk filenames now retain per-placeholder length/encoding metadata, and async WASM code generation resolves one concrete filename that is shared by the generated JavaScript and `renderManifest`.

### Optimized Phase

The optimized path is repeated interpolation of output, asset, runtime-chunk, source-map, banner, and CSS local-ident templates. The CSS path is especially hot because `localIdentName` may be rendered once per class.

### Previous Design and Bottleneck

The previous renderer scanned and potentially rebuilt the whole string once for each placeholder phase. Hash length extraction for runtime filenames also reparsed templates and collapsed placeholder-specific metadata. This made rendering approximately proportional to template length multiplied by the number of replacement phases, with multiple intermediate `Cow`/`String` values.

### New Design and Why It Is Faster

Static templates are interned as `Ustr` and compiled once into plain byte ranges plus placeholder IDs in an identity-hashed global `DashMap`; rendering then walks those segments once into a pre-sized output buffer. Filename-function results are compiled transiently, so dynamic results do not accumulate in the global compiled cache. Runtime chunk rendering consumes the compiled placeholder metadata directly, and WASM loading receives the already-resolved filename instead of reconstructing it from a module ID and hash.

### Benchmark Evidence

- **Ecosystem benchmark:** the [posted benchmark detail](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5323436568) contains current measurements but `-` for every base value, so it provides no comparative evidence. The current persistent-cache hit-rate check is positive.
- **CodSpeed:** the [final report](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5293017777) reports 3 improvements, 47 unchanged benchmarks, and 47 skipped benchmarks: `rust@create_chunk_assets` +32.51%, `bundle@css-modules-development` +29.56%, and `rust@create_full_hash` +22.21%. These are measured simulation results, but CodSpeed flags different runtime environments, so the exact percentages are not conclusive. Binary size increased by 8.13 KB (0.01%), while the Rsdoctor bundle diff found no output-size changes in its five projects.

The current PR checks pass Linux/Windows/macOS/WASM builds, E2E tests, Rust tests/checks, lint/format, Dylint, and CodSpeed. The latest [ecosystem run](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5392140618) passed 10 suites and failed `modernjs`; its SSR-fallback assertion output does not establish whether the failure is related to this change.

### Correctness Risks

1. **Per-chunk encoded full hashes can leave the loader pointing at the wrong file.** [`with_full_hash`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_runtime/src/runtime_plugin.rs#L252-L325) inspects only the global `output.filename`/`output.chunkFilename` (and CSS equivalents), while [`get_js_chunk_filename_template`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/output.rs#L538-L552) can select a per-chunk or entry override. For a split chunk using `[fullhash:base64:4]`, [`render_full_hash`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs#L27-L42) may embed the encoding of `XXXX` during the initial runtime-module hash, but the module is not marked full-hash-dependent and is not regenerated after the compilation hash exists. The emitted asset uses the final hash, so the runtime request can 404. The added tests cover global and function-valued `chunkFilename`, not per-chunk overrides.

2. **The single-pass renderer changes staged replacement behavior.** [`render_with_writer`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L230-L309) parses only the original template, and [`try_render_value`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L775-L780) appends replacement text without processing it. With template `[file]`, filename `foo-[hash].png`, and hash `abcd`, the previous staged renderer produced `foo-abcd.png`; this code leaves `foo-[hash].png`. No added case covers a later-stage placeholder introduced by an earlier replacement.

3. **The `u16` segment representation creates a broader breaking limit than filename paths.** [`Filename::from` now asserts at 65,535 UTF-8 bytes](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L373-L383), but `Filename` is also used as a generic placeholder renderer: for example, [`BannerPlugin` passes arbitrary banner content through it](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_banner/src/lib.rs#L168-L184). A valid large license banner or a function result above the limit now panics before emission. This compatibility boundary is neither documented nor covered by a checked-in consumer test.

<details>

### PR 概述

该 PR 的主要目标是优化文件名模板渲染性能，同时修复参数化哈希暴露出的两个正确性问题：运行时 chunk 文件名现在会保留每个占位符各自的长度和编码元数据；异步 WASM 在代码生成阶段解析出唯一的实际文件名，并由生成的 JavaScript 与 `renderManifest` 共同使用。

### 优化阶段

本次优化覆盖输出文件、资源、运行时 chunk、source map、banner 和 CSS local-ident 模板的重复插值。其中 CSS 路径尤其关键，因为每个 class 都可能渲染一次 `localIdentName`。

### 旧设计与瓶颈

旧渲染器会针对每个占位符阶段扫描并可能重建整段字符串。运行时文件名的哈希长度提取还会再次解析模板，并丢失各占位符独立的参数元数据。因此，渲染成本近似为模板长度乘以替换阶段数量，同时会产生多个中间 `Cow`/`String`。

### 新设计及其加速原因

静态模板会以 `Ustr` 驻留，并只编译一次，生成普通字节区间和占位符 ID，结果存入使用 identity hash 的全局 `DashMap`；渲染时只需遍历一次这些片段并写入预分配的输出缓冲区。文件名函数的结果采用临时编译，因此动态结果不会在全局编译缓存中持续累积。运行时 chunk 渲染直接消费已编译的占位符元数据；WASM 加载则接收已经解析完成的文件名，不再根据 module ID 和 hash 重新构造。

### 基准证据

- **Ecosystem benchmark：**[已发布的 benchmark 详情](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5323436568)包含当前测量值，但所有基线值都是 `-`，因此不能提供对比证据。当前 persistent-cache hit-rate 检查结果为正面。
- **CodSpeed：**[最终报告](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5293017777)显示 3 个基准提升、47 个无变化、47 个跳过：`rust@create_chunk_assets` 提升 32.51%，`bundle@css-modules-development` 提升 29.56%，`rust@create_full_hash` 提升 22.21%。这些是实际测得的 simulation 结果，但 CodSpeed 标记了运行环境不同，因此具体百分比不能视为定论。二进制体积增加 8.13 KB（0.01%），而 Rsdoctor bundle diff 在 5 个项目中未发现输出体积变化。

当前 PR 检查中的 Linux/Windows/macOS/WASM 构建、E2E 测试、Rust 测试与检查、lint/format、Dylint 和 CodSpeed 均通过。最新的 [ecosystem 运行](https://github.com/web-infra-dev/rspack/pull/15195#issuecomment-5392140618)有 10 个 suite 通过、`modernjs` 失败；其 SSR fallback 断言输出不足以确定该失败是否与本次改动相关。

### 正确性风险

1. **按 chunk 覆盖的编码 full hash 可能使加载器请求错误文件。** [`with_full_hash`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_runtime/src/runtime_plugin.rs#L252-L325)只检查全局 `output.filename`/`output.chunkFilename`（以及对应的 CSS 选项），但 [`get_js_chunk_filename_template`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/output.rs#L538-L552)可能选择按 chunk 或 entry 覆盖的模板。对于使用 `[fullhash:base64:4]` 的 split chunk，[`render_full_hash`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_runtime/src/runtime_module/get_chunk_filename.rs#L27-L42)可能在首次计算 runtime module hash 时嵌入 `XXXX` 的编码值，但该模块没有被标记为依赖 full hash，因此 compilation hash 生成后不会重新生成。发出的资源使用最终 hash，运行时请求可能因此 404。新增测试覆盖了全局和函数形式的 `chunkFilename`，但没有覆盖按 chunk 覆盖的情况。

2. **单遍渲染改变了分阶段替换行为。** [`render_with_writer`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L230-L309)只解析原始模板，[`try_render_value`](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L775-L780)会直接追加替换文本而不再处理其中的占位符。当模板为 `[file]`、filename 为 `foo-[hash].png`、hash 为 `abcd` 时，旧的分阶段渲染器会生成 `foo-abcd.png`，当前代码则保留为 `foo-[hash].png`。新增用例没有覆盖由早期替换值引入后续阶段占位符的情况。

3. **`u16` 片段表示带来的破坏性限制并不只影响文件路径。** [`Filename::from` 现在会在 UTF-8 长度超过 65,535 字节时触发断言](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_core/src/options/filename.rs#L373-L383)，但 `Filename` 同时也是通用的占位符渲染器，例如 [`BannerPlugin` 会将任意 banner 内容传入它](https://github.com/web-infra-dev/rspack/blob/df0c6350ee79d4a987bf290e910a9325dec73257/crates/rspack_plugin_banner/src/lib.rs#L168-L184)。有效的大型 license banner 或超过限制的函数结果现在会在发出资源前 panic。该兼容性边界既未记录在文档中，也没有由仓库内的消费方测试覆盖。

</details>

crates/rspack_macros_test/tests/string_enum.rs
```diff
@@ -0,0 +1,22 @@
...
+fn converts_between_enum_and_str() {
+  assert_eq!(TestStringEnum::from("first"), TestStringEnum::First);
+  assert_eq!(
+    TestStringEnum::from("multiple-words"),
+    TestStringEnum::MultipleWords
+  );
+  assert_eq!(TestStringEnum::from("unknown"), TestStringEnum::Unknown);
+  assert_eq!(TestStringEnum::First.as_str(), "first");
+  assert_eq!(TestStringEnum::MultipleWords.as_str(), "multiple-words");
+}
```
LingyuCoder, line 22:
```suggestion
```

crates/rspack_plugin_extract_css/src/plugin.rs
```diff
@@ -557,6 +557,8 @@ async fn runtime_requirement_in_tree(
   let has_hot_update = runtime_requirements.contains(RuntimeGlobals::HMR_DOWNLOAD_UPDATE_HANDLERS);
 
   if has_hot_update || runtime_requirements.contains(RuntimeGlobals::ENSURE_CHUNK_HANDLERS) {
+    let full_hash = self.options.filename.has_full_hash_digest()
+      || self.options.chunk_filename.has_full_hash_digest();
```
LingyuCoder: webpack seems just uses the `output.chunkFilename` to check full hash digest [here](https://github.com/webpack/webpack/blob/main/lib/RuntimePlugin.js#L448). Why we need to use `output.filename` to check it?

crates/rspack_macros_test/tests/string_enum.rs
intellild: This is macro test crate to test macros. I think we should keep it.

crates/rspack_plugin_extract_css/src/plugin.rs
intellild: added test in [f6c8806](https://github.com/web-infra-dev/rspack/commit/f6c88068962521192b46887a141f39aef293d938)

webpack has bugs when hmr enabled

LingyuCoder (approved):

PR #15205, opened 2026-08-15
perf(paths): intern ArcPath as a single-allocation prehashed path

stormslowly:
## Why

`ArcPath` is 24 bytes — a fat `Arc<Path>` pointer plus a cached hash — and every construction allocates, even when the same path already exists elsewhere. A build holds a lot of paths (module identifiers, file and context dependencies, watcher sets), so the same absolute path gets allocated many times over, and `==` between two allocations walks path components.

| per unique path | before | after |
| --- | --- | --- |
| handle | 24 B | 8 B |
| allocations | one per construction | one per unique path |
| `==` | component walk unless same allocation | pointer comparison |

## What

`crates/rspack_intern` is a trimmed port of rust-analyzer's [`intern`](https://github.com/rust-lang/rust-analyzer/tree/baabc5825f3f6640e99fe32887bbeced640f825e/crates/intern) crate, non-GC mode. `InternedSlice<T>` stores a header plus a slice in **one** allocation behind a thin pointer (`triomphe::ThinArc`) and takes the items by reference, so interning a known path allocates nothing. A sized `Interned<T>` would not do: `T: Sized` forces the variable-length path into a second allocation.

`ArcPath` becomes `InternedSlice<PreHashedPath>` — precomputed `FxHash` in the header, `OsStr` bytes in the tail. `Deref` yields `Path` instead of `Arc<Path>`; no call site needed the `Arc`.

Two properties are preserved deliberately:

- **`Hash` hashes the content hash, not the interned pointer.** `ArcPathSet` feeds it straight into `IdentityHasher`; pointer addresses are allocation-aligned and vary per run, which would cluster buckets and make hash-ordered output non-deterministic.
- **Equality mirrors `hash_path`'s per-platform scheme**, so a pair that compares equal can never hash differently and escape deduplication.

`ResolveContext`'s dependency sets become `ArcPathSet`. They were `FxHashSet<ResolverPath>`, which cost `Arc::from(path)` on the resolver side and another copy into `ArcPath` on the rspack side — two copies and a wasted allocation per dependency.

> [!WARNING]
> **Breaking change to `rspack_resolver`'s public API.** `ResolverPath` is removed and `ResolveContext::{file_dependencies, missing_dependencies}` change to `ArcPathSet`.

That direction (`rspack_resolver → rspack_paths`) collides with the existing `rspack_paths → rspack_cacheable → rspack_resolver` edge, and Cargo cycles are crate-level, so feature gates cannot break it. Resolved by making `rspack_paths`' `cacheable` support an optional feature that `rspack_resolver` declines — which is also what keeps the resolver building for `wasm32-unknown-unknown`, since `rspack_cacheable` pulls in `swc_core → hstr → ahash → getrandom`. The one preset `rspack_cacheable` needed the resolver for moves to `rspack_core` behind a local `AsAliasValue` adapter, and `rspack_cacheable` now re-exports `rkyv` publicly (a direct dependency is not an option — the workspace carries both 0.7 and 0.8).

Two fallouts: `rspack_paths`' `default = ["cacheable"]` does nothing because every workspace dependency declaration sets `default-features = false`, so the feature is enabled explicitly on the 13 crates that need it; and `rspack_watcher`'s tests had been getting tokio's `rt-multi-thread` transitively through the removed edge, so the watcher now declares it.

### Internal Large-Scale App verification

mem-curve-internal-large-app-2x2

> This PR improved the peak Memory about 7G. 
> But there is still 1G extra memory cost when tsconfig reference is enabled. I will investigate this later.

### Ecosystem benchmark

Release bindings, all 28 jobs, base and head interleaved over two rounds. "Beyond noise" means the delta exceeds the wider 95% confidence interval plus half the round-to-round spread.

**Peak RSS -3.14% overall — 18 jobs down beyond noise, none up.** The projects with the most modules gain the most, which is what deduplication should look like:

| job | RSS |
| --- | --- |
| `10000_production-mode_persistent-hot` | -10.76% (-74.7 MB) |
| `10000_production-mode` | -8.60% (-41.0 MB) |
| `10000_development-mode` | -7.87% (-43.3 MB) |
| `large-dyn-imports_production-mode` | -6.60% (-28.5 MB) |
| `10000_development-mode_hmr` | -6.35% (-50.6 MB) |

**Wall time -0.45% overall**, matching the callgrind cycle estimate and reproduced in both rounds. Only `10000_production-mode_persistent-hot` separates from the noise at -6.93%. 17 of 28 jobs are nominally faster, but a single job's noise band here is 1.5-9%, so wall time alone cannot resolve a half-percent effect.


stormslowly:
@codex code review

crates/rspack_intern/src/slice.rs
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+impl<T: SliceInternable> Drop for InternedSlice<T> {
+  #[inline]
+  fn drop(&mut self) {
+    // When the last handle is dropped, remove the value from the global map.
+    if ThinArc::strong_count(&self.arc) == 2 {
+      // Only `self` and the global map point to the value.
+      self.drop_slow();
```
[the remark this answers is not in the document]
stormslowly: Yes, it has a race condition here, but I like to keep it. 

Because making it atomic means taking the shard's write lock on every drop and deciding under it. ArcPath destructors are hot — every module identifier, every dependency set teardown — so that trades the measured win (-3.14% peak RSS, -0.45% cycles) against a bounded retention issue.

crates/rspack_watcher/Cargo.toml
```diff
@@ -23,6 +23,9 @@ tokio = { workspace = true, features = ["rt", "macros", "sync", "fs"] }
 
 [dev-dependencies]
 tempfile = { workspace = true }
+# The tests build a multi-threaded runtime; declare it rather than relying on some other crate
+# in the graph happening to turn the feature on.
+tokio = { workspace = true, features = ["rt-multi-thread"] }
```
hardfist, line 28: why change this?

crates/rspack_core/Cargo.toml
```diff
@@ -49,7 +49,7 @@ rspack_location            = { workspace = true }
 rspack_macros              = { workspace = true }
 rspack_napi                = { workspace = true, optional = true }
 rspack_parallel            = { workspace = true }
-rspack_paths               = { workspace = true }
+rspack_paths               = { workspace = true, features = ["cacheable"] }
```
hardfist, line 52: not necessary

hardfist (approved):

crates/rspack_core/Cargo.toml
stormslowly: It is necessary — rspack_core is the one crate that actually serializes ArcPath. Dropping it gives 141 errors from cargo check -p rspack_core and 183 from cargo check --workspace, all the trait bound ArcPath: Archive is not satisfied.

crates/rspack_watcher/Cargo.toml
stormslowly: rspack_watcher's tests build a multi-threaded runtime — tests/helpers/mod.rs:128 calls tokio::runtime::Builder::new_multi_thread() — but the crate never declared tokio's rt-multi-thread.

Before, it got this by feature merging. In this PR, I cut the merging route

hardfist:

PR #15223, opened 2026-08-17
fix(css): CSS module dashed identifier and composes edge cases

intellild:
## Summary

Fix the CSS Modules lexer/plugin boundary issues identified in the review of https://github.com/web-infra-dev/rspack/pull/14809.

### Distinguish `global` from `"global"`

The lexer now records whether a `from` source is the bare `global` keyword based on its token kind. This keeps these two forms semantically distinct:

```css
/* CSS Modules global composition */
.local {
  composes: external from global;
}

/* A quoted module request whose request string is "global" */
.local {
  composes: external from "global";
}
```

Previously, `rspack_plugin_css` stripped quotes and compared the remaining text with `global`, so the quoted request could be treated as the keyword. The plugin now consumes the semantic `from_is_global` value produced by `css-module-lexer`.

### Normalize dashed identifiers in the lexer

Dashed identifiers are now normalized by `css-module-lexer`, which removes exactly one logical `--` prefix and provides the corresponding name range to consumers.

For example:

```css
.exported {
  ----a: red;
  color: var(----a);
}
```

The logical exported name is `--a`. Previously, additional `trim_start_matches("--")` handling in `rspack_plugin_css` could reduce it to `a`. That plugin-side trimming has been removed.

The resulting flow is:

```text
CSS token
  ↓
css-module-lexer
  classify bare global / normalize dashed identifier
  ↓
Dependency semantic fields
  ↓
rspack_plugin_css
  consume the lexer result without reparsing
```

### Test coverage

This reuses the existing `configCases/css/css-modules` case ported from webpack. Its inputs already contain the upstream regression coverage:

- `----a: red` in `var-function-export.modules.css`
- `var(----a from "./var-function-export.modules.css")` in `var-function.module.css`

The existing Rspack snapshots are updated to match webpack's `--a` behavior in development, production, and disabled-option configurations. No synthetic test case or new test directory is added.

Validation:

- `cargo test -p css-module-lexer --quiet`
- `cargo clippy -p css-module-lexer -p rspack_plugin_css --all-targets -- -D warnings`
- `cd tests/rspack-test && pnpm run test -t "configCases/css/css-modules"`
- `cargo fmt --all --check`
- `git diff --check`

## Related links

- Review follow-up: https://github.com/web-infra-dev/rspack/pull/14809

## Checklist

---
_by OpenAI Codex_


intellild:
@codex review

ahabhgk (approved):

PR #15219, opened 2026-08-17
refactor(loader): schedule loader chains as execution units

intellild:
## Summary

- make `Loaders<Context>` the single owner of raw loaders, loader items, and execution chains
- expose loader execution as `Loaders::run_loaders`
- use `ArcComputed` to project lazy items/chains directly from `Arc<Loaders>`
- separate shared immutable loader metadata from per-run execution state
- schedule consecutive JavaScript loaders as one execution span while keeping native loaders as individual spans
- serialize worker options only for loaders inside the active JavaScript chain

This reduces repeated Rust ↔ JavaScript scheduling boundaries and avoids rebuilding loader items for every module build.

For example, a chain like `js-a -> js-b -> builtin:swc-loader -> js-c` is planned as `[js-a, js-b]`, `[builtin:swc-loader]`, and `[js-c]`. The first two JavaScript loaders execute in one JavaScript scheduling unit, while the native loader stays on the Rust side. When a parallel worker runs that first span, options outside `[js-a, js-b]` are excluded from structured cloning.

## Implementation flow

```text
NormalModule build
  Arc<Loaders>::run_loaders
       ↓
lazy loader_items + loader_chains
  computed once in Loaders
       ↓
ArcComputed projections
  owner: Arc<Loaders>
       ↓
LoaderContext
  per-run loader state
       ↓
chain-based pitch/normal execution
```

`Loaders` and the other `ArcComputed` owner, `ResolvedModuleOptions`, carry `PhantomPinned`. Loader-chain ranges remain compact `u8` values, with a debug assertion protecting the internal maximum-loader-count invariant.

## Validation

- `cargo fmt --all --check`
- `cargo check -p rspack_loader_runner -p rspack_core -p rspack_binding_api --all-targets --locked`
- `cargo test -p rspack_loader_runner --locked` (10 passed)
- `./node_modules/.bin/tsc --noEmit --pretty false -p packages/rspack/tsconfig.json`
- `pnpm run build:js`
- staged Rust/TypeScript format and lint hooks

## Related links

N/A

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

---
_by OpenAI Codex_

PR #15225, opened 2026-08-17
feat(loader): add in-memory loader chain cache

intellild:
## Summary

- add `Rule.use.cache` and group adjacent cached loaders into a `CacheChain` while retaining the nested JavaScript/native execution chains from #15219
- build the cache identity from the configured loader name, stable loader options, and either the loader package version or the resolved local loader file hash
- introduce a process-wide `LoaderCache` registry backed by a global `DashMap`, with one shared `LoaderCache` instance per cache directory
- encapsulate the private `new_cache::MemoryCache` storage behind `LoaderCache::get/store`; loader execution no longer sees cache facades or item storage
- use `<context>/node_modules/.cache/lodaer-cache` as the hardcoded cache-directory identity; this version remains memory-only and does not write that directory

## Example

```js
module.exports = {
  cache: false,
  module: {
    rules: [
      {
        test: /\.js$/,
        use: [
          {
            loader: require.resolve("example-loader"),
            options: { target: "modern" },
            cache: true
          }
        ]
      }
    ]
  }
};
```

For example, when an unrelated watched dependency triggers a rebuild but the input to this loader chain is unchanged, Rspack restores the cached chain result instead of running the loader again. Changing the resource content changes the input etag and executes the chain again. The loader cache is independent from the general compiler cache in this version.

For package loaders, the package name and version participate in the key. For non-package loaders, the resolved loader entry file is content-hashed. This first version intentionally does not include files imported or required by the loader in that file hash, and does not persist loader-chain entries.

## Related links

- stacked on #15219

## Validation

- `cargo check -p rspack_loader_runner -p rspack_core -p rspack_binding_api -p rspack --all-targets --locked`
- `cargo check -p rspack_core --lib --locked`
- `cargo test -p rspack_loader_runner --locked` (10 passed)
- `cargo check -p rspack_benchmark --bench benches --locked`
- targeted `loaders/loader-chain-cache` watch and incremental-watch cases (8 passed) with the general compiler cache disabled, including `parallel` cached loaders, dependency-triggered hits, source maps, additional data, and resource invalidation
- binding development build and `packages/rspack` library build

## Checklist

- [x] Tests updated.
- [x] Public option documentation updated.

---
_by OpenAI Codex_


intellild:
close in favor of #15401

PR #15235, opened 2026-08-18
ci: make sccache actually cache cargo check and clippy

stormslowly:
## Why

sccache is already enabled on the self-hosted runners, but it caches almost nothing for the workspace's own crates in the `Rust check` job. The stats that job prints said it plainly ([release run job](https://github.com/web-infra-dev/rspack/actions/runs/32117074380/job/95649052165), `cargo check` 16m49s + clippy 7m12s):

```
Compile requests        4094      Cache hits 2831 (98.03%)
Non-cacheable calls     1188
  multiple input files   470   <- every clippy compile of a workspace crate
  incremental            434   <- every cargo check compile of a workspace crate
  crate-type             238   <- proc-macro / bin & test targets (unsupported by design)
```

Dependencies hit ~98%; the workspace crates — the ones that actually change — were recompiled from scratch every run, for two independent reasons:

1. **`incremental`** — `[profile.dev] incremental = true` (Cargo.toml). Cargo passes `-C incremental` to local packages only (registry deps never get it, which is why they cache fine), and sccache refuses to cache anything built with it. `reusable-build-build.yml` and `bench-rust.yml` already set `CARGO_INCREMENTAL: 0`; this workflow was missed.
2. **`multiple input files`** — `cargo clippy` runs as `sccache clippy-driver <RUSTC> <args>` and sccache mistakes the injected rustc path for a second input file. Upstream fixed this in mozilla/sccache#2438, but the fix only skips the argument when it is the bare string `rustc`, not an absolute path. `RUSTC=rustc` makes cargo pass the bare name, which the released 0.17.0 then handles.

Both are required together: with only one set, the clippy compile is still rejected.

## What

- `CARGO_INCREMENTAL: 0` and `RUSTC: rustc` on the `Rust check` job.
- **Behaviour change:** the per-crate `cargo check --lib` now runs on every PR instead of only during a crate release. It used to be gated behind a `cargo-check-strategy` input because it was too slow to run per PR; with the compilations cached it costs **+177s in steady state**, which buys the guarantee that every published crate still builds standalone — a broken feature gate is caught on the PR instead of at release time. The input and its only caller are removed with it. See the cost caveat below.

## Results

`Run Clippy`, before and after:

| | before | after |
|---|---|---|
| `Run Clippy` | 350s | **136s** |
| non-cacheable | 610 | **276** |

Full `Rust check` job with the per-crate check included (measured on this branch; "warm" is a re-run of the same commit):

| | baseline | with env (cold) | with env (warm) |
|---|---|---|---|
| `Run Cargo Check (per crate)` | 785s | 843s | **180s** |
| `Run Clippy` | 321s | 276s | **118s** |
| compile requests executed | 2888 | 3656 | 3656 |
| cache hits | 2831 | 83 | **3595 (98.3%)** |
| non-cacheable | **1188** | **420** | **420** |
| ‑ `multiple input files` | 470 | 0 | 0 |
| ‑ `incremental` | 434 | 0 | 0 |
| ‑ `crate-type` | 238 | 370 | 370 |

768 previously rejected compilations became cacheable. The remaining 132 of the 904 are clippy's `--all-targets` test/bin targets, which fall back to `crate-type: No crate-type passed` — uncacheable by sccache's design.

The cold run is a one-off: `CARGO_INCREMENTAL` is part of the sccache hash, so the first run after this change re-populates the dependency entries.

### Cost caveat: cache entries do get evicted

The same commit was measured on three runs a day apart:

| | 08-18 | 08-19 first run | 08-19 re-run |
|---|---|---|---|
| `Run Cargo Check (per crate)` | 180s | **749s** | 177s |
| `Run Clippy` | 118s | 130s | 117s |
| cache hits / misses | 3595 / 61 | 3161 / 495 | 3595 / 61 |
| hit rate | 98.33% | 86.46% | 98.33% |

The two good runs are bit-for-bit identical (3595 / 61, same 48 names in the miss list — the `gix_*` vergen build-dependency tree plus the mimalloc static object), so 177s is the steady state, not a lucky run. The middle run is what happens after the entries written a day earlier are gone: the workspace crates miss and get re-written, and the per-crate step costs 749s. Whatever eviction the remote cache applies, that is the price of the first run after it.

Real PRs also still recompile the crates they touch plus their dependents, so a PR touching `rspack_core` pays more than 177s.

## Is it safe to cache a lint pass?

Verified on the Linux self-hosted runner (script in commit 2c40524, dropped afterwards) against a fixture whose only defect is a clippy-only lint (`needless_return`, accepted by plain rustc). All assertions passed:

| # | Assertion | Result |
|---|---|---|
| 1 | Cacheability matrix over `{CARGO_INCREMENTAL, RUSTC}` | only `0` + `rustc` hits on re-run; the other three report `multiple input files` / `incremental` and never hit |
| 2 | **Negative:** warm `cargo check` cache, then `cargo clippy -- -D warnings` | still fails (exit 101) with `needless_return` — a check artifact is never served to clippy |
| 3 | A clippy cache **hit** | replays the clippy diagnostics; lints are not silently dropped |
| 4 | `-D warnings` against a warm `-W warnings` entry | still fails |
| 5 | `RUSTC=rustc` vs unset | shares cache entries, i.e. resolves to the same compiler |
| 6 | **Equivalence:** plain clippy / `RUSTC=rustc` / sccache miss / sccache hit | byte-identical diagnostics |
| 7 | Cache keys of `cargo check` vs `cargo clippy` for the same crate | differ (`aadab357…` vs `3b44aff7…`) — the clippy-driver binary is part of the key |

## Follow-ups

- `rust_test` is blocked by `incremental` the same way; kept out of this PR to stay reviewable.
- Upstream a patch so sccache also recognises an absolute rustc path, which would remove the need for `RUSTC=rustc`.


chenjiahan (approved):
LGTM

Issue #15246, opened 2026-08-19
[Bug]: Panic "No code generation entry for css/auto|…" when tree-shaking removes a CSS-module's only JS issuer under `experiments.layers`


noyobo:
`experiments.layers`

## System Info

- `@rspack/core`: **2.1.10**
- `less-loader`: 13, `less`: 4
- Native CSS (`css/auto` module type; **`experiments.css` not set**)
- `experiments.layers: true`, entry has a `layer`
- `mode: 'production'` (tree-shaking + `sideEffects` on)
- Node 24, macOS (arm64)

## Details

With `experiments.layers` enabled and `.module.less` handled as native
`css/auto` through `less-loader`, enabling tree-shaking over a package that
declares `"sideEffects": false` makes rspack **panic during code generation**
instead of degrading gracefully:

```
thread '<unnamed>' panicked:
No code generation entry for css/auto|<abs>/less-loader/dist/cjs/index.js!<abs>/packages/components/src/Label/index.module.less|miniprogram
(existing entries: [ … many javascript/auto and css/auto identifiers, all suffixed `|miniprogram` … ])
```

The panicking identifier is a `css/auto` module (`<Component>/index.module.less`)
whose **only** importer is the component's JS module. When that JS module is an
unused export of a `sideEffects: false` package, tree-shaking removes it — but
the `css/auto` sibling is left in the module graph and the per-layer CSS code
generation still enumerates it, finding no entry → panic.

Used components are fine: both their JS and their `.module.less` get code
generation entries. Only components whose JS was shaken out hit this.

### Diagnosis (module-graph instrumentation)

At `compilation.hooks.optimizeChunkModules`, for an unused component:

```
<C>.module.less   chunks=0  active connections=0   (still present in compilation.modules)
<C>/index.tsx     chunks=0  active connections=0   (issuer already pruned)
```

The dead `css/auto` module has **zero chunks and zero active incoming
connections**, yet it is not garbage-collected before code generation, and the
layer-aware CSS codegen panics on it rather than skipping it.

## Reproduce

Minimal ingredients:

- `experiments.layers: true`; entry `{ import: […], layer: 'x' }`.
- `module.rules`: `{ test: /\.less$/, type: 'css/auto', use: ['less-loader'] }`,
  and `module.parser['css/auto'] = { namedExports: false }`.
- A nested package `"sideEffects": false` exporting several components through a
  barrel, each `import styles from './index.module.less'`.
[...]

PR #15253, opened 2026-08-20
fix: refine binding code generation result error handling

intellild:
## Summary

- Add a fallible `CodeGenerationResults::try_get` API while preserving the existing internal `get` invariant.
- Use `try_get` in the JavaScript binding so invalid code generation result lookups throw a catchable JavaScript error instead of causing a Rust panic.

For example, when a plugin queries a chunkless ModuleGraph module that has no code generation result, the binding now reports an error to JavaScript rather than aborting the process.

## Related links

- Related to #15246

## Checklist

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
@codspeedbot Is this regression real ?

SyMind:
@intellild would you like to add a test case?

SyMind (approved):

PR #15254, opened 2026-08-20
feat(watcher): support a function for the native watcher `ignored` option

stormslowly:
## Why

`watchOptions.ignored` supports a function under watchpack, but `experiments.nativeWatcher` rejected it:

```js
// before — throws under experiments.nativeWatcher
watchOptions: {
  ignored: entry => entry.includes('__generated__'),
}
// Error: NativeWatcher does not support using a function for the 'ignored' option
```

Now both watchers accept the same config, so turning on the native watcher no longer forces a config rewrite.

### The predicate crosses into `rspack_watcher` as a plain async closure

`FsWatcherIgnored` gained a `Function(IgnoredFn)` variant:

```rust
pub type IgnoredFn = Arc<dyn Fn(String) -> BoxFuture<'static, bool> + Send + Sync>;
```

`rspack_watcher` stays free of any napi dependency. This is the same shape `rspack_core` already uses for JS-backed predicates (`CleanOptions::KeepFunc`, `ModuleNoParseRule::TestFn`, `JsonParseFn`); `to_ignored_fn` in the binding is the only place that knows the closure is backed by a `ThreadsafeFunction`.

Matching watchpack, the function receives the raw entry — it is the only form whose path keeps the platform separators — and is asked about each entry on its own, like the existing `RegExp` form, so it has to cover the entries below an excluded directory itself. A throwing predicate is treated as "not ignored", so a broken config degrades to a noisier watch instead of a silently dead one.

### Event classification moved off the OS reader thread

The notify callback now only pushes `(path, kind)` into an unbounded channel. A dedicated `rspack-fs-event-filter` thread drains it one event at a time, in arrival order, and does everything that used to run inline: the `ignored` check, the `stat`-based Change→Remove normalization, the mtime staleness filter and the ancestor lookup.

Why this matters: on Linux, notify's inotify backend reads with a 1024-byte buffer and dispatches synchronously inside the read loop. The kernel queue is bounded (`max_queued_events`, 16384 by default); while the reader is busy classifying, the queue fills and the kernel drops events with `IN_Q_OVERFLOW` — silently, since the overflow marker itself never made it past `if event.paths.is_empty() { return; }`. Each event costs several `stat`s, and a JS predicate costs a full event-loop turn on top, so a large burst could starve the reader even before this PR.

Measured on Linux (16 cores), one burst of N changed module files, two attempts each:

| burst | before, `RegExp` | before, function | after, both |
|---|---|---|---|
| 6000 | ok | **events lost** | ok |
| 12000 | **events lost** | **events lost** | ok |
| 24000 | — | — | ok |

Watchpack survives the same bursts because libuv reads with a 4096-byte buffer and does not subscribe to `IN_OPEN`/`IN_CLOSE_WRITE` (41% of what notify receives, all discarded by rspack); it has no overflow recovery either, since libuv skips the `wd = -1` overflow event outright.

### The predicate is awaited, never blocked on

With classification on its own thread, the filter thread runs a current-thread tokio runtime and awaits the `ThreadsafeFunction` call; path registration awaits it on the watcher thread the same way. `is_ignored`, `PathManager::update` and `wait_for_event` became `async` accordingly. There is no `block_on` and no JS-thread re-entrancy guard: `NativeWatcher::triggerEvent` (a napi method) only enqueues, so the JS thread never ends up waiting on itself. Events injected through it go through the same classification as discovered ones.

## Tests

`NativeWatcherIgnoredFunction.test.js` drives `binding.NativeWatcher` directly:

- the predicate is reached from the watcher's own threads and the entries it rejects never register nor reach an aggregated batch;
- `triggerEvent` returns instead of wedging the JS thread. This one runs in a child process — a wedged JS thread stops the in-process test timeout from firing too, so a regression would hang the job instead of failing it. The child turns it into `expected 'SIGTERM' to be null`.

`watchCases/change-event/ignored-function-no-rebuild` covers it end to end under `watch`, `incremental watch` and `nativeWatcher`. The ignored path is a real file dependency, so a rebuild wrongly triggered by it is observable — replacing the predicate with `() => false` fails all three.

Existing `rspack_watcher` unit tests were adapted mechanically to the async signatures (`#[tokio::test]` + `.await`); no assertion changed.


crates/rspack_watcher/src/paths.rs
```diff
@@ -437,6 +437,39 @@ mod tests {
     }
   }
 
+  #[test]
+  fn test_manager_with_ignored_function() {
```
hardfist: not needed

hardfist:
@codex review

crates/rspack_binding_api/src/native_watcher.rs
```diff
@@ -7,24 +7,46 @@ use std::{
...
+      Either4::B(paths) => FsWatcherIgnored::Paths(paths),
+      Either4::C(regex) => FsWatcherIgnored::Regex(regex),
+      Either4::D(func) => FsWatcherIgnored::Function(to_ignored_fn(func)),
     }
   } else {
     FsWatcherIgnored::None
   }
 }
 
+fn to_ignored_fn(func: ThreadsafeFunction<String, bool>) -> IgnoredFn {
```
hardfist, line 30: why convert tsfn to ignoredFn

hardfist (approved):

PR #15260, opened 2026-08-21
fix(ids): make compare_modules_by_pre_order_index_or_identifier a total order

### LINKED ISSUE #15259: [Bug]: `user-provided comparison function does not correctly implement a total order` panic in production builds? (standalone reproduction attached)
## System Info

| | |
|---|---|
| `@rspack/core` | 2.1.10 |
| Node | v24.18.0 |
| OS | macOS 15.7.3, arm64 (Apple Silicon), 8 cores / 16 GB |

## Details

Production builds abort with `SIGABRT`:

```
Panic occurred at runtime. Please file an issue on GitHub with the panic info and
backtrace below: https://github.com/web-infra-dev/rspack/issues

thread 'tokio-0' (15998938) panicked at <redacted>:0:0:
user-provided comparison function does not correctly implement a total order
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The thread name varies between runs (`tokio-0`, `tokio-1`, `tokio-7`, …).

`RUST_BACKTRACE=full` prints `stack backtrace:` followed by **nothing** — the published native
binary is stripped and the panic location is `<redacted>:0:0`, so no frame is recoverable from a
released build. That is why this report has no stack trace.

Same *class* as #8337 (an inconsistent comparator in `LimitChunkCountPlugin`, fixed in #8392 by
aligning `chunk_combination.rs` with webpack's comparator), but a different site —
`LimitChunkCountPlugin` is not used here.

### Reproduction

Standalone, public dependencies only (`@rspack/core`, optionally `react`):

```bash
npm install
./run.sh 6                                  # 6 interleaved samples per configuration
CONCAT=true node repro.mjs 2 20 30          # a single sample
```

`repro.mjs` generates a synthetic project and builds it with a plain rspack config: `mode:
'production'`, `target: ['web','es5']`, `builtin:swc-loader` for `.tsx`, native CSS Modules
(`css/auto`), `experiments.layers`, one `BannerPlugin` pair, `splitChunks` with a custom
`cacheGroups`. Default size is 20 pages / 30 components, each component importing its own
`.module.css`.

The panic aborts the process, so one process is one sample. Observed hit rates on this machine:
**12/12** and **5/5** in some windows, **0/4** in others, for the exact same command.

### Important: the rate drifts, which makes ablation misleading

The reproduction rate of an *identical* command changes over minutes with no change to config,
cache, cwd or paths. That has a practical consequence for anyone bisecting this, and it burned us
repeatedly:

- **"disabled X, still panics" is reliable** → X is not necessary.
- **"disabled X, panic stopped" is not reliable** by itself. Three separate ingredients looked
  necessary at first and were then refuted by interleaved A/B sampling:
  `optimization.concatenateModules: false` (3/3 → 0/3 on one testbed, but the standalone repro
  panics 3/3 with it off), CSS Modules (0/3 once, then 5/5 with them removed under interleaving),
  and "build two projects sequentially in one process" (0/3 once, then 3/3 with a single project).
- So please sample **interleaved** (A/B/A/B…), never all-A-then-all-B.

The knobs are wired up in the reproduction (`NO_LAYERS`, `NO_ES5`, `NO_SPLIT`, `NO_BANNER`,
`NO_REACT`, `NO_CSS`, `NO_SECOND_APP`, `CONCAT`) precisely so you can re-check any of this
yourself; `./ablate.sh 3` runs them all.

### What is *not* involved

These were disabled individually while the baseline still panicked at the same rate, so they are
not necessary conditions (this direction of evidence is the trustworthy one):

- every plugin from our own build tooling (three that walk the module graph, plus two smaller ones);
- `rspack.CaseSensitivePlugin` — it compares paths case-insensitively, so it was our first suspect;
- `rspack.CircularCheckRspackPlugin`;
- `experiments.layers`;
- `optimization.minimize`;
- persistent cache — not enabled in this mode;
- `RAYON_NUM_THREADS=1`, `UV_THREADPOOL_SIZE=1` (rspack drives its own tokio pool);
- working directory and the filesystem location of the project.

### Source reading

In `crates/rspack_plugin_javascript/src/plugin/module_concatenation_plugin.rs` every sort uses a
total order (`sort_by_cached_key(get_depth)`, `sort_by_key(Reverse(len))`, `sort_unstable()`,
`sort_by_key(|(id, _)| *id)`), so we do not think the offending comparator is there. Given the
timing sensitivity, our guess is a comparator over data that concurrent work can leave in
different states between runs — but that is only a guess, and with a stripped binary we cannot
narrow it further.

## What would help

1. A build that keeps panic locations and symbols, or instructions to build `@rspack/core` locally
   with them — we will capture frames from the attached reproduction and post them.
2. Or a throwaway patch adding `debug_assert` / logging inside the comparators reachable from the
   tokio workers during `optimizeChunkModules` / code generation; we will run it and paste output.
3. Longer term: when this abort fires, naming *which* comparator violated the ordering would save
   everyone a lot of time, since released builds redact both file and line.

Happy to run anything against the reproduction and report back quickly.

---

**Reproduction repository**: [https://github.com/noyobo/rspack-total-order-panic-repro](https://github.com/noyobo/rspack-total-order-panic-repro) — the directory is
ready to push (MIT, `npm install && ./run.sh 6`).

noyobo:
`compare_modules_by_pre_order_index_or_identifier` compared by pre-order index
only when BOTH modules had one, and fell back to identifier otherwise. When the
sorted set mixes modules with and without a pre-order index (e.g. under
splitChunks / experiments.layers — `get_pre_order_index` returns `Option<u32>`),
this is not transitive: a numeric edge and an identifier edge can disagree, e.g.

  A(idx=5,id="z"), B(no idx,id="m"), C(idx=10,id="a")
  A<C (5<10), C<B ("a"<"m")  =>  A<B, but A>B ("z">"m")   x

`sort_unstable_by` aborts the process ("comparison function does not correctly
implement a total order") once it detects this (Rust >= 1.81); the failure is
non-deterministic because FxHashMap iteration order decides whether the violating
triple is visited. webpack's Array.sort runs the same logic but only yields an
unspecified order, so the port went unnoticed.

Use a single sort key: pre-order index (missing sorts last via u32::MAX) with the
identifier as a deterministic tiebreak. Preserves the intended index ordering for
the common case and is a strict total order.

Repro + verification (debug binding, same window): stock 6/6 panics, patched 0/20.
Closes #15259.

### DIFF
--- a/crates/rspack_ids/src/id_helpers.rs
+++ b/crates/rspack_ids/src/id_helpers.rs
@@ -223,13 +223,35 @@ pub fn compare_modules_by_pre_order_index_or_identifier(
   a: &Identifier,
   b: &Identifier,
 ) -> std::cmp::Ordering {
-  if let Some(a) = module_graph.get_pre_order_index(a)
-    && let Some(b) = module_graph.get_pre_order_index(b)
-  {
-    compare_numbers(a, b)
-  } else {
-    compare_ids(a, b)
-  }
+  compare_by_pre_order_index_or_id(
+    module_graph.get_pre_order_index(a),
+    a,
+    module_graph.get_pre_order_index(b),
+    b,
+  )
+}
+
+/// Ordering logic for [`compare_modules_by_pre_order_index_or_identifier`], split
+/// out as a pure function so its total-order property is unit-testable without a
+/// `ModuleGraph`.
+///
+/// Must be a *strict total order*: Rust's `sort_unstable_by` aborts the process
+/// ("comparison function does not correctly implement a total order") once it
+/// detects a violation. The previous form — compare by pre-order index only when
+/// BOTH modules have one, else fall back to identifier — is not transitive when
+/// the set mixes modules with and without a pre-order index (e.g. under
+/// splitChunks / `experiments.layers`, where `get_pre_order_index` is `None` for
+/// some modules): a numeric edge and an identifier edge can disagree. Use a single
+/// key — pre-order index (missing sorts last via `u32::MAX`) — with the identifier
+/// as a deterministic tiebreak.
+fn compare_by_pre_order_index_or_id(
+  a_index: Option<u32>,
+  a_id: &str,
+  b_index: Option<u32>,
+  b_id: &str,
+) -> std::cmp::Ordering {
+  compare_numbers(a_index.unwrap_or(u32::MAX), b_index.unwrap_or(u32::MAX))
+    .then_with(|| compare_ids(a_id, b_id))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


LingyuCoder:

LingyuCoder (approved):
The performance regression here is a necessary trade-off for correctness, so I think this can be merged. Thank you very much for this contribution.

PR #15266, opened 2026-08-21
fix: set javascript/esm moduleType for ts files whose package.json contains type="module"

JSerFeng:
## Summary

- classify `.ts` files in packages with `type: module` as `javascript/esm`
- classify `.ts` files in packages with `type: commonjs` as `javascript/dynamic`
- keep `fullySpecified` on its default `false` behavior and cover module, CommonJS, and unspecified package types


JSerFeng:
It breaks too much

PR #15273, opened 2026-08-21
fix: keep import.meta.url in createRequire when importMeta is disabled

### LINKED ISSUE #15270: [Bug]: createRequire parser folds import.meta.url into an absolute path when importMeta is false
**Version:** @rspack/core 2.1.10

**Reproduction:** https://github.com/elecmonkey/rspack-createrequire-abs-path-repro — `pnpm install && pnpm build`, read `dist/out.js`.

**Details**

With `parser.javascript = { importMeta: false, requireResolve: false, createRequire: true }`, the argument of `createRequire()` is rewritten to the module's build-time absolute file URL:

```js
// src/index.js
var __require = createRequire(import.meta.url);
__require.resolve('picomatch');
export const selfUrl = import.meta.url;
```

```js
// dist/out.js
var __require = createRequire("file:///Users/.../src/index.js");   // folded
const selfUrl = import.meta.url;                                   // preserved
```

Two reads of `import.meta.url` in one module, treated differently — `importMeta: false` is honoured everywhere except here.

Only `var` and `let` are affected; `const` emits `createRequire(import.meta.url)` correctly. `tests/rspack-test/esmOutputCases/create-require/import-meta-url-resolve-disabled` already covers the intended behaviour ("should delegate import.meta.url parsing when requireResolve is disabled", #14783 / #14813), but it declares `const req = ...`, so `var` and `let` were never covered.

In `common_js_imports_parse_plugin.rs` the delegation is gated on `declaration.kind() == VariableDeclarationKind::Const` (in `declarator()`, plus the early return in `pre_tag_created_require_declarator`). Other kinds fall through to `should_replace_create_require_argument`, which returns `true` for a plain member expression, and the argument is replaced with `evaluate_create_require_argument()`'s value — that helper resolves `import.meta.url` via `Url::from_file_path(parser.resource_data.resource())` without consulting the `importMeta` option.

**Expected**

`importMeta: false` keeps `import.meta.url` in the output regardless of the declaration kind.

lazerg:
## Summary

With `parser.javascript.importMeta` disabled, `createRequire(import.meta.url)` still had its argument folded into the module's build-time absolute file URL, but only when the binding was declared with `var` or `let`. The `const` form already delegated the argument back to the normal parser walk, so `import.meta.url` survived there. The argument replacement now also checks whether `import.meta.url` is enabled, so the declaration kind no longer changes the output.

## Related links

Fixes #15270

## Checklist


intellild (approved):

intellild:
@codex review

PR #15276, opened 2026-08-21
fix(cache): clear stale issuer updates between rebuilds

### LINKED ISSUE #14856: [Perf]: issuer_update_modules remains live during later incremental persistent-cache saves
### Reproduction

- Rspack: @rspack/core@2.1.4
- Repository: https://github.com/hardfist/rspack-repro-issuer-update-cache

The fixture builds a graph where a.js and b.js both import shared.js, then removes whichever parent Rspack selected as the initial issuer. Two subsequent rebuilds edit only an unrelated leaf. The script captures the public issuer transition and Rspack's logger trace for the make persistent-cache save span.

Observed output has the following shape (the selected initial parent may be either a.js or b.js because the graph is constructed concurrently):

~~~text
initial build              issuer=.../a.js  makeCacheSaveSpans=1
issuer-changing rebuild   issuer=.../b.js  makeCacheSaveSpans=1
first unrelated leaf edit issuer=.../b.js  makeCacheSaveSpans=1
second unrelated leaf edit issuer=.../b.js  makeCacheSaveSpans=1
BUG REPRODUCED: after shared.js changes issuer, unrelated rebuilds still reach the make-cache save path.
~~~

### Expected behavior

After the issuer transition has been consumed for one rebuild, a later rebuild that changes only an unrelated leaf should not carry the old issuer-update module set into the persistent make-cache write set. This matters for large graphs because it causes unnecessary module serialization and cache writes.

### Suspected cause

BuildModuleGraphArtifact::reset_temporary_data clears affected modules/dependencies and side-effects state but does not clear issuer_update_modules (crates/rspack_core/src/artifacts/build_module_graph_artifact.rs:222-231). The persistent make occasion then clones that set into need_update_modules before extending current affected modules (crates/rspack_core/src/cache/persistent/occasion/make/mod.rs:45-74).

The repro's public trace is available on the standard package; a debug native build can additionally expose per-module save counts.

zhsama:
## Summary

- clear per-build issuer update tracking with the other temporary module graph state
- prevent issuer-only modules from leaking into later persistent cache write sets
- add a compiler regression covering an issuer transition followed by unrelated rebuilds

## Related links

Fixes #14856

## Validation

- `pnpm run build:js`
- `cargo test -p rspack_core --lib`
- `cargo check -p rspack_core --all-targets --locked`
- `cargo fmt --all --check`
- `cd tests/rspack-test && pnpm run test -t 'compilerCases/persistent-cache-issuer-updates'`
- `cd tests/rspack-test && pnpm run test -t 'cacheCases/make/issuer-update'`

## Checklist

- [x] Tests updated.
- [x] Documentation updated (not required; internal cache lifecycle fix).


LingyuCoder (approved):

LingyuCoder:
LGTM, thanks

PR #15279, opened 2026-08-21
test(copy-plugin): cover brace expansion in glob patterns

### LINKED ISSUE #7778: [Feature]: Support for `{}` glob patterns in CopyRspackPlugin

I have a config like this:

```
      new rspack.CopyRspackPlugin({
          patterns: [
              {
                  context: path.resolve(__dirname, 'translations/'),
                  from: '{foo,bar,baz}.*.yml',
                  to: 'js/translations/[name].json',
                  toType: 'template',
                  transform: transformTranslations,
              }
          ],
      }),
```

This results in an error being output:

> ERROR in × unable to locate '/path/to/translations/{foo,bar,baz}.*.yml' glob

Ideally the glob() support for `{foo,bar}` would be present. It worked in CopyWebpackPlugin, but I think the issue is that rust's glob crate does not support it? At least it's not clear from the docs whether it does.. Anyway for now I have replaced it by multiple rules with `foo.*.yml`, `bar.*.yml`, etc. as a workaround.

kakiuwang-ui:
## Summary

When #7778 was filed, `CopyRspackPlugin` could not expand brace patterns such as `{foo,bar,baz}.*.yml`, and the issue was answered with "the rust glob does not support this match pattern yet".

That is no longer true. The copy plugin now globs through `fast_glob::glob_match` (`crates/rspack_core/src/glob_utils.rs`, `crates/rspack_plugin_copy/src/lib.rs`), and `fast-glob` does support brace expansion. `{` is already treated as a glob metacharacter by `is_glob_metacharacter`/`glob_base_dir_end`, so base-directory extraction handles brace patterns correctly too.

I verified against `@rspack/core@2.1.10` that all of these work today and that non-matching files are correctly excluded:

| pattern | copied | not copied |
| --- | --- | --- |
| `src/{foo,bar}.*.yml` | `foo.en.yml`, `bar.de.yml` | `qux.en.yml` |
| `src/{alpha,beta}.txt` | `alpha.txt`, `beta.txt` | `gamma.txt` |
| `src/{one,two}/**/*.txt` | `one/deep/x.txt`, `two/y.txt` | `three/z.txt` |

So the feature works but nothing pins it. This PR adds a `configCases` case covering the three shapes above, including the negative cases, so a future change to the glob layer cannot silently regress it.

The middle row matters most: a pattern whose only metacharacter is `{` still has to be classified as a glob rather than as a literal path, which is a distinct code path from the other two.

No source changes — test only.

## Related links

- Closes #7778

## Checklist

### DIFF
--- a/tests/rspack-test/configCases/plugins/copy-plugin-braces/index.js
+++ b/tests/rspack-test/configCases/plugins/copy-plugin-braces/index.js
@@ -1,0 +1,24 @@
+const fs = require("fs");
+const path = require("path");
+
+const read = file =>
+	fs.readFileSync(path.join(__STATS__.outputPath, file), "utf-8").trim();
+const exists = file => fs.existsSync(path.join(__STATS__.outputPath, file));
+
+it("should copy files when a brace pattern is combined with a wildcard", () => {
+	expect(read("wildcard/foo.en.yml")).toBe("from foo");
+	expect(read("wildcard/bar.de.yml")).toBe("from bar");
+	expect(exists("wildcard/qux.en.yml")).toBe(false);
+});
+
+it("should copy files when a brace is the only glob character", () => {
+	expect(read("literal/alpha.txt")).toBe("from alpha");
+	expect(read("literal/beta.txt")).toBe("from beta");
+	expect(exists("literal/gamma.txt")).toBe(false);
+});
+
+it("should copy files when a brace pattern matches a directory segment", () => {
+	expect(read("nested/src/one/deep/x.txt")).toBe("from one");
+	expect(read("nested/src/two/y.txt")).toBe("from two");
+	expect(exists("nested/src/three/z.txt")).toBe(false);
+});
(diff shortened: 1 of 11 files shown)


kakiuwang-ui:
@LingyuCoder could you approve the CI run here when you get a chance? This is my first contribution to rspack, so the workflows are waiting on manual approval and nothing has run yet.

Pinging you specifically since you answered the original issue: #7778 has effectively been fixed already by the `glob` -> `fast-glob` migration. Brace patterns work today — I verified against `@rspack/core@2.1.10` — they just have no test pinning them, which is what this PR adds.

It is test-only, no source changes: one `configCases` case covering three brace shapes (`{a,b}.*.yml`, `{a,b}.txt`, `{a,b}/**/*.txt`) plus a negative assertion for each. Should be a quick read.

hardfist (approved):

chenjiahan:
Thank you!

PR #15298, opened 2026-08-22
fix: include factory-created server actions in RSC metadata

nikhilsnayak:
## Summary

A `use server` module can export a Server Function through an expression whose result cannot be classified statically, for example:

```js
'use server'

import { validator } from 'auth'

export const action = validator(async () => {})
```

The server-actions post-pass already discovers these exports, but it only copied their reference IDs into `reference_ids_by_export_name` when compiling the client layer. On the server layer this caused `RscMeta.action_ids` to omit post-pass-discovered actions. If a module contained only those actions, `has_action` also remained false, so the server-reference imports, runtime validation, and registration annotations were not emitted.

Populate the reference map for `use server` files in both runtime layers. The existing `ensureServerActions` check continues to reject values that do not evaluate to functions at runtime.

The fixture updates cover a named-only factory export, bound method exports, and the runtime-validation path for a value whose type cannot be established statically.

## Related links

No issue filed.

## Testing

- `cargo test -p rspack_loader_swc`

## Checklist

### DIFF
--- a/crates/rspack_loader_swc/src/rsc_transforms/server_actions.rs
+++ b/crates/rspack_loader_swc/src/rsc_transforms/server_actions.rs
@@ -1560,7 +1560,7 @@ impl<'a, C: Comments> VisitMut for ServerActions<'a, C> {
       }
     }
 
-    if in_action_file && !self.config.is_react_server_layer {
+    if in_action_file {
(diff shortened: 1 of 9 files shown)


SyMind (approved):

SyMind:
@nikhilsnayak Thanks!

PR #15300, opened 2026-08-22
fix(cache): validate minimize persistent cache entries against asset identity

### LINKED ISSUE #14862: [Bug]: minimize recovery can cross-wire cached outputs between assets
### Reproduction

Repository: https://github.com/hardfist/rspack-persistent-cache-minimize-recovery-repro

This is a JavaScript repro using the public compiler API and the published `@rspack/core@2.1.4` package:

```sh
pnpm install
pnpm repro
```

The repro builds two production entrypoints (`a.js` and `b.js`) with filesystem persistent caching enabled. It then swaps the two 8-byte keys in `occasion_minimize/0.pack`, recomputes the pack integrity hash, and starts a fresh Node process against the cache.

Observed on `@rspack/core@2.1.4`:

```
recover exit: 0
dist/a.js contains A: false; contains B: true
dist/b.js contains A: true; contains B: false

BUG REPRODUCED: minimize recovery returned another asset's cached output.
```

Recovery reports success and emits valid JavaScript, but each asset receives the other asset's minimized source. The repro exits 0 after detecting the cross-wiring.

### Expected behavior

A recovered minimize entry must be validated against the exact asset/source key that produced it. If a persistent minimize entry is not usable, Rspack should invalidate/recompute it rather than silently emitting another asset's minimized output.

### Environment

- `@rspack/core@2.1.4`
- Node.js 24
- macOS arm64

raunak2007:
## Summary

`occasion_minimize` persists entries keyed only by `MinimizeCacheKey`, an 8-byte
`FxHasher` digest of `(source, options_hash, filename, is_module)`. The stored
value carries no identity of its own, and `MinimizeCacheKey::from_bytes`
validates only length — any 8 bytes parse into a well-formed key.

`MinimizeOccasion::recovery` therefore accepts whatever decodes and inserts it
under whatever key it arrived with. Nothing asks whether the value belongs to
the asset requesting it, so a key/value association that is wrong in storage
propagates into `original.set_source(...)`, and the asset is emitted with
another asset's minimized body, marked `minimized = true`, with no error.

Pack integrity hashing proves a pack was not corrupted. It cannot prove that a
given key still maps to the value written under it. Since this cache is a file
on disk that outlives the process — partially written on crash, copied between
branches, restored from CI artifacts — that gap is reachable in practice.

Strengthening the key does not close it: the association is what is wrong, not
the key, so a swap defeats any key scheme. This adds
`EntryIdentity { filename, options_hash, is_module }` to the persisted entry and
compares it on read; a mismatch is treated as a cache miss and the asset is
minimized again.

Note the `new_cache` path twenty lines above in the same function is not
affected — it keys on `asset_filename` with the hash as a separate etag, so a
bad etag is already a miss. But `experiments.newCache` defaults to `false`
(`defaults.ts:282`, asserted in `defaultsCases/default/base.js:27`), so the
unvalidated legacy path is the one that ships by default. `CachedMinimizeEntry`
is unchanged, so the `new_cache` path is untouched here.

## Verification

Against the reporter's repro, on a local build of `main`.

Before:

    dist/a.js contains A: false; contains B: true
    dist/b.js contains A: true; contains B: false
    BUG REPRODUCED: minimize recovery returned another asset's cached output.

After:

    dist/a.js contains A: true; contains B: false
    dist/b.js contains A: false; contains B: true
    Bug not reproduced: minimized outputs remained associated with their assets.

Three unit tests added, covering the cross-wired case, the matching case (cache
still hits), and stale `options_hash` / `is_module`.

### Cache compatibility

`Entry` gains a field, so packs written by an older rspack no longer decode.
`recovery` already handles decode failure by warning and skipping, so a
pre-upgrade cache degrades to a miss and rebuilds rather than erroring.

## Related links

Fixes #14862

## Checklist

### DIFF
--- a/crates/rspack_core/src/cache/persistent/occasion/minimize/mod.rs
+++ b/crates/rspack_core/src/cache/persistent/occasion/minimize/mod.rs
@@ -14,8 +14,23 @@ use crate::RayonConsumer;
 
 pub const SCOPE: &str = "occasion_minimize";
 
+/// Identity of the asset that produced a cached minimize entry.
+///
+/// `MinimizeCacheKey` folds these values into a single `u64`, which is lossy:
+/// a key that has been swapped, truncated, or collided still parses into a
+/// well-formed key. Persisting the identity beside the value lets recovery
+/// reject an entry that does not belong to the asset that requested it.
+#[cacheable]
+#[derive(Debug, Clone)]
+pub struct EntryIdentity {
+  pub filename: String,
+  pub options_hash: u64,
+  pub is_module: Option<bool>,
+}
+
 #[cacheable]
 struct Entry {
+  pub identity: EntryIdentity,
   #[cacheable(with=AsPreset)]
   pub source: BoxSource,
   pub extracted_comments: Option<ExtractedCommentsEntry>,
@@ -49,7 +64,7 @@ impl MinimizeCacheKey {
 
 #[derive(Debug, Default)]
 pub struct MinimizePersistentCache {
-  entries: FxHashMap<MinimizeCacheKey, CachedMinimizeEntry>,
+  entries: FxHashMap<MinimizeCacheKey, (EntryIdentity, CachedMinimizeEntry)>,
   /// Keys of entries that were added during this build and need to be persisted.
   dirty_keys: Vec<MinimizeCacheKey>,
 }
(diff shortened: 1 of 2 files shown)


raunak2007:
All 35 CI checks pass; CodSpeed is the only red.

On attributability — the report's own footnote says it used 73f8c9e as the base rather than my branch point b89cf5d, so unrelated commits are in the diff, and it flags differing runtime environments. 49 of 50 benchmarks are untouched, and the delta is 5µs on a 144µs microbenchmark.

On mechanism — create_module_assets is a compilation pass that runs during seal (crates/rspack_core/src/compilation/create_module_assets/mod.rs). The minimize persistent cache is only reached from SwcJsMinimizerRspackPlugin in processAssets. Neither file this PR touches is on that path, and MinimizePersistentCache's stack size is unchanged, so Compilation's layout is too.

Happy to dig further if you think it's real.

PR #15308, opened 2026-08-24
fix: pre-register modules before chunk graph splitting

### LINKED ISSUE #15136: panic: Module should be added before using when bundling lodash-es with swc-loader
## Summary

Rspack panics while processing modules from `lodash-es` through `swc-loader`.

## Error

Two instances were reported:

```
Message: Module(javascript/esm|builtin:swc-loader??ruleSet[1].rules[7].use[0]!/Users/leozang/workspace/pages/node_modules/lodash-es/_Stack.js) should be added before using
Location: crates/rspack_core/src/chunk_graph/chunk_graph_module.rs:153
```

```
Message: Module(javascript/esm|builtin:swc-loader??ruleSet[1].rules[7].use[0]!/Users/leozang/workspace/pages/node_modules/lodash-es/startsWith.js) should be added before using
error Command failed with exit code 1.
```

## Backtrace

The runtime output says the backtrace was omitted. Re-running with `RUST_BACKTRACE=1` (or `RUST_BACKTRACE=full` for source snippets) should provide the missing diagnostic details.

## Expected behavior

Rspack should handle these `lodash-es` modules without panicking, or produce a recoverable compilation error with actionable context.

## Additional context

The failing module identifiers show `builtin:swc-loader` at `ruleSet[1].rules[7].use[0]`. Exact Rspack, Node.js, loader, and project versions were not included in the report.

mvanhorn:
## Summary

Add a config-case reproduction that imports representative `lodash-es` functions while applying `builtin:swc-loader` to JavaScript ESM, so successful compilation and execution prove the reported module path no longer panics. In `build_chunk_graph`, register the complete `all_modules` set in the chunk graph before entry preparation and splitting can read or connect module state, while keeping chunk membership empty until the splitter establishes it. Remove the now-redundant post-split registration loop, preserving the special handling of weak or otherwise unconnected modules without relaxing `expect_chunk_graph_module` callers.

Rspack can panic with `Module(...) should be added before using` while production bundling walks `lodash-es` modules transformed by `builtin:swc-loader`. The panic identifies `ChunkGraph`'s missing-module assertion, and the build-chunk-graph flow currently waits until after splitting and orphan removal to register every module identifier. That ordering leaves modules which are skipped, weakly connected, or reached through a transformed ESM dependency vulnerable to chunk-graph queries before their `ChunkGraphModule` record exists. The fix must preserve the assertion as an internal invariant and prevent the invalid state instead of converting the panic into a silent empty result.

Fixes #15136

## Related links

Not applicable to this change.

## Checklist

### DIFF
--- a/crates/rspack_core/src/compilation/build_chunk_graph/mod.rs
+++ b/crates/rspack_core/src/compilation/build_chunk_graph/mod.rs
@@ -28,6 +28,14 @@ pub fn build_chunk_graph(compilation: &mut Compilation) -> rspack_error::Result<
     .copied()
     .collect::<Vec<_>>();
 
+  // make sure all module (weak dependency particularly) has a cgm
+  for module_identifier in &all_modules {
+    compilation
+      .build_chunk_graph_artifact
+      .chunk_graph
+      .add_module(*module_identifier)
+  }
+
   splitter.prepare(&all_modules, compilation)?;
 
   splitter.update_with_compilation(compilation)?;
@@ -41,14 +49,6 @@ pub fn build_chunk_graph(compilation: &mut Compilation) -> rspack_error::Result<
 
   // remove empty chunk groups
   splitter.remove_orphan(compilation)?;
-
-  // make sure all module (weak dependency particularly) has a cgm
-  for module_identifier in all_modules {
-    compilation
-      .build_chunk_graph_artifact
-      .chunk_graph
-      .add_module(module_identifier)
-  }
 
   compilation
     .build_chunk_graph_artifact
--- a/crates/rspack_core/src/compilation/build_chunk_graph/mod.rs
+++ b/crates/rspack_core/src/compilation/build_chunk_graph/mod.rs
@@ -28,12 +28,12 @@ pub fn build_chunk_graph(compilation: &mut Compilation) -> rspack_error::Result<
     .copied()
     .collect::<Vec<_>>();
 
-  // make sure all module (weak dependency particularly) has a cgm
+  // Make sure all modules (particularly weak dependencies) have a CGM before splitting.
   for module_identifier in &all_modules {
     compilation
       .build_chunk_graph_artifact
       .chunk_graph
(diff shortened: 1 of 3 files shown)


LingyuCoder:
@codex review

LingyuCoder:
Thanks for your contribution! I confirmed an orphan `ChunkGraphModule` issue with this PR when the incremental build-chunk-graph path is enabled. During a watch rebuild, a module can remain in `ModuleGraph` after its orphaned chunk is removed, while its `ChunkGraphModule` has already been deleted, resulting in `Module(.../orphan.js) should be added before using`.

Reproduction: https://github.com/LingyuCoder/rspack-orphan-chunk-graph-repro

Please run `npm run setup` followed by `npm run reproduce`. Could you take a look and try to preserve or restore the module's `ChunkGraphModule` after orphan cleanup?

mvanhorn:
Thanks for the precise diagnosis, that was the whole bug. Fixed at `9f4bcf39`.

`remove_orphan` can delete the `ChunkGraphModule` for a module that is still present in `ModuleGraph`, so the two graphs disagree afterwards. The pre-registration before splitting was not enough on its own because the cleanup runs after it. I now re-register every module after orphan cleanup, so anything still in `ModuleGraph` is guaranteed to have a CGM again.

I also fixed the config formatting in the `issue-15136` case, which was what the Lint and format check was failing on.

Verified: `cargo check -p rspack_core` passed, `cargo test -p rspack_core --lib` 120 passed / 0 failed, `cargo clippy -p rspack_core --all-targets -- -D warnings` passed, `cargo fmt --all --check` passed. One gap I should be explicit about: I could not run the JS config case locally because the environment had no registry access to install dependencies, so CI is the real signal on that one.

LingyuCoder:
LGTM, need a double check @JSerFeng

LingyuCoder (approved):

JSerFeng (approved):

JSerFeng:
Thank you

PR #15323, opened 2026-08-24
perf(resolver): collect path dependencies as lists and deduplicate once

stormslowly:
Stacked on #15306 — review that one first; this PR's diff is against its branch.

> **Status: the premise behind this change did not survive measurement.** Kept open as a draft
> because the negative result and the arithmetic below are worth recording. See *Outcome*.

## What this assumed

Resolution collects path dependencies into a `Vec` that still contains duplicates, and
`HashSet::extend` reserves the whole (still-duplicated) `size_hint`. The assumption was that this
over-reserves substantially, so collecting into a plain list and deduplicating once in place would
avoid a per-resolution hash table.

## Why that is wrong

hashbrown rounds capacity to `next_power_of_two(cap * 8 / 7)`. With ~225 unique paths per
resolution and the measured duplication factor of **1.66×**:

| | raw `cap` | `cap * 8 / 7` | buckets |
|---|---|---|---|
| deduplicated | 225 | 257 | **512** |
| as actually reserved | 374 | 427 | **512** |

**The same bucket count.** The extra allocation this PR set out to remove is not small — it is
exactly zero. Crossing into 1024 buckets needs `cap ≥ 449`, i.e. a duplication factor of **1.996×**,
and this workload is below that threshold.

This also separates two things that look like one problem:

- **#15306** targets the *container's structural* overhead — hashbrown's 7/8 load factor plus
  power-of-two bucket counts make a 512-bucket table hold 225 entries in 4640 B where the payload is
  1800 B. That is 2.6×, it is independent of duplication, and removing it is a real 716 MB.
- **This PR** targets *duplicate-driven* over-reservation, which requires a duplication factor above
  2.0× to exist at all. At 1.66× there is nothing to remove.

## What measurement showed

Three alternating A/B pairs on a large monorepo build (mimalloc, cache off, output identical down
to the byte in every run):

| pages, cumulative | base | this branch | paired Δ |
|---|---|---|---|
| pair 1 | 250400 | 258400 | +8000 |
| pair 2 | 248400 | 258700 | +10300 |
| pair 3 | 252600 | 247800 | **−4800** |

The paired difference changes sign, and the two groups overlap (base 248400–252600, branch
247800–258700). `pages` peak behaves the same way: paired Δ from −800 to +200, groups overlapping.

**No difference was measured, in either direction** — which is exactly what the arithmetic above
predicts. `reserve(374)` and `reserve(225)` produce the same allocation, so there was never anything
for this change to save, and the measurement agrees.

What this does *not* say is that the two are identical. The run-to-run spread of this pipeline on
this workload bounds what it can resolve:

| axis | base spread | branch spread | resolves |
|---|---|---|---|
| pages, peak | 0.34% | 0.76% | effects above ~1% |
| pages, cumulative | 1.68% | 4.28% | effects above ~5% |
| `committed` peak | 7.0% | | — |
| user CPU | 7.8% | | — |
| wall clock | 10.0% | | — |

So the claim is "any difference is below ~5%", not "there is no difference". (`committed` cumulative
reads as 0.00% only because its display quantum is 0.1 GiB, i.e. 1.3%.)

Figures are page counts, not bytes: mimalloc's `pages` counter records how many times a page was
taken from a segment, and pages return to the segment for reuse, so it does not convert to a byte
figure. The byte-level question is unanswerable with the tooling available — `malloc req` needs
`MI_STAT>=2`, which this build does not enable, and heaptrack deadlocks against node
(`dl_iterate_phdr` holds `_dl_load_lock`, the allocation lands in heaptrack's malloc hook,
libunwind then blocks on its own lock).

## Outcome

The premise does not hold and the measurement is consistent with that: no memory difference in
either direction, above a resolution floor of roughly 5%. There is therefore no memory argument for
this change, and the CPU side cannot supply one either — the estimated effect is under 1% of user
CPU against a 7.8% run-to-run spread.

A run against this branch's current HEAD is still in flight. By the same arithmetic it is expected
to be null as well.

### DIFF
--- a/crates/rspack_binding_api/src/resolver.rs
+++ b/crates/rspack_binding_api/src/resolver.rs
@@ -81,12 +81,12 @@ impl JsResolver {
             let mut resolve_request = ResolveRequest::from(resource);
             resolve_request.file_dependencies = resolve_dependencies
               .file_dependencies
-              .drain()
+              .drain(..)
               .map(|path| path.to_string_lossy().into_owned())
               .collect();
             resolve_request.missing_dependencies = resolve_dependencies
               .missing_dependencies
-              .drain()
+              .drain(..)
               .map(|path| path.to_string_lossy().into_owned())
               .collect();
             Ok(match simd_json::to_string(&resolve_request) {
--- a/crates/rspack_core/src/artifacts/factorization_artifact.rs
+++ b/crates/rspack_core/src/artifacts/factorization_artifact.rs
@@ -1,6 +1,6 @@
 use rspack_cacheable::cacheable;
 use rspack_error::Diagnostic;
-use rspack_paths::{InternedPath, InternedPathSet};
+use rspack_paths::{InternedPath, dedup_paths};
 use rustc_hash::FxHashMap;
 
 use crate::DependencyId;
@@ -11,7 +11,7 @@ pub struct FactorizeInfo {
   related_dep_ids: Vec<DependencyId>,
   /// These are only ever iterated, so they are stored as compact slices instead of hash sets.
   /// They must stay duplicate-free: `FileCounter::remove_files` panics when the same path is
-  /// removed twice. Building them from an `InternedPathSet` guarantees that.
+  /// removed twice. [`dedup_paths`] establishes that in `new`.
   file_dependencies: Box<[InternedPath]>,
   context_dependencies: Box<[InternedPath]>,
   missing_dependencies: Box<[InternedPath]>,
@@ -22,19 +22,22 @@ impl FactorizeInfo {
   pub fn new(
     diagnostics: Vec<Diagnostic>,
     related_dep_ids: Vec<DependencyId>,
-    file_dependencies: InternedPathSet,
-    context_dependencies: InternedPathSet,
-    missing_dependencies: InternedPathSet,
+    mut file_dependencies: Vec<InternedPath>,
+    mut context_dependencies: Vec<InternedPath>,
+    mut missing_dependencies: Vec<InternedPath>,
   ) -> Self {
     assert!(
       !related_dep_ids.is_empty(),
       "factorization should contain at least one dependency"
     );
+    dedup_paths(&mut file_dependencies);
+    dedup_paths(&mut context_dependencies);
+    dedup_paths(&mut missing_dependencies);
     Self {
       related_dep_ids,
-      file_dependencies: file_dependencies.into_iter().collect(),
-      context_dependencies: context_dependencies.into_iter().collect(),
-      missing_dependencies: missing_dependencies.into_iter().collect(),
+      file_dependencies: file_dependencies.into_boxed_slice(),
+      context_dependencies: context_dependencies.into_boxed_slice(),
+      missing_dependencies: missing_dependencies.into_boxed_slice(),
       diagnostics,
     }
   }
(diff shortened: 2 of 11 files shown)


stormslowly:
Closing: the effect is real but two orders of magnitude below what matters, and the premise it was
designed against does not hold.

## The premise does not hold

hashbrown rounds capacity to `next_power_of_two(cap * 8 / 7)`. Crossing from 512 to 1024 buckets
needs `cap >= 449`, a duplication factor of **1.996×**. This workload measures **1.66×**, so
`reserve(374)` and `reserve(225)` produce a byte-identical allocation. The over-reservation this PR
set out to remove is not small — at this duplication factor it does not exist.

Two distinct kinds of waste live in the same `HashSet`, and only one is real here:

- **structural** — the 7/8 load factor plus power-of-two bucket counts put 225 entries in 4640 B
  against 1800 B of payload, independent of duplication. That is what #15306 removes, worth 716 MB.
- **duplicate-driven** over-reservation, which needs a duplication factor above 2.0× to exist at
  all. This PR targeted that one.

## What it actually does, measured

heaptrack, one run per arm:

| | base | this branch | Δ |
|---|---|---|---|
| allocations | 389,811,574 | 389,947,206 | +135,632 (+0.035%) |
| bytes allocated | 135,797,583,310 | 134,958,007,184 | **−839,576,126 (−0.62%)** |
| peak live bytes | 6,700,825,859 | 6,691,753,023 | **−9,072,836 (−0.14%)** |

The mechanism is directly visible in the size histogram rather than inferred:

| bucket | base | this branch | Δ |
|---|---|---|---|
| ≤4K | 18,581,406 | 18,780,163 | +198,757 |
| ≤64K | 3,799,256 | 3,610,122 | −189,134 |

Near-equal and opposite, about 190k each way, against 183,320 factorizations — one table in the
4K–64K class (4624 B) traded for one vector in the ≤4K class (2992 B), per factorization.

**So it does what it was meant to do. It just does not matter:** peak live bytes fall by 0.14%, and
every process-level axis resolves at best to a few tenths of a percent — `pages` peak 0.34–0.76%,
`pages` cumulative 1.68–4.28%, `committed` peak 7.0%, user CPU 7.8%, wall clock 10.0%. Three
alternating A/B pairs found nothing on any of them, with paired differences changing sign
(`pages` cumulative +8000 / +10300 / −4800). That is the expected outcome for a 0.14% effect, not
evidence of no effect.

## The later commits on this branch make it worse

The branch HEAD replaces the `append` in `merge_dependencies` with `mem::take`, which turns out to
regress the one axis that does resolve at this scale:

```
paired Δ vs base (n=3):  pages peak +2900 / +3600 / +3900   (2.0–2.7%)
base  [144800, 145400]   (n=6, spread 600 / 0.41%)
HEAD  [147800, 149300]   non-overlapping
```

The likely cause is that `mem::take` hands the caller the collection vector *with its doubled-up
spare capacity* (cap 512 holding 374), where `append` sized it exactly, and that spare capacity then
stays live until `into_boxed_slice` in `FactorizeInfo::new`. One allocation saved, more bytes held
for longer. Not verified.

## For anyone trying this again

The measurement bounds the process-level difference below roughly 5%; it does not show there is
none. Anything under a few hundred MB is invisible to RSS-level instrumentation on this workload —
if the expected effect is smaller than that, measure allocations directly (heaptrack) or compute the
expected effect from the container's growth rule first, which in this case gives zero without
running anything.

The branch is left in place for the record.

PR #15326, opened 2026-08-25
fix(core): run afterCompile before shouldEmit and asset emission

### LINKED ISSUE #15313: [Bug]: compiler.hooks.afterCompile runs after shouldEmit and asset emission
### System Info

```text
System:
  OS: macOS 26.5.2 (arm64)
Binaries:
  Node: 24.12.0
  pnpm: 11.22.0
npmPackages:
  @rspack/core: 2.1.4
```

The same lifecycle ordering is still present on the current `main` branch.

### Details

Related downstream issue: https://github.com/rstackjs/rsbuild-plugin-type-check/issues/104

`compiler.hooks.afterCompile` currently runs after `shouldEmit`, `emit`, and `afterEmit` in Rspack.

The observed order is:

```text
afterSeal
shouldEmit
emit
afterEmit
afterCompile
done
```

This differs from webpack, where `afterCompile` runs after the compilation has been finished and sealed, but before `shouldEmit` and asset emission:

```text
afterSeal
afterCompile
shouldEmit
emit
afterEmit
done
```

It also differs from the Rspack documentation, which currently describes `afterCompile` as running before the seal phase:

https://rspack.rs/api/plugin-api/compiler-hooks#aftercompile

The cause appears to be that the JavaScript `Compiler.compile()` calls the native `#build()` first and invokes `afterCompile` only after the native build callback completes:

https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/Compiler.ts#L792-L814

The native build runs `compile_done()` immediately after `compile()`, and `compile_done()` executes `shouldRecord`, `shouldEmit`, and asset emission:

https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_core/src/compiler/mod.rs#L326-L404

This causes webpack-compatible plugins that add errors in `afterCompile` to add them too late for `optimization.emitOnErrors: false`. The compilation eventually fails and reports the errors, but invalid assets have already been written.

One example is `ts-checker-rspack-plugin`, which waits for synchronous type-checking results in `afterCompile`. In production, Rspack emits the bundle before those type errors are added, so `NoEmitOnErrorsPlugin` cannot prevent the emit.

The expected behavior is for `afterCompile` to run after seal and before `shouldRecord` / `shouldEmit`, matching webpack. A possible implementation direction is:

1. Add a native `CompilerAfterCompile` hook.
2. Invoke it between `self.compile().await?` and `self.compile_done().await?`.
3. Bridge it to `compiler.hooks.afterCompile` in the JS hooks adapter, similarly to `finishMake` and `afterEmit`.
4. Remove the post-`#build()` `afterCompile.callAsync()` from `Compiler.ts`.
5. Update the documentation to say that `afterCompile` runs after seal and before `shouldEmit`.

Plugins relying on the current undocumented post-emit timing can use `afterEmit` instead.

### Reproduce link

https://github.com/rstackjs/rsbuild-plugin-type-check/issues/104

### Reproduce Steps

Create `src/index.js`:

```js
console.log('hello');
```

Create `rspack.config.js`:

```js
const path = require('node:path');

class AfterCompileErrorPlugin {
  apply(compiler) {
    compiler.hooks.shouldEmit.tap('AfterCompileErrorPlugin', (compilation) => {
      console.log(`shouldEmit: errors=${compilation.errors.length}`);
    });

    compiler.hooks.emit.tap('AfterCompileErrorPlugin', () => {
      console.log('emit');
    });

    compiler.hooks.afterEmit.tap('AfterCompileErrorPlugin', () => {
      console.log('afterEmit');
    });

    compiler.hooks.afterCompile.tap('AfterCompileErrorPlugin', (compilation) => {
      console.log('afterCompile');
      compilation.errors.push(new Error('Error added in afterCompile'));
    });
  }
}

module.exports = {
  mode: 'production',
  entry: path.resolve(__dirname, 'src/index.js'),
  output: {
    path: path.resolve(__dirname, 'dist'),
  },
  optimization: {
    emitOnErrors: false,
  },
  plugins: [new AfterCompileErrorPlugin()],
};
```

Run:

```bash
npx rspack build
```

Rspack logs:

```text
shouldEmit: errors=0
emit
afterEmit
afterCompile
```

The build reports `Error added in afterCompile` and exits with an error, but `dist/main.js` has already been written. With webpack-compatible ordering, `afterCompile` would add the error before `shouldEmit`, and `optimization.emitOnErrors: false` would skip asset emission.

kakiuwang-ui:
## Summary

`compiler.hooks.afterCompile` ran after `shouldEmit`, `emit` and `afterEmit`, so a plugin tapping it was always called once the assets had already been written:

```text
afterSeal → shouldEmit → emit → afterEmit → afterCompile → done
```

webpack calls it right after the compilation is sealed and before it decides whether to emit, which is what lets a plugin fail the build without leaving stale output behind:

```text
afterSeal → afterCompile → shouldEmit → emit → afterEmit → done
```

The cause is that the seal *and* emit sequence both live on the Rust side inside `instance.build()`, while `afterCompile` was called from JS once that native call had already returned — so it could only ever run last.

This adds a `CompilerAfterCompile` hook to `rspack_core`, called between `compile()` and `compile_done()` in both `build_inner` and `rebuild_inner`, and exposes it to JS through the `RegisterJsTaps` interceptor the same way `finishMake` and `shouldEmit` are already wired. `Compiler#compile` no longer calls the hook itself.

`compilation.startTime` / `endTime` keep being assigned where they were, after the native build returns. That matches webpack, which also sets them in `onCompiled`, after `afterCompile` has run.

### Test fallout worth calling out

Errors constructed inside an `afterCompile` tap now carry the tap call stack used by every other natively driven hook (`QueriedHook.promise` instead of `AsyncSeriesHook.callAsync`). The `errorCases` snapshots that record that stack are updated — the only difference in all of them is the `stack` string.

`errorCases/error-read.js` shared one module-level `errors` variable between two cases that also had identical descriptions. `errorCases` run with `concurrent: true`, so the two cases race once the hook resolves at a different point. Each case now gets its own variable and a distinct description.

### Docs

The hook's docs said it was "called after the make phase and before the seal phase", which described neither the old nor the new behavior — the lifecycle diagram on the same page already showed the webpack order. Corrected in both languages.

## Related links

- Closes #15313
- Downstream report: https://github.com/rstackjs/rsbuild-plugin-type-check/issues/104
- Repro from the issue thread: https://github.com/dhouck/rspack-plugin-hook-order-repro

## Checklist


kakiuwang-ui:
The CodSpeed run flags `rust@create_module_hashes` at -3.58%. I don't think this change can reach that benchmark, so rather than leave the red check unexplained:

The timed region calls `compute_module_hashes` directly and never goes through `Compiler::build()` — `xtask/benchmark/benches/groups/compilation_stages.rs#L507`:

```rust
c.bench_function("rust@create_module_hashes", |b| {
  b.iter_batched_ref(
    || { let _compiler = compiler.borrow(); },
    |_| {
      let compiler = compiler.borrow();
      let computed = rt.block_on(async {
        compute_module_hashes(&compiler.compilation).await.unwrap()
      });
      black_box(computed);
    },
    BatchSize::PerIteration,
  );
});
```

The new `after_compile` call sits in `build_inner` / `rebuild_inner`, after `compile()` returns, so it never executes inside the measured closure. The setup (`prepare_for_module_hashes`) runs outside `bench_function` either way.

The same report also shows `rust@create_chunk_assets` moving 2.42% in the *other* direction, and that one is equally outside this change's path. Two unrelated benchmarks drifting by a similar magnitude in opposite directions reads like a codegen layout shift: this PR adds a field to `CompilerHooks` and one more `define_hook!` expansion, which is enough to perturb inlining and symbol placement in unrelated hot code.

Happy to re-run the benchmarks if that would help, or to take another look if you read it differently.

LingyuCoder:

LingyuCoder:
Thanks for the fix. The relevant order is:

```text
Rspack main:
afterSeal → shouldRecord/record → shouldEmit → emit → afterEmit → afterCompile

webpack:
shouldRecord/record → afterSeal → afterCompile → shouldEmit → emit → afterEmit

This PR:
afterSeal → afterCompile → shouldRecord/record → shouldEmit → emit → afterEmit
```

`shouldRecord`/`record` still differs from webpack, but Rspack does not expose these stages to JavaScript yet. The JS-visible hook order is now aligned with webpack, so I think this PR can be merged. We can further align the recording phase if these hooks are exposed in the future.

LingyuCoder (approved):

SoonIter:
<agent: daily-job rstack ecosystem-ci>

### Ecosystem CI regression: `rspack/rsdoctor`

**Attribution:** Confirmed source

**Impact:** Rsdoctor's child-compiler E2E report no longer contains the emitted `child-demo.js` asset, so `pnpm run test:all` fails consistently.

#### Why this PR

This PR correctly moves `afterCompile` before asset emission, but Rsdoctor's child-compiler integration currently treats `afterCompile` as its finalization point. Rspack's native Rsdoctor plugin sends the asset patch from `afterProcessAssets` through an unawaited `tokio::spawn`; with the earlier `afterCompile`, the child report can be finalized before that patch reaches JavaScript. This exposes a synchronization gap in the Rspack/Rsdoctor integration rather than a problem with PR #15276, which is only the visible status-row pivot.

#### Suggested next step

Rspack/Rsdoctor integration owners should make the native asset-patch delivery awaitable before child finalization. A fallback is to add an explicit readiness barrier or move child finalization to a later reliable hook. Verify with Rsdoctor's focused `e2e/cases/doctor-rspack/child-compiler.test.ts` and `pnpm test --stack rspack rsdoctor`.

<details>
<summary>Evidence and attribution checks</summary>

- Current run: https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/32977660229
- Tested upstream: `rspack` at `9e195410dc0da1316d627c14f61aa55de9da0dde`
- First bad / previous good: first visible red https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/32955670336 versus previous green https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/32948077897
- Failure signature: `child-compiler.test.ts:80`, expected child `chunkGraph.assets` to contain `child-demo.js`, received `false`
- Flaky/pre-existing check: the same Rsdoctor SHA `904ea04d` passed before `27c13515`; six consecutive later Rspack jobs reproduce the same signature
- Config check: not applicable
- Alternatives ruled out: #15276 only clears persistent-cache issuer bookkeeping; downstream SHA was unchanged across the green/red pivot; `27c13515` is the intervening commit and directly changes the hook used by Rsdoctor child finalization

</details>

kakiuwang-ui:
Thanks for the detailed attribution @SoonIter — the analysis matches what I see in the code, and I agree this is a synchronization gap in the Rsdoctor integration rather than a wrong hook order. Some extra detail that may help whoever picks this up.

**Confirming the mechanism.** `after_process_assets` builds the patch and then fires it without awaiting (`crates/rspack_plugin_rsdoctor/src/plugin.rs:627`):

```rust
tokio::spawn(async move {
  match hooks.borrow().assets.call(&mut RsdoctorAssetPatch { .. }).await {
    Ok(_) => {}
    Err(e) => panic!("rsdoctor send assets failed: {e}"),
  };
});

Ok(())
```

The hook returns immediately, so nothing orders the JS-side arrival of the patch against anything downstream. Before #15326, `afterCompile` happened after emission and the spawned task had effectively always landed by then; that was timing, not a guarantee.

**It is not only the assets patch.** Every patch delivery in the plugin uses the same fire-and-forget shape, and none of them retains a `JoinHandle`:

| line | hook | patch |
| --- | --- | --- |
| 303 | `CompilationOptimizeChunks` | `chunk_graph` |
| 486 | `CompilationOptimizeChunkModules` | `module_graph` |
| 533 | `CompilationModuleIds` | `module_ids` |
| 580 | `CompilationAfterCodeGeneration` | `module_sources` |
| 627 | `CompilationAfterProcessAssets` | `assets` |

So the hook-order change exposed the `assets` one first, but the other four have the same latent race — they are just further from the finalization point today.

**On the direction.** Of the two options in your report, I would favour making delivery awaitable over moving child finalization to a later hook. Moving finalization would restore the previous timing without restoring a guarantee, and would leave the other four sites exposed to any future reordering. Collecting the `JoinHandle`s into the existing compilation-scoped state (the plugin already keys maps by `compilation.id()`) and draining them at a barrier before finalization would fix all five at once.

I am happy to prepare that patch if the Rsdoctor integration owners would like — I have the context from #15326 and can verify against `e2e/cases/doctor-rspack/child-compiler.test.ts`. Equally happy to leave it to whoever owns this area; just let me know. If it turns out to be more involved than expected and this is blocking the ecosystem CI, I have no objection to #15326 being reverted in the meantime — the hook-order alignment can land again behind the fix.

For reference, the JS-visible ordering that #15326 established (from @LingyuCoder's review):

```text
before:  afterSeal → shouldRecord/record → shouldEmit → emit → afterEmit → afterCompile
webpack: shouldRecord/record → afterSeal → afterCompile → shouldEmit → emit → afterEmit
after:   afterSeal → afterCompile → shouldRecord/record → shouldEmit → emit → afterEmit
```

kakiuwang-ui:
@SoonIter Picked this up — #15399 fixes it from the Rsdoctor plugin side.

Rather than restore the previous `afterCompile` position (which restores the timing but not the guarantee), it collects the `JoinHandle` of all five patch sends in per-compilation state and drains them in a new `after_seal` tap. `AfterSealPass` runs immediately after `AfterProcessAssetsPass` and before the compiler-level `afterCompile`, so every patch is delivered before anything downstream can finalize a report. The sends still run concurrently; only the end of sealing waits.

There is a regression case in `configCases/rsdoctor/patch-delivery-barrier` — without the fix it fails with none of the five patches delivered by `afterCompile`, not just the assets one.

Happy to run the Rsdoctor E2E against it if someone can trigger the ecosystem-ci job on that branch.

PR #15337, opened 2026-08-26
feat: add bundling performance hints

JSerFeng:
## Summary

- Add opt-in diagnostics for async chunk waterfalls, embedded production source maps, large inlined assets, and top-level `this` in ES modules. Individual options override `performance.all`; all four checks are disabled by default and respect `performance.hints`.
- Preserve deeper async paths through shared chunk groups, propagate them to descendants, and avoid following cycles or reporting the same leaf twice. Treat an omitted `mode` as production for embedded source-map hints, matching the existing defaults.
- Estimate inlined assets from their generated source types, including data-URL expansion. Keep the existing lazy-loading recommendation tied to asset or entrypoint size limits.
- Add config cases for warning and error modes, individual and disabled options, source-map guards, shared and cyclic async paths, and inline asset boundaries, plus defaults snapshots and English and Chinese documentation.

## Validation

- `pnpm run build:cli:dev`
- Bundling-hint config cases: 14 passed across normal and runtime modes. Regression cases reproduce the omitted-mode and shared-path omissions before the fix and pass afterward.
- Defaults cases: 38 passed.
- `pnpm run test:type`
- `pnpm run lint:js`
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --locked -- --deny warnings`

## Related links

- Upstream reference: https://github.com/webpack/webpack/pull/21821

## Checklist


JSerFeng:
@codex review

PR #15342, opened 2026-08-26
fix(core): bound concurrent asset writes during emit

### LINKED ISSUE #15340: [Bug]: CopyRspackPlugin exhausts file descriptors on large copy patterns
### System Info

System:
  OS: Windows 11 10.0.26100
  CPU: (8) x64 Apple Silicon
  Memory: 7.90 GB / 15.99 GB
Binaries:
  Node: 24.19.0
  npm: 11.17.0
Browsers:
  Chrome: 152.0.7977.64
  Edge: Chromium (151.0.4129.107)
npmPackages:
  @rspack/core: 2.2.0 => 2.2.0

### Details

Hello! We have encountered an issue where copy-rspack-plugin emits 0KB files. Triaging showed that the issue is Windows-specific and is connected to the amount of availaible file descriptors. 

I made a synthetic reproduction project. 

### Reproduce link

https://github.com/exec-astraea/rspack-copy-emfile-windows

### Reproduce Steps

 git clone https://github.com/exec-astraea/rspack-copy-emfile-windows
 cd rspack-copy-emfile-windows
 npm install
 node synth.mjs --rounds 3

 synth.mjs generates the asset tree (default 12000 files × 1400 B, 16 MiB) so nothing large is committed, runs a build with a single copy pattern, and reports per round whether emit threw EMFILE or silently produced 0-byte files.

 Expected on 2.2.0 — EMFILE rounds: 3/3:

 rspack 2.2.0 | node v24.19.0 | win32/x64
 synthetic: 12000 files x 1400B = 16.0 MiB | rounds: 3

 round  1   10709ms  EMFILE
          × Rspack FS Error: IO error: Error: EMFILE: too many open files, open '...'
 round  2   10367ms  EMFILE
 round  3   10289ms  EMFILE

 EMFILE rounds: 3/3   0-byte rounds: 0/3

 Negative control just under the ceiling — expected clean, which isolates file count as the trigger:

 node synth.mjs --files 8192 --rounds 3

 To see the pre-2.1.6 silent-truncation behaviour on the same fixture, point it at an older core:

 RSPACK_CORE_PATH="/path/to/1.7.11/@rspack/core/dist/index.js" node synth.mjs --rounds 3

Use --files 30000 if you want every round to fail regardless of machine.

lazerg:
## Summary

A build that emits a lot of files can run out of file descriptors while writing them. The reporter hit it with `CopyRspackPlugin` copying 12000 files on Windows, where emit either failed with `Rspack FS Error: IO error: Error: EMFILE: too many open files` or left 0-byte files behind.

`Compiler::emit_assets` spawns one task per asset and starts all of them at once, so every asset's write is handed to the output file system simultaneously. The output file system is always the Node one, so each of those becomes an `fs.writeFile` and the descriptors pile up faster than libuv closes them. The copy plugin itself is not the source: its reads go through blocking `std::fs::read` on tokio workers, so they are already capped by the worker count. Copying is just an easy way to end up with that many assets.

Each emit task now takes a semaphore permit before it writes, with 15 permits, the same limit webpack uses in `Compiler.emitAssets`. Tasks above the limit wait instead of opening a file.

I reproduced it on macOS by lowering the hard descriptor limit (`ulimit -Hn 200`), since Node raises the soft limit on its own at startup. Copying 12000 files failed on every run before the change and passes after it. Peak concurrent `writeFile` calls for a 3000-asset build go from 1184 to 15.

## Related links

Fixes #15340

## Checklist

### DIFF
--- a/crates/rspack_core/src/compiler/mod.rs
+++ b/crates/rspack_core/src/compiler/mod.rs
@@ -31,6 +31,9 @@ use crate::{
   trim_dir,
 };
 
+// keeps the number of files open at once bounded, same limit webpack emits assets with
+const EMIT_ASSETS_CONCURRENCY_LIMIT: usize = 15;
+
 // should be SyncHook, but rspack need call js hook
 define_hook!(CompilerThisCompilation: Series(compilation: &mut Compilation, params: &mut CompilationParams));
 // should be SyncHook, but rspack need call js hook
@@ -447,7 +450,7 @@ impl Compiler {
       .incremental
       .passes_enabled(IncrementalPasses::EMIT_ASSETS);
 
-    let emit_results = rspack_parallel::scope(|token| {
+    let emit_results = rspack_parallel::scope_with_limit(EMIT_ASSETS_CONCURRENCY_LIMIT, |token| {
       self
         .compilation
         .assets()
--- a/crates/rspack_core/src/compiler/mod.rs
+++ b/crates/rspack_core/src/compiler/mod.rs
@@ -31,9 +31,6 @@ use crate::{
   trim_dir,
 };
 
-// keeps the number of files open at once bounded, same limit webpack emits assets with
-const EMIT_ASSETS_CONCURRENCY_LIMIT: usize = 15;
-
 // should be SyncHook, but rspack need call js hook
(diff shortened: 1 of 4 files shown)


LingyuCoder (approved):
LGTM, thanks

PR #15351, opened 2026-08-26
feat(dll): expose DelegatedPlugin and LibManifestPlugin

kakiuwang-ui:
## Summary

`DelegatedPlugin` and `LibManifestPlugin` are both listed under *Plugin compatibility gaps* in the webpack-compatibility tracking issue. Neither needed new bundler behavior — the functionality already exists, it was just not reachable from JavaScript:

- **`LibManifestPlugin`** is fully wired (Rust plugin, napi binding, `builtin-plugin` wrapper) and is already used internally by `DllPlugin`. It was simply never re-exported from `@rspack/core`.
- **`DelegatedPlugin`** is fully implemented in `rspack_plugin_dll` (`DllReferenceAgencyPlugin` applies it natively), but had no `BuiltinPluginName` entry, so there was no way to construct it from JS.

Three webpack test cases for these plugins are already checked into the repo and were skipped via `test.filter.js` (added in #12412 to document the gap). This PR un-skips all three.

### Changes

- add `RawDelegatedPluginOptions` and register `BuiltinPluginName::DelegatedPlugin`
- add the `DelegatedPlugin` JS wrapper; it supplies `compilationContext` from `compiler.options.context`, mirroring how `DllReferenceAgencyPlugin` derives it from `ctx.compiler_options.context` natively (`BuiltinPlugin::append_to` has no access to compiler options, so this is resolved on the JS side via the existing `this: Compiler` binding in `create()`)
- export `DelegatedPlugin` and `LibManifestPlugin` from `@rspack/core`
- un-skip `configCases/delegated/simple`, `configCases/delegated-hash/simple`, `configCases/plugins/lib-manifest-plugin`
- mark both plugins as fully supported in the plugin support table

### Note on the extra commit scope

`delegated-hash/simple` also asserts the webpack deprecation warning for `optimization.moduleIds: "hashed"`, which Rspack did not emit. To let that case run, this PR ports webpack's `WarnDeprecatedOptionPlugin` and applies it in the `hashed` branch of `RspackOptionsApply`. As a side effect `configCases/optimization/hashed-module-ids` now emits that warning too, hence its new `warnings.js`.

This is slightly beyond "expose the two plugins" — happy to split it out if you would prefer. Without it, `delegated-hash/simple` has to stay skipped for an unrelated reason.

## Related links

- Tracking issue: #14556 (*`DelegatedPlugin` and `LibManifestPlugin` JavaScript API exposure*)
- The skips being removed were introduced in #12412

## Checklist


LingyuCoder:
Thanks for the contribution and for taking the time to investigate these webpack compatibility gaps!

After reviewing the current use cases, we have not found a concrete scenario that requires users to instantiate `DelegatedPlugin` or `LibManifestPlugin` directly. Both plugins are already used internally by the higher-level DLL plugins, while the direct usages we could identify are limited to compatibility tests.

We would like to remain conservative about expanding the public API surface, since every exposed API becomes a long-term compatibility commitment. For this reason, we will close this PR for now.

This does not diminish the value or quality of the work here. If a concrete user or ecosystem use case emerges in the future, we would be happy to reopen this PR and continue from this implementation.

Thanks again for your contribution and understanding!

PR #15353, opened 2026-08-27
refactor(loader): schedule parallel workers with Rust MPMC

intellild:
## Summary

- replace Tinypool's JavaScript-side task scheduling with one process-wide, unbounded Rust MPMC dispatcher
- expose generic `dispatchWorkerTask` / `recvWorkerTask` primitives outside the JS loader module
- let every JS worker run its own receive loop: receive one native task, execute it locally, complete or fail its native handle, then receive the next task
- keep MessagePorts per worker instead of allocating async/sync channels for every loader invocation
- remove worker registration, consumer bookkeeping, pool IDs, callback retention, and N-API cleanup hooks
- keep the task envelope type-agnostic for the current single loader task; it can evolve to an enum and JS-side task-type dispatch when a second worker task is introduced

A received `WorkerTask` directly owns the Rust `WorkerJob`. Calling `complete` or `fail` resolves the original dispatcher waiter. If the handle is dropped without finishing, `WorkerJob::drop` reports failure automatically, so no separate task registry or worker teardown protocol is needed.

This PR intentionally keeps the existing `getWorkerLoaderContext()` DTO and main-thread RPC behavior. It does not include the native-backed `JsLoaderContext`, `NormalModule` ownership/accessor, or loader-options refactor. Removing that DTO serialization and its copies belongs in the follow-up loader-context PR.

Example: with `babel-loader` configured as `parallel: { maxWorkers: 8 }`, Rust enqueues serialized loader jobs into one unbounded MPMC channel. All eight workers await `recvWorkerTask()`; whichever worker receives a job runs the loader, calls `WorkerTask.complete()`, and immediately waits for its next job. APIs that require the main thread continue through that worker's persistent RPC ports.

## Related links

N/A

## Validation

- `pnpm run build:cli:dev`
- `pnpm run lint:js`
- `cargo fmt --all --check`
- `cargo clippy -p rspack_tasks -p rspack_napi -p rspack_binding_api --all-targets -- -D warnings`
- `cd tests/rspack-test && pnpm run test -t "configCases/loader-parallel"` (52 passed)

## Checklist

- [x] Tests updated (existing loader-parallel coverage is sufficient).
- [x] Documentation updated (not required; no public configuration change).

---
_by OpenAI Codex_


intellild:
duplicated

PR #15355, opened 2026-08-27
perf(bench): collect mimalloc memory in benchmark setup

stormslowly:
## Why

Between two measured iterations a benchmark leaves the allocator dirty: thread-local free lists and retired pages are laid out according to the *previous* iteration's allocation and free order. The next iteration then walks a different allocator fast path, which was the suspected source of run-to-run variance.

This calls `mi_collect(true)` in every `iter_batched` / `iter_batched_ref` setup closure, so each measured iteration starts from a comparable heap state.

Setup is outside the measured region — `codspeed-criterion-compat` only wraps the routine with `start_benchmark()` / `end_benchmark()` (`src/compat/bencher.rs:127-139`), and criterion's walltime bencher excludes setup from timing — so the collect cost is not counted.

## What

- `rspack_benchmark::collect_memory()` — thin wrapper over `libmimalloc_sys::mi_collect(true)`, no-op on wasm.
- `libmimalloc-sys` added with the `extended` feature (that feature is what exposes the `mi_collect` binding); feature set kept aligned with the existing `mimalloc` dependency.
- Call inserted as the first statement of all 21 setup closures across `compilation_stages`, `build_chunk_graph`, `bundle`, `persistent_cache`, `scan_dependencies` and `walltime`.

## Measured effect: no stability gain

Ran the simulation benchmark 5× on this branch and 5× on its base (`faba242cd6`) as independent CI runs, then read the per-benchmark instruction counts straight out of the uploaded callgrind dumps (50 benchmarks, matching the 50 CodSpeed reports).

| | bit-identical across the 5 runs | CV median | CV mean | CV max |
|---|---|---|---|---|
| with `mi_collect` | 16/50 | 0.0079% | 0.133% | 1.21% |
| base | 16/50 | 0.0065% | 0.134% | 1.93% |

Paired per benchmark: with-collect is tighter on 18, base is tighter on 16, 16 ties. Restricted to the 25 benchmarks that are not already deterministic (`max CV > 0.01%`): 12 vs 13. That is a coin flip, so allocator carry-over is not what drives the residual noise.

The noise that remains is concentrated in the small `rust@<stage>` benchmarks (a few million Ir); every `sources@*`, `bundle@*` and `build_*` benchmark is already deterministic. `rust@create_chunk_assets` is the worst case in both arms (1.21% / 1.93%), which is the benchmark previously traced to glibc `memcpy` ifunc dispatch — something `mi_collect` cannot address.

One real effect did show up: collecting shifts the measured level slightly **down** (22 of the 28 benchmarks that moved by more than 0.01%, up to −0.78%, mean −0.086%). That is a one-off bias absorbed at the next baseline, not a variance reduction.

Wall-clock cost is nil: the `Run benchmark (simulation)` step took 727s here versus a 759s median over the last 22 main runs on the same runner type.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,6 +1217,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "01334b89b69ff726750c5ce5073fc8bd860e99aa9a8fc5ca11b04730e3aee97a"
 
 [[package]]
+name = "cty"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b365fabc795046672053e29c954733ec3b05e4be654ab130fe8f1f94d7051f35"
+
+[[package]]
 name = "dashmap"
 version = "6.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2592,6 +2598,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "667f4fec20f29dfc6bc7357c582d91796c169ad7e2fce709468aefeb2c099870"
 dependencies = [
  "cc",
+ "cty",
  "libc",
 ]
 
@@ -3957,6 +3964,7 @@ version = "0.102.0"
 dependencies = [
  "async-trait",
  "codspeed-criterion-compat",
+ "libmimalloc-sys",
  "mimalloc",
  "rayon",
  "rspack",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -63,6 +63,7 @@ itoa                = { version = "1.0.18", default-features = false }
 json                = { version = "0.12.4", default-features = false }
 json-escape-simd    = { version = "3.1.1", default-features = false }
 json-strip-comments = { version = "3.1.1", default-features = false }
+libmimalloc-sys     = { version = "0.1.44", default-features = false }
 lightningcss        = { version = "1.0.0-alpha.71", default-features = false, features = ["serde"] }
 linked-hash-map     = { version = "0.5.6", default-features = false }
 md4                 = { version = "0.10.2", default-features = false }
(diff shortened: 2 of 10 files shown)


stormslowly:
Closing: measured, and it does not improve stability.

I ran the simulation benchmark 5× on this branch and 5× on its base (`faba242cd6`) as independent CI runs, and read the per-benchmark instruction counts straight out of the uploaded callgrind dumps (50 benchmarks, matching the 50 CodSpeed reports).

| | bit-identical across the 5 runs | CV median | CV mean | CV max |
|---|---|---|---|---|
| with `mi_collect` | 16/50 | 0.0079% | 0.133% | 1.21% |
| base | 16/50 | 0.0065% | 0.134% | 1.93% |

Paired per benchmark: with-collect is tighter on 18, base is tighter on 16, 16 ties. Restricted to the 25 benchmarks that are not already deterministic (`max CV > 0.01%`), it is 12 vs 13 — a coin flip. Allocator carry-over between iterations is not what drives the residual noise.

Where the noise actually lives: entirely in the small `rust@<stage>` benchmarks of a few million Ir. Every `sources@*`, `bundle@*` and `build_*` benchmark is already bit-identical across runs. The worst case in both arms is `rust@create_chunk_assets` (1.21% / 1.93%), the benchmark previously traced to glibc `memcpy` ifunc dispatch — which `mi_collect` cannot address, and which CI already pins as far as it can via `GLIBC_TUNABLES`.

The one real effect is a small systematic shift, not a variance reduction: of the 28 benchmarks that moved by more than 0.01%, 22 moved down (max −0.78%, mean −0.086%). That is a one-off bias absorbed at the next baseline.

Cost was not the problem — the `Run benchmark (simulation)` step took 727s here versus a 759s median over the last 22 main runs on the same runner type. But since there is no measurable benefit, this is not worth a new `libmimalloc-sys` dependency and 21 call sites.

PR #15361, opened 2026-08-27
fix: report diagnostics for unsupported context modes

intellild:
## Summary

Report diagnostics for unsupported context modes instead of panicking or silently accepting invalid values.

- implement `FromAstExpr` for `ContextMode` so `import.meta.webpackContext` options use the existing `AstObject` diagnostic path
- reuse the typed AST conversion for `webpackMode` magic comments while preserving the dynamic import-specific allowed modes
- report unsupported evaluated modes passed to `require.context`
- remove the panicking `From<&str> for ContextMode` implementation

For example:

```js
import.meta.webpackContext("./dir", { mode: "lazi" });
require.context("./dir", false, /\.js$/, "lazi");
import(/* webpackMode: "lazi" */ "./module");
```

Previously, the first call could panic, while the other invalid modes silently fell back. After this change:

- `import.meta.webpackContext` emits a warning and falls back to `sync`
- `require.context` emits an unsupported-mode error
- an invalid `webpackMode` emits a magic-comment warning and uses the normal dynamic import fallback

The updated flow for object options is:

```text
AstObject
  extracts mode as ContextMode
       ↓
ContextMode::from_ast_expr
  validates the static string
       ↓
from_ast_object_with_diagnostics
  collects the diagnostic and applies the default
       ↓
Javascript parser
  emits the warning without panicking
```

Tests run:

- `pnpm run test -t "import-meta-context-unsupported-mode"`
- `pnpm run test -t "require-context-unsupported-mode"`
- `pnpm run test -t "magic_comment_warning"`

## Related links

- Supersedes and reuses the applicable changes from #14366.

## Checklist

---
_by OpenAI Codex_


crates/rspack_plugin_javascript/src/parser_plugin/require_context_dependency_parser_plugin.rs
```diff
@@ -34,6 +44,19 @@ impl<'p, 'a> JavascriptParserPlugin<'p, 'a> for RequireContextDependencyParserPl
...
+        let mut error: Error = create_traceable_error(
+          "Unsupported mode".into(),
+          format!(
+            r#"`mode` expected "sync", "eager", "weak", "async-weak", "lazy" or "lazy-once", but received: "{}"."#,
+            mode_expr.string()
+          ),
+          parser.source.to_string(),
+          expr.args[3].expr.span().into(),
+        );
+        error.hide_stack = Some(true);
```
LingyuCoder, line 58: webpack seems set the fallback mode to lazy [here](https://github.com/webpack/webpack/blob/394a05f14acfe5d447783e353d1d150d0957e36b/lib/dependencies/ImportParserPlugin.js#L534-L547)

intellild: `require` is 'sync' by default.

LingyuCoder (approved):

PR #15363, opened 2026-08-27
[POC] perf(loader): dispatch parallel loaders from Rust

intellild:
## Summary

- Dispatch parallel JavaScript loader tasks directly from the Rust loader runner through a generic unbounded MPMC worker queue.
- Keep persistent JavaScript workers independent from loader-specific registration: each worker receives one native task, executes it, completes it, then calls `recv` again.
- Expose loader context and the canonical `NormalModule` through native-backed accessors so parallel execution does not copy the module or bulk context state through the JavaScript main thread.
- Keep `Rule.use.options` in a compiler-owned map on the JavaScript main thread. Native code carries only a `u32` handle; a worker requests the value over its existing `MessagePort`, ordinary values use structured clone, and functions continue to use the synchronous `SharedArrayBuffer + Atomics` bridge.

Execution flow:

```text
Rust loader runner
  enqueue native task
       ↓
generic unbounded MPMC queue
       ↓
persistent JS worker recv loop
       ↓
native-backed loader context accessors
       ↓
run pitch/normal loader and complete native task
```

Loader options flow:

```text
Rule.use.options
  stored in main-isolate compiler map
       ↓ u32 handle only
Rust loader metadata / worker task
       ↓
worker requests handle over MessagePort
       ↓
structured clone ordinary values
function handles synchronously call the main isolate
```

For example, a Babel loader cache hit no longer requires the main JavaScript thread to materialize and forward the parallel loader payload or serialize the options object through V8. A non-parallel loader still receives the original options object identity, while a parallel loader receives a worker-local structured clone with `Map`, typed arrays, URLs, and function options preserved by the bridge.

## Related links

- Follow-up implementation plan is included in `task.md`.

## Checklist

- [x] Tests updated.
- [x] Documentation updated.

Validation:

- `cargo check -p rspack_binding_api -p rspack_core -p rspack_loader_runner -p rspack`
- `cargo test -p rspack --test loader --no-run`
- `pnpm run build:js`
- `pnpm exec rs check --type-check`
- `pnpm run test -t "configCases/loader-parallel"` (62 tests passed)

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15364, opened 2026-08-27
fix(css): ignore exportsOnly in local ident hashes

### LINKED ISSUE #14786: [Bug]: CSS Modules local identifiers differ when `exportsOnly` changes
### System info

- Rspack: 2.1.3
- Node.js: 24.15.0
- Platform: macOS arm64

### Minimal reproduction

Compile the same entry twice with builtin CSS Modules enabled. The only relevant difference between the client and SSR configurations is `generator.exportsOnly`:

```js
const common = {
  experiments: {
    css: true,
  },
  module: {
    rules: [
      {
        test: /\.module\.css$/,
        type: 'css/module',
        generator: {
          localIdentName: '[local]-[hash:base64:6]',
        },
      },
    ],
  },
};

// client build
{
  ...common,
  module: {
    rules: [
      {
        test: /\.module\.css$/,
        type: 'css/module',
        generator: {
          exportsOnly: false,
          localIdentName: '[local]-[hash:base64:6]',
        },
      },
    ],
  },
}

// SSR build
{
  ...common,
  target: 'node',
  module: {
    rules: [
      {
        test: /\.module\.css$/,
        type: 'css/module',
        generator: {
          exportsOnly: true,
          localIdentName: '[local]-[hash:base64:6]',
        },
      },
    ],
  },
}
```

Import a class named `external` from the same `style.module.css` file in both builds.

### Actual behavior

The generated local identifiers differ:

- client (`exportsOnly: false`): `external-WlNYTF`
- SSR (`exportsOnly: true`): `external-UmUtc0`

If `exportsOnly` is forced to `false` in both builds, both sides generate `external-WlNYTF`.

This reproduces with a regular `.module.css` import, so it is not specific to a framework or SFC integration.

### Expected behavior

`exportsOnly` should control whether CSS assets are emitted, but should not affect CSS Modules local identifier generation. The SSR and client compilations must produce identical class names for the same resource and local name.

### Impact

SSR markup contains a class name that does not match the client CSS output. This can cause missing styles and hydration mismatches in applications that use builtin CSS Modules with `exportsOnly: true` for the server build.

---
_by OpenAI Codex_

mattcosta7:
## Summary

When building for SSR with CSS handling, the hashes end up differing because `exportsOnly: true | false` impacts the hash generation. 

fixes https://github.com/web-infra-dev/rspack/issues/14786

- exclude `exportsOnly` from CSS module local identifier hash inputs
- keep `exportsOnly` in generator/runtime hashing where it affects emitted output
- cover web (`exportsOnly: false`) and node (`exportsOnly: true`) builds with matching local-ident snapshots

## Related links

N/A

## Validation

- `cargo check -p rspack_plugin_css`
- verified web/node local-ident snapshots are byte-for-byte identical in both runtime modes
- full binding-backed config test could not run locally because the container repeatedly terminated `rspack_core` compilation with `SIGTERM`, including with one Cargo job and one rustc thread

## Checklist


LingyuCoder:
Thanks. Some test cases are failing due to the hash changes and need to be updated.

intellild:
Please update tests

LingyuCoder (approved):

PR #15372, opened 2026-08-28
feat(core): add compilation.hooks.afterOptimizeChunkIds

### LINKED ISSUE #10015: [Feature]: compilation.hooks.afterOptimizeChunkIds

We use this to generate an extra runtime module to preload all chunks.

https://github.com/awesome-webextension/webpack-target-webextension/blob/8606cbde3b63ac0d90069ca8755790fe6b3ecce0/lib/webpack5/ServiceWorkerPlugin.js#L43

Like webpack

kakiuwang-ui:
## Summary

Adds `compilation.hooks.afterOptimizeChunkIds`, which webpack calls right after chunk ids are assigned. Rspack already runs the chunk id assignment (`ChunkIdsPass`), but exposed no JS hook at that point, so plugins had no place to read the final chunk ids before code generation.

The concrete use case from the issue is [`webpack-target-webextension`](https://github.com/awesome-webextension/webpack-target-webextension/blob/8606cbde3b63ac0d90069ca8755790fe6b3ecce0/lib/webpack5/ServiceWorkerPlugin.js#L43), which taps this hook to emit an extra runtime module that preloads every chunk.

Implementation follows the existing `afterOptimizeModules` wiring end to end:

- `rspack_core`: new `CompilationAfterOptimizeChunkIds` hook, called at the end of `ChunkIdsPass` after `chunk_by_ukey` and the named chunk ids artifact are written back.
- `rspack_binding_api`: new `RegisterJsTapKind::CompilationAfterOptimizeChunkIds`, register/interceptor plumbing, `skip = true` so the native side stays free of the tap when nothing on the JS side uses it.
- `@rspack/core`: `hooks.afterOptimizeChunkIds` as a `SyncHook<[Iterable<Chunk>]>`, called with `compilation.chunks`.

`ChunkIdsPass` runs on every seal (incremental only controls artifact recovery inside the pass), so the hook fires on rebuilds too.

`optimizeChunkIds` is intentionally left out: rspack's chunk id assignment is driven by the internal `chunkIds` hook, so the "after" notification is the part that had no equivalent. Happy to add the rest of the chunk id hook chain in a follow-up if that is wanted.

## Related links

Closes #10015

## Checklist

Test coverage:

- `tests/rspack-test/compilerCases/hooks-after-optimize-chunk-ids.js` — the hook is called once with chunks whose ids are already assigned.
- `tests/rspack-test/configCases/hooks/after-optimize-chunk-ids/` — end-to-end with an async chunk: asserts the hook runs before `processAssets` and that the ids it sees match the ids in the final stats.

Docs: added `afterOptimizeChunkIds` to the compilation hooks reference (en + zh) and moved it out of the "not supported" class in the pipeline diagram.

Locally: full `tests/rspack-test` suite is 9159 passed / 2 failed, both failures being `configCases/url/self`, which fails on this machine regardless of the change because the checkout path contains non-ASCII characters that `new URL(import.meta.url)` percent-encodes.


kakiuwang-ui:
@hardfist could you approve the CI run and take a look when you have a moment? All 36 checks are green.

Quick summary of the shape, since it spans both sides:

- `rspack_core`: new `CompilationAfterOptimizeChunkIds` hook, fired at the end of `ChunkIdsPass` once `chunk_by_ukey` and the named chunk ids artifact are written back.
- `rspack_binding_api`: register/interceptor plumbing following the existing `afterOptimizeModules` wiring, with `skip = true` so the native side stays free of the tap when nothing on the JS side taps it.
- `@rspack/core`: `hooks.afterOptimizeChunkIds` as `SyncHook<[Iterable<Chunk>]>`.

It closes #10015, whose concrete consumer is `webpack-target-webextension` tapping this hook to emit a runtime module that preloads every chunk.

One deliberate scoping decision worth your call: `optimizeChunkIds` is left out, because rspack's chunk id assignment is driven by the internal `chunkIds` hook, so only the "after" notification had no equivalent. Happy to add it too if you would rather land the pair together.

website/docs/en/api/plugin-api/compilation-hooks.mdx
```diff
@@ -564,6 +564,22 @@ Called after the tree optimization, at the beginning of the chunk modules optimi
   </CollapsePanel>
 </Collapse>
 
+## `afterOptimizeChunkIds`
+
```
LingyuCoder, line 569: need to add an ApiMeta tag to this new hook like `<ApiMeta addedVersion="2.2.2" />`

kakiuwang-ui: Added `<ApiMeta addedVersion="2.2.2" />` to the `afterOptimizeChunkIds` section on both the English and Chinese pages (e26e928). Placed it directly under the heading, above the existing `Read-only` badge.

website/docs/en/api/plugin-api/compilation-hooks.mdx
```diff
@@ -564,6 +565,24 @@ Called after the tree optimization, at the beginning of the chunk modules optimi
...
+<ApiMeta addedVersion="2.2.2" />
```
LingyuCoder:
```suggestion
<ApiMeta addedVersion="2.2.3" />
```

website/docs/zh/api/plugin-api/compilation-hooks.mdx
```diff
@@ -565,6 +566,24 @@ type ExecuteModuleContext = {
   </CollapsePanel>
 </Collapse>
 
+## `afterOptimizeChunkIds`
+
+<ApiMeta addedVersion="2.2.2" />
```
LingyuCoder:
```suggestion
<ApiMeta addedVersion="2.2.3" />
```

LingyuCoder (approved):
Thanks

LingyuCoder:
Thanks again for the contribution.

After further team discussion, we decided to revert `compilation.hooks.afterOptimizeChunkIds` in [#15471](https://github.com/web-infra-dev/rspack/pull/15471). The use case in #10015 can already be implemented with `additionalTreeRuntimeRequirements`, while exposing this hook would make the current chunk ID pipeline a public boundary and constrain planned Rust-side refactoring.

```js
compilation.hooks.additionalTreeRuntimeRequirements.tap(NAME, (chunk, set) => {
  const entryPoint = compilation.entrypoints.get(entry)
  const entryChunk = entryPoint?.getEntrypointChunk()
  if (!entryChunk || chunk.id !== entryChunk.id) return

  const ids = new Set([...getReachableChunks(entryPoint)].map(c => c.id))
  set.add(compiler.webpack.RuntimeGlobals.ensureChunkHandlers)
  compilation.addRuntimeModule(entryChunk, new RuntimeModule([...ids]))
})
```

Thank you for your work and understanding.

PR #15375, opened 2026-08-28
fix(binding): preserve function-valued chunk filename templates

SyMind:
## Summary

- Return the original JavaScript filename function from `Chunk.filenameTemplate` while preserving existing string and `undefined` behavior.
- Track filename function references per compiler with weak Rust ownership, cleaning up expired references after compilation and releasing them when the compiler closes.
- Support filename functions originating from entry options, split chunks, `addEntry`, and `addInclude`, with updated types, documentation, and integration coverage.

## Checklist


SyMind:
@codex review

PR #15393, opened 2026-08-29
perf(ids): optimize deterministic module id assignment

LingyuCoder:
## Summary

- precompute the webpack-compatible FNV prefix state in the existing parallel full-module-name pass
- cache module pre-order indices before sorting instead of looking them up from the module graph in every comparison
- reserve the module ID collections and materialize each numeric `ModuleId` only once
- preserve the existing range selection, UTF-16 hashing, salt, collision retry, and deterministic ordering behavior

On the React 10k production fixture with a warm persistent cache:

- `DeterministicModuleIdsPlugin`: 11.905 ms -> 6.490 ms median (45.5% reduction)
- complete `module ids` pass: 12.340 ms -> 7.646 ms median (38.0% reduction)
- all 19 cold-build output file names and SHA-256 hashes are unchanged

## Related links

N/A

## Testing

- `cargo test -p rspack_util number_hash`
- `cargo check -p rspack_util -p rspack_ids`
- `pnpm run test:unit` (9158 passed, 4 skipped; CLI 90 passed, 1 skipped)

## Checklist

### DIFF
--- a/crates/rspack_ids/src/deterministic_module_ids_plugin.rs
+++ b/crates/rspack_ids/src/deterministic_module_ids_plugin.rs
@@ -1,15 +1,17 @@
 use derive_more::Debug;
 use rayon::prelude::*;
 use rspack_core::{
-  ChunkGraph, Compilation, CompilationModuleIds, ModuleIdsArtifact, Plugin,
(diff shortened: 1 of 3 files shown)


chenjiahan (approved):
👍

PR #15399, opened 2026-08-31
fix(rsdoctor): await patch delivery before sealing completes

kakiuwang-ui:
## Summary

Follow-up to the `rspack/rsdoctor` ecosystem-ci regression reported by @SoonIter on #15326 ([comment](https://github.com/web-infra-dev/rspack/pull/15326#issuecomment-5440018521)): Rsdoctor's child-compiler E2E report stopped containing the emitted `child-demo.js` asset.

The hook reorder in #15326 exposed the problem rather than causing it. In `crates/rspack_plugin_rsdoctor/src/plugin.rs` **all five patch deliveries to the JS side are fire-and-forget `tokio::spawn`**: the `JoinHandle` is dropped, nothing awaits the send, and a failure inside the task `panic!`s on a detached task where nobody observes it.

| producing hook | patch |
| --- | --- |
| `CompilationOptimizeChunks` (stage 9999) | `hooks.chunk_graph` |
| `CompilationOptimizeChunkModules` (stage 9999) | `hooks.module_graph` |
| `CompilationModuleIds` (stage 9999) | `hooks.module_ids` |
| `CompilationAfterCodeGeneration` (stage 9999) | `hooks.module_sources` |
| `CompilationAfterProcessAssets` (stage 9999) | `hooks.assets` |

Rsdoctor treats `afterCompile` as its finalization point, so moving `afterCompile` earlier meant the asset patch consistently lost a race it had always been able to lose. Restoring the previous ordering would restore the timing but not the guarantee — the other four sends have the same shape and are latent instances of the same bug.

## What this does

Records every spawned send in a per-compilation registry and drains it in a new `CompilationAfterSeal` tap. `AfterSealPass` runs immediately after `AfterProcessAssetsPass` (`crates/rspack_core/src/compilation/run_passes.rs`) — after the last producer and before the compiler-level `afterCompile` — so the report is complete by the time anything downstream reads it.

The sends still run concurrently, so the build does not block on report serialization; only the end of sealing waits. No napi, binding or TS change is needed — `after_seal` already exists as a compilation hook.

One behaviour change worth calling out: an error from a send is now propagated as a build error instead of `panic!`ing on a detached task where it was silently swallowed.

## Test

New `configCases/rsdoctor/patch-delivery-barrier` taps all five JS-side hooks with an async tap that yields before recording delivery, then asserts on `afterCompile` that every patch has arrived.

Verified discriminating — on `main` without the Rust change it fails with none of the five delivered:

```
× AssertionError: expected [] to deeply equal [ 'assets', 'chunkGraph', …(3) ]
```

With the fix, all 10 `configCases/rsdoctor/*` cases pass; the full `rs test` suite is 9158 passed / 2 failed, both being the pre-existing `configCases/url/self` failure caused by the non-ASCII characters in my local checkout path.

I can rebase or revert #15326 instead if you would rather fix this from the other side, but this looked like the more durable place.

### DIFF
--- a/crates/rspack_plugin_rsdoctor/src/plugin.rs
+++ b/crates/rspack_plugin_rsdoctor/src/plugin.rs
@@ -1,18 +1,19 @@
 use std::{
   fmt,
-  sync::{Arc, LazyLock},
+  future::Future,
+  sync::{Arc, LazyLock, Mutex},
 };
 
 use atomic_refcell::AtomicRefCell;
 use rspack_collections::IdentifierMap;
 use rspack_core::{
   BuildModuleGraphArtifact, ChunkGroupUkey, Compilation, CompilationAfterCodeGeneration,
-  CompilationAfterProcessAssets, CompilationId, CompilationModuleIds,
+  CompilationAfterProcessAssets, CompilationAfterSeal, CompilationId, CompilationModuleIds,
   CompilationOptimizeChunkModules, CompilationOptimizeChunks, CompilationOptimizeDependencies,
   CompilationParams, CompilerCompilation, ExportsInfoArtifact, ModuleIdsArtifact,
   OptimizationBailoutItem, Plugin, SideEffectsOptimizeArtifact,
 };
-use rspack_error::{Diagnostic, Result};
+use rspack_error::{Diagnostic, Result, error};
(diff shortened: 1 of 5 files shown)


LingyuCoder (approved):
LGTM, thanks

PR #15305, opened 2026-08-23
feat(core): add compilation.runtimeTemplate

kakiuwang-ui:
## Summary

Adds `compilation.runtimeTemplate`, aligning with webpack's `RuntimeTemplate`.

Plugin authors currently have to hand-roll the environment branches when they emit runtime code (`() => …` vs `function() { … }`, `var [a, b] = c;` vs `var a = c[0];`, `for(const x of y)` vs `y.forEach(…)`, template literal vs string concatenation). `compilation.runtimeTemplate` derives all of that from `output.environment`, so generated code stays valid for the configured target.

The environment dependent branches mirror `RuntimeTemplate` in `crates/rspack_core/src/runtime_template.rs`, so JS-side generated code matches what Rspack's own Rust code generation produces for the same options.

What is available:

- Environment checks: `isIIFE`, `isModule`, `isNeutralPlatform`, `supportsConst`, `supportsArrowFunction`, `supportsAsyncFunction`, `supportsOptionalChaining`, `supportsForOf`, `supportsDestructuring`, `supportsBigIntLiteral`, `supportsDynamicImport`, `supportsEcmaScriptModuleSyntax`, `supportTemplateLiteral`, `supportNodePrefixForCoreModules`
- Code generation: `renderNodePrefixForCoreModule`, `returningFunction`, `basicFunction`, `expressionFunction`, `emptyFunction`, `concatenation`, `destructureArray`, `destructureObject`, `iife`, `forEach`
- Missing module handling: `throwMissingModuleErrorBlock`, `throwMissingModuleErrorFunction`, `missingModule`, `missingModuleStatement`, `missingModulePromise`
- Properties: `compilation`, `outputOptions`, `globalObject`, `contentHashReplacement`

Deliberate omissions, documented on the class and in the docs:

- The methods that render references to other modules (`moduleId`, `moduleRaw`, `moduleExports`, `moduleNamespace`, `moduleNamespacePromise`, `importStatement`, `exportFromImport`, `blockPromise`, `asyncModuleFactory`, `syncModuleFactory`, `weakError`, `runtimeConditionExpression`) need read access to the module graph and the code generation results, which Rspack keeps on the Rust side.
- `comment()` needs `compilation.requestShortener`, which Rspack does not have.
- `defineEsModuleFlagStatement()` needs runtime globals as webpack identifier strings, while Rspack resolves them from a numeric `RuntimeGlobals` enum whose final identifier depends on the runtime mode.

[...]

PR #15400, opened 2026-08-31
feat(javascript): add JavascriptModulesPlugin renderContent hook

### LINKED ISSUE #10013: [Feature]: JavascriptModulesPlugin.getCompilationHooks().renderContent hook

When porting https://github.com/awesome-webextension/webpack-target-webextension/ to rspack, one of features cannot be implemented in rspack because of missing this hook.

We added a hook to wrap some code outside of the render content.

https://github.com/awesome-webextension/webpack-target-webextension/blob/8606cbde3b63ac0d90069ca8755790fe6b3ecce0/lib/webpack5/ServiceWorkerPlugin.js#L74-L89

As what webpack have.

kakiuwang-ui:
## Summary

Closes #10013.

Adds `JavascriptModulesPlugin.getCompilationHooks(compilation).renderContent`, a `SyncWaterfallHook<[Source, RenderContext]>` matching webpack's hook of the same name. It is called with the rendered content of a JavaScript chunk before that chunk's init fragments are injected, which is the only point where a plugin can wrap the chunk content as a whole.

## Why

@Jack-Works filed this while porting [webpack-target-webextension](https://github.com/awesome-webextension/webpack-target-webextension/) to Rspack. Its `ServiceWorkerPlugin` wraps the service-worker entry chunk in a `try`/`catch` so a throwing service worker still lets the console open, and the code [already probes for the hook](https://github.com/awesome-webextension/webpack-target-webextension/blob/8606cbde3b63ac0d90069ca8755790fe6b3ecce0/lib/webpack5/ServiceWorkerPlugin.js#L74-L89):

```js
const hooks = javascript.JavascriptModulesPlugin.getCompilationHooks(compilation)
if (hooks.renderContent) {
  // rspack don't have it
  hooks.renderContent.tap(name, (source, context) => {
    if (entryChunk !== context.chunk) return source
    return new sources.ConcatSource('try {\n', source, '\n} catch (e) { Promise.reject(e) }')
  })
}
```

There is no way to express this with the hooks Rspack exposes today — `processAssets` runs after hashing and would have to re-parse the emitted asset, and none of the render-stage hooks reach JavaScript.

## What changed

- **`crates/rspack_plugin_javascript`** — new `JavascriptModulesRenderContent` hook, called from all three render paths at the position webpack calls it (after the chunk body is assembled / after `renderChunk`, before init fragments are added and before `render`):
  - `render_webpack_main` (`plugin/mod.rs`)
  - `render_rspack_main` (`plugin/runtime_context.rs`)
  - `render_chunk` (`plugin/mod.rs`)
- **`crates/rspack_binding_api`** — `JavascriptModulesRenderContent` tap kind and interceptor. The source crosses the binding through the existing `JsSourceToJs` round-trip (same mechanism `compilation.hooks.runtimeModule` already uses). The register is `skip = true`, so nothing crosses the boundary when no plugin taps the hook.
- **`packages/rspack`** — the `SyncWaterfallHook` on `getCompilationHooks()` and its tap registration. When a tap returns the source it was handed, the JS side returns `undefined` so Rust keeps the original `BoxSource` rather than rebuilding it from its string form.

### Deliberate difference from webpack

`RenderContext` only carries `chunk`. webpack also puts `chunkGraph`, `moduleGraph`, `runtimeTemplate`, `codeGenerationResults` and `runtimeRequirements` on it; in Rspack all of those are reachable from the `compilation` the hooks were retrieved from (`compilation.runtimeTemplate` landed in #15305), so only the per-call value is sent across the binding. Documented in both the en and zh hook pages.

## Test

`configCases/hooks/javascript-modules-render-content` covers both render paths — the runtime chunk and a named async chunk — and asserts three things:

- the returned `Source` is what the chunk is built from (the wrapper is present in the emitted file **and** executes, so it is not merely appended afterwards),
- a pass-through tap that returns its input does not disturb the source the next tap receives,
- `context.chunk` is identity-equal to `compilation.entrypoints.get('main').getEntrypointChunk()`, which is exactly the comparison the consumer plugin makes.

It runs under both `Config` and `RuntimeModeConfig`, so `render_webpack_main` and `render_rspack_main` are both exercised.

Full `rs test`: 44 files passed, 2 failed — both `configCases/url/self`, a pre-existing local failure caused by the non-ASCII characters in my checkout path. `cargo clippy` and `cargo fmt` clean on both touched crates.


website/docs/en/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +23,58 @@ Called when computing the chunk hash for JavaScript chunks.
...
 </Collapse>
+
+## `renderContent`
+
+`SyncWaterfallHook<[Source, RenderContext]>`
+
+Called with the rendered content of a JavaScript chunk, before the chunk's init
+fragments are injected and before the chunk is written out. Taps must return a
+`Source`, which becomes the content the following taps and the rest of the
+rendering pipeline see.
```
LingyuCoder:
```suggestion
Called with the rendered content of a JavaScript chunk, before the chunk's init fragments are injected and before the chunk is written out. Taps must return a `Source`, which becomes the content the following taps and the rest of the rendering pipeline see.
```

website/docs/zh/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +23,55 @@ import { Collapse, CollapsePanel } from '@components/Collapse';
...
+```ts
+type RenderContext = {
+  chunk: Chunk;
+};
+```
+
+webpack 的 render context 上还有 `chunkGraph`、`moduleGraph`、`runtimeTemplate`、
+`codeGenerationResults` 和 `runtimeRequirements`。在 Rspack 中这些都可以从获取 hooks 时用的
+`compilation` 上拿到，因此只有每次调用才会变化的 `chunk` 会跨越 binding 传递。
+
```
LingyuCoder:
```suggestion
```

website/docs/en/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +23,58 @@ Called when computing the chunk hash for JavaScript chunks.
...
+type RenderContext = {
+  chunk: Chunk;
+};
+```
+
+webpack additionally exposes `chunkGraph`, `moduleGraph`, `runtimeTemplate`,
+`codeGenerationResults` and `runtimeRequirements` on the render context. In
+Rspack those are reachable from the `compilation` the hooks were retrieved from,
+so only the per-call `chunk` is carried across the binding.
+
```
LingyuCoder:
```suggestion
```

website/docs/zh/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +23,51 @@ import { Collapse, CollapsePanel } from '@components/Collapse';
     <HashType />
   </CollapsePanel>
 </Collapse>
+
+## `renderContent`
+
```
LingyuCoder, line 29: need to add an ApiMeta tag to this new hook like `<ApiMeta addedVersion="2.2.2" />`

kakiuwang-ui:
Thanks — added `<ApiMeta addedVersion="2.2.2" />` to the `renderContent` section on both the English and Chinese pages, and applied the same one-line reflow you made on the English intro sentence to its Chinese counterpart. Your batched suggestions are in as of 41322ea.

On the reporter side: I asked @Jack-Works on #10013 whether a `RenderContext` carrying only `chunk` covers webpack-target-webextension. He confirmed it works for them — the plugin already feature-detects `hooks.renderContent`, so it picks the hook up automatically once this lands.

website/docs/en/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +24,52 @@ Called when computing the chunk hash for JavaScript chunks.
     <HashType />
   </CollapsePanel>
 </Collapse>
+
+## `renderContent`
+
+<ApiMeta addedVersion="2.2.2" />
```
LingyuCoder:
```suggestion
<ApiMeta addedVersion="2.2.3" />
```

website/docs/zh/api/plugin-api/javascript-modules-plugin-hooks.mdx
```diff
@@ -22,3 +24,52 @@ import { Collapse, CollapsePanel } from '@components/Collapse';
...
+<ApiMeta addedVersion="2.2.2" />
```
LingyuCoder:
```suggestion
<ApiMeta addedVersion="2.2.3" />
```

LingyuCoder (approved):
Thanks

LingyuCoder:
After an in-depth discussion within the team, we’ve decided to revert this hook in [#15471](https://github.com/web-infra-dev/rspack/pull/15471).

The main concern is that this API introduces several architectural issues. For example, the `modern-module` rendering path does not go through this hook, so the API cannot provide consistent behavior across different output modes. In addition, using the hook requires transferring the full chunk source through the Rust–JavaScript binding. This can introduce significant performance overhead when the hook is actively used, especially when multiple chunks are rendered.

Most asset-level transformations already use `processAssets`, and Rspack has implemented a series of optimizations for that path. `processAssets` can also fully satisfy the use case described in the original issue. For example:

```js
compilation.hooks.processAssets.tap(
  {
    name: PLUGIN_NAME,
    stage: compiler.rspack.Compilation.PROCESS_ASSETS_STAGE_ADDITIONS,
  },
  () => {
    const chunk = compilation.entrypoints
      .get(entryName)
      ?.getEntrypointChunk();

    for (const file of chunk?.files ?? []) {
      const asset = compilation.getAsset(file);
      if (asset?.info.javascriptModule === undefined) continue;

      compilation.updateAsset(
        file,
        source =>
          new compiler.rspack.sources.ConcatSource(
            "try {\n",
            source,
            "\n} catch (e) { Promise.reject(e) }",
          ),
      );
    }
  },
);
```

Therefore, we recommend using `processAssets` for this transformation instead of introducing a new rendering hook.

Thank you again for the contribution and for helping us explore this compatibility gap. We really appreciate the work and discussion around this proposal.

PR #15404, opened 2026-08-31
fix: fallback eval devtools for modern-module

JSerFeng:
## Summary

- Reject eval-based devtools when `library.type` is `modern-module` before devtool plugin selection.
- Change `eval` to `false`, and strip only the `eval-` prefix from `eval-*-source-map` so modifiers such as `nosources` and `cheap-module` are preserved.
- Emit an infrastructure warning for each automatic fallback.
- Add ESM output coverage for both fallback paths, including normalized options, warnings, emitted top-level code, and source-map artifacts.

## Related links

N/A

## Testing

- `pnpm run build:js`
- `cd tests/rspack-test && pnpm exec rs test run EsmOutput.test.js` (516 passed)
- `pnpm exec rs check`
- `git diff --check`

## Checklist

### DIFF
--- a/packages/rspack/src/builtin-plugin/EsmLibraryPlugin.ts
+++ b/packages/rspack/src/builtin-plugin/EsmLibraryPlugin.ts
@@ -5,10 +5,11 @@ import type {
   RspackOptionsNormalized,
 } from '../config';
 import WebpackError from '../lib/WebpackError';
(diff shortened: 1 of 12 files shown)


LingyuCoder:
need to handle "eval"

LingyuCoder (approved):

PR #15409, opened 2026-08-31
fix(cache): skip non-serializable entries during filesystem cache writes

### LINKED ISSUE #15406: fix(cache): skip non-serializable entries during filesystem cache writes
## Trigger

This only affects filesystem-backed new_cache entries. It occurs when an entry has been accepted by the in-memory cache, but its deferred encoder later fails during FileCacheStrategy::after_all_stored.

For module cache, a representative trigger is cache: { type: persistent } with experiments.new_cache.module enabled, where a completed module build contains a dependency or other payload that cannot be serialized by the cache codec, for example process-local state or an unsupported dynamic cacheable type. Memory cache hits are unaffected because they do not encode values.

## Current behavior

The idle write batch uses try_for_each over every pending entry. If encoding one entry fails, after_all_stored returns an error before writing the batch. Valid entries in the same batch are not persisted, and the pending batch can be retried and fail again.

Before the typed module-cache refactor, module cache encoded its own value before storing it and skipped that module entry on failure. Moving encoding behind the generic cache boundary removes that module-specific recovery path.

## Expected behavior

FileCacheStrategy should encode pending entries independently: log and discard only an entry whose encoder fails, while continuing to persist the other valid entries, validator data, and metadata. This should be implemented as a generic filesystem-cache behavior, not as a module-cache exception.

lazerg:
## Summary

`FileCacheStrategy::after_all_stored` encodes every pending entry inside a `try_for_each` over the write batch, so one entry whose encoder fails aborts the whole batch: the valid entries, the validator data and the metadata are all dropped, and the next idle store retries the same batch and fails again. Before module cache moved its encoding behind the generic cache boundary it encoded its own value and skipped just that module, and that recovery path went away with the refactor.

Each entry is now encoded independently. A failing encoder is traced with the key that caused it and the entry is dropped; everything else in the batch is written as before. The skipped entries are counted so `after_all_stored` can warn once with the total and report an accurate `Stored cache (N items)`, rather than warning per entry inside the rayon loop. Errors from `batch.put` still propagate, since those are real database failures rather than a single unserializable value.

## Test

`non_encodable_entry_does_not_discard_the_batch` in `crates/rspack_core/src/new_cache/file_cache_strategy.rs` stores one entry with a working encoder, one with an encoder that always errors, and some metadata, then checks the good entry and the metadata are readable from the database afterwards. Without the change it fails on `after_all_stored` returning the encoder error.

`cargo test -p rspack_core --lib` is 120 passed, `cargo clippy -p rspack_core --all-targets` and `cargo fmt --check` are clean.

## Related links

Closes #15406

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4271,6 +4271,7 @@ dependencies = [
  "swc_experimental_ecma_ast",
  "swc_experimental_ecma_parser",
  "swc_experimental_ecma_semantic",
+ "tempfile",
  "tokio",
  "tracing",
  "triomphe",
--- a/crates/rspack_core/Cargo.toml
+++ b/crates/rspack_core/Cargo.toml
@@ -85,6 +85,7 @@ turbo-persistence = { workspace = true }
 
 [dev-dependencies]
 pretty_assertions = { workspace = true }
+tempfile          = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/rspack_core/src/new_cache/file_cache_strategy.rs
+++ b/crates/rspack_core/src/new_cache/file_cache_strategy.rs
@@ -222,16 +222,14 @@ impl FileCacheStrategy {
       let pending_items = entries.len()
         + if validator.is_some() { 1 } else { 0 }
         + if meta.is_some() { 1 } else { 0 };
-      let logger = &self.logger;
-      let skipped_items = &AtomicUsize::new(0);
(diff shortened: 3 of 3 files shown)


hardfist:
Thanks for your contribution, but this is part of the cache refactoring effort and isn’t suitable as an external contribution at this time.

PR #15431, opened 2026-09-02
feat(css): support CSS import and asset URL externals

### LINKED ISSUE #15102: [Feature]: Support externalsType: 'css-import' (and 'css-url') for CSS externals

**There is currently no way to keep a CSS `@import` external in Rspack.** Native CSS handling
resolves and inlines every `css-import` dependency unconditionally, and `externals` cannot
intercept it because it only applies to the JavaScript module graph.

### Use case

We build a React component library with Rslib (Rspack under the hood). It has a stylesheet
entry that depends on a shared design-token package which ships a plain `dist/index.css`:

```css
/* src/style.css — the library's CSS entry */
@import '@acme/tokens';

.button {
  color: var(--acme-color-primary);
}
```

`@acme/tokens` is a **runtime `dependency`** of our library, so every consumer already has it
installed, and many consumers import it directly themselves. We want the published
`dist/style.css` to keep the `@import` so that:

1. The token CSS is not duplicated when the consumer also imports `@acme/tokens`.
   In our real case this is **46 KB of duplicated CSS**.
2. The webfonts referenced by the token package's `@font-face` rules are not re-emitted into
   our package. Today Rspack copies two `.woff2` files (**194 KB**) into our `dist/`, and the
   browser downloads them twice under different hashed URLs because the consumer's own copy of
   `@acme/tokens` emits the same fonts again.
3. The token package can be upgraded by the consumer without republishing our library.

### Why the existing config surfaces don't work

```js
output: {
  externals: [/^@acme\//],  // no effect on CSS
}
```

`externals` never sees the request — the `css-import` dependency is resolved inside the CSS
pipeline. There is no `externalsType` value that maps to a CSS `@import` either, so even a
function-form external has no valid type to return.

The only workaround is a post-build script that strips the import from the source and prepends
`@import '@acme/tokens';` back onto the emitted file. That works, but the build output no
longer corresponds to the source, and it breaks silently whenever the CSS entry changes.

### webpack parity

webpack supports exactly this via `externalsType: 'css-import'`. It is **undocumented on
webpack.js.org**, but it is present in webpack's public types and implementation — the
`ExternalsType` union in `webpack/types.d.ts` (verified in 5.104.0, 5.106.2, 5.107.2 and
5.108.4) reads:

```ts
type ExternalsType =
  | "import"
  | "var"
  // …
  | "node-commonjs"
  | "asset"
  | "asset-url"
  | "css-import"
  | "css-url";
```

The corresponding union in `@rspack/core@1.7.8` (`dist/config/types.d.ts`) has none of the
last four:

```ts
export type ExternalsType = 'var' | 'module' | 'assign' | 'this' | 'window' | 'self' | 'global'
  | 'commonjs' | 'commonjs2' | 'commonjs-module' | 'commonjs-static' | 'amd' | 'amd-require'
  | 'umd' | 'umd2' | 'jsonp' | 'system' | 'promise' | 'import' | 'module-import' | 'script'
  | 'node-commonjs' | 'commonjs-import';
```

This gap is not listed in the webpack-compatibility tracking issue #14556, which does cover
other externals parity items (`{ interop }` external requests, `amd-async`,
`externalsPresets.nodeModules`).

---

Add `'css-import'` (and ideally `'css-url'`, for `url()` references) to `ExternalsType`,
matching webpack's behaviour.

### Config

```js
// rspack.config.js
module.exports = {
  experiments: { css: true },
  externals: [
    ({ request, dependencyType }, callback) => {
      if (dependencyType === 'css-import' && request.startsWith('@acme/')) {
        return callback(null, request, 'css-import');
      }
      callback();
    },
  ],
};
```

The object form should work too, scoped by dependency type:

```js
externals: {
  byLayer: undefined,
  '@acme/tokens': 'css-import @acme/tokens',
}
```

### Input

```css
/* src/style.css */
@import '@acme/tokens';

.button {
  color: var(--acme-color-primary);
}
```

### Current output (Rspack 1.7.8)

```css
/* ~46 KB of @acme/tokens rules inlined here, plus its @font-face rules  */
/* and two .woff2 files emitted into the output directory                */
.button{color:var(--acme-color-primary)}
```

### Expected output with the feature

```css
@import '@acme/tokens';
.button{color:var(--acme-color-primary)}
```

The emitted `@import` keeps its position at the top of the stylesheet (per CSS spec, `@import`
must precede other rules), and no assets from the external package are emitted.

`'css-url'` would behave analogously for `url()` references, leaving the URL untouched instead
of resolving it into an emitted asset.

### Environment

- `@rspack/core` 1.7.8 (via `@rslib/core` 0.20.0)
- Reference implementation: webpack 5.104.0+

intellild:
## Summary

- add `asset`, `asset-url`, and `css-import` to the public `ExternalsType` options
- resolve an asset external as `asset-url` when it is consumed by CSS `url()`, while keeping JavaScript `new URL()` behavior as `asset`
- rename the internal URL source type from `SourceType::CssUrl` to `SourceType::AssetUrl`
- extend `StringEnum` with:
  - a newtype fallback that preserves unknown strings
  - serde-style `rename_all` rules, with `snake_case` as the default and per-variant `rename` taking precedence
- derive the complete `SourceType` string mappings with `rename_all = "kebab-case"`
- preserve external CSS URLs in generated CSS instead of emitting a JavaScript wrapper
- port webpack's `configCases/externals/asset-url` coverage

The deprecated webpack `css-url` alias is intentionally not supported.

## Example

```js
module.exports = {
  externalsType: "asset-url",
  externals: {
    "external-image": "https://example.com/image.png"
  }
};
```

Given:

```css
.logo {
  background: url(external-image);
}
```

Rspack now keeps the external URL in the emitted CSS. Before this fix, the external was treated as a JavaScript source and could not be generated correctly for the CSS consumer.

The new `StringEnum` forms keep mappings concise and custom source types round-trippable:

```rust
#[string_enum(rename_all = "kebab-case")]
enum SourceType {
  #[string_enum(rename = "javascript")]
  JavaScript,
  AssetUrl,
  #[string_enum(fallback)]
  Custom(Ustr),
}
```

Unknown strings are converted to `Custom(value)`, and `as_str()` returns the preserved value.

## Validation

- `pnpm run test -t 'externals/asset-url'`
- `pnpm run test -t 'custom-url'`
- `pnpm run test -t 'css/import'`
- `cargo test -p rspack_macros_test`
- `cargo clippy -p rspack_macros -p rspack_macros_test -p rspack_core -p rspack_plugin_asset -p rspack_plugin_externals --all-targets -- -D warnings`
- webpack gap tracker: no webpack-only files remain for `configCases/css/import` and `configCases/externals/asset-url`

## Checklist

- [x] Tests added or updated
- [x] Relevant tests pass
- [x] Deprecated `css-url` is not exposed

---
_by OpenAI Codex_

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4536,6 +4536,7 @@ dependencies = [
 name = "rspack_macros"
 version = "0.102.2"
 dependencies = [
+ "cow-utils",
  "heck",
  "proc-macro2",
  "quote",
--- a/crates/rspack_core/src/external_module.rs
+++ b/crates/rspack_core/src/external_module.rs
@@ -433,6 +433,13 @@ fn resolve_external_type<'a>(
         "module"
       }
     }
+    "asset" | "asset-url" | "css-url" => {
+      if dependency_meta.source_type == Some(SourceType::CssUrl) {
+        "asset-url"
+      } else {
+        "asset"
+      }
+    }
 
     import_or_module => import_or_module,
   }
@@ -1100,17 +1107,10 @@ impl Module for ExternalModule {
   }
 
   fn source_types(&self, _module_graph: &ModuleGraph) -> &[SourceType] {
-    if self.external_type == "asset"
-      && self
-        .dependency_meta
-        .source_type
-        .is_some_and(|t| t == SourceType::CssUrl)
-    {
-      EXTERNAL_MODULE_CSS_URL_SOURCE_TYPES
-    } else if self.external_type == "css-import" {
-      EXTERNAL_MODULE_CSS_SOURCE_TYPES
-    } else {
-      EXTERNAL_MODULE_JS_SOURCE_TYPES
+    match self.resolve_external_type() {
+      "asset-url" => EXTERNAL_MODULE_CSS_URL_SOURCE_TYPES,
+      "css-import" => EXTERNAL_MODULE_CSS_SOURCE_TYPES,
+      _ => EXTERNAL_MODULE_JS_SOURCE_TYPES,
     }
   }
 
@@ -1228,7 +1228,7 @@ impl Module for ExternalModule {
 
     let mut cgr = CodeGenerationResultBuilder::default();
     let (request, external_type) = self.get_request_and_external_type();
-    match self.external_type.as_str() {
+    match self.resolve_external_type() {
       "asset" if request.is_some() => {
         let request = request.expect("request should be some");
         cgr.add(
@@ -1240,6 +1240,12 @@ impl Module for ExternalModule {
           ))
           .boxed(),
         );
+        cgr
+          .data_mut()
+          .insert(CodeGenerationDataUrl::new(request.primary().to_string()));
+      }
+      "asset-url" if request.is_some() => {
+        let request = request.expect("request should be some");
         cgr
           .data_mut()
           .insert(CodeGenerationDataUrl::new(request.primary().to_string()));
--- a/crates/rspack_core/src/external_module.rs
+++ b/crates/rspack_core/src/external_module.rs
@@ -433,7 +433,7 @@ fn resolve_external_type<'a>(
         "module"
       }
     }
-    "asset" | "asset-url" | "css-url" => {
+    "asset" | "asset-url" => {
       if dependency_meta.source_type == Some(SourceType::CssUrl) {
         "asset-url"
       } else {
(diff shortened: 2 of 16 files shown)


intellild:
@codspeedbot is this regression real ?

intellild:
@codex review

LingyuCoder (approved):

Issue #8055, opened 2024-10-05
Does rspack `normalModuleLoader` plan to align `beforeLoaders` and other hooks with webpack 5?


gaoachao:
### webpack `NormalModuleCompilationHooks`:
![Image](https://github.com/user-attachments/assets/5bbcaf16-3928-4429-acb2-8505019ceb1e)

### rspack js side `NormalModuleHooks`: 
![Image](https://github.com/user-attachments/assets/75537720-1c62-4bd3-bafa-eb1b2f7f4668)

## My Confusion
In addition, I found some code about `before_loaders` on the rust side, but the definition here is a litter different from webpack. Does it only need to be exported on the js side before it can be used? Please forgive me if my understanding is wrong.
![Image](https://github.com/user-attachments/assets/c7552026-e93c-4577-9ac1-2a44f9def62c)

## Background
My webpack project use this hooks to add some custom loader context.
```
compiler.hooks.compilation.tap(PLUGIN_NAME, (compilation) => {
    NormalModule.getCompilationHooks(compilation).beforeLoaders.tap(PLUGIN_NAME, (loaders, mod) => {
```

OskarLebuda:
Is there any possibilities to expose `beforeLoader`?
This is part of https://github.com/Trinovantes/vue-ssr-assets-plugin that I would like to move into Rspack. Without that this cannot be done.

kkaaddff:
Actually, rspack js side 

> NormalModuleBooks: loader

 was not executed correctly.
`class Plugin {
  apply(compiler) {
    compiler.hooks.compilation.tap('Plugin', (compilation) => {
      compiler.webpack.NormalModule.getCompilationHooks(compilation).loader.tap(
        'Plugin',
        (_loaderContext, module) => {
          // There is no execution here
        }
      )
    })
  }
}`

peterczg:
Any Progress?

OskarLebuda:
bump

kakiuwang-ui:
Opened #15433, which exposes `NormalModule.getCompilationHooks(compilation).beforeLoaders` to JS.

The Rust-side `NormalModuleBeforeLoaders` hook you spotted was indeed already there and already called from `NormalModule::build` — it just had no napi interceptor and nothing tapping it, so there was no way to reach it from JS. The PR wires that up, and the webpack test for this hook (`configCases/loaders/pr-14384`, previously skipped with a TODO pointing at this gap) now runs.

[...]

PR #15433, opened 2026-09-02
feat(core): expose NormalModule beforeLoaders hook to JS

kakiuwang-ui:
## Summary

Exposes `NormalModule.getCompilationHooks(compilation).beforeLoaders` to JS, so plugins can adjust a module's loader list right before its loaders run.

`NormalModuleBeforeLoaders` already exists in `rspack_core` and is already called from `NormalModule::build`, but nothing has ever tapped it and there was no napi interceptor, so the hook was unreachable from JS. This PR wires it end to end.

### Rust

The hook becomes `SeriesBail` and returns the loader list the tap wants the module built with:

```rust
define_hook!(NormalModuleBeforeLoaders: SeriesBail(compiler_id: CompilerId, module: &mut NormalModule) -> Vec<BeforeLoadersItem>, tracing=false);
```

- `BeforeLoadersItem::Kept(index)` — an entry the tap left alone. It reuses the already resolved `BoxLoader` and its `LoaderRunnerOptions`, so untouched loaders are not resolved a second time and keep their cache identity.
- `BeforeLoadersItem::Added(ModuleRuleUseLoader)` — an entry the tap added or rewrote. It goes through `resolve_each_with_options`, the same path `NormalModuleFactory` uses for loaders coming from `module.rules`, so `resolveLoader` and `builtin:` loaders work here too.

`BuildContext` gains `loader_resolver_factory` for that resolution — it previously only carried the module resolver factory.

### Binding

`JsBeforeLoadersArgs { loaders: Vec<JsLoaderItem>, module: ModuleObject }` in, `Array<number | JsAddedLoaderItem>` out. Registered with `skip = true`, so no napi crossing happens when nobody taps the hook, and the JS tap returns `undefined` when it changed nothing, which skips the Rust-side rebuild entirely.

### JS

```ts
beforeLoaders: liteTapable.SyncHook<[LoaderItem[], NormalModule]>;
```

`taps/normalModule.ts` converts between the binding's loader request strings and webpack's `{ loader, options, ident, type }` shape, reusing `parseResourceWithoutFragment` for the way in and `createRawModuleRuleUses` for the way back.

One subtlety worth calling out for review: **an ident inherited from the config is dropped when an entry is rewritten.** The ident is the key the rule's own options object is registered under in `compiler.__internal__ruleSet.references`; reusing it for rewritten options would overwrite that rule for *every other module*. The `mutate.js` case in the new test caught exactly this.

### Deliberate difference from webpack

webpack passes a third `loaderContext` argument. Rspack does not, because a module's loader context is only created once the loader runner starts, which is after this hook has run. Documented in the warning box on the new docs page.

## Related links

#8055 — three people asked for this on that issue; @OskarLebuda is blocked on it for moving [vue-ssr-assets-plugin](https://github.com/Trinovantes/vue-ssr-assets-plugin) to Rspack.

`tests/rspack-test/configCases/loaders/pr-14384` is the webpack test for this hook. It was skipped with `test.filter.js` saying *"Rspack does not export `NormalModule` and support `beforeLoaders` from JS side"*; `NormalModule` is exported now, and with this PR the rest of it works, so the filter is removed and the case runs.

## Notes

- I put `<ApiMeta addedVersion="2.2.3" />` on the docs page since the repo is at 2.2.2 and 2.2.3 is what the hooks merged this cycle got. Happy to change it if the target release is different.
- `NormalModule::request` is derived from the loader list at creation time and is not recomputed when a tap changes the loaders. webpack behaves the same way, so I left it alone.

## Checklist


kakiuwang-ui:
Pushed 8e2b7f4 hardening the JS conversion layer. Three defects, the first of which was a correctness bug in exactly the place the original comment warned about:

- **A tap that rebuilds an entry rather than assigning to it** (`loaders[0] = { ...loaders[0], options }`) drops the non-enumerable snapshot the ident guard relied on, but keeps the `ident` — so the guard did not fire and `resolveStringifyLoaders` overwrote `references[ident]`, corrupting the options every other module matched by that rule receives. The set of configured idents is now collected *before* the taps run and checked against that instead.
- **`JsLoaderItem::from(&Arc<dyn Loader>)` derives its fields by splitting the loader identifier on `'|'`**, which drops the loader type entirely for JS loaders (`type` was always `null`, contradicting the docs) and truncates the path of any identifier containing a `'|'`. Rather than change that conversion, which `module.loaders` also depends on, the hook now passes a dedicated `JsBeforeLoadersLoaderItem` built from `Loader::r#type()` and the module's loader options. Generated idents no longer contain a `'|'` either, since module identifiers embed one.
- **`Rule.use[].cache` and `Rule.use[].parallel` were silently dropped** from any loader a tap rewrote. Both are carried over from the snapshot now, without adding either to the public `LoaderItem` type.

The test gained the rebuilt-entry case — which reproduced the first bug — plus a `finishModules` assertion that the rule's options object survived the build, and a `.cjs` loader asserting `type: "commonjs"`.

One thing left as documentation rather than code: mutating `options` **in place** is invisible to this hook. That matches webpack, where the object is shared across every module the rule matched, but Rspack additionally keeps the cache key the loader was resolved with, so `Rule.use[].cache` could serve output produced with the previous options. Detecting it would mean serializing every loader's options for every module, which seemed the wrong trade for an opt-in cache; the docs now tell taps to replace `options` instead.

Local: full `rspack-test` run is 9190 passed / 2 failed, both `configCases/url/self` failing for an unrelated reason on my machine (the checkout path contains non-ASCII characters, which `new URL(import.meta.url)` percent-encodes).

kakiuwang-ui:
@hardfist gentle ping on this one — no rush if it is just queued, I mostly want to check the routing is right.

Status: 37/37 green, 9 skipped, nothing outstanding on my side. The change turns the existing (defined and already-called, but never tapped) `NormalModuleBeforeLoaders` Rust hook into a `SeriesBail` that JS can tap, so `module.rules`-resolved loaders can be added, replaced, reordered or left alone before the loader runner starts. It un-skips `configCases/loaders/pr-14384`, whose `test.filter.js` TODO points at exactly this gap.

Two design points worth a second opinion, since they are the parts I could not decide alone:

- `BeforeLoadersItem::Kept(index)` reuses the already-resolved `BoxLoader` + `LoaderRunnerOptions` rather than re-resolving, so loader cache identity survives a tap that leaves entries untouched. Added loaders go through `resolve_each_with_options`, the same path as `module.rules`, so `resolveLoader` and `builtin:` keep working.
- Deliberately unlike webpack, the tap does not receive a third `loaderContext` argument — in Rspack the loader context does not exist until the loader runner starts. This is called out in a warning box in the en/zh docs.

Happy to split anything out or change the shape if you would rather it landed differently.

kakiuwang-ui:
I saw #15471 and the reasoning on #15400 / #15372. Rather than resolve the conflict here first, I would rather ask you to apply the same standard to this PR now — the conflict is a direct consequence of the revert, and I do not want to spend the round trip if the answer is that this hook should not exist either.

Taking your three criteria honestly:

**Performance — the criticism applies to me as well.** The register is `skip = true` and a tap that changes nothing returns `undefined`, so the untapped path is free and the "no rebuild" path is free. But that is exactly the untapped measurement you called out; I did not measure the cost of an active tap, which for this hook is one crossing per built module plus the loader list conversion.

**Architecture — partly.** It is one call site rather than several, and the hook is already defined and invoked on the Rust side, so there is no new ordering contract. But it is still a synchronous boundary on the module build path, and that is your call to make.

**Replaceability — this is the one I would like your read on, because I could not find a substitute.** The hook rewrites an *input* to the build (which loaders run on this module) rather than an output, so `processAssets` does not apply. What I do not know is whether `module.rules` with `resourceQuery` covers the real cases — please correct me if it does.

Two pieces of evidence that it may not:

- The repo's own `tests/rspack-test/configCases/loaders/pr-14384/test.filter.js` says the gap is *"tracked here: https://github.com/orgs/web-infra-dev/projects/9"*, so the team has already scoped this at some point. That board is not visible to me — if it has since been decided against, that answers the question outright.
- In #8055, @OskarLebuda says this blocks porting [vue-ssr-assets-plugin](https://github.com/Trinovantes/vue-ssr-assets-plugin) ("Without that cannot be done"). Its use ([VueSsrAssetsServerPlugin.ts](https://github.com/Trinovantes/vue-ssr-assets-plugin/blob/master/src/server/VueSsrAssetsServerPlugin.ts#L38)) reads the already-resolved, loader-prefixed `normalModule.request` to decide whether to inject one extra loader, specifically for vue-loader's generated `?vue&type=script&setup=true` sub-requests, which the comment above it says bypass the normal loader chain.

If your read is that this is replaceable, I am happy to close it and, if useful, send a docs note showing the supported way instead. If you would rather it land, say so and I will rebase onto the revert and re-run CI.

kakiuwang-ui:
Rebased onto main (`7972a55`) — the conflict was only the `use crate::{...}` block in `normal_module.rs`, where the dependency-ownership refactors changed the import list around the four names this PR adds. Nothing in the hook itself moved.

Verified after the rebase: `cargo check` and `clippy` clean, `napi build` regenerates the committed `napi-binding.d.ts` unchanged, and the full `rspack-test` suite is 9274 passed / 2 failed — both `configCases/url/self`, which fails on my machine for an unrelated reason (the checkout path contains non-ASCII characters that `new URL(import.meta.url)` percent-encodes). `configCases/loaders/before-loaders` and the unskipped `configCases/loaders/pr-14384` both pass.

To be clear about where this stands: I rebased so the PR is not sitting in a conflicted state, not to push for a merge. My question from #15433 (comment) is still the one that matters — whether the standard applied in #15471 rules this hook out too, and specifically whether `module.rules` can reach the case in #8055 where the decision depends on the already-resolved, loader-prefixed request. If it can, I am happy to close this and send a docs note instead.

intellild:
Can you bench the beforeLoaders hook with more than 10000 modules?
You can let ai agents write scripts that generate the codebase, just like https://github.com/intellild/rspack-bench

kakiuwang-ui:
@intellild Benchmarked it, using your harness — I added a `before-loaders` case alongside `loader-hook`, same fixture generator (10,000 plain JS modules, binary-tree imports, one noop JS loader each), same conditions, one validation + one warmup + 5 measured builds per group in fresh Node processes with rotating order. Both sides are `--profile release` builds of the same checkout.

Four groups: published 2.2.3 untapped, and my branch untapped / with a pass-through tap / with a tap that appends one loader to every module.

Times in ms, median [min, max]:

| Group | Build | make | Tap calls | Appended loader ran |
|---|---:|---:|---:|---:|
| 2.2.3, untapped | 475.93 [471.44, 488.10] | 404.35 | 0 | — |
| branch, untapped | 500.24 [499.12, 523.41] | 426.65 | 0 | — |
| branch, pass-through tap | 580.74 [577.11, 585.65] | 507.50 | 10000 | — |
| branch, appends one loader | 643.71 [632.99, 683.02] | 567.73 | 10000 | 10000× |

| Tap | Overhead vs untapped | Ratio | Per module |
|---|---:|---:|---:|
| pass-through | +80.49 ms | 1.161× | 8.0 µs |
| appends one loader | +143.47 ms | 1.287× | 14.3 µs |

**On whether the untapped path is free.** Your criticism of my last two PRs was that I measured only the untapped path, so I want to be precise about what this number is and is not. `branch untapped / 2.2.3 = 1.051×` is *not* attributable to me — it spans 18 commits of main. So I built plain `main` (`debd9a68`) at the same release profile and ran the untapped groups again:

| Untapped | Build median | vs 2.2.3 |
|---|---:|---:|
| plain main | 501.75 [496.04, 510.54] | 1.0605× |
| this branch | 500.24 [499.12, 523.41] | 1.0511× |

−1.51 ms, well inside the spread. The ~6% gap from 2.2.3 is main's, not this PR's. That is what `skip = true` buys: with no tap registered the boundary is never crossed.

**On the tapped path, which is the number that matters.** 8.0 µs per module to cross and hand back an unchanged list; 14.3 µs when the list actually changes and every module is re-resolved. For reference, your recorded `loader-hook` run measured +377 ms for an *empty* tap at the same module count — 37.7 µs per module. This hook is cheaper per crossing because it carries a loader list rather than constructing a `LoaderContext`, but it is the same class of cost, and at 1.16× for a no-op tap I am not going to argue it is negligible. If the answer is that no per-module JS boundary is acceptable regardless of price, that is a coherent position and I would rather hear it than keep polishing this.

**On replaceability**, which was my open question. #15591 landed while this was waiting, and `taps/normalModule.ts` now writes `context.loaderItems` back from `loaderContext.loaders` after the JS tap — which looked like it might make `beforeLoaders` unnecessary. Your new `hook-before-loaders` fixture confirms the timing is right (`loaderIndex === 0`, nothing executed yet). So I tested whether a `loader` tap can change the chain, on plain main:

```js
hooks.loader.tap('P', (context, module) => {
  context.loaders.push({ /* valid LoaderObject shape, loaderItem included */ });
});
```

The JS array goes from 1 to 2, the build succeeds, **and the added loader never runs** — the emitted module is unchanged, with no error or warning. `merge_loader_context` zips `to.loader_items` against `from.loader_items`, so anything the JS side appended past the original length is dropped, and only `normal_executed` / `pitch_executed` / `data` are carried back, never the path. Same for removing or reordering.

So the two hooks share a moment but not a capability: `loader` can change the context, `beforeLoaders` changes which loaders exist. That is the case in #8055 — deciding from the already-resolved, loader-prefixed request whether to inject a loader.

Worth flagging separately regardless of what happens to this PR: **that silent drop is a footgun.** A plugin mutating `context.loaders` gets no signal that it did nothing. Happy to file it.

The bench case is four files (`config.mjs`, `run-build.mjs`, `bench.mjs`, `loaders/added.cjs`) — say the word and I will send it to rspack-bench as a PR. One thing I had to change from `loader-hook`: the mutate group cannot assert the chain through `module.identifier`, because `NormalModule::request` is fixed at creation and an appended loader never appears there (webpack behaves the same). It counts invocations of the appended loader instead, which is why the table above reports it ran 10000 times.

Branch is rebased onto `debd9a68` (`bd0619fa`); the add/add conflict with #15591 on `taps/normalModule.ts` is resolved by putting both registers in one `createNormalModuleHooksRegisters`. Full suite 9306 passed / 2 failed, both `configCases/url/self`, which fails on my machine for an unrelated reason. Your `configCases/loader/hook-before-loaders` passes on this branch.

intellild:
@codex review

packages/rspack/src/taps/normalModule.ts
```diff
@@ -2,11 +2,130 @@ import binding from '@rspack/binding';
...
+    ) as LoaderItem['options'];
+    parallel = compiler.__internal__ruleSet.references.get(
+      `${ident}$$parallelism`,
+    ) as Snapshot['parallel'];
+  }
+  const loaderItem: TrackedLoaderItem = {
+    loader: path,
+    options,
+    ident,
+    type: item.type ?? null,
```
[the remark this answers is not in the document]
kakiuwang-ui: Right — fixed in cd3200c. Checked it against webpack 5.110.3 first: `NormalModuleFactory` resolves the type to `undefined` when there is no `.mjs`/`.cjs` extension and no `descriptionFileData`, and `types.d.ts` declares `LoaderItem.type?: null | string`. `LoaderObject` already normalizes the binding's empty type to `undefined` for `cjs-loader-type`, so `beforeLoaders` was the odd one out.

The conversion now passes the binding's `undefined` through, `LoaderItem.type` is optional to match webpack, and the test asserts `undefined` (confirmed it fails with `+ null` on the previous build) and the en/zh docs say `undefined`. `normalCases/loaders/cjs-loader-type` still passes.

crates/node_binding/napi-binding.d.ts
```diff
@@ -659,6 +659,18 @@ export const EXPECTED_RSPACK_CORE_VERSION: string
...
+ * on the module are handed back as their index instead, so that their resolved
+ * loader and cache options can be reused.
+ */
+export interface JsAddedLoaderItem {
+  loader: string
+  options?: string
+  cache: boolean
+  optionsCacheKey: string
+}
+
```
intellild: Reuse `RawModuleRuleUse` and adapt the type in `packages/rspack/src/taps/normalModule.ts`

Refer to `registerNormalModuleLoaderTaps`

intellild (changes requested):

intellild:
@codex review

crates/node_binding/napi-binding.d.ts
intellild: There is a binary size limitation for single PR, you can check if other `raw` types can be reused.
These `raw` types are private apis that won't break users.

kakiuwang-ui: Done in 51344ca. `JsAddedLoaderItem` was field-for-field identical to `RawModuleRuleUse` — and `createRawModuleRuleUses` was already producing exactly that on the JS side, so I was copying its result into a second shape for no reason. The tap returns the value directly now.

Following your second note I looked for others: `JsBeforeLoadersLoaderItem` is gone too, replaced by `JsLoaderItem`, the type `registerNormalModuleLoaderTaps` already passes. One caveat so it does not surprise you in review — I still build it from the loader rather than through `JsLoaderItem::from(&Arc<dyn Loader>)`, because that impl derives its fields by splitting the identifier on `'|'` and drops `Loader::r#type()`, so the type would always come across empty. The absent type now crosses as the empty string, which is the binding's existing convention and what `LoaderObject` normalizes away.

That leaves `JsBeforeLoadersArgs` as the only type this PR adds, for the tap's single argument. I did not find an existing one that fits a `(loaders, module)` pair — if you would rather it reused something or passed the two separately, say which and I will change it.

packages/rspack/src/taps/normalModule.ts
```diff
@@ -2,11 +2,130 @@ import binding from '@rspack/binding';
...
+      : undefined;
+  // A loader a plugin added or rewrote still has to go through the same
+  // stringify + resolve path as the loaders coming from `module.rules`.
+  const [use] = createRawModuleRuleUses(
+    {
+      loader: item.loader,
+      options: item.options ?? undefined,
+      ident,
+      cache: original?.cache ?? inherited?.cache,
+      parallel,
```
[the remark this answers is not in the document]
kakiuwang-ui: Confirmed and documented rather than fixed, in 51344ca — here is the reasoning, in case the other call is preferred.

The report is accurate: `loadLoader` branches on `loader.type === 'module'` and also keys its module cache on it, and the return path resolves a rewritten entry through `resolve_loader`, which re-derives the type from the extension and the nearest `package.json`. `ModuleRuleUseLoader` has no type field, so there is nowhere for an override to travel.

Honouring it would mean adding one to `ModuleRuleUseLoader` and threading it through resolution — a change to a core config struct shared by every rule, for a case with no reported user. That cuts against the direction of the review this PR is already under, which is to remove types from the boundary rather than add fields to shared ones. So the en/zh docs now state that `type` is always derived when a loader is resolved, for a rewritten entry as much as for an added one, that Rspack therefore ignores what a tap writes there, and that the supported way to change how a loader is loaded is its file extension or its `package.json` `type`.

@intellild if you would rather have webpack parity here I am happy to add the field instead.

intellild (approved):

PR #15439, opened 2026-09-02
feat(core): expose import attributes on NormalModuleFactory ResolveData

### LINKED ISSUE #8770: [Feature]: support get import attributes in NormalModuleFactory resolve hook

In order to better support [Import Attributes](https://github.com/tc39/proposal-import-attributes), we need to support `attributes` in NormalModuleFactory resolve hook. 
```diff
export type ResolveData = {
	contextInfo: ContextInfo;
	context: string;
	request: string;
+       attributes?: Record<string, any>;
	fileDependencies: string[];
	missingDependencies: string[];
	contextDependencies: string[];
	createData?: CreateData;
};
```

By the way, this feature is already supported in webpack, but the key name is `assertions`.

https://github.com/webpack/webpack/blob/5e09d0e05f243d8d2c14c9fe140eba32b7aaeac7/types.d.ts#L12442

```diff
compiler.hooks.compilation.tap(
  'MyPlugin',
  (compilation, { normalModuleFactory }) => {
    normalModuleFactory.hooks.resolve.tap('MyPlugin', resolveData => {
+      console.log(resolveData.attributes);
    });
  },
);
```

kakiuwang-ui:
## Summary

Closes #8770.

`NormalModuleFactory`'s JS hooks receive a `ResolveData` that has never carried the import attributes of the request being resolved, so a plugin cannot tell `import data from './data.json' with { type: 'json' }` apart from a plain `import data from './data.json'`.

The data is already there on the Rust side. `ModuleFactoryCreateData` holds the triggering dependency, and `Dependency::get_attributes()` is read a few lines away in `NormalModuleFactory::calculate_module_rules` to match `module.rules[].with`. Nothing on that path was ever handed to JS.

This adds `attributes?: Record<string, string>` to `JsResolveData`, sourced from the first dependency of the factory call — the same `data.dependencies[0]` that `resolve_normal_module` already uses for `dependency_type` / `category` / `phase`.

```js
normalModuleFactory.hooks.beforeResolve.tap('MyPlugin', (resolveData) => {
  if (resolveData.attributes?.type === 'json') {
    // ...
  }
});
```

It is populated on `beforeResolve`, `factorize`, `resolve` and `afterResolve` (all four go through `JsResolveData::from_nmf_data`), and on the `NormalModuleReplacementPlugin` replacer, which uses the same constructor. It is `undefined` when the request carries no attributes.

## Alignment with webpack

webpack 5.110.3 exposes the same thing under the same name — `ResolveData.attributes`, read from `dependency.attributes` in `NormalModuleFactory#create`. (It was `assertions` when #8770 was filed and has since been renamed, so the field name the issue proposed is now also the webpack-parity one.)

## Read-only

`update_nmf_data` ignores `attributes`, matching the existing "only supports update request for now" limitation for the rest of the struct. Verified end-to-end: assigning `resolveData.attributes = { type: 'rewritten' }` in `beforeResolve` leaves `factorize` / `resolve` / `afterResolve` seeing the declared attributes, and so does `delete resolveData.attributes`. There is a regression case for this.

Since `JsResolveData` round-trips back through `FromNapiValue`, the field does still have to hold a string map when the hook returns; replacing it with values of another type fails the build with `Failed to convert JavaScript value ... on JsResolveData.attributes`. That is the same failure mode as every other field on this struct, and the docs say so.

## Tests

`configCases/hooks/normal-module-factory-attributes` covers, across all four hooks:

- a static import with two attributes (`with { type: 'custom', flavor: 'spicy' }`)
- a dynamic import (`import('./dynamic.js', { with: { level: '2' } })`), which goes through a different dependency type
- an import with no attributes — `undefined`
- the read-only guarantee described above

The plugin records what each hook saw and emits it as a JSON asset; the bundle reads it back and asserts, following `configCases/hooks/create-module`. Deliberately breaking one expected value fails the case, so the assertions are live.

`configCases/**` runs under both `Config.*` and `RuntimeModeConfig.*`, so this is exercised in both runtime modes.

## Out of scope

webpack's `ResolveData` also carries `phase` (`dependency.get_phase()` exists in rspack too, and is read at the same call site). That is a separate gap and is not part of this PR.

## Checklist

Local: full `rspack-test` suite green apart from the pre-existing `configCases/url/self` failure caused by the non-ASCII checkout path on my machine (9196 passed / 2 failed, both that case). `cargo clippy -p rspack_binding_api --all-targets` clean, `pnpm run test:type` and `pnpm run lint:js` pass.

### DIFF
--- a/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
+++ b/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
@@ -3,6 +3,7 @@ description: 'NormalModuleFactory is used by the Compiler to generate modules (N
 ---
 
 import ResolveDataType from '../../types/resolve-data.mdx';
+import { ApiMeta } from '@components/ApiMeta';
 import { Collapse, CollapsePanel } from '@components/Collapse';
 
 # NormalModuleFactory
@@ -16,6 +17,29 @@ NormalModuleFactory.hooks.someHook.tap(/* ... */);
 ```
 
 All hooks inherit from `Tapable`. In addition to `tap()`, you can also use `tapAsync()` and `tapPromise()`, depending on the type of the hook.
+
+## Import attributes
+
+<ApiMeta addedVersion="2.2.3" />
+
+`ResolveData.attributes` holds the [import attributes](https://github.com/tc39/proposal-import-attributes) of the dependency that triggered the resolution, for example `{ type: 'json' }` for `import data from './data.json' with { type: 'json' }`. It is available on all four hooks below, and is `undefined` when the request carries no attributes.
+
+```js
+compiler.hooks.compilation.tap(
+  'MyPlugin',
+  (compilation, { normalModuleFactory }) => {
+    normalModuleFactory.hooks.beforeResolve.tap('MyPlugin', (resolveData) => {
+      if (resolveData.attributes?.type === 'json') {
+        // ...
+      }
+    });
+  },
+);
+```
+
+:::warning
+`attributes` is read-only. Assigning to it inside a hook has no effect on how the module is resolved or built.
+:::
 
 ## `beforeResolve`
 
--- a/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
+++ b/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
@@ -22,6 +22,8 @@ All hooks inherit from `Tapable`. In addition to `tap()`, you can also use `tapA
 
 <ApiMeta addedVersion="2.2.3" />
 
+**Type:** `Readonly<Record<string, string>>`
+
 `ResolveData.attributes` holds the [import attributes](https://github.com/tc39/proposal-import-attributes) of the dependency that triggered the resolution, for example `{ type: 'json' }` for `import data from './data.json' with { type: 'json' }`. It is available on the [`beforeResolve`](#beforeresolve), [`factorize`](#factorize), [`resolve`](#resolve) and [`afterResolve`](#afterresolve) hooks, and is `undefined` when the request carries no attributes.
 
 ```js
@@ -36,10 +38,6 @@ compiler.hooks.compilation.tap(
   },
 );
 ```
-
-:::warning
-`attributes` is read-only: assigning to it inside a hook has no effect on how the module is resolved or built. It still has to hold a string map when the hook returns, so replacing it with values of another type fails the build.
-:::
 
 ## `beforeResolve`
 
--- a/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
+++ b/website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
@@ -22,7 +22,7 @@ All hooks inherit from `Tapable`. In addition to `tap()`, you can also use `tapA
 
 <ApiMeta addedVersion="2.2.3" />
 
-`ResolveData.attributes` holds the [import attributes](https://github.com/tc39/proposal-import-attributes) of the dependency that triggered the resolution, for example `{ type: 'json' }` for `import data from './data.json' with { type: 'json' }`. It is available on all four hooks below, and is `undefined` when the request carries no attributes.
+`ResolveData.attributes` holds the [import attributes](https://github.com/tc39/proposal-import-attributes) of the dependency that triggered the resolution, for example `{ type: 'json' }` for `import data from './data.json' with { type: 'json' }`. It is available on the [`beforeResolve`](#beforeresolve), [`factorize`](#factorize), [`resolve`](#resolve) and [`afterResolve`](#afterresolve) hooks, and is `undefined` when the request carries no attributes.
 
 ```js
 compiler.hooks.compilation.tap(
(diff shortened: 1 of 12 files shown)


kakiuwang-ui:
A note on the red CodSpeed check, since it is the only failing one — all of the repo's own CI is green.

CodSpeed reports `rust@create_module_ids` at 3.3 ms → 3.4 ms (-2.39%). This PR cannot reach that benchmark:

- The only Rust file it changes is `crates/rspack_binding_api/src/normal_module_factory.rs`; everything else is docs and a test case.
- `cargo tree -p rspack_benchmark` is 2134 lines and contains no `rspack_binding_api` (and no `node_binding`), so the benchmark binary does not link the changed code, let alone execute it. `JsResolveData::from_nmf_data` is only reachable from a JS tap interceptor.

The report also carries its own `Different runtime environments detected` warning, which lines up with `.github/workflows/ci.yml`: `Rust Bench` runs on `ubuntu-22.04` for every PR, while `Rust Bench Walltime` runs on `physical-1-exclusive` only on push to `main` / `workflow_dispatch`. The baseline therefore mixes two environments while a PR produces only the ubuntu one.

I re-ran the benchmarks (6cc60d6) and got exactly the same 3.3 ms → 3.4 ms. Reproducing against an unchanged baseline is consistent with the baseline being the source of the delta rather than variance on the HEAD side.

Happy to adjust anything here if you would rather see it handled differently.

kakiuwang-ui:
Flagging a routing gap rather than asking for speed: this PR has no reviewer requested on it, while its sibling #15433 got one automatically. @LingyuCoder — you reviewed the closest prior art (#15400, the `renderContent` binding surface), so you are probably the right person, but happy to be pointed elsewhere.

Status: 36 of the repo's own checks green, 9 skipped. The only red is the CodSpeed check, which I do not believe is attributable — see my comment above; the same benchmark reports a byte-identical `3.3 ms → 3.4 ms` on my #15445, which changes a disjoint set of crates.

The change itself is 16 lines of Rust: `data.dependencies.first().get_attributes()` was already used in `calculate_module_rules` to match `module.rules[].with`, it just never crossed into `JsResolveData`. All four NMF hooks plus `NormalModuleReplacementPlugin` pick it up at once, with no TS-side change.

One thing worth confirming before it lands, since it changes what the issue asked for: the issue (from 2024-12) proposes the name `assertions`, following webpack at the time. **webpack has since renamed it to `attributes`** (verified against 5.110.3), so `attributes` is now the aligned name and that is what this PR uses. Flagging it so it does not read as inventing a name.

website/docs/zh/api/plugin-api/normal-module-factory-hooks.mdx
```diff
@@ -17,6 +18,29 @@ NormalModuleFactory.hooks.someHook.tap(/* ... */);
 
 所有钩子均继承自 `Tapable`，除了 `tap()`，你还可以使用 `tapAsync()` 和 `tapPromise()`，具体取决于钩子的类型。
 
+## 导入属性
+
+<ApiMeta addedVersion="2.2.3" />
+
```
LingyuCoder, line 24:
```suggestion

**类型：** `Readonly<Record<string, string>>`

```

website/docs/zh/api/plugin-api/normal-module-factory-hooks.mdx
```diff
@@ -17,6 +18,29 @@ NormalModuleFactory.hooks.someHook.tap(/* ... */);
...
+        // ...
+      }
+    });
+  },
+);
+```
+
+:::warning
+`attributes` 是只读的：在钩子中对它赋值不会影响模块的解析与构建。但钩子返回时它仍必须是一个字符串映射，替换成其他类型的值会导致构建失败。
+:::
```
LingyuCoder:
```suggestion
```

website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
```diff
@@ -17,6 +18,29 @@ NormalModuleFactory.hooks.someHook.tap(/* ... */);
...
+      }
+    });
+  },
+);
+```
+
+:::warning
+`attributes` is read-only: assigning to it inside a hook has no effect on how the module is resolved or built. It still has to hold a string map when the hook returns, so replacing it with values of another type fails the build.
+:::
+
```
LingyuCoder:
```suggestion
```

website/docs/en/api/plugin-api/normal-module-factory-hooks.mdx
```diff
@@ -17,6 +18,29 @@ NormalModuleFactory.hooks.someHook.tap(/* ... */);
 
 All hooks inherit from `Tapable`. In addition to `tap()`, you can also use `tapAsync()` and `tapPromise()`, depending on the type of the hook.
 
+## Import attributes
+
+<ApiMeta addedVersion="2.2.3" />
+
```
LingyuCoder, line 24:
```suggestion

**Type:** `Readonly<Record<string, string>>`

```

LingyuCoder:
LGTM, thanks, cc @9aoy

LingyuCoder (approved):

9aoy (approved):

PR #15447, opened 2026-09-03
chore(deps): upgrade Rslib to 1.0.0

Timeless0911:
## Summary

Upgrade Rslib from `1.0.0-rc.2` to `1.0.0`. Add explicit `.js` extensions to the shared binding declaration for NodeNext compatibility.

## Related links

- https://github.com/web-infra-dev/rslib/releases/tag/v1.0.0

## Checklist


packages/rspack/rstack.browser.config.ts
```diff
@@ -249,7 +249,10 @@ function replaceDtsPlugin(): RsbuildPlugin {
             .replaceAll(
               `declare module '@rspack/binding'`,
               `declare module '${relativeBindingDts}'`,
-            );
+            )
+            // Copied declarations bypass Rslib's redirect pass.
+            // Keep explicit extensions for NodeNext consumers.
+            .replaceAll('from "./napi-binding"', 'from "./napi-binding.js"');
```
chenjiahan: Should we update /crates/node_binding/binding.d.ts directly?

Timeless0911: Yes, updated it directly in 8ff3951063. `crates/node_binding/binding.d.ts` is the shared hand-written source published by `@rspack/binding` and copied into `@rspack/browser`, so this keeps a single source of truth and removes the browser-only post-processing. I also verified the copied declaration is byte-identical and resolves under both NodeNext and Bundler.

chenjiahan (approved):

PR #15470, opened 2026-09-04
chore: add Rspack API assessment skill

LingyuCoder:
## Summary

- Add a Rspack-specific workflow for assessing proposed public APIs, including hooks.
- Apply strict admission criteria based on existing alternatives, ecosystem compatibility, performance, and architectural cost.
- Produce a concise fixed-format report with a compatibility example only when an existing API or hook is a viable substitute.
- Keep implicit invocation disabled.

## Related links

None.

## Checklist

- [x] Tests updated (not required; validated with the skill validator).
- [x] Documentation updated (the new skill is self-contained).


chenjiahan (approved):
👍

PR #15471, opened 2026-09-04
fix: revert afterOptimizeChunkIds and renderContent hooks

LingyuCoder:
## Summary

This reverts #15400 and #15372, removing:

- The `JavascriptModulesPlugin` `renderContent` hook.
- `compilation.hooks.afterOptimizeChunkIds`.
- Their associated bindings, tests, and documentation.

Both demonstrated use cases can be implemented with existing Rspack APIs while preserving the required output. Keeping these hooks would introduce permanent Rust–JavaScript boundaries and incomplete public contracts whose architectural and active-path costs outweigh the convenience they provide.

#### Replaceability

The required behavior—wrapping a JavaScript entry with additional code—can be implemented through the existing `processAssets` hook at `PROCESS_ASSETS_STAGE_ADDITIONS`, using `updateAsset` and `ConcatSource`.

This approach preserves source maps, requires no source reparsing, and still runs before real-content hashing. It executes later than `renderContent` and requires a small plugin adaptation, but preserves the observable output without introducing another rendering hook ([asset API](https://github.com/web-infra-dev/rspack/blob/04133c96b3a09ea6b84fdb1762a9d546e92f0c4d/packages/rspack/src/Compilation.ts#L672-L697), [hash stage](https://github.com/web-infra-dev/rspack/blob/04133c96b3a09ea6b84fdb1762a9d546e92f0c4d/crates/rspack_plugin_real_content_hash/src/lib.rs#L54-L68)).

#### Performance

When tapped, the binding converts the complete chunk source into a JavaScript representation and serializes its source map before the plugin can filter the chunk. It then synchronously waits for the JavaScript callback and converts the returned source back into Rust.

This makes the active cost proportional to the rendered source size and can increase peak memory usage. The provided benchmark only covered the untapped path, so it does not measure these conversion and synchronization costs ([binding implementation](https://github.com/web-infra-dev/rspack/blob/04133c96b3a09ea6b84fdb1762a9d546e92f0c4d/crates/rspack_binding_api/src/plugins/interceptor.rs#L1963-L1979), [benchmark](https://github.com/web-infra-dev/rspack/pull/15400#issuecomment-5473937131)).

#### Architecture

The hook adds a synchronous Rust–JavaScript boundary to a hot rendering path and must be invoked consistently by every renderer. Its implementation already requires multiple call sites, while the `modern-module` renderer bypasses those sites entirely ([bypassed renderer](https://github.com/web-infra-dev/rspack/blob/04133c96b3a09ea6b84fdb1762a9d546e92f0c4d/crates/rspack_plugin_javascript/src/plugin/impl_plugin_for_js_plugin.rs#L631-L657)).

It also exposes only a limited rendering context, making the public contract incomplete while still committing every future renderer to supporting its ordering and source-conversion semantics.

#### Replaceability

The required behavior—reading finalized chunk IDs and adding a runtime module—can be implemented inside the existing `additionalTreeRuntimeRequirements` hook.

That hook already runs after chunk ID assignment, supports `addRuntimeModule`, and completes before hashing. Calculating the relevant IDs there preserves the same emitted runtime behavior with a small plugin rewrite ([existing API](https://github.com/web-infra-dev/rspack/blob/d683791119be32ffa32787f502904027fe0a9faf/packages/rspack/src/Compilation.ts#L295-L305), [existing coverage](https://github.com/web-infra-dev/rspack/blob/d683791119be32ffa32787f502904027fe0a9faf/tests/rspack-test/configCases/runtime/add-runtime-module-stage/rspack.config.js#L43-L52)).

#### Performance

An active tap introduces a synchronous Rust–JavaScript crossing in the compilation pipeline. Exposing the chunks also materializes wrapper objects proportional to the number of chunks being inspected.

The reported benchmark only measured builds without an active tap and therefore does not cover the traversal, wrapper allocation, or synchronization costs incurred by actual use ([binding implementation](https://github.com/web-infra-dev/rspack/blob/1b0a8304a47aed32c5807cb82cb54f8d77bfd648/crates/rspack_binding_api/src/plugins/interceptor.rs#L875-L880), [benchmark](https://github.com/web-infra-dev/rspack/pull/15372#issuecomment-5449816536)).

#### Architecture

Chunk ID optimization is a Rust-owned global compilation phase. Exposing it as a public synchronous JavaScript boundary adds a long-term ordering and lifecycle contract to the core pipeline.

The hook is also exposed as read-only through getter-based chunk wrappers, so it does not represent the full semantics of the underlying phase. Since the required behavior can be expressed at the narrower runtime-requirements boundary, exposing the broader global phase is unnecessary.

## Related links

- Reverts #15400
- Reverts #15372

## Checklist


chenjiahan (approved):

chenjiahan (commented):
Can we explain the reason for the revert in the PR description?

LingyuCoder:
> Can we explain the reason for the revert in the PR description?

updated

PR #15512, opened 2026-09-07
perf(javascript): borrow AST ranges in parser walker

LingyuCoder:
## Summary

- Replace temporary AST handle vectors in the parser walker with borrowed `TypedSubRange` iteration.
- Keep traversal order and parser behavior unchanged; this PR only changes walker data structures.
- This is the first Step 3 optimization stacked on #15506. The next layer is #15513.

## Related links

- #15506
- #15513

## Checklist

- [x] Tests updated (not required; behavior is unchanged).
- [x] Documentation updated (not required).

PR #15513, opened 2026-09-07
perf(javascript): borrow parser plugin arguments

LingyuCoder:
## Summary

- Pass call and constructor arguments to parser plugins through borrowed AST ranges.
- Add a small borrowed argument view where CommonJS parsing needs indexed access.
- Keep plugin matching behavior unchanged; this PR only changes argument data structures.
- This is the second Step 3 layer, stacked on #15512. The next layer is #15514.

## Related links

- #15512
- #15514

## Checklist

- [x] Tests updated (not required; behavior is unchanged).

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/amd/amd_define_dependency_parser_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/amd/amd_define_dependency_parser_plugin.rs
@@ -6,7 +6,7 @@ use rspack_core::{
   RuntimeRequirementsDependency, get_context,
 };
 use rspack_intern::Atom;
-use rspack_util::SpanExt;
+use rspack_util::{SpanExt, swc::AstSubRangeExt};
 use rustc_hash::FxHashMap;
 use swc_next_ecma_ast::{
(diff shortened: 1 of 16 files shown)


LingyuCoder:
## Performance comparison with #15512

CodSpeed compares both PRs against `main` instead of comparing this stacked PR directly with #15512. Comparing the reported HEAD measurements gives:

| Benchmark | #15512 | #15513 | Difference |
| --- | ---: | ---: | ---: |
| `scan_dependencies@three_module` | 13.7 ms | 13.7 ms | neutral |
| `build_module_graph` | 452.7 ms | 452.6 ms | ~0.02% faster |
| `threejs-production-sourcemap` | 179.7 ms | 179.5 ms | ~0.11% faster |
| `threejs-development` | 119.2 ms | 119.0 ms | ~0.17% faster |

The overall reported result also changes from **-1.09%** on #15512 to **-0.95%** on this PR, an improvement of approximately **0.14 percentage points**.

Therefore, relative to #15512, this PR **does not introduce a performance regression**. The results are performance-neutral to slightly improved.

The exact improvement should still be treated as directional rather than significant because the reports used different `main` revisions and CodSpeed detected different runtime environments.

hardfist (approved):

PR #15514, opened 2026-09-07
perf(javascript): streamline AST range traversal

LingyuCoder:
## Summary

Traverse side-effects analysis, call evaluation, IIFE parameters, and Rstest arguments directly over AST ranges, removing temporary node vectors. Reuse `AstSubRangeExt::nodes`, `first`, and `second` throughout JavaScript parser plugins, walkers, evaluators, magic comments, and the Rslib/Rstest parser integrations to encapsulate range-slot decoding. Consume template expressions sequentially, and build React directive metadata without intermediate string/span or reference vectors.

Expose the iterator's existing `DoubleEndedIterator` capability so property lookup and string concatenation evaluation retain reverse traversal. Preserve array holes, positional pairing, traversal order, and short-circuit behavior. Member-call evaluation still rejects spread arguments before invoking evaluation hooks.

## Related links

- #15513
- #15515

## Checklist

- [x] Tests updated (not required; behavior is unchanged).

### DIFF
--- a/crates/rspack_plugin_javascript/src/magic_comment.rs
+++ b/crates/rspack_plugin_javascript/src/magic_comment.rs
@@ -4,7 +4,10 @@ use regex::Regex;
 use rspack_core::{ContextMode, DependencyRange};
 use rspack_error::{Diagnostic, Error, Severity};
 use rspack_regex::RspackRegex;
-use rspack_util::{SpanExt, swc::RspackComment as Comment};
+use rspack_util::{
+  SpanExt,
+  swc::{AstSubRangeExt, RspackComment as Comment},
+};
 use rustc_hash::{FxHashMap, FxHashSet};
 use swc_next_allocator::Allocator;
 use swc_next_ecma_ast::{
@@ -414,7 +417,7 @@ fn expr_to_str<'a>(ast: &'a Ast<'a>, expr: Expr) -> Option<Cow<'a, str>> {
     ExprData::TemplateLiteral(template)
       if template.expressions(ast).is_empty() && template.quasis(ast).len() == 1 =>
     {
-      let element = template.quasis(ast).get_node(ast, 0)?;
+      let element = ast.first(template.quasis(ast))?;
       Some(Cow::Borrowed(ast.get_utf8(element.raw(ast))))
     }
     _ => None,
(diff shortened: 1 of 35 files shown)


LingyuCoder:
## Performance comparison with #15513

Updated for #15514 at [`78ea976`](https://github.com/web-infra-dev/rspack/commit/78ea9762827c061b7746d39108f64aa59dc64482), using the completed CodSpeed simulation run from September 8, 2026.

CodSpeed still falls back to `main` because the actual `release/2.3.0` base has no successful benchmark run. Its headline is therefore not a direct comparison with #15513. Comparing the four HEAD measurements published in the [#15513 report](https://github.com/web-infra-dev/rspack/pull/15513#issuecomment-5567291805) and the [latest #15514 report](https://github.com/web-infra-dev/rspack/pull/15514#issuecomment-5567203055) gives:

| Benchmark | #15513 (`8914072`) | #15514 (`78ea976`) | Reduction in simulated duration |
| --- | ---: | ---: | ---: |
| `scan_dependencies@three_module` | 13.7 ms | 13.6 ms | ~0.73% |
| `build_module_graph` | 452.6 ms | 449.7 ms | ~0.64% |
| `threejs-production-sourcemap` | 179.5 ms | 179.2 ms | ~0.17% |
| `threejs-development` | 119.0 ms | 118.9 ms | ~0.08% |

Calculated as `(#15513 HEAD - #15514 HEAD) / #15513 HEAD`, using the rounded values displayed in the reports. These are simulation measurements, not wall-clock timings.

The overall reported result changes from **-0.95%** on #15513 to **-0.58%** on the latest #15514: `-0.58 - (-0.95) = +0.37 percentage points`. This replaces the previous comparison of -0.95% versus -0.67% (+0.28 percentage points).

**No regression is observed in these four matched measurements relative to #15513; they indicate a modest improvement**, particularly in dependency scanning and module graph construction. The regression labels against fallback `main` do not establish a regression introduced by this PR.

The magnitude remains directional: #15513 used `main@f03bd2f`, #15514 used `main@6d647a7`, and CodSpeed flagged different runtime environments. Consequently, the 0.37-percentage-point headline difference is **not** a controlled measurement of the PR's overall speedup, and these four published rows do not establish a benchmark-wide no-regression guarantee.

hardfist (approved):

PR #15530, opened 2026-09-07
fix: restore side-effect imports after metadata changes

### LINKED ISSUE #14855: [Bug]: incremental rebuild does not restore a side-effect-only import after sideEffects is re-enabled
### Reproduction

- Rspack: @rspack/core@2.1.4
- Repository: https://github.com/hardfist/rspack-repro-incremental-side-effects

The repro uses one compiler instance with incremental make enabled and changes a package metadata file between builds:

1. sideEffects: true — the side-effect marker is present.
2. sideEffects: false — the side-effect-only import is removed.
3. sideEffects: true again — the incremental bundle still omits the marker.
4. A cold build with sideEffects: true includes the marker again.

Observed output:

~~~text
incremental: initial sideEffects=true  marker=true
incremental: sideEffects=false         marker=false
incremental: sideEffects=true again    marker=false
cold: final sideEffects=true           marker=true
BUG REPRODUCED: the incremental bundle stays stale after re-enabling side effects.
~~~

### Expected behavior

Changing package sideEffects metadata back to true should make the incremental result match a cold build and retain the side-effect-only module.

### Suspected area

The stale state appears to cross the make/finish/optimize boundary used for side-effect connection evaluation. Relevant code paths are crates/rspack_core/src/artifacts/side_effects_state_artifact.rs, crates/rspack_core/src/compilation/finish_modules/mod.rs, and crates/rspack_core/src/compilation/build_module_graph/graph_updater/repair/lazy.rs.

The repository contains the complete runnable fixture and a cold-build control.

zhsama:
## Summary

When a package's metadata flips from `sideEffects: true` to `false` and back to `true`, an incremental build keeps omitting the side-effect-only module from the bundle, while a cold build includes it.

Root cause:

- With `sideEffects: false`, `FactorizeResultTask` make-skips the `import "./pkg/side.js"` dependency and calls `dep.set_lazy()`.
- `ModuleDependency::weak()` defaults to `self.lazy().is_some()`, and the code splitter skips weak dependencies when building the chunk graph.
- When `sideEffects` is switched back to `true`, the dependency is re-factorized and the module is created and connected again, but the stale `lazy_make` flag is never cleared, so the connection is still treated as weak and the module never reaches a chunk.

Fix: when factorization proceeds with a module, clear the lazy flag on evaluation-phase `EsmImport` dependencies. This mirrors the exact condition under which the make-skip path sets the flag, so it only affects dependencies that were skipped in a previous build. Dependencies that arrive here through lazy-barrel are already unset by `process_unlazy_dependencies`, and `EsmImport` dependencies are never marked lazy at parse time.

## Related links

fix https://github.com/web-infra-dev/rspack/issues/14855

Verified with the reproduction repository from the issue: the incremental result now matches the cold build. The added `compilerCases/side-effects-metadata-rebuild` cases (cache on/off, `incremental` make-only/all, shared vs. non-shared target module) fail on unmodified `main` and pass with this change.

## Checklist


zhsama:
I investigated the `rust@create_module_hashes` regression reported by CodSpeed (-2.28%).

Before attempting a local A/B comparison, I ran two A/A control series on the same baseline binary, with six independent processes per series. The largest absolute changes between paired runs were 1.81% and 2.76%. Neither series met the preselected 1% stability threshold, so I did not proceed with the formal A/B comparison.

These were native Criterion measurements on macOS arm64, whereas CI uses CodSpeed CPU simulation on Linux. The local variation therefore cannot directly explain or invalidate the CI result. CodSpeed also warns that the baseline and PR were measured in different runtime environments.

At this point, I would treat the regression as unconfirmed rather than dismiss it as noise. A comparison under matched runtime environments is still needed to determine whether this change causes a reproducible slowdown.

PR #15531, opened 2026-09-07
ci: probe self-hosted tool cache volume (diagnostic, do not merge)

stormslowly:
## Why

On self-hosted Linux runners, `actions/setup-node`'s "Adding to the cache ..." step (copying the extracted Node tree, 4,797 files / 192MB, file by file into `$RUNNER_TOOL_CACHE`) takes anywhere from 0.7s to 278s on different pods, while extracting the same tarball into `$RUNNER_TEMP` on the same pod is consistently 1-2s. With `timeout-minutes: 5` on the pnpm setup step this kills jobs: 7 job failures today across `ci.yml` and `ci-diff.yml` (e.g. run 34119614737).

This PR temporarily replaces `ci.yml` with a 6-sample probe job that records mount layout, cgroup io limits, and small-file write latency for `$RUNNER_TOOL_CACHE` vs `$RUNNER_TEMP` / `$HOME` / `/tmp`, then runs the real pnpm setup and copies the node tree both ways for timing.

Heavy jobs in `ci-diff.yml` / `ci-rust.yaml` are disabled on this branch to save runner time. Will be closed after data is collected.

### DIFF
--- a/.github/workflows/ci-diff.yml
+++ b/.github/workflows/ci-diff.yml
@@ -40,6 +40,7 @@ permissions:
 jobs:
   build:
     name: Build Bindings
+    if: ${{ false }}
     uses: ./.github/workflows/reusable-build.yml
     with:
       target: x86_64-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


stormslowly:
Probe complete; findings recorded offline. Closing.

PR #15544, opened 2026-09-08
perf(javascript): reduce AST decoding and Atom materialization

LingyuCoder:
## Summary

Reduce repeated AST decoding and Atom materialization in JavaScript dependency scanning. The follow-up changes from #15547 are consolidated here, retaining separate commits for the two areas.

### AST decoding and walker adaptation

- Use a narrow identifier-name check for non-computed member properties; preserve the computed-property classification rules.
- Construct statement classifications directly from SWC Next node-kind data, removing intermediate statement/declaration classifications.
- Reuse borrowed identifier names across internal pattern callbacks and callee classification within call walking. Read compatibility-hook AST names only after the hook-name check matches.

### Atom materialization

- Borrow identifier/root names in evaluator, member-root and meta-property hooks, and use borrowed lookups in CommonJS/createRequire, dynamic import, ESM export, compatibility and URL plugins.
- Borrow configuration/property names for comparisons and binding names during temporary duplicate-name counting; create owned Atoms only for names that must be retained.
- Defer Atom creation until a purity or inner-graph result is accepted, or an evaluated/deferred result needs to own its name.
- Reuse existing Atoms from scope enumeration, accepted configured names and default exports instead of interning them again.

This follows #15515 and targets `release/2.3.0`. Parser behavior, name-based duplicate counting, local-binding snapshots, shadowing, hook ordering, WTF-8 conversion and deferred-purity decisions are preserved. No caches, unsafe code or Atom hashing/collection changes are introduced. Scope metadata lookup elimination, semantic/SymbolId reuse, concatenation identifier changes and comment storage remain out of scope.

### Local performance

Measured `rust@scan_dependencies@three_module` with native ARM64 Docker and standard Valgrind Callgrind, using the same image, fixture, compiler, build flags and scan boundary:

| Version | Instructions |
| --- | ---: |
| Original baseline: #15515 (`51576f7f688a`) | 40,120,353 |
| AST decoding changes (`15fc2a7212`) | 39,318,142 |
| Combined changes (`c65664b063`) | 38,669,201 |
| Combined change relative to the original baseline | **-1,451,152 (-3.62%)** |

The Atom-materialization commit contributes -648,941 instructions (-1.65%) relative to the AST-decoding commit. The combined percentage above is calculated directly from instruction counts, not by adding percentages.

These are the existing local measurements; no benchmark rerun was needed to consolidate the unchanged commits. The release base used for this PR has the same Rust source as the measured baseline; the additional changes are CI workflow configuration only. The intermediate source patch checksum was verified against its commit. These are dependency-scanning instruction counts, not wall-clock build times or CodSpeed cloud percentages. Parsing and semantic construction are outside the measured boundary.

Environment: `linux/arm64`, Rust `1.97.0-nightly (e8e4541ff)`, Valgrind `3.19.0`, `three@0.183.2/build/three.module.js`. Docker image ID: `sha256:ad4bf3d1931a328cfebd58092b5a9af81c2fcf9fb04b06c33882126246275b52`.

### Local checks

The following checks passed for the combined commit `c65664b063` before consolidation:

- Full dev build and Rust/JavaScript formatting.
- JavaScript-plugin Clippy with `-D warnings`.
- Rust library tests: 48 passed.
- `test:unit` passed; rspack-test: 47 files, 9,221 passed and 4 skipped.
- No snapshots or test expectations changed. One existing scope test converts the newly returned `&Atom` to `&str` to keep its original assertion.

An additional workspace `--all-features` Clippy check fails in the unchanged `rspack_cacheable` crate: enabling `noop` leaves `PortablePath` without the `Archive` implementation required by its adapters. This also reproduces with `cargo clippy -p rspack_cacheable --all-targets --all-features`, without compiling the modified JavaScript parser. No unrelated fix is included.

## Related links

- #15515
- #15547 (consolidated into this PR)
- Original migration plan: #15331

## Checklist

- [x] Tests updated (only an API type adaptation; assertions unchanged).


LingyuCoder:
image

hardfist (approved):

PR #15547, opened 2026-09-08
perf: defer parser Atom materialization

LingyuCoder:
## Summary

Stacked on #15544. This is the Atom-materialization portion of the SWC Next follow-up work.

- Borrow identifier/root names in evaluator, member-root and meta-property hooks, and use borrowed lookups in CommonJS/createRequire, dynamic import, ESM export, compatibility and URL plugins.
- Borrow configuration/property names for comparisons. Keep side-effects binding names borrowed during temporary duplicate-name counting, and create owned Atoms only for names that must be retained.
- Defer Atom creation until a purity or inner-graph result is accepted, or an evaluated/deferred result needs to own its name.
- Reuse existing Atoms from scope enumeration, accepted configured names and default exports instead of converting them back to strings and interning them again.

These paths scale with identifier occurrences and bindings during dependency scanning. Avoiding eager interning, atom cloning/dropping and temporary owned keys reduces per-node work without adding a cache or changing Atom hashing/collection internals. Existing owned-Atom queries retain their fast path.

Scope metadata lookup elimination, semantic/SymbolId reuse, concatenation identifier changes, comment storage and general AST decoding changes are deliberately excluded. The existing name-based duplicate counting, local-binding snapshot, shadowing, hook ordering, WTF-8 conversion and deferred-purity decisions are preserved.

### Local performance

`rust@scan_dependencies@three_module`, native ARM64 Docker + standard Valgrind Callgrind:

| Version | Instructions |
| --- | ---: |
| #15544 (`15fc2a7212`) | 39,318,142 |
| This PR (`c65664b063`) | 38,669,201 |
| Change relative to #15544 | **-648,941 (-1.65%)** |

The candidate was measured once. The baseline reuses the previously measured #15544 source: its patch checksum was verified against the parent commit, and the same image, native architecture, fixture, compiler flags and measurement boundary were retained. No performance reruns or cloud comparison were used.

Environment: `linux/arm64`, Rust `1.97.0-nightly (e8e4541ff)`, Valgrind `3.19.0`, `three@0.183.2/build/three.module.js`. Image ID: `sha256:ad4bf3d1931a328cfebd58092b5a9af81c2fcf9fb04b06c33882126246275b52`. This is the instruction-count change for dependency scanning, not wall-clock build time; parsing and semantic construction are outside the measured boundary.

### Local checks

- Full dev build, Rust/JavaScript formatting and JavaScript-plugin Clippy with `-D warnings` passed.
- Rust library tests: 48 passed.
- `test:unit` passed; rspack-test: 47 files, 9,221 passed and 4 skipped.
- No snapshots or test expectations changed. One existing scope test converts the newly returned `&Atom` to `&str` to keep its original assertion.
- The additional workspace `--all-features` Clippy check still fails in the unchanged `rspack_cacheable` crate with the same `noop`/`PortablePath: Archive` incompatibility recorded on #15544. No unrelated fix is included.

## Related links

- #15544
- Original migration plan: #15331

## Checklist

- [x] Tests updated (only an API type adaptation; assertions unchanged).

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
@@ -212,17 +212,20 @@ pub fn is_create_require_import(
   let Some(specifier) = create_require_import_specifier(parser, source) else {
     return false;
   };
-  export_name.is_some_and(|export_name| export_name == &specifier)
+  export_name.is_some_and(|export_name| export_name == specifier)
 }
 
 #[inline(never)]
-fn create_require_import_specifier(parser: &JavascriptParser, source: &Atom) -> Option<Atom> {
+fn create_require_import_specifier<'a>(
+  parser: &'a JavascriptParser,
+  source: &Atom,
+) -> Option<&'a str> {
   let option = parser.javascript_options.create_require_option()?;
   let (specifier, module) = option.split_once(" from ")?;
   (!specifier.is_empty()
     && !module.is_empty()
     && (source.as_ref() == module || (module == "module" && source.as_ref() == "node:module")))
-    .then(|| specifier.into())
+    .then_some(specifier)
 }
 
 #[inline(never)]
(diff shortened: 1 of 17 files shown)


LingyuCoder:
Consolidated into #15544. Its branch now includes c65664b063, retaining both commits for review, and its description includes the combined local IR result (-3.62% relative to the original baseline). GitHub automatically marked this PR as merged when its commit reached its base branch; it is no longer open. This consolidation did not merge #15544 into release/2.3.0.

PR #15566, opened 2026-09-08
chore(docs): fix standalone configuration examples and type declarations

pangxie1991:
## Summary

Make configuration examples usable as written in both English and Chinese documentation.

- Initialize `__dirname` from `import.meta.url` and add missing Node.js imports in the entry, HTTP build, and resolver examples. The previous `.mjs` snippets throw `ReferenceError` when loaded.
- Separate the four `output.clean` alternatives into standalone configurations. In the previous example, repeated `output` keys silently overwrite the preceding alternatives.
- Fix the full-width union operator in `BundlerInfo` and parenthesize the callback union in `EntryDescription.publicPath` so the documented declarations parse as TypeScript.

## Validation

- Loaded all 18 added or modified `.mjs` examples with Node.js 24.10.0; checked for duplicate object keys.
- Checked the 4 changed TypeScript declarations with TypeScript 6.0.3; callback declarations were also checked against the exported Rspack types.
- Reproduced the original missing-variable, duplicate-key, and type-syntax problems.
- Prettier 3.9.6 and `git diff --check` passed for the changed files. New internal links and anchors were checked where applicable.
- Verified all 6 merge orders of the three split PRs: no conflicts, and the combined tree exactly reproduces the original documentation changes on the updated base.
- The full documentation site build was not run; website dependencies are not installed. Validation tooling is separate and is not included in the PR.

## Related links

- [Configuration types](https://github.com/web-infra-dev/rspack/blob/bba0032cb676be776dc8a09eb65c1d197c930db3/packages/rspack/src/config/types.ts)
- Extracted from #15565. This PR targets `main` independently; no other split PR is required first.

## Checklist

- [x] Tests updated (or not required; documentation only).
- [x] Documentation updated (English and Chinese).

### DIFF
--- a/website/docs/en/config/entry.mdx
+++ b/website/docs/en/config/entry.mdx
@@ -52,6 +52,9 @@ You can also use the [path module](https://nodejs.org/api/path.html) in Node.js 
 
 ```js title="rspack.config.mjs"
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
 export default {
   entry: path.join(__dirname, './src/index.js'),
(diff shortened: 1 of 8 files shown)


chenjiahan (approved):
LGTM

chenjiahan (approved):

PR #15592, opened 2026-09-09
refactor(loader): replace should-yield hooks with execution kind

intellild:
## Motivation

Split the runtime-selection refactor out of #15401 so it can be reviewed independently from LoaderChain scheduling and chain caching.

The loader runner currently asks an async plugin/hook whether it should yield for each loader. The implementation mainly distinguishes native loaders from JavaScript loaders by their request prefix, even though that execution kind is already determined when the loader is resolved.

## Changes

- Introduce `LoaderExecutionKind::{Native, JavaScript}` and cache this small Copy enum in `LoaderItem`.
- Native loaders use the default execution kind; `JsLoader` explicitly selects JavaScript.
- Replace the runner's `should_yield` hook dispatch with a direct execution-kind check.
- Remove `LoaderRunnerPlugin::should_yield`, `NormalModuleLoaderShouldYield`, and their registration/adapter.
- Keep the actual `loader_yield` bridge. Move the existing no-pitch check to the JavaScript loader scheduler so known pitch-less loaders still skip the JavaScript callback.

For example, in `js-loader -> builtin:swc-loader -> js-loader`, the runner uses the JavaScript bridge for the JS loaders and directly runs the native loader, without asking a separate should-yield hook at each step.

This PR does not introduce LoaderChain, change loader-cache behavior/storage, split per-module loader state, or include the loader-hook-position work from #15591.

## Stack

1. This PR: execution kind and removal of the should-yield hook, based on `main`.
2. #15602: reusable `Loaders` metadata and per-run state, based on this branch (one commit).
3. #15401: LoaderChain scheduling and caching, based on #15602.

## Validation

- `pnpm run build:cli:dev`: passed.
- `cargo test -p rspack_loader_runner --locked`: 10 passed.
- `cargo clippy -p rspack_binding_api --all-targets -- -D warnings`: passed.
- `cargo fmt --all --check` and `git diff --check`: passed.
- `pnpm test --project base -t loader`: 563 passed, including mixed native/JS and parallel pitching cases. Two failures are from the pre-existing local, untracked `rspack-issue-6457-mixed` context-identity experiment, which is not included in either PR.

---
_by OpenAI Codex_


intellild:
@codex review

SyMind (approved):

PR #15597, opened 2026-09-09
refactor(javascript): reuse semantic and member analysis results

LingyuCoder:
## Summary

Reuse semantic and member-analysis results across side-effects analysis, the parser walker, and the evaluator. Stacked on #15595, ultimately targeting `release/2.3.0`.

- **Side effects:** use existing semantic bindings and declaration counts instead of collecting all top-level names and building a duplicate-name map/set. Retain export-alias/default-export handling and distinguish top-level pure functions from shadowing bindings using resolved callee symbols.
- **Identifier evaluation:** reuse immutable semantic resolution across the hook/fallback path. Mutable variable state is still read again after hooks, so alias/tag changes remain visible.
- **Member walking and evaluation:** extract member-chain information once, share the result with member-callee evaluation, and combine normal-member/await-import analysis. Preserve optional-chain handling, plugin evaluation, and synthetic AST identity/ranges.
- **Hook dispatch:** use already-resolved qualified member names directly, avoiding another variable-name lookup. Consolidate remaining Atom/string name queries in a parser helper, while keeping direct identifier semantic lookup.
- Add compatibility coverage for shadowed/duplicate declarations, require aliases, DefinePlugin qualified names, computed/optional members, and await-import reads/calls.

The two commits remain grouped by responsibility: side-effects declaration reuse, followed by walker/evaluator result reuse. The combined PR changes 18 files (362 insertions / 371 deletions). Existing test assertions and snapshots are unchanged. There is no persistent evaluator cache or caching of mutable binding state across hooks.

### Performance

Local Docker + Valgrind Callgrind, `rust@scan_dependencies@three_module`:

| Version | Instructions (IR) |
| --- | ---: |
| Parent #15595 (`0ca2c88dd2`) | 34,406,832 |
| Side-effects-only revision (`5d0562b8c3`) | 34,263,698 |
| Combined PR (`f8c8aa492b`) | 32,466,923 |

The combined PR reduces scan IR by **1,939,909 (-5.64%) versus #15595**. Adding walker/evaluator reuse reduces IR by a further **1,796,775 (-5.24%) versus the side-effects-only revision**, replacing the previous standalone -0.42% PR result.

The candidate was measured once; the two verified earlier measurements were reused. All use the same native Linux ARM64 image, nightly-2026-04-16, Valgrind 3.19.0, build flags, and Three.js fixture. This measures dependency scanning, excluding parsing and semantic construction; it is not a whole-build wall-time claim.

## Related links

- Parent: #15595
- Original migration/optimization work: #15331

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/side_effects_parser_plugin.rs
@@ -8,17 +8,17 @@ use rspack_util::{
   SpanExt,
   swc::{AstSubRangeExt, RspackComments},
 };
-use rustc_hash::FxHashMap;
+use rustc_hash::FxHashSet;
 use swc_next_ecma_ast::{
   ArgumentData, ArrowFunctionBodyData, ArrowFunctionExpression, AssignmentExpression,
   AssignmentOperator, AssignmentTargetData, Ast, BindingPattern, BindingPatternData,
   CallExpression, Class, ClassElement, ClassElementData, CommentKind, Decl, DeclData,
   ExportDefaultDeclarationKindData, Expr, ExprData, ForStatementInitData, FormalParameterItemData,
-  FormalParameterPatternData, Function, GetSpan, ImportDeclarationSpecifierData,
-  MethodDefinitionKind, ModuleExportName, ModuleExportNameData, Program, PropertyKey,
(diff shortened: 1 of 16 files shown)


LingyuCoder:
@codex review

LingyuCoder:
@codex review

CPunisher (approved):

PR #15599, opened 2026-09-09
fix: omit build-local paths from sourcemap content hash

cellison-figma:
## Motivation

Rspack’s code generation hash currently hashes each generated `Source` structurally. For `SourceMapSource`, that structural hash includes the full source map, including path-like fields such as `sources`, `sourceRoot`, and `file`. In sandboxed or remote-execution CI environments, those fields can contain absolute worker paths like `/mnt/engflow/worker/work/N/exec/...`, where `N` varies between otherwise identical builds. Those path differences do not affect emitted JavaScript bytes, but they feed into per-module codegen hashes. The tainted module hashes then feed chunk content hashes, causing unstable asset filenames and reducing cross-build asset reuse.

## Changes

This change makes codegen hashing deterministic for that case by:
- sorting codegen `SourceType` entries before hashing, avoiding hash-map iteration-order instability
- hashing only content-relevant source-map fields for `SourceMapSource`
- deliberately excluding path-like source-map metadata from the source hash while preserving it in emitted source maps

This keeps debug/source-map output unchanged, while ensuring content hashes reflect emitted content and stable source-map content rather than ephemeral build-machine paths.

The added regression test confirms builds with identical emitted code and different sandbox source-map paths produce the same source hash.

### DIFF
--- a/crates/rspack_core/src/artifacts/code_generation_results.rs
+++ b/crates/rspack_core/src/artifacts/code_generation_results.rs
@@ -343,7 +343,9 @@ impl CodeGenerationResultBuilder {
     hash_salt: &HashSalt,
   ) {
     let mut hasher = RspackHasher::with_salt(hash_function, hash_salt);
-    for (source_type, source) in self.value.sources.as_ref() {
+    let mut entries = self.value.sources.as_ref().iter().collect::<Vec<_>>();
+    entries.sort_unstable_by_key(|(source_type, _)| **source_type);
+    for (source_type, source) in entries {
       source_type.hash(&mut hasher);
       std::hash::Hash::hash(source, &mut hasher);
     }
@@ -365,7 +367,9 @@ impl CodeGenerationResultBuilder {
   ) {
     let mut hasher = RspackHasher::with_salt(hash_function, hash_salt);
     runtime_hash.hash(&mut hasher);
-    for source_type in self.value.sources.as_ref().keys() {
+    let mut source_types = self.value.sources.as_ref().keys().collect::<Vec<_>>();
+    source_types.sort_unstable();
+    for source_type in source_types {
       source_type.hash(&mut hasher);
     }
     self.value.data.update_hash(&mut hasher);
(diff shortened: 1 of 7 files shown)


LingyuCoder:
@codex review

SyMind:
@cellison-figma Thanks for the PR. Two points before we dive into the implementation:
1. Current behavior is aligned with webpack. webpack also mixes the loader-provided source (including the full source map, with path fields like sources/sourceRoot) into the module build hash, and from there into chunk content hashes, so it has the same instability in sandboxed CI environments. The webpack team's official answer to this class of issue (webpack/webpack#14905) is: use optimization.realContentHash, which is enabled by default in production mode and computes hashes from the final emitted bytes — machine-specific map paths can't taint it. Changing the hash semantics of rspack-sources would make Rspack diverge from webpack here, which is something we're very cautious about.
2. What's the underlying goal? If it's stable output filenames, realContentHash should already cover it — is there a reason it doesn't work for you (e.g. non-production mode, or a setup that requires realContentHash: false)? Or is the real goal reusing the persistent build cache across different worker paths? If so, that's a broader problem: module identifiers, resource paths, and loader paths all embed absolute paths too, so excluding source-map fields from this one hash won't make the cache portable — that deserves a dedicated design discussion rather than a field-level hash change.

SyMind:
Hi @cellison-figma, thanks for the detailed write-up — we understand the issue now. The root cause is that loader-provided source maps keep machine-specific absolute paths in path-like fields (`sources`, `sourceRoot`, `file`), and those paths leak into module sources and from there into module/chunk hashes, which breaks build determinism in sandboxed CI environments.

Instead of changing the hash semantics of `rspack-sources` (which would diverge from webpack, as discussed above), we're fixing this at the source in #15671: module source maps are made context independent by normalizing the path-like fields before they ever reach codegen hashes. That addresses both the content-hash instability and the build-cache portability problem in one place.

Let's continue the discussion in #15671 — feedback there is very welcome!

cellison-figma:
@SyMind Thank you for looking into this. The approach in #15671 does address the machine-specific path leak. Closing to continue discussion there.

PR #15604, opened 2026-09-10
fix: normalize HttpUriPlugin allowedUris before matching

SyMind:
## Motivation

`HttpUriPlugin` compares string `allowedUris` rules without normalizing them as URLs. A host-only rule can therefore match a URL with userinfo or a different hostname that shares the permitted hostname prefix, allowing requests outside the intended allowlist.

Align this behavior with [webpack’s fix](https://github.com/webpack/webpack/commit/c5100702335a9cdcb75558ccd80def2329bd4abf).

## Changes

- Parse and normalize requested URLs and string rules before prefix matching. Ignore invalid string rules and match regular expressions against the normalized URL. Initial requests and redirects continue to share the same policy check.
- Add regression cases for valid normalized rules, regex matching, allowed redirects, userinfo bypasses, hostname-prefix bypasses, path boundaries, invalid rules, and disallowed redirects. The security cases use a mock HTTP client and fail if a disallowed request reaches it.
- Update the English and Chinese documentation and existing HTTP fixtures. Scheme-only strings such as `'https://'` must be replaced with complete URLs or an intentional regular expression.

Validation:

- `pnpm run build:cli:dev`
- `pnpm --dir tests/rspack-test run test -t configCases/build-http` — 14 passed
- `pnpm --dir tests/rspack-test run test -t configCases/asset-modules/http-redirect-security` — 2 passed
- `cargo clippy -p rspack_plugin_schemes --lib -- --deny warnings`
- Rust formatting, configuration/documentation formatting, and `git diff --check`

### DIFF
--- a/crates/rspack_plugin_schemes/src/http_uri/mod.rs
+++ b/crates/rspack_plugin_schemes/src/http_uri/mod.rs
@@ -282,7 +282,22 @@ impl HttpUriOptionsAllowedUris {
   }
 
   pub fn is_allowed(&self, uri: &str) -> bool {
-    self.conditions.try_match(uri)
+    let Ok(parsed_uri) = Url::parse(uri) else {
+      return false;
+    };
+
+    // Normalize both URLs so a host-only rule includes the trailing slash and
+    // cannot match a different host through userinfo or a shared hostname prefix.
+    let matches = |condition: &AssetCondition| match condition {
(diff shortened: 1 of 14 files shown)


SyMind:
@codex review

ahabhgk (approved):

PR #15608, opened 2026-09-10
refactor: replace FreezeLock blocking locks with atomic borrows

hardfist:
## Motivation

Avoid blocking locks when accessing build metadata, following Servo's shared-lock approach.

## Changes

Use atomic borrow guards during construction and release/acquire publication after freezing. Add a safe `get_unchecked()` that asserts publication and returns a direct reference for seal-specific reads. Preserve metadata identity and existing freeze points.

## Open questions

`Module::build()` returns before metadata is final: `succeedModule` can emit assets, make completion merges `importModule` assets, cache preparation writes `BuildInfo.snapshot`, and failed rebuilds can restore `BuildMeta`. `importModule` also runs codegen during make, before freezing is guaranteed.

The implementation therefore still needs guarded reads and field-specific `&self` updates. We need to settle whether finalization can finish while the module is exclusively owned, before freezing and sharing it. Until that boundary is resolved, read/write separation relies on runtime borrow checks, with conflicts panicking instead of waiting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4213,6 +4213,7 @@ dependencies = [
  "arc-swap",
  "async-recursion",
  "async-trait",
+ "atomic_refcell",
  "bitflags 2.13.1",
  "concat-string",
  "cow-utils",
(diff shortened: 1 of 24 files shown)


SyMind:
@codex review

SyMind (dismissed):

PR #15609, opened 2026-09-10
fix: enforce HttpUriPlugin lockfile integrity in frozen builds

SyMind:
## Motivation

`experiments.buildHttp` records SHA-512 integrity in its lockfile but accepts cached content without verifying it. For example, editing a cached module while leaving the lockfile unchanged still allows that module into a production bundle. When the content cache is missing or disabled, a download can also replace an existing locked entry without enforcing a frozen policy.

## Changes

- Implement `frozen` across the public options and native binding, defaulting it to `true` in production mode, as webpack does. Frozen builds reject missing entries, corrupted or missing cached content, and changes to the downloaded content, resolved URL, or content type before updating the lockfile or content cache.
- Preserve the original URL's locked entry through redirects and enforce it even when the cache is missing or disabled. Send cached ETags only to the locked resolved URL and retain cached content through intermediate redirects for target 304 responses. Keep webpack's unfrozen cache repair, line-ending repair, upgrade, and `no-cache`/`ignore` behavior.
- Support older lockfiles that record redirected imports only under their final URL: follow allowed redirects to an existing locked entry, validate its content, and leave the committed lockfile and cache unchanged in frozen mode. Preserve the total redirect limit across the lookup and subsequent requests; an existing original-URL entry remains authoritative.
- Parse the stored `resolved` URL and propagate lockfile/cache errors. Clone lock entry metadata before asynchronous reads so resource I/O does not hold the lockfile mutex, and serialize lockfile updates within the plugin to preserve concurrently fetched entries.
- Add 47 integrity scenarios and concurrent-update coverage using the existing config-case runners, and document the behavior in English and Chinese. Set `frozen: false` explicitly in the URL policy fixtures so their mocked requests can exercise allowlist and redirect checks without a pre-existing lockfile.

Production builds using `buildHttp` now require an existing trusted lockfile and content cache unless caching is disabled. Populate or intentionally update them with `frozen: false`; enable `upgrade: true` when checking remote updates.

Validation on the main-based branch:

- `pnpm run build:cli:dev` and `pnpm run build:binding:dev`
- `cargo clippy -p rspack_plugin_schemes --all-targets -- --deny warnings`
- `pnpm --dir tests/rspack-test run test -t 'configCases/build-http'` — 18 tests passed across the normal and runtimeMode suites.
- `pnpm --dir tests/rspack-test run test -t 'http-redirect-security'` — 2 tests passed across the normal and runtimeMode suites.
- Commit hooks passed Rust formatting, JavaScript lint/format checks, and documentation spelling checks.


SyMind:
@codex review

SyMind:
@codex review

SyMind:
@codex review

ahabhgk (approved):

chenjiahan:
Defaulting `frozen` to `true` in production is a breaking change: existing configurations now fail, as shown by [eco CI](https://github.com/rstackjs/rstack-ecosystem-ci/actions/runs/34571989659/job/103175986218)

SyMind:
@chenjiahan I don't think this is a breaking change. Looking at the actual eco CI failure:

```
https://www.unpkg.com/the-answer has an outdated lockfile entry, but lockfile is frozen
Lockfile: LockfileEntry { resolved: "https://www.unpkg.com/the-answer", integrity: "sha512-qy9nd...", ... }
Received: LockfileEntry { resolved: "https://www.unpkg.com/the-answer@1.0.0/dist/the-answer.js", integrity: "sha512-qy9nd...", ... }
```

The failure comes from a pre-existing defect in the committed lockfile, not from the new `frozen` default:

1. **The entry was recorded incorrectly in the first place.** `https://www.unpkg.com/the-answer` responds with a 302 redirect to `https://www.unpkg.com/the-answer@1.0.0/dist/the-answer.js`, so the lockfile should record the final resolved URL. The committed entry instead records the original URL as `resolved`, which never accurately described what was fetched.

2. **The lockfile was previously generated but never actually enforced.** Rspack wrote integrity entries without validating the downloaded or cached content against them, so an inaccurate entry like this could sit unnoticed indefinitely. Frozen mode now checks the entry and surfaces that latent inconsistency — note the integrity hashes are identical, meaning the bundled content has not changed at all; only the stale metadata is rejected.

So the eco CI failure exposes a bug that already existed in the lockfile, rather than a regression introduced by this PR. The correct fix for the example is to regenerate the lockfile with an unfrozen build and commit it. This also matches webpack's long-standing behavior: webpack has defaulted `experiments.buildHttp.frozen` to `true` in production mode since the feature was introduced, and rejecting outdated entries in frozen builds is the intended semantics there.

PR #15613, opened 2026-09-10
docs: correct Lightning CSS unusedSymbols option type

HashemKhalifa:
## Motivation

The English Lightning CSS minimizer documentation declares `unusedSymbols` as `Set<String>`, but the public option accepts `string[]`. Copying the documented collection type produces an invalid configuration.

## Changes

Correct the declaration to `unusedSymbols?: string[]`, matching the public implementation and the existing Chinese documentation. This changes documentation only.

Validation: the website production build and affected-file formatting check pass. With `@rspack/core` 2.2.3, a public-API type check accepts an array and rejects a `Set`; a CSS build using `unusedSymbols: ['drop']` removes `.drop` and preserves `.keep`. The rendered page contains the corrected declaration.


chenjiahan (approved):
Thank you!

PR #15616, opened 2026-09-10
feat: share persistent cache cross child compilers

ahabhgk:
## Motivation

With `experiments.newCache`, parent and child compilers inherit the same persistent cache directory but create separate native cache instances. Their database writes and cleanup can interfere with each other, losing cache entries or removing SST files that another compiler still needs. They need to coordinate access to the shared storage while keeping each compiler's cache entries and metadata separate.

## Changes

- Share one native `Cache` through the JavaScript `Cache` object, and use `CompilerCache` views to scope cache keys by compiler path.
- Drive idle work and shutdown through the JavaScript run/watch lifecycle so child compilations use the parent's shared cache. Continue recording build dependencies and compiler metadata at the end of each build/rebuild.
- Store compiler metadata as regular entries in each compiler's cache namespace, replacing the single global metadata record.
- Add a regression case that closes and recreates a parent compiler using the same persistent cache directory, then verifies that both the parent and its child compiler restore their modules from disk.

### DIFF
--- a/crates/node_binding/napi-binding.d.ts
+++ b/crates/node_binding/napi-binding.d.ts
@@ -276,6 +276,14 @@ export declare class ExternalModule {
   _emitFile(filename: string, source: JsSource, assetInfo?: AssetInfo | undefined | null): void
 }
 
+/** One shared cache, initialized from the first compiler that uses it. */
+export declare class JsCache {
+  constructor()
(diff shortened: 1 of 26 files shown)


ahabhgk:
@codex review

hardfist (approved):

PR #13147, opened 2026-02-27
feat(javascript): align scope termination and DCE with webpack

LingyuCoder:
## Summary

- Align JavascriptParser `scope.terminated` tracking and control-flow analysis with webpack, including `if`/`try`/`switch`/loop blocks and function/class scopes.
- Integrate ConstPlugin `unused_statement` handling to eliminate dead control flow while preserving hoisted `var` / function declarations and matching webpack's `with`-statement semantics.
- Skip unreachable dynamic `import()` calls after terminating statements and deduplicate inline eval parse warnings per-source to fix `dead-code-elimination` and DefinePlugin diagnostics test cases.

## Related links

- Config case: `tests/rspack-test/configCases/parsing/dead-code-elimination`
- Diagnostics case: `tests/rspack-test/diagnosticsCases/module-parse-failed/define_plugin_warning`
- Tree shaking case: `tests/rspack-test/treeShakingCases/rollup-unused-inner-functions-and-classes`

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

---

> [!NOTE]
> **High Risk**
> Touches core JavaScript parsing control-flow tracking (`terminated`) and dependency collection, which can change the module graph and emitted output for `if`/`try`/`switch`/loop edge cases. Regressions are most likely around dead-code paths, hoisting, and dynamic import/require resolution semantics.
> 
> **Overview**
> Implements webpack-style *scope termination* tracking in `JavascriptParser` via a new `terminated` state (return vs throw) and updates statement walking (`if`/`try`/`switch`/loops/with) to propagate or clear termination correctly across branches and scopes.
> 
> Introduces a new parser plugin hook `unused_statement` and wires `ConstPlugin` to replace unreachable statements with a webpack-matching `// removed by dead control flow` stub while preserving hoisted `var` (and non-strict function) declarations via the refactored `get_hoisted_declarations`.
> 
> Prevents resolving/adding dependencies for unreachable `import()` and `require()` calls after `return`/`throw` in non-top-level scopes, and updates related tests/snapshots (dead-code elimination, stats handling, and tree-shaking output).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 910b42e8dfd32a5707375f6ab007456120f685f8. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #15618, opened 2026-09-10
fix(javascript): do not terminate the enclosing scope after a do-while loop

### LINKED ISSUE #15617: [Bug]: statements after a do…while whose body ends in throw/return are removed as dead control flow
### System Info

```
  System:
    OS: macOS 26.6.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 9.40 GB / 48.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.16.0 - ~/.local/state/fnm_multishells/64812_1789032444855/bin/node
    npm: 11.13.0 - ~/.local/state/fnm_multishells/64812_1789032444855/bin/npm
    pnpm: 10.17.1 - ~/.local/state/fnm_multishells/64812_1789032444855/bin/pnpm
  Browsers:
    Chrome: 152.0.7977.83
    Safari: 26.6.2
  npmPackages:
    @rspack/core: 2.2.3 => 2.2.3
```

### Details

Since rspack 2.0.0 the JavaScript parser's dead-control-flow elimination treats the code **after** a `do { … } while (cond);` loop as unreachable whenever the last statement of the loop body is a `throw` or `return` — even when the body contains a `break` (or `continue`) that makes the following code reachable. Every statement after the loop is replaced with `// removed by dead control flow\n{}`.

This happens in the parser itself, so it reproduces with `optimization.minimize: false` and without any loader (also reproduced with `builtin:swc-loader`, `target: es2020`). rspack 1.x and webpack 5 keep the code. swc is not involved.

Minimal trigger:

```js
export function f(x, c, after) {
  do {
    if (x()) break;
    throw new Error('e');
  } while (c());
  after();        // rspack 2.x: removed by dead control flow
  return 'done';  // rspack 2.x: removed by dead control flow
}
```

Output of `@rspack/core` 2.2.3 (`mode: 'production'`, `optimization.minimize: false`, no loader):

```js
function f(x, c, after) {
  do {
    if (x()) break;
    throw new Error('e');
  } while (c());
  // removed by dead control flow
{}
  // removed by dead control flow
{}
}
```

No warning or error is emitted; the function silently returns `undefined`.

#### Version matrix

Same input and config for all rows:

| Tool | Result |
|---|---|
| `@rspack/core` 1.7.11 | ✅ code kept |
| `@rspack/core` 2.0.0 | ❌ removed |
| `@rspack/core` 2.2.3 (latest) | ❌ removed |
| webpack 5.109.2 | ✅ code kept |
| `@swc/core` 1.11.24 / 1.15.43, transform (es2020/es5) and minify (compress+mangle) | ✅ code kept |

#### Which shapes trigger it

Loop bodies followed by `AFTER(); return 2;`:

| Shape | rspack 2.x |
|---|---|
| `do { if (x()) break; throw new Error('e'); } while (c());` | ❌ |
| `do { if (x()) continue; throw new Error('e'); } while (c());` | ❌ |
| `do { if (x()) break; return 1; } while (c());` | ❌ |
| `do { if (x()) break; if (y()) { throw a; } else { throw b; } } while (c());` | ❌ |
| `do { const r = x(); if (r === 1) break; throw new Error('e'); } while (c());` | ❌ |
| `do { if (x()) { after1(); } break; } while (c());` (body ends in `break`) | ✅ |
| `while (c()) { if (x()) break; throw new Error('e'); }` | ✅ |
| `for (;;) { if (x()) break; throw new Error('e'); }` | ✅ |
| `for (const i of [1]) { if (x()) break; throw new Error('e'); }` | ✅ |
| `outer: { if (x()) break outer; throw new Error('e'); }` | ✅ |
| `switch (x()) { case 1: break; default: throw new Error('e'); }` | ✅ |
| `do { throw new Error('e'); break; } while (c());` | removed — correct, genuinely unreachable |

Only `do…while` is affected, and only when the body's final statement terminates (`throw`/`return`).

#### Root cause (pointer)

[`crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs`](https://github.com/web-infra-dev/rspack/blob/main/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs)

`while`, `for`, `for…in` and `for…of` walk their bodies inside `in_block_scope(false, …)`, which saves/restores `self.terminated` and only propagates termination when `in_executed_path` is `true`:

```rust
fn walk_while_statement(&mut self, stmt: &WhileStmt) {
  self.in_block_scope(false, |this| {
    this.walk_expression(&stmt.test);
    this.walk_nested_statement(&stmt.body);
  });
}
```

`do…while` is the only loop that walks its body directly, so a `throw`/`return` at the end of the body sets `self.terminated` on the *enclosing* scope, and the `unused_statement` hook (ConstPlugin) removes everything after the loop:

```rust
fn walk_do_while_statement(&mut self, stmt: &DoWhileStmt) {
  self.walk_nested_statement(&stmt.body);
  self.walk_expression(&stmt.test);
}
```

webpack's parser, which this logic was ported from, does enter a block scope for `do…while` ([`lib/javascript/JavascriptParser.js`](https://github.com/webpack/webpack/blob/main/lib/javascript/JavascriptParser.js)):

```js
walkDoWhileStatement(statement) {
  const oldScope = this._enterBlockScope();
  this.walkNestedStatement(statement.body);
  this.walkExpression(statement.test);
  this._exitBlockScope(oldScope, false);
}
```

The fix is most likely a one-liner: wrap the do-while body and test in `in_block_scope(false, …)` exactly like `walk_while_statement`. I can open a PR for this.

#### Regression origin

Introduced by #13147 "feat(javascript): align scope termination and DCE with webpack" (merged 2026-03-02, first released in 2.0.0). Still present on `main` as of 2026-09-10.

#### Not a duplicate of

- #15335 / #15343: a labeled `do…while` with `break` miscompiled by `SwcJsMinimizerRspackPlugin`; that one needs minification and was fixed by bumping the swc minifier. This report reproduces with minification disabled and is in rspack's own parser.

#### Real-world impact

A tokenizer in a published library parses `name:op=arg,op2;name2:op`-style input with an outer `while (parser.hasMore())` and an inner `do { …; if (endOfFilter) break; if (endOfOperator) continue; throw new SyntaxError(...); } while (parser.hasMore());` followed by `result.push(filter)`. Under rspack 2.x the `push` is removed and the function silently returns `[]` for every input — no warning, no error, only wrong runtime behaviour.

### Reproduce link

https://github.com/nest-aka-swan/rspack-do-while-dead-control-flow-repro

### Reproduce Steps

```
git clone https://github.com/nest-aka-swan/rspack-do-while-dead-control-flow-repro
cd rspack-do-while-dead-control-flow-repro
npm install
npm run repro
```

Actual output:

```
@rspack/core 2.2.3: after() NOT called, f() returned undefined, marker in bundle: true
@rspack/core 1.7.11: after() called, f() returned "done", marker in bundle: false
```

Expected for every version: `after() called, f() returned "done", marker in bundle: false`.

`dist/2.2.3/out.js` shows the two `// removed by dead control flow` markers in place of `after();` and `return 'done';`. The config is `mode: 'production'`, `devtool: false`, `optimization.minimize: false`, no loaders.

nest-aka-swan:
## Motivation

Fixes #15617.

Since 2.0.0 (#13147) the parser's dead-control-flow elimination removes every statement after a `do { … } while (cond);` loop whose body ends in `throw`/`return`, even when the body contains a `break`/`continue` that makes the following code reachable:

```js
do {
  if (x()) break;
  throw new Error('e');
} while (c());
after();       // removed by dead control flow
return 'done'; // removed by dead control flow
```

`while`, `for`, `for…in` and `for…of` walk their bodies inside `in_block_scope(false, …)`, which restores `terminated` on exit and only propagates it when the scope is on an always-executed path. `walk_do_while_statement` was the only loop that walked its body directly, so a terminating statement at the end of the body marked the *enclosing* scope as terminated. webpack's `walkDoWhileStatement` enters a block scope for `do…while` too.

## Changes

- `walk_do_while_statement` now walks the body and test inside `in_block_scope(false, …)`, matching `walk_while_statement` and webpack.
- Added cases to `configCases/parsing/dead-code-elimination` covering `break` + `throw`, `continue` + `throw`, and `break` + `return` bodies.

Verified locally with `pnpm --dir tests/rspack-test run test -t "configCases/parsing/dead-code-elimination"`; the new cases fail before the fix (the statements after the loop are replaced by `// removed by dead control flow`) and pass after.

### DIFF
--- a/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
+++ b/crates/rspack_plugin_javascript/src/visitors/dependency/parser/walk.rs
@@ -1869,8 +1869,10 @@ impl JavascriptParser<'_> {
   }
 
   fn walk_do_while_statement(&mut self, stmt: &DoWhileStmt) {
-    self.walk_nested_statement(&stmt.body);
-    self.walk_expression(&stmt.test);
+    self.in_block_scope(false, |this| {
+      this.walk_nested_statement(&stmt.body);
+      this.walk_expression(&stmt.test);
+    });
   }
 
   fn walk_block_statement(&mut self, stmt: &BlockStmt) {
--- a/tests/rspack-test/configCases/parsing/dead-code-elimination/index.js
+++ b/tests/rspack-test/configCases/parsing/dead-code-elimination/index.js
@@ -134,6 +134,52 @@ it("Shouldn't hoist const/let from dead for-in branch", () => {
 	expect(key).toBe("start");
 });
 
+it("Shouldn't eliminate statements after a do-while loop whose body ends in throw (#15617)", () => {
+	let afterCalled = false;
+	const f = (x, c) => {
+		do {
+			if (x()) break;
+			throw new Error("e");
+		} while (c());
+		afterCalled = true;
+		return "done";
+	};
+	expect(
+		f(
+			() => true,
+			() => false
+		)
+	).toBe("done");
+	expect(afterCalled).toBe(true);
+
+	const g = (x, c) => {
+		let i = 0;
+		do {
+			if (x(i++)) continue;
+			throw new Error("e");
+		} while (c(i));
+		afterCalled = "g";
+		return "done";
+	};
+	expect(
+		g(
+			() => true,
+			i => i < 2
+		)
+	).toBe("done");
+	expect(afterCalled).toBe("g");
+
+	const h = x => {
+		do {
+			if (x()) break;
+			return 1;
+		} while (false);
+		return 2;
+	};
+	expect(h(() => true)).toBe(2);
+	expect(h(() => false)).toBe(1);
+});
+
 if (true) {
 	return;
 }


LingyuCoder (approved):
LGTM, thanks

PR #15630, opened 2026-09-11
docs: add Extension.js to the ecosystem page

cezaraugusto:
## Motivation

hey folks, I maintain Extension.js, a framework for building browser extensions built on Rspack

the Rspack 2.0 post already name-checks it in the ecosystem list:
https://rspack.rs/blog/announcing-2-0

14 of the other projects from the post are under Community integrations, but Extension.js never made it across. this PR adds it

## Changes

adds an Extension.js entry to the ecosystem page, English and Chinese. entry sits after Docusaurus so the existing order holds

what it actually does, in case it helps with the wording: Extension.js reads the entry points declared in a `manifest.json` (background, content scripts, popup, options, and so on) and feeds them into the Rspack entry graph, then emits one build per browser. one source tree, output for Chrome, Edge, Firefox and Safari. running on `@rspack/core` 2.x

- https://extension.js.org
- https://github.com/extension-js/extension.js

happy to trim the blurb or drop a tag if you'd rather keep entries shorter

### DIFF
--- a/website/docs/en/guide/start/ecosystem.mdx
+++ b/website/docs/en/guide/start/ecosystem.mdx
@@ -65,6 +65,15 @@ Rspack is part of Rstack, the fast, unified JavaScript toolchain for developers 
 
 Docusaurus supports Rspack as the bundler since v3.6, see [Docusaurus Faster](https://docusaurus.io/blog/releases/3.6#docusaurus-faster) for details.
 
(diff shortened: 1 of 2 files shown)


chenjiahan (approved):
Thanks for adding this! Happy to have Extension.js in the Rspack ecosystem ❤️

PR #15632, opened 2026-09-11
fix(loader): preserve string BOM across native boundaries

intellild:
## Motivation

A loader-produced string is transported through the native bridge as a Buffer. Without preserving its original type, the next JavaScript loader treats it as binary input and strips its leading BOM.

For example, in normal execution order `producer (JS) -> builtin:test-passthrough-loader -> consumer (JS)`, returning the string `"\uFEFFhello"` incorrectly delivers `"hello"` to the non-raw consumer.

## Changes

- Populate the existing Rust-to-JavaScript `__internal__utf8Hint` from the native content variant.
- Restore string content before normal JavaScript loader execution, preserving its leading BOM.
- Keep the existing BOM stripping behavior for Buffer-to-string conversion and preserve BOM bytes for raw loaders.
- Add regression coverage for JS-only and JS/native/JS chains, parallel on/off, string/Buffer outputs, and normal/raw consumers.

## Validation

- Rust/JavaScript build and clippy passed.
- BOM regression tests: 17 passed.
- Raw/string loader regression tests: 8 passed.
- Disabling the JavaScript hint read reproduces BOM loss in the new regression test; restoring it makes the test pass.

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
@codex review

SyMind (approved):

PR #15633, opened 2026-09-11
perf(css): reduce magic comment processing overhead

intellild:
## Motivation

CSS ignore magic comments introduced in #15314 add comment scanning and allocation work during dependency extraction. Reduce that overhead while keeping CSS and JavaScript on the same magic comment parser and preserving parsing and diagnostic behavior.

For example, in `background: /* webpackIgnore: true */ /* ordinary comment */ url(./image.png)`, dependency extraction previously scanned backward for the comment group and CSS then split the raw text into comments again. Reusing comment ranges recorded by the lexer avoids these repeated scans. Ordinary comments are filtered before creating the parser allocator.

## Changes

- Record comment ranges during lexing and use them directly when extracting CSS magic comments.
- Move optional comment metadata out of every balanced-stack entry into a separate table, reducing `BalancedItem` from 24 to 12 bytes.
- Find each magic comment candidate once and create the allocator only when there is a candidate.
- Replace intermediate CSS/JS comment vectors with iterators and omit deduplication for the already unique CSS comment ranges.
- Extend the existing CSS ignore case to cover adjacent comments, embedded comment markers, image-set URLs, pure CSS modules, and comments separated from URLs by another value.

CSS and JavaScript continue to use the shared full magic comment parser. No CSS-specific parser or boolean-only parsing path is introduced.

Validation: `pnpm run build:cli:dev`; Clippy for `css-module-lexer`, `rspack_plugin_css`, and `rspack_plugin_javascript` with all targets and warnings denied; 313 lexer tests; 7 shared parser tests; 321 CSS integration tests; 4 JS magic comment and 2 require-resolve-ignore integration tests. All passed. End-to-end performance impact still needs CodSpeed validation.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15635, opened 2026-09-11
perf(loader): lazily bridge boxed loader contexts and sources

intellild:
## Motivation

Crossing the Rust/JavaScript loader boundary currently materializes the loader context and splits source content from its source map, even when JavaScript does not consume them. Builtin loaders also pass content and maps separately, losing the opportunity to preserve native source graphs between loaders.

For example, a JavaScript pitch loader that returns no output should leave the native source untouched. A chain of builtin loaders should retain `BoxSource` until a JavaScript normal loader actually needs a string or Buffer and a separate source map. When a native consumer needs owned text, consuming a uniquely owned raw buffer can reuse its allocation instead of copying the whole file. No benchmark speedup is claimed.

## Changes

- Turn `JsLoaderContext` into a N-API class directly owning `Option<Box<LoaderContext>>`. The JavaScript loader runner receives and returns the same Box; return-value conversion takes it on the JavaScript thread, leaving retained instances empty.
- Batch mutable fields in a `napi(object)` state. The JavaScript facade reads and updates this state locally and writes it back once, while source content and maps remain lazy.
- Keep ordinary native loader hooks on `&mut LoaderContext`. JavaScript hooks exchange a state snapshot and return the complete state without moving the native context. Preserve hook-installed properties, loader object identity, and native/JavaScript hook ordering.
- Carry `BoxSource` through builtin loaders and normal module processing. Add consuming Source access that reuses uniquely owned leaf allocations, including after cache restoration, and use SIMD UTF-8 validation. Remove full-content copies in NormalModule and SWC and per-module static runtime copies in React/Preact refresh.
- Preserve native source graphs when JavaScript produces no output, and retain invalid UTF-8 bytes in mapped sources and cache serialization.
- Update generated bindings, WASI exports, regression cases, and English and Chinese architecture documentation.

### Validation

- Native binding and JavaScript package builds passed.
- Clippy with `-D warnings` passed for the binding API, loader runner, sources, builtin loader crates, and HMR.
- Rust tests passed for `rspack_loader_runner`, `rspack_sources`, and `rspack_cacheable_test` (135 tests plus 11 doc tests). Pointer checks verify allocation reuse before and after cache restoration and shared-source preservation.
- Loader regressions passed (351 tests), and source-map regressions passed (184 tests), using the final native build.
- Rust formatting, JavaScript lint/format checks, documentation spelling checks, and `git diff --check` passed.
- CodSpeed has not yet been rerun for these changes.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15636, opened 2026-09-11
perf(loader): use one-shot refs for js values

stormslowly:
## Motivation

Two sources of overhead when carrying loader `additionalData` across the JS/Rust boundary.

## Changes

1. **Use `ThreadsafeOneShotRef` instead of `ThreadsafeJsValueRef`.** `ThreadsafeJsValueRef` creates a `JsCallback`/TSFN per referenced value, so every referenced value adds a libuv async handle and its reference-management allocations to the queue. `ThreadsafeOneShotRef` shares a single deletion TSFN instead. `ThreadsafeJsValueRef` is removed; `loader_context_state` from #15591 moves to the same owned reference.

2. **Stop producing `additionalData` where nothing consumes it.** Once normal loader execution has exhausted the chain (`loaderIndex < 0`), the JS runner returns `undefined` rather than creating an unused Node-API reference. The unused `NormalModuleAdditionalData` hook and the CSS extract pitch's fourth callback argument are removed as well; CSS dependencies keep using `__internal__setParseMeta`.

A yield with loaders still pending continues to carry `additionalData`, including a yield to a builtin loader; pitch behavior is otherwise preserved.

## Validation

- Rebuilt Rust bindings and JS packages with `pnpm run build:cli:dev`.
- Passed `hook-before-loaders`, `rspack-issue-13654`, `additional-data`, `rspack-issue-5600`, `loader-cache`, and `css-extract` test selections.


SyMind:
@codex review

SyMind:
@codex review

SyMind (approved):

PR #15638, opened 2026-09-11
refactor: stream new cache writes directly into database batches

ahabhgk:
## Motivation

Flushing the new filesystem cache currently retains all serialized entries in a `Vec` before starting database writes. Large flushes therefore keep the entire encoded batch in memory, even though the database can flush its buffers as entries arrive.

## Changes

Stream parallel encoding results directly into a single database write batch, with validator data included in the same commit. Count entries as they are consumed and skip empty commits.

In my benchmark, cold-start peak native memory decreased by **1.10 GiB (16.5%)**, and memory footprint decreased by **8.5%**. Cold-cache write time fell from **2.935 s to 2.007 s (−31.6%)**.

Total RSS did not decrease, and warm starts showed no clear memory benefit.

### DIFF
--- a/crates/rspack_core/src/new_cache/db/mod.rs
+++ b/crates/rspack_core/src/new_cache/db/mod.rs
@@ -3,10 +3,6 @@ mod noop;
 #[cfg(not(target_family = "wasm"))]
 mod turbo;
 
-use rspack_error::Result;
-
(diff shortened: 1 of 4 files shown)


ahabhgk:
@codex review

intellild (approved):

PR #15639, opened 2026-09-11
chore(deps): upgrade swc_core to 78.0.0

chenjiahan:
## Motivation

Update SWC dependencies while keeping the native binding buildable.

## Changes

Upgrade `swc_core` to `78.0.0`, `swc_ecma_minifier` to `61.0.6`, and `@swc/plugin-remove-console` to `14.0.0`.

Pin `swc_core` and `swc` (`76.0.0`) because the `78.0.1` / `76.0.1` release pair fails to compile: `swc` references an AMD `module_root` field missing from its published dependencies.

Preserve the release size optimizations for the renamed `react_compiler` crates.

## Related Links

- [Upstream AMD moduleRoot change](https://github.com/swc-project/swc/pull/11964)
- [@swc/plugin-remove-console 14.0.0](https://www.npmjs.com/package/@swc/plugin-remove-console/v/14.0.0)


chenjiahan:
### Bundle size changes

The changes come from SWC’s updated minification behavior:

- **React benchmarks: +59 B each.** SWC now preserves unused `Set` construction and array spreading when it cannot prove iteration is free of side effects.
- **popular-libs: +129 B.** More conservative property hoisting keeps React Hook Form’s validation rules as an object instead of separate variables.
- **ui-components: −46 B.** Improved Unicode escape handling produces smaller Chinese template strings in Vant.

The React and popular-libs deltas were reproduced with identical minifier input.

### Binary size changes

The stripped Linux x64 CI binding increased by **372 KiB (0.53%)**.

The main cause is switching from the React Compiler fork to the official packages. The fork preserved unsupported AST nodes directly; the official implementation uses a JSON round trip, bringing AST deserialization code back into the binary. SWC minifier updates also contribute.

The increase includes **229 KiB of machine code**, plus relocation tables, unwind metadata, and constants. Existing size optimization settings were preserved. This measures the CI profile; release builds use fat LTO and may differ.

### Ecosystem CI failures

Both failures are unrelated to this SWC upgrade:

- **examples:** The HTTP import lockfile records a URL that differs from the final redirect URL. Base-branch PR #15609 enabled frozen lockfiles by default in production and added stricter validation, exposing the stale entry.
- **rspress:** `skills-package-manager install` failed with `ENOTEMPTY` while removing a temporary Git directory, before the Rspress build started.

PR #15642, opened 2026-09-11
refactor(javascript): optimize inner graph analysis

LingyuCoder:
## Summary

- Allocate module-local, contiguous `TopLevelSymbol` IDs and store symbol data, usage operations and graph state in a single `Vec`. Keep slot zero for the global symbol; do not depend on SWC symbol IDs being dense.
- Resolve named declaration bindings through the existing Semantic data and reuse the immediate resolution when attaching tags. Preserve mutable aliases/tags, scope restoration and name-based fallback for synthetic names or inactive scope owners.
- Borrow and copy the symbol stored in a tag instead of cloning a temporary boxed payload.
- Reuse drained HashSet capacity during usage propagation and borrow the global edge set when extending other symbols. Keep HashSet throughout; no Small/Large collection variant or propagation-algorithm rewrite.

- Remove the separate `pure_declarators` HashSet and derive the pure-expression dependency gate from the initializer when processing a tracked declarator. Preserve the function/literal exclusions and the separate class-initializer path.

Evaluator payloads and dependency source locations are outside this PR. Existing test assertions and snapshots are unchanged.

## Performance

Local Docker + Valgrind, native Linux ARM64, Rust nightly 2026-04-16, Valgrind 3.19.0, fixed fixture and `rust@scan_dependencies@three_module`:

| Version | Instructions |
| --- | ---: |
| Release base / #15626 | 29,446,715 |
| Before pure-declarator cleanup (`6325d6427c`) | 28,646,837 |
| This PR, including pure-declarator cleanup | 28,636,993 |
| Total reduction vs release | **809,722 / 2.75%** |

The pure-declarator cleanup saves an additional **9,844 instructions / 0.034%** relative to the previous PR head: a small incremental improvement, not a substantial new speedup. It also removes 26 net lines of code.

Each implementation version was measured once. The previous measurements were reused after verifying source and environment equivalence: the release base (`0e963bc1af`) matches the measured #15626 tree, and the previous PR head (`6325d6427c`) matches all 23,148 tracked files in its measured snapshot. The new measurement uses the same image, tool versions, fixture and build settings, with only the two InnerGraph source files changed. This is scan instruction count, not full-build wall time.

## Validation

- Development binding build.
- Full `test:unit`: 9,229 Rspack tests and 90 CLI tests passed; existing skips unchanged; binding TypeScript checks passed.
- `cargo test -p rspack_plugin_javascript --lib --locked`: 48 passed.
- `cargo clippy -p rspack_plugin_javascript --all-targets --locked -- --deny warnings`.
- `cargo fmt --all --check` and `cargo shear --locked --deny-warnings -p rspack_plugin_javascript`.
- Standalone differential check of the before/after propagation function: 20,040 randomized and structured graph cases, including cycles, self-edges, dense fan-out, global states and deferred pure checks. Dependency outputs, deferred-check multisets and final graph states matched.

The pure-declarator follow-up was rebuilt and the full unit suite, Rust tests, Clippy, format and cargo shear passed again on the current source. The earlier 20,040-case differential check covers the propagation implementation, which this follow-up does not change.

## Related links

- Follow-up to #15331, #15595 and #15626.
- Base branch: `release/2.3.0`.

## Checklist

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/inner_graph/plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/inner_graph/plugin.rs
@@ -606,20 +606,6 @@ impl<'p, 'a> JavascriptParserPlugin<'p, 'a> for InnerGraphParserPlugin {
           .inner_graph
           .decl_with_top_level_symbol
           .insert(decl.span(ast), v);
-
-        if !matches!(
-          ast.expr_data(init),
-          ExprData::Function(_)
-            | ExprData::ArrowFunctionExpression(_)
-            | ExprData::StringLiteral(_)
-            | ExprData::NumericLiteral(_)
-            | ExprData::BigIntLiteral(_)
-            | ExprData::BooleanLiteral(_)
-            | ExprData::NullLiteral(_)
-            | ExprData::RegExpLiteral(_)
-        ) {
-          parser.inner_graph.pure_declarators.insert(decl.span(ast));
-        }
       }
     }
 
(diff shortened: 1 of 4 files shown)


LingyuCoder:

LingyuCoder:
@codex review

CPunisher (approved):

PR #15656, opened 2026-09-12
fix: drop trailing comma when rewriting require.resolve calls

### LINKED ISSUE #15655: require.resolve() with a trailing comma emits invalid JavaScript
### System Info

```
rspack 1.7.11
node 26.8.1
macOS 26.6.2 (arm64)
```

### Details

`require.resolve(...)` written as a multi-line call with a trailing comma compiles to invalid JavaScript. rspack replaces only the `require.resolve` callee with a comment, leaving the argument list as a parenthesized expression — and a parenthesized expression may not contain a trailing comma, so the emitted bundle throws `SyntaxError: Unexpected token ')'` when it is parsed.

Input:

```js
// src/index.js
const id = require.resolve(
  "./foo.js",
);
console.log(id);
```

Emitted:

```js
const id = /*require.resolve*/(
  "./src/foo.js",
);
console.log(id);
```

`new vm.Script(output)` on that bundle fails with `SyntaxError: Unexpected token ')'`.

Dropping the trailing comma in the source produces valid output, so the bug is purely the loss of call-expression semantics during the rewrite. Since prettier and biome both add that trailing comma by default, published packages hit it — we ran into it with `mocha@11`, whose `lib/nodejs/parallel-buffered-runner.js` contains:

```js
const DEFAULT_WORKER_REPORTER = require.resolve(
  "./reporters/parallel-buffered",
);
```

Bundling any module that reaches that file makes the whole output unparseable.

### Reproduce link

Minimal reproduction (no repo needed):

```
mkdir -p /tmp/rspack-repro/src && cd /tmp/rspack-repro
printf 'module.exports = 1;\n' > src/foo.js
printf 'const id = require.resolve(\n  "./foo.js",\n);\nconsole.log(id);\n' > src/index.js
cat > rspack.config.js <<'CFG'
module.exports = {
  mode: "development",
  devtool: false,
  target: "node",
  entry: "./src/index.js",
  output: { path: __dirname + "/dist", filename: "out.js" },
};
CFG
npm i @rspack/core@1.7.11
npx rspack build
node dist/out.js
```

### Reproduce Steps

1. Run the commands above.
2. `node dist/out.js` fails with `SyntaxError: Unexpected token ')'`.
3. Inspect `dist/out.js` — the `require.resolve` call has been rewritten to `/*require.resolve*/(\n  "./src/foo.js",\n)`.

Expected: the whole call expression is replaced by the resolved module id, as webpack does, so the trailing comma disappears with it.

perbergland:
## Summary

`require.resolve(...)` is compiled by replacing only the callee with a `/*require.resolve*/` comment and the argument with the resolved module id. That leaves the original argument list behind as a parenthesized expression:

```js
// in
const id = require.resolve(
  "./foo.js",
);

// out (before this PR)
const id = /*require.resolve*/(
  "./src/foo.js",
);
```

A trailing comma is valid in a call argument list but a syntax error inside a parenthesized expression, so the emitted bundle fails to parse with `SyntaxError: Unexpected token ')'` — taking down the whole chunk, not just that module. Prettier and biome both add that trailing comma to multi-line calls by default, so published packages hit it; the reporter ran into it via `mocha@11`'s `lib/nodejs/parallel-buffered-runner.js`.

This PR rewrites everything between the end of the argument and the end of the call to a bare `)` whenever that span contains a comma:

```js
// out (after this PR)
const id = /*require.resolve*/(
  "./src/foo.js");
```

The call has exactly one argument by the time this runs, so any comma in that span is either the trailing comma or inside a comment that is fine to drop along with it.

`import.meta.resolve` goes through the same callee-to-comment rewrite and had the identical defect, so it is fixed in the same place via a shared helper in `visitors::dependency::util`. `require()` and `import()` are unaffected — the former keeps call syntax (`__webpack_require__(...)`), the latter replaces the whole expression.

### On webpack parity

This is a deliberate divergence from webpack, which has the same defect (reproduced on webpack 5.105.0 and 5.110.3). The issue's original "as webpack does" expectation was retracted in a [follow-up comment](https://github.com/web-infra-dev/rspack/issues/15655#issuecomment-5644639304). The output otherwise keeps webpack's `/*require.resolve*/(<id>)` shape — only the trailing comma goes away.

## Testing

New config case `tests/rspack-test/configCases/parsing/require-resolve-trailing-comma` covering `require.resolve`, `require.resolveWeak`, `import.meta.resolve`, a context (non-static) request, and a trailing comma preceded by a comment.

Verified it fails before the fix with the reported `SyntaxError: Unexpected token ')'` and passes after, and that the issue's own repro now runs instead of throwing. The full `base` project suite passes.

## Related links

- Closes #15655

## Checklist

---

### DIFF
--- a/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/common_js_imports_parse_plugin.rs
@@ -38,6 +38,7 @@ use crate::{
     CallHooksName, ExportedVariableInfo, JavascriptParser, StatementPath, TagInfoData,
     VariableDeclaration, VariableDeclarationKind, VariableInfo, VariableInfoFlags, context_reg_exp,
     create_context_dependency, create_traceable_error, expr_name, get_non_optional_part,
+    resolve_call_trailing_comma_range,
   },
 };
 
@@ -1565,6 +1566,10 @@ impl CommonJsImportsParserPlugin {
     let loc = parser.to_dependency_location(range);
     let require_resolve_header_dependency =
       BoxDependency::new(RequireResolveHeaderDependency::new(range, loc));
+
+    if let Some(range) = resolve_call_trailing_comma_range(parser, call_expr) {
+      parser.add_presentational_dependency(Arc::new(ConstDependency::new(range, ")".into())));
+    }
 
     if param.is_conditional() {
       for option in param.options() {
--- a/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
+++ b/crates/rspack_plugin_javascript/src/parser_plugin/import_meta_plugin.rs
@@ -39,6 +39,7 @@ use crate::{
     AllowedMemberTypes, ExportedVariableInfo, ExprRef, JavascriptParser, MemberExpressionInfo,
     RootName, context_reg_exp, create_context_dependency, create_traceable_error, expr_name,
     get_non_optional_member_chain_from_expr, member_property_to_atom,
+    resolve_call_trailing_comma_range,
   },
 };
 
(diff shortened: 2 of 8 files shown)


LingyuCoder:
@codex review

LingyuCoder (approved):
LGTM, thanks

PR #15658, opened 2026-09-13
perf: add a mimalloc allocation fast path

LingyuCoder:
## Motivation

Avoid the overhead of mimalloc's aligned allocation API for allocations with ordinary alignment requirements, and make the cached-source hash benchmarks measure actual lookups.

## Changes

Implement the shared mimalloc adapter in `rspack_allocator`. Route requests with alignment up to `align_of::<usize>()` and no larger than the allocation size through `mi_malloc` / `mi_zalloc`, retaining the aligned fallback and existing deallocation and reallocation behavior. Register the binding's global allocator in `rspack_binding_api`, preserving its allocator selection, while benchmarks register their own wrapper around the shared implementation.

Separate cached-source lookups from single and batched HashMap lifecycles. Execute both lookups and include allocation and destruction in the lifecycle measurements, using new benchmark names for the changed workloads.


LingyuCoder:
@codspeedbot explain the regression of sources@cached_source_hash

PR #15670, opened 2026-09-14
fix(rslib): render externalized new URL targets as plain relative URLs

elecmonkey:
## Motivation

Under `parser.javascript.url: 'new-url-relative'`, `URLDependencyTemplate` emits a placeholder that `replace_static_url_placeholders` resolves from the target's `CodeGenerationDataFilename`. External modules carry no such data, so an externalized `new URL()` target hits `unreachable!()` and panics. #14900 fixed the same problem for `new Worker(new URL(...))`.

## Changes

New `crates/rspack_plugin_rslib/src/url_external.rs`, modelled on `worker_external.rs`: `ExternalURLDependencyTemplate` writes the external request literal into the `new URL(...)` arguments, and `cutout_url_externals` marks those connections inactive so the target stays out of the chunk graph. Limited to `JavascriptParserUrl::NewUrlRelative`, since the other modes render `__webpack_require__(id)` and need the module in the graph.


elecmonkey:
@codex review

JSerFeng (approved):

PR #15674, opened 2026-09-14
refactor(core): share BuildContext across module graph tasks

intellild:
## Motivation

Module graph tasks repeatedly carry the same compiler services as separate fields and assemble them into factory inputs and build contexts. This duplicates plumbing and introduces avoidable handle clones and per-module context construction.

For example, when an entry imports `a.js` and `b.js`, both task chains use the same compiler options, resolver factory, plugin driver, and filesystem services. They can share one context created by `TaskContext`, while keeping each module's dependencies and build state separate.

## Changes

- Construct an `Arc<BuildContext>` in `TaskContext` and share it through factorization, factory creation, and module build tasks.
- Make `ModuleFactoryCreateData` hold the shared build context and add a constructor for dependency-specific inputs, preserving the existing context, request, and issuer-layer selection.
- Change `Module::build` to accept `&BuildContext`, including module implementations and the runtime module macro. `BuildContext` does not implement `Clone`, preventing accidental whole-context copies.
- Keep `RunnerContext` unchanged. Clone the cache facade and filesystem-info handle only when transferring them into the loader runner. The build-time runtime template is read-only; code generation retains its separate mutable template.

This is an internal refactor with no intended bundling behavior changes.

### Validation

- `cargo fmt --all --check` and `git diff --check`.
- Clippy with `-D warnings` for core and the affected plugin crates.
- `pnpm run build:binding:dev`.
- Focused base test runs reported 287 passed, covering loader/importModule, context modules, layers, issuers, externals, module federation, DLLs, memory-cache recovery and reuse, module concatenation, CSS loaders, and RSC concatenation. Storage and native watcher tests were excluded.

---
_by OpenAI Codex_


intellild:
@codex review

ahabhgk (approved):

PR #15686, opened 2026-09-15
perf(watch): defer internal Watchpack timestamp collection

### LINKED ISSUE #15680: [Bug]: MultiCompiler extracts Watchpack timestamp maps twice for the same rebuild
## Versions and problem

Reproduced with Rspack 2.2.4 / its bundled Watchpack on Node 24.19.0 and Linux. In a large client/server development graph, one source edit produces three compilations (client, server, then a client update for server-data revalidation). It performs **six full timestamp-map sweeps: three in the aggregate callback and three in `getInfo()`**.

In one real three-edit capture, counters observed 18 sweeps (9 aggregate + 9 getInfo), taking about 3049 ms total wall time across the three edits. Watcher rearming separately took about 3644 ms. These are nested diagnostic spans, not additive end-to-end CPU totals. No actual-application latency improvement has been measured for the proposed timestamp change.

## Causal chain

1. [`NodeWatchFileSystem.ts`](https://github.com/web-infra-dev/rspack/blob/v2.2.4/packages/rspack/src/node/NodeWatchFileSystem.ts) pauses its Watchpack instance after `aggregated`, purges changed/removed filesystem cache entries, and eagerly calls `fetchTimeInfo()` / `collectTimeInfoEntries()` before invoking its callback.
2. [`MultiCompiler.ts`](https://github.com/web-infra-dev/rspack/blob/v2.2.4/packages/rspack/src/MultiCompiler.ts) coordinates pending/blocked compilers. [`Watching.ts`](https://github.com/web-infra-dev/rspack/blob/v2.2.4/packages/rspack/src/Watching.ts) accumulates changes, but a blocked `#invalidate()` does not consume the supplied timestamp Maps.
3. When the scheduler admits the compiler, `#go()` has no timestamp Maps and calls `pausedWatcher.getInfo()`. That extracts another full fresh pair of Maps.

`collectTimeInfoEntries()` traverses the watched files/directories rather than just the changed paths. Existing directory timestamps can be cached, so this is **not** a claim that every file is newly `stat`ed on every sweep. Large dependency/context graphs still incur substantial traversal and Map allocation.

## Proposed boundary

Keep the current Watchpack backend, filesystem purging, collected change/removal sets and scheduling. Defer timestamp extraction until the existing `Watching.#go()` consumption boundary. It already obtains fresh Maps through `pausedWatcher.getInfo()` when maps are absent. Do not memoize Maps across events or assume paused watchers cannot observe additional changes.

The source draft below adds a private, non-enumerable Symbol capability to callbacks created by `Watching`:

- `NodeWatchFileSystem` uses the capability to deliver changed/removed Sets without eagerly building Maps.
- Ordinary callback and `watch()` signatures remain compatible.
- A custom watch filesystem replacing/wrapping the callback gets the original eager Map contract because its replacement has no private capability.
- A transparent forwarding wrapper preserves the capability and can use the optimization.

Files changed: `packages/rspack/src/util/fs.ts`, `packages/rspack/src/node/NodeWatchFileSystem.ts`, and `packages/rspack/src/Watching.ts`.

A callback wrapper deliberately copying all descriptors/Symbols could copy the private capability and bypass its newly wrapped callback. The exact internal capability design needs maintainer review; arbitrary reflection-based wrappers are not claimed compatible.

## Behavioral evidence

A real two-compiler fixture uses at most six modules per compiler and covers ten cases: initial build, shared edit through MultiCompiler scheduling, context file add/delete, missing import create/delete, burst edits, an edit while a compiler is held with its watcher paused, synthetic invalidate, and a web-only revalidation-file analogue.

| Rspack 2.2.4 fixture | Result records | Aggregate sweeps | getInfo sweeps | Total |
| --- | ---: | ---: | ---: | ---: |
| Original direct | 20 | 17 | 19 | 36 |
| Deferred direct | 20 | 0 | 19 | 19 |
| Deferred through custom callback wrapper | 20 | 17 | 19 | 36 |

These are results from the exact public standalone script below, executed on Node 24.19.0 against isolated stock 2.2.4 and copies with the included bundled patch applied without fuzz. All ten cases and final outputs passed; each arm rearmed 21 times. The wrapped case asserted Maps on all 17 legacy callbacks. All watchers and compilers closed and all three processes exited successfully. Burst/event aggregation can change callback counts between runs; no fixture wall-time speedup is claimed.

The fixture's revalidation-file analogue is not actual Router loader-data/generation validation. Full application behavior, timestamp accuracy transitions, symlinks/platform differences and broader custom filesystem integrations still need coverage. The candidate has not been installed in the application.

## Minimal reproduction

In a new standalone directory containing the script and bundled patch below:

```sh
pnpm add -D @rspack/core@2.2.4
node watch-timeinfo-repro.mjs direct
pnpm patch @rspack/core@2.2.4 --edit-dir .rspack-edit
(cd .rspack-edit && patch --fuzz=0 -p1 < ../watch-timeinfo.patch)
pnpm patch-commit .rspack-edit
node watch-timeinfo-repro.mjs direct
node watch-timeinfo-repro.mjs wrapped
```

The standalone script contains its own counter hooks and has been runtime-verified in all three arms against unpacked 2.2.4 copies. The included bundled patch was applied with `--fuzz=0`; the package-manager installation commands themselves were not rerun in this isolated validation. Counter hooks forward the original callback unchanged and never request timestamp maps merely to inspect them. The TypeScript source draft was parsed but has not received a full upstream typecheck.

The runtime results are scoped to version 2.2.4. Source inspection of current [`main` NodeWatchFileSystem.ts](https://github.com/web-infra-dev/rspack/blob/main/packages/rspack/src/node/NodeWatchFileSystem.ts) on 2026-09-14 (file blob `6c51d66e379e6fe10e26e94c6cea637e226cf739`) still shows eager timestamp extraction in the aggregate callback. The full reproduction has not been run on `main`.

<details>
<summary>watch-timeinfo-repro.mjs</summary>

```js
// Self-contained Rspack 2.2.4 / Watchpack regression fixture (six modules/compiler).
import fs from 'node:fs/promises';
import path from 'node:path';
import { createRequire } from 'node:module';
import assert from 'node:assert/strict';
import os from 'node:os';
const [mode = 'direct'] = process.argv.slice(2);
const variant = 'installed';
if (!['direct','wrapped'].includes(mode)) throw Error('Expected direct or wrapped');
const root = await fs.mkdtemp(path.join(os.tmpdir(),'rspack-watch-timeinfo-'));
const source = path.join(root,'src');
await fs.mkdir(path.join(source,'extras'),{recursive:true});
for (const filename of ['optional.js','extras/new.js']) await fs.rm(path.join(source,filename),{force:true});
await fs.writeFile(path.join(source,'shared.js'), "module.exports = 'initial';\n");
await fs.writeFile(path.join(source,'hdr.js'), "module.exports = 'hdr-initial';\n");
await fs.writeFile(path.join(source,'extras/base.js'), "module.exports = 'base';\n");
const entry = `const shared = require('./shared.js');
let optional = 'absent'; try { optional = require('./optional.js'); } catch {}
const extras = require.context('./extras', false, /\\.js$/);
module.exports = { shared, optional, extras: extras.keys().sort(), hdr: HDR };
`;
await fs.writeFile(path.join(source,'web.js'),entry.replace('HDR',"require('./hdr.js')"));
await fs.writeFile(path.join(source,'node.js'),entry.replace('HDR',"'node'"));
const resolve = createRequire(path.join(process.cwd(),'package.json'));
const corePath = resolve.resolve('@rspack/core');
const rspack=await import(corePath); const require=createRequire(corePath);
const Watchpack = require('../compiled/watchpack/index.js');
const counts = { aggregate: 0, getInfo: 0, other: 0, rearms: 0 };
let origin = 'other';
function scope(kind, fn, self, args) {
  const previous = origin; origin = kind;
  try { return Reflect.apply(fn, self, args); } finally { origin = previous; }
}
const timeout = Watchpack.prototype._onTimeout;
Watchpack.prototype._onTimeout = function (...args) { return scope('aggregate', timeout, this, args); };
const collect = Watchpack.prototype.collectTimeInfoEntries;
Watchpack.prototype.collectTimeInfoEntries = function (...args) { counts[origin]++; return Reflect.apply(collect, this, args); };
const rearm = Watchpack.prototype.watch;
Watchpack.prototype.watch = function (...args) { counts.rearms++; return Reflect.apply(rearm, this, args); };
const compilerWatch = rspack.Compiler.prototype.watch;
rspack.Compiler.prototype.watch = function (...args) {
  const fileSystem = this.watchFileSystem, watch = fileSystem.watch;
  fileSystem.watch = function (...watchArgs) {
    const handle = Reflect.apply(watch, this, watchArgs);
    const getInfo = handle.getInfo;
    handle.getInfo = function (...infoArgs) { return scope('getInfo', getInfo, this, infoArgs); };
    return handle;
  };
  return Reflect.apply(compilerWatch, this, args);
};
const api = { async start() {}, async stop() { await fs.writeFile(path.join(root,'counter-result.json'),JSON.stringify(counts,null,2)); } };
await api.start(root);
const results={variant,mode,cases:[],builds:[],wrappedCallbacks:0,maxModules:0};
const latest=new Map(); let fatal; let hold; let watcher;
const sleep=ms=>new Promise(resolve=>setTimeout(resolve,ms));
async function until(predicate,label) {
  const deadline=Date.now()+15000;
  while(Date.now()<deadline){if(fatal)throw fatal;if(predicate())return;await sleep(20)}
  throw Error(`Timed out: ${label}; latest=${JSON.stringify([...latest])}`);
}
function recordStats(stats) {
  for(const child of stats.stats??[stats]) {
    const name=child.compilation.name;
    results.maxModules=Math.max(results.maxModules,child.compilation.modules.size);
    if(child.hasErrors()){fatal=Error(JSON.stringify(child.toJson({all:false,errors:true}).errors));return}
    const file=path.join(root,'out',name,'main.js');delete require.cache[file];
    const value=require(file); latest.set(name,value);
    results.builds.push({name,value,modified:[...(child.compilation.compiler.modifiedFiles??[])].map(p=>path.relative(source,p)),removed:[...(child.compilation.compiler.removedFiles??[])].map(p=>path.relative(source,p))});
  }
}
const plugin={apply(compiler){
  if(mode==='wrapped') {
    const original=compiler.watchFileSystem;
    compiler.watchFileSystem={watch(...args){
      const callback=args[5];
      args[5]=(...values)=>{
        if(!values[0]){assert(values[1] instanceof Map);assert(values[2] instanceof Map);results.wrappedCallbacks++}
        return callback(...values);
      };
      return original.watch(...args);
    }};
  }
  compiler.hooks.beforeCompile.tapPromise('fixture-hold', async()=>{
    if(hold&&compiler.name==='web'&&!hold.entered){hold.entered=true;await hold.promise}
  });
}};
const configs=['web','node'].map(name=>({name,context:source,entry:`./${name}.js`,mode:'development',target:'node',devtool:false,cache:false,
  output:{path:path.join(root,'out',name),filename:'main.js',library:{type:'commonjs2'}},
  infrastructureLogging:{level:'none'},plugins:[plugin],watchOptions:{aggregateTimeout:15},
}));
const compiler=rspack.rspack(configs);
try {
  watcher=compiler.watch({aggregateTimeout:15},(error,stats)=>{if(error)fatal=error;else try{recordStats(stats)}catch(e){fatal=e}});
  await until(()=>latest.size===2,'initial');
  assert(results.maxModules<100); results.cases.push('initial');
  async function bothShared(value){await until(()=>['web','node'].every(name=>latest.get(name)?.shared===value),value);await sleep(80)}
  await fs.writeFile(path.join(source,'shared.js'),"module.exports = 'edit';\n");await bothShared('edit');results.cases.push('multicompiler-shared-edit');
  await fs.writeFile(path.join(source,'extras/new.js'),"module.exports = 'new';\n");await until(()=>['web','node'].every(name=>latest.get(name)?.extras.includes('./new.js')),'context-add');await sleep(80);results.cases.push('context-add');
  await fs.unlink(path.join(source,'extras/new.js'));await until(()=>['web','node'].every(name=>!latest.get(name)?.extras.includes('./new.js')),'context-delete');await sleep(80);results.cases.push('context-delete');
  await fs.writeFile(path.join(source,'optional.js'),"module.exports = 'present';\n");await until(()=>['web','node'].every(name=>latest.get(name)?.optional==='present'),'missing-add');await sleep(80);results.cases.push('missing-to-present');
  await fs.unlink(path.join(source,'optional.js'));await until(()=>['web','node'].every(name=>latest.get(name)?.optional==='absent'),'missing-delete');await sleep(80);results.cases.push('present-to-missing');
  await fs.writeFile(path.join(source,'shared.js'),"module.exports = 'burst-one';\n");await fs.writeFile(path.join(source,'shared.js'),"module.exports = 'burst-final';\n");await bothShared('burst-final');results.cases.push('burst');
  let release;hold={entered:false,promise:new Promise(resolve=>{release=resolve})};hold.release=()=>release();
  await fs.writeFile(path.join(source,'shared.js'),"module.exports = 'held-first';\n");await until(()=>hold.entered,'held compile');
  await fs.writeFile(path.join(source,'shared.js'),"module.exports = 'held-final';\n");await sleep(60);release();await bothShared('held-final');hold=undefined;results.cases.push('edit-while-compiler-paused');
  const previous=results.builds.length;await new Promise((resolve,reject)=>watcher.invalidate(error=>error?reject(error):resolve()));await until(()=>results.builds.length>previous,'synthetic invalidation');results.cases.push('synthetic-invalidate');
  await fs.writeFile(path.join(source,'hdr.js'),"module.exports = 'hdr-next';\n");await until(()=>latest.get('web')?.hdr==='hdr-next','web-only HDR analogue');await sleep(80);results.cases.push('web-only-hdr-file');
  results.final=Object.fromEntries(latest);assert(results.maxModules<100);results.status='passed';
} finally {
  hold?.release();
  if(watcher)await new Promise((resolve,reject)=>watcher.close(error=>error?reject(error):resolve()));
  await new Promise((resolve,reject)=>compiler.close(error=>error?reject(error):resolve()));
  await api.stop();
  await fs.writeFile(path.join(root,'fixture-result.json'),JSON.stringify(results,null,2));
}
console.log(JSON.stringify({root,...results,builds:results.builds.length,counts}));
```

</details>

<details>
<summary>upstream-source.patch</summary>

```diff
--- a/packages/rspack/src/util/fs.ts
+++ b/packages/rspack/src/util/fs.ts
@@ -700,6 +700,23 @@
   timestamp?: number;
 }
 
+// Only callbacks created by Watching opt into deferred timestamp extraction.
+// A custom watch filesystem wrapping the callback retains the eager contract.
+export const deferredWatchCallback = Symbol('deferredWatchCallback');
+
+export type WatchCallback = ((
+  error: Error | null,
+  fileTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
+  contextTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
+  changedFiles: Set<string>,
+  removedFiles: Set<string>,
+) => void) & {
+  [deferredWatchCallback]?: (
+    changedFiles: Set<string>,
+    removedFiles: Set<string>,
+  ) => void;
+};
+
 export interface WatchFileSystem {
   watch(
     files: Iterable<string> & {
@@ -716,13 +733,7 @@
     },
     startTime: number,
     options: WatchOptions,
-    callback: (
-      error: Error | null,
-      fileTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
-      contextTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
-      changedFiles: Set<string>,
-      removedFiles: Set<string>,
-    ) => void,
+    callback: WatchCallback,
     callbackUndelayed: (fileName: string, changeTime: number) => void,
   ): Watcher;
 
--- a/packages/rspack/src/node/NodeWatchFileSystem.ts
+++ b/packages/rspack/src/node/NodeWatchFileSystem.ts
@@ -12,6 +12,7 @@
 import { createRequire } from 'node:module';
 import util from 'node:util';
 import type Watchpack from 'watchpack';
+import { deferredWatchCallback, type WatchCallback } from '../util/fs';
 
 import type {
   FileSystemInfoEntry,
@@ -46,13 +47,7 @@
     missing: Iterable<string>,
     startTime: number,
     options: Watchpack.WatchOptions,
-    callback: (
-      error: Error | null,
-      fileTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
-      contextTimeInfoEntries: Map<string, FileSystemInfoEntry | 'ignore'>,
-      changedFiles: Set<string>,
-      removedFiles: Set<string>,
-    ) => void,
+    callback: WatchCallback,
     callbackUndelayed: (fileName: string, changeTime: number) => void,
   ): Watcher {
     if (!files || typeof files[Symbol.iterator] !== 'function') {
@@ -119,6 +114,11 @@
           fs.purge?.(item);
         }
       }
+      const deferred = callback[deferredWatchCallback];
+      if (deferred) {
+        deferred(changes, removals);
+        return;
+      }
       const { fileTimeInfoEntries, contextTimeInfoEntries } = fetchTimeInfo();
 
       callback(
--- a/packages/rspack/src/Watching.ts
+++ b/packages/rspack/src/Watching.ts
@@ -12,7 +12,7 @@
 import type { Compilation, Compiler } from '.';
 import { Stats } from '.';
 import type { WatchOptions } from './config';
-import type { FileSystemInfoEntry, Watcher } from './util/fs';
+import { deferredWatchCallback, type FileSystemInfoEntry, type WatchCallback, type Watcher } from './util/fs';
 
 type PendingWatchDelta = { added: Set<string>; removed: Set<string> };
 
@@ -117,7 +117,7 @@
       missing,
       this.lastWatcherStartTime,
       this.watchOptions,
-      (
+      Object.defineProperty<WatchCallback>((
         err,
         fileTimeInfoEntries,
         contextTimeInfoEntries,
@@ -138,7 +138,13 @@
           removedFiles,
         );
         this.onChange();
-      },
+      }, deferredWatchCallback, {
+        value: (changedFiles: Set<string>, removedFiles: Set<string>) => {
+          // #go reads current timestamps when the scheduler admits this build.
+          this.#invalidate(undefined, undefined, changedFiles, removedFiles);
+          this.onChange();
+        },
+      }),
       (fileName, changeTime) => {
         if (!this.#invalidReported) {
           this.#invalidReported = true;
```

</details>

<details>
<summary>watch-timeinfo.patch</summary>

```diff
--- a/dist/index.js
+++ b/dist/index.js
@@ -9351,6 +9351,7 @@
         ];
     }
 }
+const deferredWatchCallback = Symbol('deferredWatchCallback');
 class NodeWatchFileSystem {
     inputFileSystem;
     watcherOptions;
@@ -9390,6 +9391,8 @@
                 for (let item of changes)fs.purge?.(item);
                 for (let item of removals)fs.purge?.(item);
             }
+            const deferred = callback[deferredWatchCallback];
+            if (deferred) return deferred(changes, removals);
             let { fileTimeInfoEntries, contextTimeInfoEntries } = fetchTimeInfo();
             callback(null, fileTimeInfoEntries, contextTimeInfoEntries, changes, removals);
         }), this.watcher?.watch({
@@ -12008,10 +12011,12 @@
         });
     }
     watch(files, dirs, missing) {
-        this.pausedWatcher = void 0, this.watcher = this.compiler.watchFileSystem.watch(files, dirs, missing, this.lastWatcherStartTime, this.watchOptions, (err, fileTimeInfoEntries, contextTimeInfoEntries, changedFiles, removedFiles)=>{
+        this.pausedWatcher = void 0, this.watcher = this.compiler.watchFileSystem.watch(files, dirs, missing, this.lastWatcherStartTime, this.watchOptions, Object.defineProperty((err, fileTimeInfoEntries, contextTimeInfoEntries, changedFiles, removedFiles)=>{
             if (err) return this.compiler.fileTimestamps = void 0, this.compiler.contextTimestamps = void 0, this.compiler.modifiedFiles = void 0, this.compiler.removedFiles = void 0, this.handler(err);
             this.#invalidate(fileTimeInfoEntries, contextTimeInfoEntries, changedFiles, removedFiles), this.onChange();
-        }, (fileName, changeTime)=>{
+        }, deferredWatchCallback, { value: (changedFiles, removedFiles)=>{
+            this.#invalidate(void 0, void 0, changedFiles, removedFiles), this.onChange();
+        } }), (fileName, changeTime)=>{
             this.#invalidReported || (this.#invalidReported = !0, this.compiler.hooks.invalid.call(fileName, changeTime)), this.onInvalid();
         });
     }
```

</details>

LingyuCoder:
## Motivation

Fixes #15680.

`NodeWatchFileSystem` eagerly collects Watchpack timestamp Maps when an aggregate event arrives. If `Watching` is suspended or waiting for MultiCompiler scheduling, it retains the changed/removed paths but discards those Maps. Starting the rebuild then collects fresh Maps again through `getInfo()`, duplicating traversal and allocation.

## Changes

Defer timestamp collection only for callbacks created and registered internally by `Watching`. A module-private `WeakSet` records callback identity without adding properties to the function or introducing another notification callback. `NodeWatchFileSystem` skips eager collection for these callbacks and invokes the same callback with absent timestamps. The existing `Watching.#go()` path obtains current timestamps through `getInfo()` when compilation starts.

### API compatibility

- Keep the public `WatchFileSystem.watch()` signature unchanged. External callers of `compiler.watchFileSystem.watch()` still receive real file and context timestamp Maps in their callbacks.
- A wrapper forwarding the original internal callback can retain the optimization. Replacing the callback with a new function falls back to eager Maps, including wrappers that copy all property descriptors and Symbols.
- Preserve filesystem purging, change/removal accumulation, and scheduling. Read fresh timestamps at compilation start instead of caching Maps across events, since paused Watchpack instances can still record changes.
- Leave the native watcher backend unchanged.

### Validation

- Built JS packages/type declarations and the development native binding; focused lint and formatting checks passed.
- 19 watch-related Compiler/MultiCompiler tests passed, including six new cases for immediate rebuilds, suspend/resume, queued late changes and fresh timestamps, ordinary and property-copying wrappers, and direct external `watchFileSystem.watch()` calls.
- Disabling the optimization made the queued and suspended regression cases fail on the extra eager collection, while the compatibility cases continued to pass.
- Ran the issue's standalone fixture against the local source build in direct and wrapped modes. All ten scenarios passed in each mode. Direct mode recorded 0 aggregate collections and 19 `getInfo()` collections. On macOS the fixture used realpath-normalized temporary paths and a wait after the initial build. No application-level latency improvement is claimed.


stormslowly (approved):
Verified the root cause: `Watching.#invalidate` drops the eagerly collected timestamp Maps on every early return, and MultiCompiler hits that path unconditionally because `isBlocked` (MultiCompiler.ts:453) returns true for an idle `'done'` node, while the resume path `watching.invalidate()` carries no timestamps. So the aggregate-time sweep is 100% wasted there.

Deferring rather than memoizing is the right call — a paused Watchpack keeps recording into `aggregatedChanges` (`_onChange` writes regardless of `paused`), so cached Maps would go stale. Using callback identity instead of a call-site flag is also the right trade-off: a wrapper that replaces the callback falls back to the eager path automatically, and a WeakSet is not defeated by descriptor-copying wrappers the way a Symbol property would be.

Pushed one follow-up commit renaming the helpers, since they track callback identity rather than performing any deferral:
- `deferWatchTimeInfo` → `markInternalCallback`
- `canDeferWatchTimeInfo` → `isInternalCallback`
- `DeferredWatchCallback` → `OptionalTimeInfoWatchCallback`

PR #15688, opened 2026-09-15
fix: prevent premature multi compiler done notifications

### LINKED ISSUE #15682: [Bug]: MultiCompiler done accounting misses a child restart without a new invalid hook
## Problem

MultiCompiler resets a child's completed flag on `invalid`, but not on `watchRun`. If a child starts another compilation after its previous `done` without a new `invalid` hook, MultiCompiler can publish aggregate `done` while that child is running, using the child's previous Stats.

The reproduction executes Rspack 2.2.4's actual `MultiCompiler` and `MultiStats` classes with controlled child hooks and Stats payloads. It demonstrates the constructor's accounting behavior. It does **not** run the watch scheduler or a native compiler, so reachability through a real watch build remains unproven by this fixture.

## Tested environment

Rspack 2.2.4, Node.js 24.19.0. No dev server or browser is involved.

## Reproduction

Save the script below as `repro.mjs` in an empty directory:

```sh
npm init -y
npm install --save-dev @rspack/core@2.2.4
node repro.mjs
```

After both children complete an initial build, the script emits this schedule:

1. Web `invalid`; Node `invalid`.
2. Web `watchRun`; Node `watchRun`.
3. Web `invalid` while running; Web `done`.
4. Web `watchRun` again, with no intervening `invalid`.
5. Node `done`.
6. Web `done`.

**Actual:** aggregate `done` fires at step 5 with the previous Web Stats, then again at step 6.

**Expected for this schedule:** after Web starts again, it should no longer count as complete; publish the combined result at step 6.

## Candidate change and result

In the [MultiCompiler constructor](https://github.com/web-infra-dev/rspack/blob/v2.2.4/packages/rspack/src/MultiCompiler.ts#L128-L150), share the existing idempotent reset between `invalid` and `watchRun`:

```diff
-      compiler.hooks.invalid.tap('MultiCompiler', () => {
+      const markNotDone = () => {
         if (compilerDone) {
           compilerDone = false;
           doneCompilers--;
         }
-      });
+      };
+      compiler.hooks.invalid.tap('MultiCompiler', markNotDone);
+      compiler.hooks.watchRun.tap('MultiCompiler', markNotDone);
```

The script compares the installed baseline with this one-block transformation loaded in memory. For the schedule above, the baseline has one premature publication and the candidate has none. Initial-build, ordinary single-child update, and duplicate-`invalid` controls produce identical results in both versions. All assertions passed in the original run.

This is a tested candidate for the controlled hook schedule, not a validated fix for a native watcher race. Real-compiler coverage is still needed for dependent children, limited parallelism, edits during compilation, errors and close/re-watch. Resetting on `watchRun` also does not account for a queued rerun that has not started yet; it does not make aggregate `done` a guarantee that no future work is pending.

<details>
<summary>Self-contained Node 24 reproduction</summary>

```js
import assert from 'node:assert/strict';
import { createHash } from 'node:crypto';
import { readFile, writeFile } from 'node:fs/promises';
import { createRequire, registerHooks } from 'node:module';
import { pathToFileURL } from 'node:url';

// With Node 24, run after a local install of @rspack/core@2.2.4.
// Optional existing package resolution location is never written to results.
const require = createRequire(process.env.RSPACK_PACKAGE_JSON || import.meta.url);
const corePath = require.resolve('@rspack/core');
const version = require('@rspack/core/package.json').version;
assert.equal(version, '2.2.4');
const original = await readFile(corePath, 'utf8');
const digest = value => createHash('sha256').update(value).digest('hex');
const before = `            }), compiler.hooks.invalid.tap('MultiCompiler', ()=>{
                compilerDone && (compilerDone = !1, doneCompilers--);
            });`;
const after = `            });
            const markNotDone = ()=>{
                compilerDone && (compilerDone = !1, doneCompilers--);
            };
            compiler.hooks.invalid.tap('MultiCompiler', markNotDone);
            compiler.hooks.watchRun.tap('MultiCompiler', markNotDone);`;
assert.equal(original.split(before).length - 1, 1, 'Exact baseline block must match once');
const baseline = await import(pathToFileURL(corePath).href);
const candidateURL = `${pathToFileURL(corePath).href}?done-reset-candidate`;
let transformed = 0;
const loader = registerHooks({
  load(url, context, nextLoad) {
    const result = nextLoad(url, context);
    if (url !== candidateURL) return result;
    const source = typeof result.source === 'string' ? result.source : Buffer.from(result.source).toString();
    assert.equal(source, original, 'Only the exact installed module may be transformed');
    transformed++;
    return { ...result, source: source.replace(before, after) };
  },
});
let candidate;
try {
  candidate = await import(candidateURL);
} finally {
  loader.deregister();
}
assert.equal(transformed, 1);
assert.notEqual(baseline.MultiCompiler, candidate.MultiCompiler);

// Controlled fake child hooks. The actual installed MultiCompiler and MultiStats
// classes execute; no Watching, watch graph, or native compilation is invoked.
class ControlledHook {
  taps = [];
  tap(name, callback) { this.taps.push({ name, callback }); }
  call(...args) { for (const { callback } of this.taps) callback(...args); }
}
function runCase(MultiCompiler, name, actions) {
  const states = Object.fromEntries(['web', 'node'].map(name => [name, {
    name, building: false, generation: 0,
    hooks: Object.fromEntries(['done', 'invalid', 'run', 'watchRun', 'beforeCompile', 'shutdown', 'infrastructureLog'].map(name => [name, new ControlledHook()])),
  }]));
  const multi = new MultiCompiler(Object.values(states));
  const events = [];
  const publications = [];
  multi.hooks.done.tap('Capture', stats => {
    publications.push({
      at: events.length,
      activeChildren: Object.values(states).filter(child => child.building).map(child => child.name),
      generations: Object.fromEntries(stats.stats.map(stat => [stat.name, stat.generation])),
    });
  });
  for (const [childName, type] of actions) {
    const child = states[childName];
    events.push({ child: childName, type });
    if (type === 'watchRun') {
      assert.equal(child.building, false);
      child.building = true;
      child.generation++;
      child.hooks.watchRun.call(child);
    } else if (type === 'done') {
      assert.equal(child.building, true);
      child.building = false;
      child.hooks.done.call({ name: childName, generation: child.generation });
    } else {
      assert.equal(type, 'invalid');
      child.hooks.invalid.call('component.js', 1);
    }
  }
  assert(Object.values(states).every(child => !child.building));
  return { name, events, publications };
}
const initial = [['web', 'watchRun'], ['node', 'watchRun'], ['web', 'done'], ['node', 'done']];
const normal = [...initial, ['web', 'invalid'], ['web', 'watchRun'], ['web', 'done']];
const duplicateInvalid = [...initial, ['web', 'invalid'], ['web', 'invalid'], ['web', 'watchRun'], ['web', 'done'], ['node', 'invalid'], ['node', 'watchRun'], ['node', 'done']];
const silentRestart = [...initial,
  ['web', 'invalid'], ['node', 'invalid'], ['web', 'watchRun'], ['node', 'watchRun'],
  ['web', 'invalid'], ['web', 'done'], ['web', 'watchRun'], ['node', 'done'], ['web', 'done'],
];
const cases = { initial, normal, duplicateInvalid, silentRestart };
const report = {
  rspack: version, node: process.version,
  implementation: 'Actual installed MultiCompiler and MultiStats; controlled fake child hooks and Stats payloads',
  candidate: 'Exact one-block in-memory module transformation; only watchRun done-flag reset added',
  installedSourceSha256: digest(original),
  nativeCompilationInvoked: false, watchGraphInvoked: false,
  arms: Object.fromEntries([['baseline', baseline], ['candidate', candidate]].map(([name, module]) => [name,
    Object.fromEntries(Object.entries(cases).map(([caseName, actions]) => [caseName, runCase(module.MultiCompiler, caseName, actions)])),
  ])),
};
for (const [name, count] of [['initial', 1], ['normal', 2], ['duplicateInvalid', 3]]) {
  assert.deepEqual(report.arms.baseline[name], report.arms.candidate[name]);
  assert.equal(report.arms.baseline[name].publications.length, count);
  assert(report.arms.baseline[name].publications.every(item => item.activeChildren.length === 0));
}
assert.deepEqual(report.arms.baseline.silentRestart.publications, [
  { at: 4, activeChildren: [], generations: { web: 1, node: 1 } },
  { at: 12, activeChildren: ['web'], generations: { web: 2, node: 2 } },
  { at: 13, activeChildren: [], generations: { web: 3, node: 2 } },
]);
assert.deepEqual(report.arms.candidate.silentRestart.publications, [
  { at: 4, activeChildren: [], generations: { web: 1, node: 1 } },
  { at: 13, activeChildren: [], generations: { web: 3, node: 2 } },
]);
assert.equal(await readFile(corePath, 'utf8'), original, 'Installed source must remain identical');
report.assertionsPassed = true;
report.installedSourceUnchanged = true;
await writeFile(new URL('./results.json', import.meta.url), JSON.stringify(report, null, 2) + '\n');
console.log(JSON.stringify({ assertionsPassed: true, controlsIdentical: 3,
  baselinePrematurePublications: 1, candidatePrematurePublications: 0,
  installedSourceUnchanged: true }));
```

</details>

LingyuCoder:
## Motivation

Fixes #15682.

MultiCompiler can emit aggregate `done` with an outdated child's Stats while that child is rebuilding. The completion counter only resets on `invalid`, which can be deduplicated when the watch graph requeues an outdated build.

## Changes

Keep the existing constructor-level aggregation and child hook order. Reset each child's completion flag whenever the graph starts it, even without another public `invalid` notification. Check watcher ownership before resetting so that scheduling a closed dependent cannot erase its last completion. Likewise, preserve completion when a busy child rejects a concurrent run. Before publishing aggregate `done`, also check the graph for pending, blocked, queued or outdated nodes.

This preserves asynchronous child `done` interleaving, aggregate error attribution, and the original `failed`, callback and `afterDone` paths. Independently started child compilers still produce aggregate notifications, including after the graph's watchers have closed. Failed graphs immediately release their aggregation check, and closed or replaced watchers are excluded individually while remaining active watchers still constrain publication. No changes to `Compiler` or `Watching` are needed.

Validation:

- Native binding and JavaScript builds passed; source lint and formatting checks passed.
- Full `test:unit` passed, including all 49 MultiCompiler tests.
- Regression cases cover outdated rebuilds, limited parallelism, dependencies, reentrant invalidation, repeated runs, watch recovery/restart, partial watch setup failures followed by independent child runs, partially closed outdated watchers with other watchers still active, and closing during asynchronous hooks.
- Exact event-order tests cover normal completion and aggregate failures in both run and watch modes, including reversed asynchronous child completion order and independently started children. Standalone reproductions also match the baseline event sequences.


LingyuCoder:

### PR Overview

Bug fix for #15682. `MultiCompiler` could publish an aggregate `done` carrying an outdated child's `Stats` while that child was already rebuilding: the constructor-level `doneCompilers` counter is reset only by the child `invalid` hook, and `Watching` deduplicates `invalid` notifications while the watch graph requeues an outdated build. The change installs a per-graph readiness gate (`#isGraphReady`), resets a child's completion flag when a graph restart actually starts it, and releases the gate on graph failure and after run-mode completion. It adds `tests/rspack-test/multiCompilerCases/aggregate-done.js` with 23 cases covering rebuilds, parallelism, dependencies, reentrant invalidation, watch recovery/restart, partial watch-setup failures, closed/detached watchers, and exact event ordering. CI is green for lint and unit tests on Linux/macOS/Windows (Node 22/24); CodSpeed reports no performance change; review by hardfist is requested.

### Problem and Trigger

`Watching` fires a child's `invalid` hook at most once per cycle (`#invalidReported`), so a child can go `done -> pending -> blocked -> running` (or `running -> running-outdated -> blocked -> running`) without another public `invalid`. Its completion flag stays set, and when the last remaining child finishes, `doneCompilers === this.compilers.length` publishes `hooks.done` with the rebuilding child's stale `Stats`. The reproduction in #15682 exercises exactly this silent-restart path.

### Previous Design and Root Cause

Aggregation was purely counter-based in the constructor: `done` increments, `invalid` decrements, and publication ignores the run graph's node states entirely. The counter cannot know that a child is queued to rebuild, so deduplicated-invalid restarts produced premature publications. webpack main still uses the same counter-only aggregation with no readiness gate, so this PR is deliberate hardening beyond webpack rather than a port of an upstream fix.

### Fix Design

- `#resetCompilerDone: WeakMap<Compiler, () => void>` holds the per-child reset. The graph's `run` closure invokes it when the graph starts the node — in watch mode only when the graph still owns the watcher (`compiler.watching === watching`), and in run mode only when `!compiler.running` so a rejected concurrent run retains the active run's completion.
- `#isGraphReady` is installed per graph and consulted by the constructor's `done` tap: publication requires `!errored` and every graph-owned node in `done | running | starting`; nodes with detached/undefined watchers are exempt because they no longer represent work owned by the graph.
- The gate is released on graph error and after run-mode completion, with a closure-identity guard so a callback that starts another graph cannot clear the new gate.

Tracing the state machine confirms the main paths are correct: the gate blocks exactly when the counter is "stale-full" while a node is `pending/blocked/queued/running-outdated`, and in every graph-owned path that suppressed publication is followed by a fresh one after the rebuild (the `run` closure resets the flag before the rebuild starts). The interleaving with asynchronous child `done` hooks and the `failed`, callback, and `afterDone` ordering is preserved, and the added exact-event-order tests match the traced sequences.

### Correctness Risks

1. **A detached middle node in a dependency chain can silence aggregate `done` indefinitely (main risk).** The gate exempts only nodes whose *own* watcher is detached. Consider `A <- B <- C` (B depends on A, C depends on B). Close B's watcher, then invalidate A: the cascade marks B and C `blocked`. When A's `done` tap brings the counter to `compilers.length`, the gate blocks on C (`blocked`, watcher still graph-owned). `nodeDone(A)` enqueues B, but the watch-mode `run` closure early-returns for the detached B — no reset, no rebuild, and `running` stays unbalanced — so B is stuck in `running`, C is never enqueued, no further `done` taps occur, and `running` never returns to 0. The publication is skipped and nothing ever publishes again; the multi watch handler also stops firing. In this configuration no child is actually rebuilding (B is closed by user intent and C's inputs are unchanged), so the baseline publication was semantically fine here and this change converts it into permanent silence. The added "closed dependent watcher" case only closes a *leaf* dependent, so this is uncovered. Suggested fix: treat a detached node as terminal at dequeue time — mark it `done` keeping its last `result`, decrement `running`, and enqueue its children — or exempt nodes transitively blocked behind a detached ancestor in `isGraphReady`. (The unbalanced `running` on the early-return path predates this PR, but the gate now amplifies it into suppressed publications.)

2. **`starting` counts as ready.** This is only safe because the flag reset runs synchronously inside the graph's `run` closure before any hook can interleave, so no `done` tap can fire while a node is `starting` with a full counter. Harmless in practice, but excluding `starting` from the ready set (or a comment stating the synchronous-window invariant) would be stricter and self-documenting.

3. **Run-mode release sits inside `if (stats.length > 0)`.** On today's success path `stats.length` is always positive, so the gate is always released; moving `releaseGraph()` next to the `running === 0 && all done` check would be more robust if a future path ever yields zero stats.

4. **Minor:** after `MultiWatching.close()` the stale gate closure stays installed until the next graph replaces it. The detachment exemption makes it behaviorally benign, but releasing it on close would be easier to reason about.

5. **Nit:** `aggregate-done.js` is CJS while its siblings are `.mjs`; the harness picks it up fine, but matching the directory convention would be cleaner. Extracting the detached-watcher check in `isGraphReady` into a named helper would also improve readability.

The added per-publication cost is one `nodes.every(...)` (O(number of children)) plus one WeakMap lookup per graph restart — negligible and off the hot path. CodSpeed reports 50 untouched benchmarks with no alteration, and the size-limit check is unchanged. No material performance risk.

<details>

### PR 概述

针对 #15682 的 bug 修复。`MultiCompiler` 可能在某个子编译器已开始重建时，仍用其过期 `Stats` 发布聚合 `done`：构造函数层的 `doneCompilers` 计数器只由子编译器 `invalid` 钩子复位，而 `Watching` 会去重 `invalid` 通知，同时 watch 图会重新入队过期构建。该改动为每个图安装就绪门控（`#isGraphReady`），在图重启真正启动某个子节点时复位其完成标记，并在图失败及 run 模式完成后释放门控。新增 `tests/rspack-test/multiCompilerCases/aggregate-done.js`，共 23 个用例，覆盖重建、并发度、依赖、重入式 invalidate、watch 恢复/重启、部分 watch 初始化失败、已关闭/脱离的 watcher 以及精确事件顺序。CI 在 Linux/macOS/Windows（Node 22/24）上 lint 与单测全绿；CodSpeed 报告性能无变化；已请求 hardfist 审查。

### 问题与触发条件

`Watching` 每个周期最多触发一次子编译器 `invalid` 钩子（`#invalidReported`），因此子编译器可能在没有新的公开 `invalid` 的情况下经历 `done -> pending -> blocked -> running`（或 `running -> running-outdated -> blocked -> running`）。其完成标记保持为 true，当最后一个子编译器完成时，`doneCompilers === this.compilers.length` 就会携带正在重建的子编译器的过期 `Stats` 发布 `hooks.done`。#15682 中的复现正是这一静默重启路径。

### 原设计与根因

原实现完全基于构造函数中的计数器聚合：`done` 递增、`invalid` 递减，发布条件完全无视运行图的节点状态。计数器无法知道"某个子节点已排队等待重建"，因此去重 invalid 的重启会产生过早发布。webpack main 目前仍是同样的纯计数器聚合（没有就绪门控），所以本 PR 是在 webpack 之上的一次有意加固，而非移植上游修复。

### 修复设计

- `#resetCompilerDone: WeakMap<Compiler, () => void>` 保存每个子编译器的复位闭包。图的 `run` 闭包在图启动该节点时调用它——watch 模式下仅在图仍拥有该 watcher 时（`compiler.watching === watching`），run 模式下仅在 `!compiler.running` 时，从而被拒绝的并发 run 不会抹掉活跃 run 的完成状态。
- 每个图安装一个 `#isGraphReady`，构造函数中的 `done` 钩子发布前检查它：要求 `!errored` 且每个由图拥有的节点处于 `done | running | starting`；watcher 已脱离或为 undefined 的节点被豁免，因为它们不再代表图拥有的工作。
- 图出错和 run 模式完成时释放门控，并以闭包身份比较防止"回调中再启动新图"误清新图的检查。

对状态机的推演确认主路径正确：门控恰好在"计数器已满但某节点处于 `pending/blocked/queued/running-outdated`"时拦截，且在所有由图拥有的路径中，被拦截的发布都会由重建后的新鲜发布补上（`run` 闭包在重建开始前复位标记）。与异步子 `done` 钩子及 `failed`、callback、`afterDone` 的交错顺序保持不变，新增的精确事件顺序用例与推演序列一致。

### 正确性风险

1. **依赖链中间的脱离节点可能永久静默聚合 `done`（主要风险）。** 门控只豁免自身 watcher 已脱离的节点。考虑 `A <- B <- C`（B 依赖 A，C 依赖 B）。关闭 B 的 watcher 后 invalidate A：级联会把 B、C 置为 `blocked`。当 A 的 `done` 钩子把计数推到 `compilers.length` 时，门控因 C（`blocked` 且 watcher 仍归图所有）而拦截。`nodeDone(A)` 入队 B，但 watch 模式的 `run` 闭包对脱离的 B 提前返回——不复位、不重建，`running` 也不平衡——于是 B 永远卡在 `running`，C 永远不会入队，之后不再有 `done` 钩子触发，`running` 也回不到 0。这次发布被跳过且之后再无任何发布；multi watch handler 同样停止触发。在该配置下其实没有任何子编译器正在重建（B 是用户主动关闭，C 的输入也未变化），因此基线版本在这里的发布语义上是正确的，本改动把正确发布变成了永久静默。新增的"关闭依赖子 watcher"用例只关闭了叶子依赖，未覆盖此场景。建议修复：在出队时把脱离节点视为终态——置为 `done` 并保留最后 `result`，递减 `running`，再入队其子节点；或在 `isGraphReady` 中豁免"被脱离祖先传递阻塞"的节点。（提前返回路径上 `running` 不平衡在改动前就存在，但门控现在把它放大成了发布被吞。）

2. **`starting` 被当作就绪态。** 这之所以安全，仅因为标记复位在图 `run` 闭包内同步执行、任何钩子都无法穿插；实际上不存在"计数已满且某节点处于 `starting` 时触发 `done` 钩子"的窗口。实践中无害，但把 `starting` 从就绪集合中排除（或加注释说明同步窗口不变式）会更严格、更自明。

3. **run 模式的释放写在 `if (stats.length > 0)` 内部。** 当前成功路径上 `stats.length` 恒为正，所以门控总能被释放；把 `releaseGraph()` 移到 `running === 0 && 所有节点 done` 判定旁会更稳健，防止未来出现零 stats 路径。

4. **小问题**：`MultiWatching.close()` 之后旧图闭包仍挂着，直到下一个图替换它。脱离豁免使其行为上无害，但在 close 时释放会更易推理。

5. **细节**：`aggregate-done.js` 是 CJS，而同目录其他文件都是 `.mjs`；harness 能正常发现它，但沿用目录约定会更整洁。把 `isGraphReady` 中的脱离检查提取为具名辅助函数也能提升可读性。

### 性能回归风险

每次发布检查新增的开销是一次 `nodes.every(...)`（O(子编译器数)）加每次图重启一次 WeakMap 查找，均可忽略且不在热路径上。CodSpeed 报告 50 个基准未受影响/无变化，size-limit 检查无变化。没有实质性性能风险。

</details>

PR #15689, opened 2026-09-15
fix(rstest): resolve manual mocks next to package entries

9aoy:
## Motivation

Rstest does not discover manual mocks next to resolved package entries when mocking a package without a factory.

For example, given this structure:

```text
project/
├── node_modules/
│   └── pkg/
│       ├── package.json       # main: "dist/index.js"
│       └── dist/
│           ├── index.js
│           └── __mocks__/
│               └── index.js
└── example.test.js
```

```js
import { value } from 'pkg';

rs.mock('pkg');
```

Previously, Rstest looked for `<root>/__mocks__/pkg` and fell back to automocking when it was missing. It did not discover `pkg/dist/__mocks__/index.js`.

Related: https://github.com/web-infra-dev/rstest/issues/1850

## Changes

When a non-relative request has no project-level manual mock, look for a mock next to the resolved module entry before falling back to automocking.

For `rs.mock('pkg')`, the lookup order is:

```text
<root>/__mocks__/pkg.js
  → <resolved package>/dist/__mocks__/index.js
  → automock
```

The project-level mock still takes precedence when both exist.

For adjacent mocks, first resolve the entry's full filename (including `resolve.extensionAlias`). Only if it is missing, retry with the file stem using `resolve.extensions`. For example, an `index.mjs` entry tries `__mocks__/index.mjs`, then `__mocks__/index`. This preserves format-specific mocks while allowing `.mjs` and `.cjs` entries to share `__mocks__/index.js`. This extension fallback is an Rstest-specific addition to the Jest-style adjacent lookup.

This also supports linked workspace packages:

```text
project/
├── node_modules/
│   └── pkg → ../packages/pkg
└── packages/
    └── pkg/
        ├── package.json       # main: "src/index.js"
        └── src/
            ├── index.js
            └── __mocks__/
                └── index.js
```

With no project-level mock, `rs.mock('pkg')` uses `packages/pkg/src/__mocks__/index.js`.

Source aliases use the same fallback. If `@/foo` resolves to `<root>/src/foo.js`:

```text
rs.mock('@/foo')

<root>/__mocks__/@/foo.js
  → <root>/src/__mocks__/foo.js
  → automock
```

Relative requests retain their existing lookup behavior:

```text
rs.mock('./src/foo')

<root>/src/__mocks__/foo.js
  → automock
```

No configuration option is added. The fallback checks the resolved entry's immediate `__mocks__` directory; it does not search ancestor directories. Mock resolution errors other than missing files remain visible.

Regression tests cover npm packages, workspace symlinks, source aliases, package subpaths, root mock precedence, and missing-mock fallback.

Validated with the Rspack mock config cases and Clippy. The original adjacent lookup behavior was independently verified with seven passing Jest 30.2.0 tests; the cross-extension fallback is not claimed as Jest behavior. Added Rspack fixtures cover `.mjs` and `.cjs` fallback and format-specific mock precedence. The CommonJS fixture explicitly selects the `require` condition for mock target dependencies, which currently use the ESM dependency category even for `mockRequire`.


crates/rspack_plugin_rstest/src/plugin.rs
```diff
@@ -255,16 +255,46 @@ impl RstestPlugin {
...
+      )
+    {
+      let (resolve_result, resolve_dependencies) = resolver
+        .resolve_with_context(data.context.as_ref(), stripped)
+        .await;
+      data.add_file_dependencies(resolve_dependencies.file_dependencies);
+      data.add_missing_dependencies(resolve_dependencies.missing_dependencies);
+
+      if let Ok(ResolveResult::Resource(resource)) = resolve_result
+        && let (Some(parent), Some(file_name)) = (resource.path.parent(), resource.path.file_name())
```
fi3ework, line 286: The sibling lookup uses `file_name()`, so the mock must carry the same extension as the resolved entry (`dist/index.js` → `dist/__mocks__/index.js`). That matches Jest's `existsSync` check, but it is the only extension-sensitive lookup in RstestPlugin:

- `./foo` → `__mocks__/foo`
- `./dir` → `dir/__mocks__/index` (via `resolve_directory_mock_target`, which joins `main_file` without an extension)
- `pkg` → `<manualMockRoot>/pkg`

All three defer the extension to `resolve.extensions`. Rstest also sets `extensionAlias['.js'] = ['.js', '.ts', '.tsx']`, so today an `index.js` entry already matches `__mocks__/index.ts`, while an `index.mjs` / `index.cjs` entry only matches the exact extension. That asymmetry shows up with dual packages: `exports` conditions resolve to `index.mjs` or `index.cjs` depending on the dependency category, and users would need one mock file per condition.

Suggest using `file_stem()` here so the rule is the same for every path ("same name as the mocked entry, extension resolved by `resolve.extensions`"):

```rust
&& let (Some(parent), Some(stem)) = (resource.path.parent(), resource.path.file_stem())
{
  resolved_request = parent.join("__mocks__").join(stem).to_string();
```

Every file the current exact match finds is still found (`.ts .tsx .mjs .js .jsx .json .cjs` are all in Rstest's default list); the only loss is `.mts` / `.cts` entries, which users can restore by extending `resolve.extensions`. Worth a fixture for an `.mjs`/`.cjs` entry with a single `__mocks__/index.js` either way.

9aoy: Updated in 01c87e4ec5. The adjacent lookup now tries the full filename first (with the existing extensionAlias behavior), then retries with file_stem() through resolve.extensions only on a missing-file result. Other resolution errors do not trigger the retry.

Keeping the first lookup preserves format-specific mocks: when both __mocks__/index.mjs and __mocks__/index.js exist, an index.mjs entry still selects index.mjs regardless of resolve.extensions order. When the format-specific mock is absent, .mjs and .cjs entries can share index.js.

Added fixtures for both extensions and for exact-match precedence. The mock config cases and Clippy pass. One separate limitation surfaced while testing: mockRequire target dependencies currently use the ESM category, so the CommonJS fixture explicitly selects the require condition for that category. This change covers extension fallback, not that existing condition-selection issue.

fi3ework (approved):

PR #15694, opened 2026-09-15
perf(core): back BoxModule with UniqueArc

intellild:
## Motivation

Publishing a built module converts `Box<dyn Module>` into `Arc<dyn Module>`, allocating again and moving the module. For example, a `NormalModule` finishes loader execution in a `Box` and pays this cost when it is published into the module graph.

## Changes

- Back `BoxModule` with `UniqueArc<dyn Module>`, preserving unique ownership throughout `Module::build` and `RunnerContext`. Publish into `Arc<dyn Module>` or `ModuleRef` with `UniqueArc::into_arc`, retaining the allocation.
- Add opt-in `#[cacheable_dyn(unique_arc)]` deserialization. Archived modules construct `UniqueArc<ConcreteModule>` directly, without an intermediate boxed module.
- Route both `BoxModule` and `ModuleRef` through the same `deserialize_unique` entry point. The `AsUniqueArc` adapter preserves Arc's archive format and registers restored allocations in the rkyv pool, so repeated references still share one module.
- Generate only the unique deserialization interface for opted-in traits. Remove their old raw-pointer deserializers and keep pointer metadata for shared reference restoration. Default `cacheable_dyn` behavior is unchanged.
- Keep `ModuleExt` as an extension trait rather than a `Module` supertrait, avoiding unnecessary boxed conversion methods in module vtables.
- Preserve the boxed archive representation of `BoxModule` and adapt field borrows without cloning module data.

Example allocation flow:

```text
NormalModule::boxed() / archived module deserialization
  allocate UniqueArc<NormalModule> once
       ↓
BoxModule
  mutate during Module::build / RunnerContext
       ↓
ModuleRef::from / Arc::from
  publish the same allocation via UniqueArc::into_arc

ModuleRef cache restoration
  same deserialize_unique entry point
       ↓
UniqueArc::into_arc + rkyv pool
  repeated references reuse the restored allocation
```

For example, restoring two cached `ModuleRef` handles to one module now uses the same unique factory as `BoxModule` and returns two Arc handles to one allocation. This removes the second set of module restoration functions from the binary.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --locked --offline -- --deny warnings`
- `cargo test --locked --offline -p rspack_cacheable_test` — 49 passed.
- `cargo test --locked --offline -p rspack_macros_test --test box_module --test runtime_module` — 5 passed.
- Tests cover allocation identity, mutation after cache restoration, exactly one owning allocation for aligned dynamic objects, shared reference identity, pool ownership/release, and rejection of pending cyclic references.
- Built the native binding with the CI profile, target, features, trim-paths and unwind-table flags; compared stripped binaries in the same source/build directories:
  - baseline `fe573cf`: 72,233,832 bytes
  - before unified restoration: 72,279,784 bytes (**+44.875 KiB**)
  - this version: 72,263,144 bytes (**+28.625 KiB**)
  - unified restoration saves **16.25 KiB**. All 80 old module `deserialize_dyn` functions are absent; the 83 unique factories retain the same total code size.
- The local native toolchain differs from CI's cross toolchain; CI will confirm its size.
- Ran Compiler, MultiCompiler and Hook JS integration suites against the rebuilt binding: **170 passed, 4 skipped**. Loaded the binding only in the workspace package to avoid overriding the test runner's separate Rspack version.

Storage and native watcher tests were not run.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15696, opened 2026-09-15
refactor(core): share compilation build context

intellild:
## Motivation

`Compilation` and module build tasks keep duplicate copies of the same build inputs. Each module graph task context and concatenated-module build reconstructs a `BuildContext`, cloning its handles and creating another module template. For example, compiling many concatenated module groups creates a separate context for each group even though they belong to the same compilation.

## Changes

- Store one `Arc<BuildContext>` on `Compilation` and move its compiler/compilation IDs, options, input filesystem, filesystem snapshot service, resolver factory, and plugin driver into that context. Expose borrowed accessors for existing consumers to keep the migration small; retain direct access where disjoint field borrows are required.
- Reuse the compilation's context in module graph tasks and concatenated-module builds.
- Derive a fresh context for each new compilation, retaining compiler-wide inputs while resetting the compilation ID and filesystem snapshots. Rebuilds must not reuse cached file metadata from the previous compilation.
- Give build-time module execution its own derived context with the build-time plugin driver and execution runtime template. Keep the parent context unchanged and continue dispatching `executeModule` through the parent plugin driver.

## Validation

- `cargo clippy --workspace --all-targets --locked -- --deny warnings`
- `cargo fmt --all --check` and `git diff --check`
- `pnpm run build:js` and `pnpm run build:binding:dev`
- Focused Rstest runs against the rebuilt binding: 42 executed import-module cases, 48 concatenation cases, 2 runtime-mode module-execution cases, and 3 HMR cases passed. The import-module group also reported 2 case-filter placeholders. Coverage includes parallel/recursive imports and both standard and Rspack runtime modes.
- Storage and native-watcher tests were not run.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15731, opened 2026-09-16
perf: use SsoHashSet for small module chunk memberships

SyMind:
## Motivation

`ChunkGraphModule::chunks` allocates a hash table even when a module belongs to only one or a few chunks. Inspired by rustc's [SsoHashSet](https://doc.rust-lang.org/beta/nightly-rustc/rustc_data_structures/sso/set/struct.SsoHashSet.html), keep these small memberships inline.

In the [main CodSpeed baseline](https://app.codspeed.io/web-infra-dev/rspack/runs/6aaaa6551a58cd679b6e0f78), `connect_chunk_and_module` accounts for 7.14% of the `rust@build_chunk_graph` simulation profile; chunk-set insertion accounts for 3.85%, including 2.99% in hash-table growth.

## Changes

- Introduce reusable `rspack_collections::SsoHashSet<T>`, storing up to four elements inline and promoting to `FxHashSet` on the fifth distinct insertion. Chunk graph and plugin consumers use `SsoHashSet<ChunkUkey>` directly, without a `ModuleChunks` alias. Large sets retain their allocation on removal/clear to avoid repeated promotion.
- Support non-`Copy`, non-`Clone` elements by moving them during promotion. Add borrowed lookup/removal, `Extend`, and `FromIterator`.
- Preserve membership checks, deduplication, removal, subset checks, and borrowed/owned iteration. Iteration order remains unspecified.
- Retain the same representation in module-concatenation and SplitChunks snapshots. Convert to a hash set only when preparing a chunk combination. `Clone` is needed to preserve memberships while mutating the graph and to retain independent concatenation snapshots; small clones do not allocate.
- Adapt the split-chunks and ESM-library consumers without changing JavaScript APIs or configuration.
- Cover 1/4/5/9 memberships, promotion, and removal/reconnection through SplitChunks in the existing config-case harness. The case uses `shared.js` directly as its entry and validates memberships in compilation hooks, without a separate runtime test entry.

Validation of the generic implementation (latest alias removal additionally rebuilt the binding and passed focused Clippy, formatting, 2 membership and 202 SplitChunks regression tests):
- Rebuilt the development native binding; JavaScript packages were already built and are unchanged.
- On the current 64-bit target, `SsoHashSet<ChunkUkey>` remains 32 bytes and `ChunkGraphModule` remains 96 bytes.
- New regression case: 2 harness tests passed (8 configurations in each runtime mode).
- SplitChunks config cases: 202 passed.
- Scope-hoisting config cases: 20 passed.
- Chunk HMR cases: 77 passed.
- Rust formatting and focused Clippy with all targets and warnings denied passed.

Remote validation for `0372199` completed successfully: [PR CI](https://github.com/web-infra-dev/rspack/actions/runs/35169654647), [dispatched simulation and walltime CI](https://github.com/web-infra-dev/rspack/actions/runs/35169656542), and the CodSpeed performance check passed.

[Latest CodSpeed comparison against main](https://app.codspeed.io/web-infra-dev/rspack/runs/compare/6aaaa6551a58cd679b6e0f78..6aab4276b36bf4a2cf12fd45):

| Benchmark | Mode | Base → head | Result |
| --- | --- | --- | --- |
| `rust@build_chunk_graph` | CPU simulation | 26.2 → 25.2 ms | +3.58% efficiency in the PR report |
| `rust@split_chunks` | CPU simulation | 1.6 → 1.6 ms (rounded) | About -1%; below the 2% threshold, classified untouched |
| `rust@create_concatenate_module` | CPU simulation | 13.7 → 13.5 ms | About +1%; classified untouched |
| `bundle@threejs-10x-development` | Walltime | 199.3 → 197.5 ms | Classified untouched |
| `bundle@threejs-10x-production-sourcemap` | Walltime | 468.2 → 460.7 ms | Classified untouched |

The latest PR simulation report contains 1 improvement, 49 untouched benchmarks, 47 skipped benchmarks, and no detected regressions. The combined simulation/walltime comparison contains 1 improvement and 96 untouched benchmarks, with no detected regressions. Simulation values are simulated timings, not elapsed walltime; the 3.58% improvement applies to the chunk-graph benchmark, not total bundling time.

[Compared with the previous specialized implementation at `77e7d7c`](https://app.codspeed.io/web-infra-dev/rspack/runs/compare/6aab2dd48f7bb17f4d1b9066..6aab4276b36bf4a2cf12fd45), all 97 benchmarks are classified untouched. Generalizing the collection and removing the aliases introduced no detected significant regression.

CodSpeed flags different runtime environments for the comparison against main and identifies the self-hosted walltime environment as unknown, so small differences should be interpreted cautiously. The main comparison uses `d4cd073` because `f077888` has no benchmark run; their diff contains website files only.


JSerFeng:
@codspeedbot explain why performance improved

JSerFeng (approved):

PR #15735, opened 2026-09-17
fix(binding): release loader TSFNs on compiler close

SyMind:
## Motivation

Follow up on #15636 with regression coverage for libuv handle growth. The loader runner getter's raw TSFN was unreferenced but never released, leaving an async handle behind for each compiler. The last rebuild could also retain cached JS hook taps after close.

## Changes

- Manage the loader runner getter with a napi-rs weak TSFN and release it, together with the cached loader runner, from `JsLoaderRspackPlugin`'s close hook. Compiler lookup continues to use the existing weak-reference table.
- Clear cached JS hook taps in the binding's close finalizer, including when a close hook fails.
- Extend the existing TSFN lifecycle suite with isolated Node processes that count `process.report.getReport().libuv` handles, including inactive and unreferenced handles. Cover repeated compiler lifecycles, repeated builds, loader errors, skipped emit, and 4-versus-64-module scaling through a JS → builtin → JS loader chain.
- Verify the detector rejects deliberately retained unreferenced MessageChannel handles and observes recovery after closing them. Require three consecutive samples within the warmed baseline, with bounded cleanup waits and child-process timeouts.

These tests guard resource counts; they do not impose CPU, memory, or event-loop latency thresholds.

Validation:

- `pnpm run build:cli:dev`
- `pnpm --dir tests/rspack-test run test -t 'compilerCases/tsfn-lifecycle'`
- `pnpm --dir tests/rspack-test run test -t 'configCases/loader'`
- `cargo fmt --all --check`
- The lifecycle test reproduced an extra async handle before the fix; the rebuild test also exposed retained hook-cache handles.


SyMind:
@codex review

intellild (approved):

PR #15740, opened 2026-09-17
perf(javascript): share ESM resource identifiers per import statement

LingyuCoder:
## Motivation

Every ESM dependency builds its own resource identifier in `new()`, even when the dependencies come from the same import statement: one `ESMImportSideEffectDependency` per import plus one `ESMImportSpecifierDependency` per usage site. In React 10k that is 107,623 identifier constructions from ~40k import statements while only ~10k distinct request strings exist; `esm|react` alone is rebuilt about 10k times. Each construction allocates a `String`, copies the request text and looks the result up in the Ustr interner, and the result is identical for the whole statement.

## Changes

- Build the identifier once per import statement in the `import` hook and keep it on the parser, keyed by the declaration node id.
- Store it in `ESMSpecifierData` so every specifier dependency created from that binding reuses it.
- Pass the precomputed `ResourceIdentifier` into `ESMImportSideEffectDependency::new`, `ESMImportSpecifierDependency::new` and `ESMExportImportedSpecifierDependency::new` instead of rebuilding it. Export paths keep their existing per-statement/per-specifier construction.

The identifier content is produced from the same request, phase and attributes, so behavior is unchanged.

### Validation

- `cargo fmt --check`, `cargo clippy -p rspack_plugin_javascript --all-targets`, `cargo test -p rspack_plugin_javascript` (48 passed).
- `pnpm run build:binding:dev` then `pnpm run test:unit`: 48 test files, 9,344 passed / 4 skipped; CLI 90 passed; binding type test passed.
- Before/after builds emit byte-identical React 10k output (35 dist files, same sha256).

### Local IR result

React 10k, single worker, production, no minify, cache disabled. Fat-LTO symbolized AArch64 Linux diagnostic bindings built from clean `b8904ce6a5` and from the same source plus this change; Callgrind strict make interval (make entry to finishMake entry), one run per version, no warmups, self-cost sums match the Callgrind summaries. This is IR, not walltime.

| Metric | Baseline | Candidate | Change |
| --- | ---: | ---: | ---: |
| React 10k strict make IR | 8,229,247,912 | 8,203,722,712 | -25,525,200 (-0.310%) |
| `create_resource_identifier_for_esm_dependency` calls | 107,623 | 50,024 | -53.5% |
| that function's self IR | 16,651,421 | 3,823,784 | -77.0% |
| `Ustr::from` calls / self IR | 600,973 / 22.8M | 370,576 / 15.9M | -38.3% / -30.3% |

Largest family deltas are generic allocation (-9.17M), Atom/Ustr interning (-6.77M) and the JS plugin/walker (-4.39M); semantic and parser families are unchanged. Fat-LTO inlining reattribution shifts some symbol-level costs (for example `to_dependency_location` has an identical call count but +8.88M self IR), so the whole-make total is the load-bearing number. The measurement used `b8904ce6a5` as the parent baseline; this branch is rebased on the current `release/2.3.0`.


LingyuCoder:
image

@codspeedbot explain the regression of rust@create_chunk_hashes and the improvement of rust@scan_dependencies@three_module

CPunisher (approved):

PR #15744, opened 2026-09-17
test branch , do not merge

intellild:
## Motivation

Revert #15719 (`2b1dd37161ad269f6eb8bf988b1fd32717639c77`) on a test branch to exercise the native loader hook implementation again. **This draft is for testing only. Do not merge.**

For example, with `CssExtractRspackPlugin` tapping the loader hook, modules using only `builtin:swc-loader` (or no loaders) will again enter JavaScript for the hook. This intentionally restores the behavior, including the performance overhead, that #15719 mitigated.

## Changes

- Restore native loader hook registration, retained loader context state, and the `hook-before-loaders` regression cases.
- Remove the `hook-context-initialization` cases introduced by #15719, including their subsequently renamed `.mjs` configuration.
- Preserve subsequent parallel-loader custom `buildInfo` support while resolving the loader-runner import conflict.

## Validation

- `cargo fmt --all --check` passed.
- Restored loader and `.mjs` test configuration passed `node --check`.
- `git diff --cached --check` passed before committing.
- Build and integration tests could not be completed: `pnpm run build:cli:dev` triggered dependency installation, which encountered `ENOTFOUND` for `registry.npmjs.org`. Generated binding declarations are restored from the reverted commit and have not been regenerated locally.

---
_by OpenAI Codex_


intellild:
Confirmed the new benchmark works

PR #15749, opened 2026-09-17
perf(css-extract): set loader data during plugin apply

intellild:
## Motivation

`CssExtractRspackPlugin` registers a `thisCompilation` callback and a `NormalModule` loader hook just to put its runtime flag on each loader context. The loader runner already copies `compiler.options.loader` onto the context with `Object.assign`, including enumerable symbol properties.

For example, `new CssExtractRspackPlugin({ runtime: false })` currently assigns `{ runtime: false }` through the loader hook for every module. Storing that data once during `apply()` lets the existing context initialization copy it, avoiding the extra hook callback and per-module data allocation.

## Changes

- Write `{ runtime: this.options.runtime !== false }` directly to `compiler.options.loader[pluginSymbol]` during `apply()`.
- Remove the `thisCompilation` / `loader.tap` registration and unused `NormalModule` import.

## Validation

- `pnpm run build:js` — passed, including declaration generation.
- `pnpm exec rs check packages/rspack/src/builtin-plugin/css-extract/index.ts` — passed.
- `pnpm --dir tests/rspack-test run test --project base -t css-extract` — 12 test files passed; 57 tests reported passed, including 2 existing filtered placeholders. Covers extraction, HMR with `runtime: false`, watch, and incremental cases.
- Integration tests used the rebuilt JavaScript with an existing local native binding from worktree `ce35` (commit `189709c3d737`); the native binding was not rebuilt at this PR's base commit.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15752, opened 2026-09-17
feat: file system info path classification

ahabhgk:
## Motivation

The new cache's snapshot path handling differs from webpack's `FileSystemInfo`, which can cause incorrect cache reuse or invalidation. For example, an immutable directory can also match a sibling with the same prefix, and searching upward for `package.json` can select an internal package instead of the managed package root.

Context snapshot creation also fails on a symlink whose target directory is missing because it resolves the link with `canonicalize`.

## Changes

- Classify and cache paths in `FileSystemInfo` with webpack's unmanaged → immutable → managed precedence, directory boundaries for string matchers, and the first regexp capture as the managed root. Update the new cache's default `managedPaths` regexp accordingly.
- Align managed package discovery and metadata handling with webpack, including scoped packages, nested `node_modules`, missing packages, grouping directories, and fallback to normal snapshots when package metadata is unavailable.
- Apply the same classification to file, context, and missing dependencies and recursive context reads, and retain the corresponding package metadata dependencies.
- Add `read_link` to the filesystem abstraction and Node bindings so context snapshots can read and normalize symlink targets even when the target does not exist.
- Port webpack's `FileSystemInfo` unit tests and add the `newCacheCases/snapshot/mixed-paths` integration case covering immutable boundaries, unmanaged overrides, managed/scoped/nested packages, and contexts without package metadata.

Remaining webpack compatibility gaps, including symlink read retries and shared snapshot optimization, are recorded with expected-panic tests; missing APIs have TODO explanations.

Validation: `pnpm run build:binding:dev` and `cargo test -p rspack_core --lib file_system_info::tests` passed. The Rust suite reports 163 passed and 0 failed, including 41 expected-panic cases for outstanding compatibility work.


ahabhgk:
@codex review

intellild (approved):

PR #15755, opened 2026-09-18
perf: preallocate chunk graph collections

SyMind:
## Motivation

Chunk graph construction grows several maps from empty even though module, entry, and async block counts are already available. Reserving capacity up front can avoid repeated allocation and rehashing while populating these maps.

## Changes

- Reserve module, chunk, chunk group, entrypoint, and CodeSplitter maps directly in `build_chunk_graph`, using existing collection lengths without extra counting passes or temporary sets.
- Estimate chunk groups as entries plus async blocks, and allow one additional runtime chunk per entry. These are upper-bound estimates; shared names and inlined or unreachable blocks can leave unused capacity.
- Expose thin `reserve(additional)` methods on the chunk containers and preserve existing entries when reserving capacity.

Local CodSpeed measurements for `rust@build_chunk_graph`, comparing main at `1032815020` with this change at `10695fbed6`, using the same ARM64 Linux Docker container (8 vCPUs), Rust toolchain, and `codspeed` build profile:

| Mode | Main | PR | CodSpeed speedup |
| --- | ---: | ---: | ---: |
| [Simulation](https://app.codspeed.io/web-infra-dev/rspack/runs/compare/6aacb41c4177a00d11fb69ee..6aacb66ac3556e8d3da3bcf1) | 157.69 ms | 148.85 ms | 5.94% |
| [Walltime, first pair](https://app.codspeed.io/web-infra-dev/rspack/runs/compare/6aacb93ac3556e8d3da3bd0a..6aacba4c4177a00d11fb6a72) | 9.00 ms | 8.27 ms | 8.72% |
| [Walltime, repeat pair](https://app.codspeed.io/web-infra-dev/rspack/runs/compare/6aacba8a4177a00d11fb6a75..6aacbaad4177a00d11fb6a7b) | 8.62 ms | 8.21 ms | 5.00% |

Simulation values are modeled CPU time. Walltime values are CodSpeed's minimum of 100 samples per run; medians also decreased in both pairs (9.70 → 8.94 ms and 9.42 → 8.85 ms). Walltime ran with `CODSPEED_PROFILER_ENABLED=false` and `CODSPEED_ISOLATION=false` because this container lacks perf permissions and systemd CPU isolation. These VM measurements are local reference results, not a full-build speedup claim.

The simulation flamegraph shows `ChunkGraph::add_module` decreasing from about 7.6 ms to 3.9 ms, with its previous `reserve_rehash` hotspot removed. The benchmark's chunk-count and diagnostic assertions pass in both modes.

Validation: Rust formatting, development binding and JS builds, and targeted Clippy for `rspack_core` with warnings denied all pass. The existing `configCases/chunk`, `configCases/entry/`, `configCases/worker/`, and `hotCases/chunk` suites pass (316 tests, including incremental and HMR coverage).


SyMind:

PR #15756, opened 2026-09-18
perf(core): replace BuildMeta FreezeLock with atomic fields

intellild:
## Motivation

`BuildMeta` contains small scalar values, but accessing them goes through `FreezeLock`, including a read lock before freezing and a publication-state check afterward. Store these values directly in modules while retaining failed-rebuild recovery through shared module references.

For example, reading a module's export configuration can load one atomic value containing both `exports_type` and `default_object`, without obtaining a metadata guard. Existing frozen reads already avoid the RwLock; end-to-end compilation performance has not yet been benchmarked.

## Changes

- Replace `FreezeLock<BuildMeta>` with directly owned `BuildMeta` across module implementations and return `&BuildMeta` from the module interface.
- Use `AtomicCell<Option<bool>>` for independent flags, preserving `None` versus `Some(false)`. Store the related export fields in one aligned atomic cell with combined reads/writes and compile-time lock-free checks on supported targets.
- Implement generic `RspackHash` and cache-conversion support for `AtomicCell<T>`. Preserve the existing metadata hash inputs and flattened JSON field names, including omission of absent optional fields.
- Derive Serde serialization through a reusable `AtomicCell<T>` adapter, omitting absent flags and flattening the export pair into the original JSON fields in the same order.
- Clone `BuildMeta` into independent cells for failed-rebuild recovery, concatenated modules and DLL manifests. No separate metadata snapshot type is needed. Ordinary setters still require `&mut self`; shared-reference restoration stays in the post-repair phase, separate from cloning, hashing and caching.
- Remove the BuildMeta-freezing API. Retain `FreezeLock::freeze_with` with a method-local `#[allow(unused)]`, since BuildMeta no longer calls it.

For example, when a module with top-level await fails to rebuild, recovery restores the previous metadata values from an independent `BuildMeta` clone. Restoring the export configuration updates both related fields in one atomic operation.

## Validation

- `cargo test -p rspack_macros_test --test rspack_hash --offline`: 3 tests covering generic atomic hashing, metadata hash/JSON compatibility, cache round trips, and independent clones.
- `cargo clippy --workspace --all-targets --tests --locked --offline -- -D warnings`
- After the Serde refactor: all 3 Rust compatibility tests and 19 DLL integration cases passed with the rebuilt native binding.
- `cargo fmt --all --check`
- Development native binding build; JavaScript packages built before integration tests.
- 40 focused integration cases passed, run in separate groups: failed-rebuild metadata recovery, JSON, scope hoisting, DLL generation/consumption, and the runtime-mode DLL regression.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15760, opened 2026-09-18
chore: add CLAUDE.md that imports AGENTS.md

stormslowly:
## Motivation

Claude Code only reads `CLAUDE.md` and does not pick up `AGENTS.md` yet (see https://github.com/anthropics/claude-code/issues/6235), so the project instructions in `AGENTS.md` are ignored when contributors use Claude Code.

## Changes

Add a root `CLAUDE.md` containing only `@AGENTS.md`, which imports `AGENTS.md` so it stays the single source of truth.


chenjiahan (approved):

CPunisher:
Now we can remove it https://x.com/trq212/status/2101009392611278961

PR #15764, opened 2026-09-18
refactor: dispatch JavaScript loaders through a shared channel

intellild:
## Motivation

JavaScript loaders currently require a lazily retrieved runner, TSFN ownership and cache cleanup, and a separate `start_yielding` hook chain. Nested `this.importModule()` calls make this synchronization particularly delicate: an outer loader must be able to await another module while its loaders continue executing.

For example, when loader A awaits `importModule()` for a module using loader B, serializing runner execution behind A would prevent B from completing. A shared task queue lets JavaScript start B while A remains pending.

## Changes

- Dispatch JavaScript loader tasks through a shared Tokio mpsc channel and return each result through a oneshot channel. The JavaScript receive loop starts loaders without awaiting their completion.
- Route tasks by numeric compiler ID through a `Map<number, WeakRef<Compiler>>`. Remove entries after compiler close, with a finalization registry for garbage collection.
- Stop the receive loop when all builds are idle so a pending receive does not keep Node alive.
- Invoke the dispatcher from `JsLoader::pitch/run`, removing `_runLoader`, the runner getter and TSFN cache, and the `start_yielding` hook chain. Retain the per-compiler no-pitch optimization.
- Add cases for concurrent compilers with nested loaders, error and close isolation, repeated builds, and natural process exit with an unclosed compiler.

## Validation

- `pnpm run build:binding:dev` and `pnpm run build:js`
- `cargo test -p rspack_loader_runner --offline` — 10 passed
- `cargo clippy -p rspack_binding_api --offline -- -D warnings`
- `cargo fmt --all --check`, targeted JS lint/format checks, and `git diff --check`
- Focused integration suites passed: loader-channel, libuv-handles, object-lifecycle, compiler-close, config/normal loader cases, loader-import-module, and the import-module race-condition case.

Full-suite and WASI runtime validation were not run; storage and native watcher tests were skipped.

---
_by OpenAI Codex_


intellild:
Close for no benefit

PR #15767, opened 2026-09-18
fix: skip entry modules without JavaScript in the chunk startup

### LINKED ISSUE #15766: [Bug]: an entry that lists a stylesheet next to a script calls a module the bundle no longer defines (2.2.4+)
### System Info

macOS arm64, Node 24.18.1, @rspack/core 2.2.3 vs 2.2.6 with the matching @rspack/binding

### Details

since 2.2.4 an entry that lists a stylesheet next to a script, `['./src/plain.js', './src/styles.css']`, emits a startup that calls the stylesheet's module id, but no factory exists for it, so the bundle throws on load

```
TypeError: e[249] is not a function
```

2.2.3 is fine. bisected by swapping only @rspack/core in the same install: 2.2.4, 2.2.5 and 2.2.6 all show it. #14879 made such a stylesheet a CSS-only module, and the bootstrap in `rspack_plugin_javascript` still walks every entry module. in development mode the stylesheet, being last, also receives the entry's exports object. webpack guards that loop with `getModuleSourceTypes(entryModule).has("javascript")`

2.2.6 output

```js
(()=>{"use strict";var e={829(){console.log("page")}};e[829](),e[249]()})();
```

2.2.3 output

```js
(()=>{"use strict";var r={249(r){r.exports={}},829(){console.log("page")}},e={};function o(t){...}o(829),o(249)})();
```

### Reproduce link

https://github.com/cezaraugusto/rspack-css-entry-startup

### Reproduce Steps

```bash
git clone https://github.com/cezaraugusto/rspack-css-entry-startup
cd rspack-css-entry-startup
npm install
npm test
```

the test builds the same two files with 2.2.3 and 2.2.6 through package aliases, prints both bundles and runs them. config is `experiments.css: true`, a `type: "css"` rule and production mode, nothing else

cezaraugusto:
## Motivation

since 2.2.4 an entry that lists a stylesheet next to a script, `['./index.js', './style.css']`, emits a startup that calls the stylesheet's module id, but no factory exists for it, so the page throws `TypeError: e[249] is not a function` on load. #14879 made such a stylesheet a CSS-only module, and the bootstrap still walks every entry module. fixes #15766, repro: https://github.com/cezaraugusto/rspack-css-entry-startup

## Changes

the bootstrap, the inlined startup and the `render_startup` hook only see entry modules with a JavaScript source type, in both the legacy and the runtime template. same filter as webpack's `renderBootstrap` and as `generate_entry_startup` in the runtime plugin. new config case `css/entry-with-stylesheet` fails on main and passes here. `css/entry-css-imports-css-module` covers the other side, a stylesheet entry that pulls a css module into the chunk, so the chunk renders JavaScript with no JavaScript entry module and the startup stays empty


LingyuCoder:
@intellild PTAL

intellild (approved):

intellild:
@codex review

PR #15782, opened 2026-09-20
perf: emit assets natively when the output filesystem is plain node:fs

stormslowly:
## Motivation

`compiler.outputFileSystem` defaults to `node:fs`, yet every emitted asset still crosses the language boundary: Rust copies the asset into a Node `Buffer` and the write is finished on the JS main thread through a threadsafe function, which then calls `fs.writeFile` — the very filesystem Rust could have used directly.

Measured on a 2000-chunk project (6 independent cold builds each, medians):

| | plain `node:fs` (before) | native (after) |
| --- | --- | --- |
| main thread busy during emit | 133ms | **8ms** |
| peak RSS during emit | +22.6MB | **+2.5MB** |
| emit wall time | 165ms | 136ms |

The main-thread number is the point: during that window no JS plugin, dev-server request or timer can run. Emit wall time itself is within noise — repeated runs put it on either side of the JS path — so this is about not paying for a round trip that buys nothing, not about writing files faster.

## Changes

The output filesystem argument of `JsCompiler` is now optional; passing `undefined` makes Rust use its own `NativeFileSystem`. `Compiler.ts` passes `undefined` when `outputFileSystem` is still the `node:fs` module, so anything custom (memfs under dev-server, `graceful-fs`, test harnesses) keeps the existing bridge.

One behavior gap had to be closed: `output.clean: true` on a fresh build hits `remove_dir_all` on a directory that does not exist. The JS `rmrf` helper swallows `ENOENT`; `NativeFileSystem` reports it. The cleanup now ignores `NotFound`.

## Tests

`configCases` (3870), `Watch`/`NativeWatcher`/`Incremental-watch`/`Compiler`/`MultiCompiler` (1062) — green locally.


stormslowly:
close due to breaking the Binding compatible rule.

PR #15790, opened 2026-09-20
feat: cache context modules with filesystem snapshots

ahabhgk:
## Motivation

`ContextModule` cannot reuse the new module build cache, so unchanged context requests still rebuild across compiler restarts. Support caching context modules and restoring their dependencies and async blocks through the same cache lifecycle as `NormalModule`.

## Changes

- Create filesystem snapshots for context modules and validate them in `need_build`, using `cache.snapshot.contextModule` independently of the normal-module snapshot strategy. Respect module cacheability and value dependency versions when checking cache entries.
- Exclude the resolver callback from serialization and consume it during a fresh build. Valid cache entries reuse the cached module; invalid entries build the module created by the factory.
- Add `newCacheCases` covering all six context modes, empty contexts, async-block restoration, file additions, removals and content changes, and timestamp-only invalidation. Assert builds and cache hits across compiler restarts with the default incremental settings.


ahabhgk:
@codex review

intellild (approved):

PR #15808, opened 2026-09-21
perf: optimize placeholder and Node external matching

intellild:
## Motivation

Generated placeholders have fixed markers, but their replacement uses regexes or repeatedly initialized string searchers. For example, `RSPACK_AUTO_URL_STATIC_PLACEHOLDER_42` only needs a literal-prefix search followed by decimal-ID parsing.

The same marker is searched across many short CSS identifiers. Its searcher must be initialized once and reused so that preprocessing does not outweigh the cost of searching.

Node externals also allocate builtin-name strings and compile `^node:` where a synchronous request predicate is sufficient.

## Changes

- Add `rspack_util::placeholder::PlaceholderFinder`, backed by `memchr::memmem::Finder`, and hold fixed-marker scanners in global `LazyLock` values. Support parsed markers, numeric IDs, literal ranges, and literal replacement while preserving byte offsets and overlapping invalid-prefix handling.
- Migrate runtime globals, static asset and worker URLs, ESM chunk references, SRI hashes, Rstest flags, native CSS and CSS extract placeholders, and JavaScript/ESM asset public paths.
- Add lazy replacement that computes the replacement text once after the first match, then continues the same iterator. CSS module IDs use this to avoid a separate presence scan; unchanged strings remain borrowed.
- Compute the output-relative public path once per rendered source. Preserve CSS extract replacement order and `ReplaceSource` offsets.
- Extract `is_node_external(request)`, reusing `starts_with("node:")` and builtin-name matching, retaining Yarn PnP's `pnpapi`, and invoking it through a synchronous external predicate.
- Remove unused regex and `cow-utils` dependencies. Extend integration fixtures for repeated URLs after Unicode text, unresolved SRI markers, and Node-prefixed externals.

For example, thousands of CSS class names containing `__RSPACK_PLUGIN_CSS_MODULE_ID__` share one precomputed searcher. Each identifier is scanned once, and module-ID normalization is skipped if there is no marker. Two automatic public-path markers in `css/main.css` both become `../`, with that path computed once.

For Node externals, both `fs` and `node:fs` remain external. Unknown `node:` names remain external for forward compatibility; an ordinary request such as `./node:local` still goes through module resolution.

## Validation

- Rebuilt the native binding and passed formatting plus Clippy (`--all-targets -- --deny warnings`) for all eight affected crates.
- 618 focused configuration cases passed: CSS/CSS extract/CSS loader (244), asset/asset URL (170), worker (44), SRI (24), modern-module (30), runtime (94), and Rstest (12), including applicable runtime-mode variants.
- Differential validation passed: 120,000 parsed-marker scans against the previous regexes and 37,268 literal replacements against `str::replace`, plus lazy-callback and concurrent shared-scanner checks.
- Local release microbenchmark using 1,024 identifiers from the CSS fixtures: the original `contains/cow_replace` path took 610.5 ns/replacement; the reusable object took 457.9 ns/replacement (five-round medians). These are local walltime measurements, not full-build CodSpeed results.
- Node externals: 284 configuration cases passed across externals (122), target (48), and library (114), with 795 request classifications matching the previous implementation.

## Rust memory comparison

Compared main `cb9a731` with this PR's `61bde4b` using 1,000 generated fixture units per configuration, 32 CSS rules per stylesheet, and five fresh-process cold builds per version/configuration. The table reports medians.

Both native builds use identical benchmark-only allocation counters around mimalloc and identical build settings. Peak memory counts live Rust heap allocation sizes; allocation volume counts cumulative allocation requests. These measurements exclude V8, allocator overhead/fragmentation, stack memory, and direct C allocations.

| Scenario | Main peak (MiB) | PR branch peak (MiB) | Peak change | Allocation volume change |
|---|---:|---:|---:|---:|
| Runtime templates | 39.32 | 39.19 | -0.33% | -0.47% |
| Native CSS + SRI | 248.10 | 246.75 | -0.54% | -1.84% |
| CSS extraction: auto path + SRI | 171.21 | 171.26 | +0.03% | -1.61% |
| CSS extraction: relative path | 173.85 | 173.65 | -0.11% | -1.63% |
| Relative URLs | 38.39 | 38.37 | -0.06% | -0.39% |
| Static URLs and workers | 41.13 | 38.15 | **-7.25%** | **-4.34%** |
| Modern-module output | 105.05 | 102.01 | **-2.89%** | -1.61% |
| ESM with preserved asset imports | 77.26 | 74.27 | **-3.87%** | -1.90% |
| Rstest mock hoisting | 35.68 | 35.71 | +0.08% | -0.08% |
| Node externals | 32.94 | 32.84 | -0.29% | -0.27% |

The URL/Worker and two ESM configurations have non-overlapping peak ranges across these five samples. CSS, Rstest, and Node external peak ranges overlap, so their small peak differences are inconclusive. All ten cumulative allocation-volume medians decrease.

All 100 measured builds and 20 audit builds passed with identical output hashes across both versions and all rounds. Coverage includes the runtime-template global replacement, all four CSS extract placeholder forms, CSS module IDs/text exports, static URLs/workers, ESM references, SRI, Rstest flags, and Node externals.

[Full report, sample ranges, raw measurements, and reproduction details](https://github.com/intellild/rspack-bench/blob/93e2e68534a5a4b89755ef61ee32a6c87a7c3d8e/cases/placeholder-memory/results/pr15808-2026-09-22/report.md).

---
_by OpenAI Codex_


intellild:
@codex review

PR #15810, opened 2026-09-21
perf: avoid redundant parallel loader input cloning

SyMind:
## Motivation

Parallel loaders currently call `structuredClone` to validate every task before Tinypool sends it to a worker. The successful path therefore serializes the full loader context and arguments twice: once for validation and once for `postMessage`.

In the stable metadata benchmark with four loader workers, removing the redundant validation clone reduced the paired `compiler.run` wall time by 13.9% (121.38 ms to 103.14 ms; candidate/baseline 95% interval 0.814–0.900).

Benchmark harness and reproduction steps: [SyMind/rspack-parallel-loader-bench](https://github.com/SyMind/rspack-parallel-loader-bench).

Tinypool scheduler fix submitted upstream: [tinylibs/tinypool#142](https://github.com/tinylibs/tinypool/pull/142).

## Changes

- Let Tinypool's actual `postMessage` perform the successful-path serialization.
- Run the existing field-level cloneability diagnostics only when the input send reports `DataCloneError`, including Node 20's missing-transferable error.
- Keep draining Tinypool's pending queue when a queued task fails `postMessage`, so one invalid loader input cannot block later builds.
- Close both sides of task-specific message channels on send failure.
- Add coverage proving successful input is serialized once, function and `MessagePort` options retain precise diagnostics, and an unclonable loader result is not misclassified as an input error.

Validation:

- `pnpm run build:js`
- `pnpm exec rs check packages/rspack/src/loader-runner/service.ts`
- `pnpm --dir tests/rspack-test run test -t 'configCases/loader-parallel'` (64 passed)
- The same 64-case suite with `RSPACK_LOADER_WORKER_THREADS=1` on Node 22.22.3 and Node 24.15.0
- Node 20.20.2 direct compilation of the new clone/error cases


SyMind:
@codex review

PR #15816, opened 2026-09-21
fix(css): align CSS Modules behavior with webpack css-loader fixtures

intellild:
## Motivation

Port webpack's `css/css-loader` fixtures to exercise Rspack's built-in CSS Modules support across `link`, `text`, `css-style-sheet`, and `style` exports. Comparing the generated snapshots exposed missing nested selectors, incorrect ICSS overwrites, duplicate CSS modules, and undefined self-composed exports.

## Changes

- Copy the upstream fixtures and adapt the configuration and snapshot assertions to the Rspack harness, including an equivalent template for webpack's `localIdentName` callback.
- Resume selector scanning after consuming `composes`. For example, `.bar { composes: foo; &.baz { color: blue; } }` previously left `baz` unscoped and unexported; it now receives a local name.
- Apply last-declaration-wins semantics to ICSS exports: `:export { value: first; value: second; }` now exports `second` instead of `first second`. Treat `--`-prefixed ICSS keys as literal names and let only the configured export convention create aliases. For example, with `as-is`, `:export { --color: red; --color: blue; }` exports only `{ "--color": "blue" }`, without an extra `color: "red blue"` export or contamination of an independent `color` export. Regression cases cover repeated declarations within and across blocks, camel-case conventions, module concatenation, and all four export types; real CSS custom-property exports remain covered.
- Reuse CSS modules reached through JS and composition when their export types match, while retaining distinct CSS import rendering roles and export-type overrides. This avoids emitting duplicate styles for the same module.
- Preserve the current CSS import rendering role for same-module compositions. For example, `@import "./self.module.css"` followed by `composes: self from "./self.module.css"` used to create a second module without the import suffix, producing different default class names and duplicate CSS rules. Matching the resolved request and retaining the original render-condition/export-type suffix makes the composition resolve to the existing module. Tests cover unconditional, media, and layer imports across four export types, with module concatenation and module layers enabled and disabled.
- Expand same-module composition references during code generation with a visited-export guard, following webpack's cycle detection. A class composing itself now exports its class name once instead of appending `undefined`. Pure ICSS cycles with no concrete value retain their existing runtime resolution. Preserve external references as runtime reads: combining a self-composition with `composes: token from "./tokens.js"` retains the JS-provided token and external module evaluation. Unlike webpack's fully static resolution, this preserves Rspack's existing dynamic reexports, as supported by css-loader with JS-compatible loader settings.
- Add regression coverage for both composition orders, indirect and mutual same-file references, pure ICSS cycles, dynamic JS values, and external CSS style injection across all four export types with module concatenation enabled and disabled.
- Correct two existing reexport snapshot entries that concatenated an automatic ICSS alias with the actual CSS variable name; the emitted CSS remains unchanged.
- Add 157 snapshots. A subagent compared all 156 class maps and 1,990 CSS AST nodes against webpack, excluding hashes, equivalent escaping, quotes, and serialization differences. Existing Rspack behavior for undeclared animation names and complete transitive composition is preserved and noted in the test.

Validation:
- Built the JS packages and native binding.
- CSS config tests in normal and Rspack runtime modes: **238 passed**; the 157 css-loader snapshots are unchanged, with only the two reexport entries described above corrected.
- `cargo test -p css-module-lexer`: **287 passed**.
- `cargo clippy -p css-module-lexer -p rspack_core -p rspack_plugin_css --all-targets -- -D warnings`: passed; reran CSS clippy after the review fixes.
- `cargo fmt --all --check`: passed.

---
_by OpenAI Codex_


intellild:
@codex review

intellild:
@codex review

intellild:
@codx review

PR #15821, opened 2026-09-21
perf: reduce exports info and parser memory footprint

LingyuCoder:
## Motivation

The exports info artifact is one of the largest native structures retained while building a real app: on a 20k-module fixture it is the single biggest allocation site (75.8 MiB peak, 9.7 MiB still alive when the build process exits). The parser shows a second, unrelated site: `JavascriptParser::new` pre-allocates vectors that are moved into the scan result, so their capacity is a per-module cost that stays alive for the rest of the build.

Three things dominate the footprint:

- `ExportsInfoData::exports` is a `BTreeMap<Atom, ExportInfoData>`. A `BTreeMap` node allocates `11 * size_of::<ExportInfoData>()` (1512 bytes here) as soon as the first named export is inserted, and nearly every module has at least one, while most modules only export a handful of names.
- `ExportInfoData` keeps `target` and `used_in_runtime` as inline hash maps. They cost roughly 40 bytes each in every export info (two of them live in every exports info, plus one per named export) but are only populated for a minority of exports.
- The parser's `dependencies`, `blocks` and `presentational_dependencies` vectors are pre-allocated for 64 entries each (1024 + 512 + 1024 bytes per module), while most modules have a handful of dependencies, no async blocks and no presentational dependencies.

## Changes

- Add `NamedExports`, an inline `SmallVec<[(Atom, ExportInfoData); 1]>` kept sorted by name, and use it for `ExportsInfoData::exports`. The common case (0-1 named exports) allocates nothing, the key order matches the previous `BTreeMap` iteration order, and the container exposes the operations the call sites used (`get`, `get_mut`, `insert`, `remove`, `iter`, `values`, ...).
- Move the entries back into a `BTreeMap` once a module has more than 16 named exports. Inserting into a sorted array memmoves the trailing entries: free at a handful of exports, quadratic at a few thousand, and a CSS module exposes one export per class. Modules that outgrow the inline vector keep the `O(log n)` insert they had before.
- Box `target`, `used_in_runtime` and `can_inline_provide` on first use, which drops `size_of::<ExportInfoData>()` from 128 to 56 bytes.
- Start the parser vectors at 16/4/8 and let the rare large module grow.

No behavior change is intended: key ordering is preserved and the boxed maps are created on first mutable access.

## Validation

- heaptrack on a 20k-module production build (same binary otherwise): allocations of the 1512-byte `BTreeMap` node drop from 20,121 to 27; peak native heap 530.5 -> 489.2 MiB (-7.8%); bytes still alive at process exit 410.8 -> 378.8 MiB (-7.8%).
- Real app A/B (8 warm production builds per side, alternating): peak RSS 15.39 -> 15.22 GiB (-1.13%); 20-30s window -1.17% (t ~ -2.1); 30-40s window -1.54% (t ~ -3.7). Dev server steady-state RSS 13.18 -> 12.98 GiB (-1.5%). The emitted asset list and per-file sizes are identical between the two sides.
- react-10k, 20k modules with the cache disabled: the parser change is -22.4 MiB peak RSS (-2.37%) and -18.9 MiB dev server steady state (-1.40%) with the V8 heap unchanged; the exports container keeps the production peak flat (median -6 MiB over 8 alternating builds per side, inside the run-to-run spread) and the dev server steady state drops by 36 MiB (-2.7%) over two paired runs.
- Large-module insert cost: a module with 25k named exports inserted in random order builds in 148 ms with the spill container, versus 332 ms with a pure sorted vector and 157 ms before this PR.
- `cargo fmt --check` and `cargo check -p rspack_binding_api` pass locally; CI covers the test suites.


LingyuCoder:

LingyuCoder:

LingyuCoder:
@codspeedbot explain the improvement

stormslowly (approved):

PR #15823, opened 2026-09-22
fix(rstest): resolve mock APIs through the runtime registry

9aoy:
## Motivation

Rstest's previous `importMock` transform handled a literal request entirely at compile time. It resolved the request to a manual mock path and generated a direct dependency on the `__mocks__` file. That meant the API could load a manual mock file, but it could not honor a mock factory registered through `rs.mock` or `rs.doMock` when the test later called `rs.importMock`: the generated code had already bypassed the mock registry.

The previous transform also did not handle a variable request. A call such as `rs.importMock(request)` remained a runtime API call and Rstest reported that it had not been transformed. This is the pattern used by runtime loaders in real projects.

## Changes

This PR changes the flow in two related parts:

1. **Registry lookup for static mock APIs**
   - Keep the bundle dependency and the normal mock registration.
   - Generate static `importMock` through the request-keyed runtime registry.
   - Let a registered factory win over the manual `__mocks__` module, while retaining the manual mock as the fallback.
   - Preserve compatibility with older Rstest runtimes through guarded code generation.

2. **Limited runtime support for variable requests**
   - Rewrite variable `rs.importMock(request)` and `rs.requireMock(request)` calls when the corresponding option is enabled.
   - Append the source module origin so the runtime can resolve the request in the context where it was written.
   - Preserve the distinct async `importMock` and synchronous `requireMock` semantics.
   - This supports requests that are represented in the bundle's mock registry; it does not load arbitrary modules that were never bundled or registered.

The Rstest mock runtime template is updated together with the Rspack transform. It provides the request-keyed registry and the `importMock` / `requireMock` runtime helpers consumed by the generated code. The Rstest plugin enables the two corresponding Rspack options, while Rspack keeps them opt-in for compatibility with older Rstest templates.

Tests: `cargo fmt --all -- --check`; `cargo clippy -p rspack_plugin_rstest --all-targets -- -D warnings`; `cargo check -p rspack_plugin_rstest -p rspack_binding_api`; `pnpm --config.verifyDepsBeforeRun=false run build:binding:dev`; `pnpm --config.verifyDepsBeforeRun=false --dir tests/rspack-test run test -t "configCases/rstest/mock-dynamic-import-external"` (1 passed).


9aoy:
@codex review

9aoy:
@codex review

